PR #27, opened 2015-05-04
Do we need While/For?

lukewagner:
This is more a question than a proposal:

I know the current polyfill has while/for/do-while (b/c I was lazy and it makes it easier to generate valid asm.js) but I remember Ben had mentioned thinking all we needed was unconditional loops.  The counter-argument would be if there was some win from having the higher-level while/for/do-while ops (e.g. size or avoiding depending on backend control-flow optimizations).  Does the inclusion of these reflect more recent thinking?

It's fine if we leave this open until we can do real experiments.  Perhaps we'll find different control flow primitives (that don't simply reflect a source-level primitives) that better capture compile-time control flow optimizations.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -35,8 +35,7 @@ are statements.
 
   * Block - a fixed-length sequence of statements
   * If - if statement
-  * While - while statement
-  * For - for statement
+  * Loop - unconditional loop statement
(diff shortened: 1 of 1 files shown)


Nic30:
Was loop unrotate actually implemented in some form in WebAssembly or anywhere else?

PR #47, opened 2015-05-07
Add not-equal comparisons.

sunfishcode:
(x != y) can be synthesized as ((x == y) ^ 1), or we can reverse the arms
of a select or the blocks of an if-else, but this feels needlessly
minimalistic for such a common operation.

(I do agree though that excluding greater-than and greater-than-or-equal
makes sense, because those can be supported just by swapping the operands).


jfbastien:
Ping? I'm adding this to LLVM since we seem to have agreement. I'd like to make sure we indeed agree!

titzer:
No problem, just go ahead and add them.

PR #191, opened 2015-06-18
New Future Feature: Shared Memory ACL

Marqin:
I hereby propose to add Shared Memory ACL to Future Features:
* Shared memory ACL - thread that allocates memory pool can grant other threads
  read or read-write access to that pool, and without that grant no other thread
  could access that memory. It should be also possible for allocator thread to
  revoke given accesses ( any time ).

I also request for your comments on that.

### DIFF
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -44,6 +44,10 @@ possible to use a non-standard ABI for specialized purposes.
 * `madvise(MADV_DONTNEED)`.
 * Shared memory, where a physical address range is mapped to multiple physical
   pages in a single WebAssembly module as well as across modules.
+* Shared memory ACL - thread that allocates memory pool can grant other threads
+  read or read-write access to that pool, and without that grant no other thread
+  could access that memory. It should be also possible for allocator thread to
+  revoke given accesses ( any time ).
 
 ## More expressive control flow
 


jfbastien:
This is currently not the design we're going for. We're going to an [extensible web manifesto](https://extensiblewebmanifesto.org) approach where we'll expose the basic capabilities needed to implement things like what you suggest. This approach fits nicely with the [SharedArrayBuffer](https://docs.google.com/document/d/1NDGA_gZJ7M7w1Bh8S0AoDyEqwDdRh4uSoTPSNn77PFk/) proposal, which in turn maps nicely onto the C++11 memory model.

I believe that what you suggest can be implemented on top of this, and could be offered as a user-level library. WDYT?

Keep in mind that we have many usecases to meet, including supporting existing C++ code efficiently. Forcing ACL onto these code bases isn't something we can do, but we can have an opt-in library that offers that capability.

I'm assuming that you're talking about threads within the same process (in WebAssembly, a module and its shared objects share a heap, I call that a process). Shared memory across processes is an entirely different thing.

lukewagner:
I can see this feature being useful to catch bugs.  If paired with a form of portable signal handling, it could also allow advanced optimizations.  One concern is that, to be efficient it seems like the engine would require memory protection + signal handling support from the platform.

Marqin:
> I believe that what you suggest can be implemented on top of this, and could be offered as a user-level library. WDYT?

If that's the best way to do that, and if engine would respect and enforce that - so no other process /thread in browser ( not talking about those system-wide, because that would be hard if not impossible to block eg. root ) would really be able to peek.

> I'm assuming that you're talking about threads within the same process  [... ] Shared memory across processes is an entirely different thing.

I was thinking about both.

> I can see this feature being useful to catch bugs. 

My intention was security, but that's also good scenario.

jfbastien:
> I can see this feature being useful to catch bugs. If paired with a form of portable signal handling, it could also allow advanced optimizations. One concern is that, to be efficient it seems like the engine would require memory protection + signal handling support from the platform.

Agreed. That's the type of thing we want to enable as discussed in Tooling.md. It's pretty similar to sanitizers, we should make sure we expose the features required to do this, but I don't think we should mandate it be done on every implementation.

jfbastien:
@Marqin said:
> > I can see this feature being useful to catch bugs.
>
> My intention was security, but that's also good scenario.

WebAssembly secures users from buggy and malicious `.wasm` files, it doesn't enforce security of developer's own code. It instead offers features which allow developers to use the right tooling to secure their own application.

I filed #205 because this isn't clear at all in the current design! Thanks for pointing this out.

lukewagner:
#288 added several specific memory protection features to [FutureFeatures](https://github.com/WebAssembly/design/blob/master/FutureFeatures.md#finer-grained-control-over-memory) which I think addresses the use cases.  Feel free to reopen if there is something else to discuss.

PR #195, opened 2015-06-18
Point to the Streams Standard as that's actively being worked on

annevk:


jfbastien:
You'll have to join the [W3C Community Group](https://www.w3.org/community/webassembly) before we can merge this.

annevk:
No thanks!

MikeHolman:
Ok, then I'll close this. I'm sure we will update when we start investing work in this area.

domenic:
Are we allowed to file issues without being in the community group? Or will those get closed too?

MikeHolman:
@domenic see #199 (we're kind of still figuring this out)

sideshowbarker:
So as indicated in other updates now shown here, in the interest of getting this resolved, I duplicated @annevk’s patch at #211. (From the resolution that’s emerged at #199 it seems like this, being a minor typo-level change, could be merged even with a patch from someone who’s not a member of the CG. But in this case it’s moot since I am already a member of the CG and have signed off on the CLA.)

annevk:
Thanks @sideshowbarker! I didn't realize that removing my fork made merging this impossible. At the top it still shows exactly the change I proposed and such. Weird.

PR #199, opened 2015-06-18
Clarify CG membership rules

lukewagner:
Adrian Batement informed me that CG membership is necessary for issues that end up contributing to the spec as well, so update Contributing.md.

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -11,10 +11,10 @@ Interested in participating? We suggest you start by:
 2. Joining the IRC channel `irc://irc.w3.org:6667/#webassembly`.
 
 With that background understood and communication set up, feel free to
-[file issues][] in the WebAssembly design repository. Please join the
-[W3C Community Group](https://www.w3.org/community/webassembly/) before sending pull
-requests: it provides the legal framework that protects the work in this
-repository.
+join the [W3C Community Group](https://www.w3.org/community/webassembly/)
+and then [file issues][] or submit pull requests to the WebAssembly design
+repository. The W3C Community Group provides the legal framework that protects
+the work in this repository.
 
 As WebAssembly moves forward we expect to form an official standards body, which
 will have its own contribution process to the specification.


MikeHolman:
Yeah, he mentioned that in email to us as well... I'm not happy about it though, especially when it's something like "oh this link is broken". Is there any way we can enforce this? Having to cross-check the W3C CG list any time someone new comments feels unsustainable. Like what are you supposed to do if you have a pull request and someone points out a typo? Do you not fix it because they weren't part of the CG and then wait for someone else to realize it once everyone has forgotten that this non-member said anything?

BrendanEich:
Common sense and (IANAL) my experience with open source say you're ok taking typo and punctuation fixes, other de-minimus stuff. Anything covered by copyright or IPR could be an issue, but now you need real legal advice :-|.

kripken:
I accept that CG membership is necessary to contribute to the **spec**, but so far we just have a **design** repo, right?

I don't see any harm in accepting pulls to the design repo. It won't be submitted to a standards body. Eventually, CG members will write a spec based on input from the design repo (and other places), and we need to be sure about who writes that formal document up. But so far the design repo is just an informal place for us to throw around ideas, isn't it?

MikeHolman:
For this to actually be sustainable and to minimize the amount of headache for us to figure out legal stuff, I think we need a bot that you have to request access from by linking your GitHub with your W3C CG membership. And then the bot will give access to open issues/submit PRs here. Is that level of granularity possible?

MikeHolman:
@kripken but the problem is that things from the design repo will likely go into the spec in some form or another, even if no text is copy/pasted. And if opening issues against a spec can be considered a legal problem, contributing actual text to a the spec's precursor seems even more likely to be an issue.

kripken:
I suppose we need to get lawyers involved then, as @BrendanEich says.

lukewagner:
@kripken Well we (the companies that joined the W3C CG so far) *have* had lawyers discuss this and this is what they've said to us; that's why we have a CG in the first place.

kripken:
Apologies, not to be annoying, but I have to ask: Did the lawyers specifically say that a design - not spec - repo, i.e., a repo not for actual work on the spec, but for open and casual discussion on a future spec, is something that needs CG membership?

(I say casual, because it is on github; anyone can open an issue or a pull, just like anyone can join irc; and we have plenty of open issues that are just random thoughts and discussion, going nowhere near any future spec.)

Or are we interpreting a lawyer's statement about "materials that go directly into the spec need CG membership", as applying to the design repo?

Again, sorry if this seems annoying, I get the impression you have already looked into this carefully, but I want to understand where the interpretation of this repo comes from (directly from lawyers, or indirectly).

adrianba:
As @BrendanEich say, proposing editorial changes is fine without membership.

The issue is the IP potentially impacted by ideas that make it into the spec. What we want to avoid is Anna having a patent on something, filing an issue proposing it be added, and then having Bob write it into the spec. We need to make sure that Anna has accepted the contribution license since she proposed the idea.

It shouldn't be a big deal. W3C Community Groups were specifically designed to be low friction but to provide this protection. It provides a little more traceability than just following pull requests, which are the things you worry about more in OSS software projects where copyright and distribution is the main issue.

kripken:
I see, thank you.

adrianba:
> Did the lawyers specifically say that a design - not spec - repo, i.e., a repo not for actual work on the spec, but for open and casual discussion on a future spec, is something that needs CG membership?

Things that eventually land in the spec may be initially proposed in the design document. We'd want those to be covered. It's easiest if the people making the contributions are just a member of the CG then nobody has to worry about it. We may have to make judgements (and potentially involve lawyers) if someone makes a very interesting contribution to the design that makes it into the spec and doesn't want to join and make the contribution commitment.

darobin:
It so happens that I am developing a tool to handle precisely this sort of issue, which we're likely going to face regularly in an upcoming CG. The currently planned target date is July 1st; ping me if you want to be in the loop.

Mind you, the first version will still involve *some* manual work because we won't immediately have a good link between GitHub IDs and W3C IDs. But the thinking is that the vast majority of contributions will come, as in all open projects, from a small number of contributors. And since each contributor only needs be okayed once, the number of manual okaying ought to decrease pretty fast.

Once that works I will look into properly bridging GitHub IDs with membership in a CG (or more generally in a group) but this requires changes to things that I don't have control over so I can't give an immediate ETA.

This legal stuff is annoying but we kinda need it to keep the Web open. We're trying to progressively automate it to within an inch of its life (though I can't commit to automating your employers' lawyers just yet).

jfbastien:
Thanks @darobin, this sounds great. @lukewagner and I discussed using github's [new premissions](https://github.com/blog/2020-improved-organization-permissions), but were waiting for things to die down a bit before doing this. It seems like it would fit well with what you suggest?

darobin:
@jfbastien We've also applied to take part in the new permissions stuff, but we haven't yet had our access granted so we don't know exactly how it will pan out. That said, I'm pretty sure that it will tie in nicely with what we're building so it certainly is an avenue worth exploring. Also, we're always happy to pick up cool ideas from people so any useful way you find to apply it will certainly interest us.

johnku1:
@darobin let me know how progress goes. It sounds like a great idea!

domenic:
All specs I've participated in, even in the most patent-restrictive pay-to-play venues, have public mailing lists that I can raise issues and discuss on. If wasm restricts the GitHub issue tracker to collaborators only, will there be a public mailing list we can use?

lukewagner:
@domenic That's a good point.  So, if I understand @adrianba correctly (both above and from a separate email thread), the only CG-membership barrier is concrete ideas that wind up in the spec (and I expect there would be no difference between issues vs. mailing list in this regard).  (Is that right @adrianba?)  Thus, I think I was overzealous (in the heat of 10 parallel activities yesterday) in saying "before filing issues".  Sorry!  Perhaps the text could instead say "before proposing specific changes to the design repo, whether via PR or Issue"?

domenic:
I still don't understand. On those mailing lists I can request features, point out bugs, suggest solutions, note broken links and their correct destination, etc. Will there be such a mailing list for wasm?

lukewagner:
There was general interest in using issues for all those things (expecting them to be equivalent, but better ergonomics).  This is all new to me so a question: for W3C CGs with mailing lists, what happens if someone proposes a concrete change to the spec (the one that we're being advised to require CG membership for)?  It seems like we should be able to use the same rules for Issues in GH.

adrianba:
> I still don't understand. On those mailing lists I can request features, point out bugs, suggest solutions, note broken links and their correct destination, etc.

You can propose features but someone should get your IPR commitment before they are included. When this doesn't happen it is a bug in the process. Groups should be very suspicious of someone making a feature proposal who isn't prepared to sign an IP declaration.

In the past I have given feedback to a WG on behalf of Microsoft that we weren't a member of with the promise to make an RF commitment on the contribution if the group accepts the feedback. It's the group's responsibility to police this.

darobin:
@domenic As Adrian says, this is in no way specific to here. It also is not a blanket statement against all contributions, certainly not against reporting bugs. It is just about substantive features.

Asking people to either join or sign a royalty-free commitment applies just the same on mailing lists. The only reason you aren't aware of it is because 1) it is handled in direct communication so as no to annoy the people on the list, and 2) some stuff probably falls through the cracks. You work for a member who's participating in the relevant groups so you'd never notice this anyway.

Again, I know this is annoying, but I've been in groups with patent issues and frankly I'll take this annoyance over that "fun" any time. We'll automate as much as possible, too — we'll keep you posted.

In conclusion, there's only one rule: if someone provides substantive input into the specification (directly as a PR or indirectly in discussion), just make sure they're under the CLA. This is typically the chairs' responsibility (sorry!). Typically, the overwhelming majority of contributions are from people you know who are under the CLA, so it shouldn't be that much of a hassle.

lukewagner:
Given that, sounds like I can close this PR and we'll just be vigilant for substantive contributions that make it into the repo.

jfbastien:
@domenic on the topic of mailing lists: our standards folks suggested we stick to IRC and github instead of also having a mailing list.

domenic:
Sure, I prefer those as well, but it seems like GitHub is going to be a pretty restrictive place for this community, so I was wondering if there would be a less-restrictive mailing list.

(Despite Robin saying that I never had to deal with this because I work for Google, I've been contributing for much longer than that. Maybe I just fell through the cracks, repeatedly and consistently, for three years.)

lukewagner:
I could be over-rotating here, but what I got out of this discussion is that people can keep filing issues w/o having a priori joined the CG and it's on the core team ("Workshop") to barrier substantial additions to the repo on CG membership.

MikeHolman:
@lukewagner That's kind of what I've got out of this as well. Seems like a common sense philosophy will work reasonably well.

domenic:
That sounds reasonable. If it's the case, then perhaps #195 could be reopened and merged as a sign of good faith?

lukewagner:
I actually tried to earlier when we all confirmed that minor edits didn't need membership but it seems the repo was deleted :)

sideshowbarker:
> I actually tried to earlier when we all confirmed that minor edits didn't need membership but it seems the repo was deleted :)

In the interest of getting #195 resolved, I’ve gone ahead and duplicated it at #211.

PR #200, opened 2015-06-18
Update AstSemantics.md

titzer:
This is mostly some wordsmithing on AstSemantics.md and a bit of moving some sentences to a more logical flow.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -200,8 +200,8 @@ is why it isn't the specified default.)
 
 ### Out of bounds
 
-The ideal semantics is for out-of-bounds accesses to trap, but the performance
-implications and ability to polyfill to JS are not yet fully clear.
+The ideal semantics is for out-of-bounds accesses to trap, but the 
+implications are not yet fully clear.
 
 There are several possible variations on this design being discussed and
 experimented with. More measurement is required to understand the associated tradeoffs.
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -146,7 +146,7 @@ nested. This guarantees that all resulting control flow graphs are well-structur
 Programs address the heap by using integers that are interpreted as unsigned byte indexes
 starting at `0`.
 Accesses to the heap at indices larger than the size of the heap are considered out-of-bounds,
-and a module may optionally define that out-of-bounds includes small indices close to `0` (see https://github.com/WebAssembly/design/issues/204).
+and a module may optionally define that out-of-bounds includes small indices close to `0` (see [discussion] (https://github.com/WebAssembly/design/issues/204)).
 Out-of-bounds access is considered a program error, and the semantics are discussed below.
 Each heap access is annotated with a *Memory type* and the presumed alignment of the index.
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -1,28 +1,31 @@
 # Abstract Syntax Tree Semantics
 
-The Abstract Syntax Tree (AST) has a basic division between statements and
-expressions. Expressions are typed; validation consists of simple, bottom-up,
-`O(1)` type checking.
+WebAssembly code is represented as an abstract syntax tree
+that has a basic division between statements and
+expressions. Each function body consists of exactly one statement.
+All expressions and operations are typed, with no implicit conversions or
+overloading rules.
+
+Verification of WebAssembly code requires only a single pass with constant-time
+type checking and well-formedness checking.
 
 Why not a stack-, register- or SSA-based bytecode?
-* Smaller binary encoding: [JSZap][], [Slim Binaries][].
+* Trees allow a smaller binary encoding: [JSZap][], [Slim Binaries][].
 * [Polyfill prototype][] shows simple and efficient translation to asm.js.
 
   [JSZap]: https://research.microsoft.com/en-us/projects/jszap/
   [Slim Binaries]: https://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.108.1711
   [Polyfill prototype]: https://github.com/WebAssembly/polyfill-prototype-1
 
-Each function body consists of exactly one statement.
-
-The operations available in the AST are defined here in language-independent
-way but closely match operations in many programming languages and are
-efficiently implementable on all modern computers.
+WebAssembly offers a set of operations that are language-independent but closely
+match operations in many programming languages and are efficiently implementable
+on all modern computers.
 
 Some operations may *trap* under some conditions, as noted below. In the MVP,
 trapping means that execution in the WebAssembly module is terminated and
 abnormal termination is reported to the outside environment. In a JS
-environment, such as a browser, this would translate into a JS exception being
-thrown. If developer tools are active, attaching a debugger before the
+environment such as a browser, a trap results in throwing a JS exception.
+If developer tools are active, attaching a debugger before the
 termination would be sensible.
 
 Callstack space is limited by unspecified and dynamically varying constraints.
@@ -90,20 +93,19 @@ perform saturation, trap on overflow, etc.
 
 ## Addressing local variables
 
-All local variables occupy a single index space local to the function.
-Parameters are addressed as local variables. Local variables do not have
-addresses and are not aliased in the globals or the heap. Each function has a
-fixed, pre-declared number of local variables. Local variables have *Local
-types* and are initialized to the appropriate zero value for their type at the
-beginning of the function, except parameters which are initialized to the values
+Each function has a fixed, pre-declared number of local variables which occupy a single
+index space local to the function. Parameters are addressed as local variables. Local
+variables do not have addresses and are not aliased in the globals or the heap. Local
+variables have *Local types* and are initialized to the appropriate zero value for their
+type at the beginning of the function, except parameters which are initialized to the values
 of the arguments passed to the function.
 
   * `get_local`: read the current value of a local variable
   * `set_local`: set the current value of a local variable
 
-The details of index space for local variables and their types needs
-clarification, e.g. whether locals with type `int32` and `int64` must be
-contiguous and separate from others, etc.
+The details of index space for local variables and their types will be further clarified,
+e.g. whether locals with type `int32` and `int64` must be contiguous and separate from 
+others, etc.
 
 ## Control flow structures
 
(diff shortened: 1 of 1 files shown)


AstSemantics.md
```diff
@@ -198,12 +200,11 @@ is why it isn't the specified default.)
 
 ### Out of bounds
 
-The ideal semantics is for out-of-bounds accesses to trap. A module may
-optionally define that "out of bounds" includes low-memory accesses.
+The ideal semantics is for out-of-bounds accesses to trap, but the performance
+implications and ability to polyfill to JS are not yet fully clear.
```
sunfishcode: Polyfill considerations were specifically moved out [to a separate document](Polyfill.md).

AstSemantics.md
```diff
@@ -350,7 +352,7 @@ The same operations are available on 64-bit integers as the ones available for
 Floating point arithmetic follows the IEEE-754 standard, except that:
  - The sign bit and significand bit pattern of any NaN value returned from a
    floating point arithmetic operation other than `neg`, `abs`, and `copysign`
-   are computed nondeterministically. In particular, the "NaN propagation"
+   are not specified. In particular, the "NaN propagation"
```
sunfishcode: "Not specified" makes it unclear whether implementations are allowed to change their behavior over time. We changed this text to say "nondeterministic" since there seems to be a consensus that running programs could migrate from one CPU to another with a different implementation, and there doesn't seem to be any clean way to handle that in the semantics.

AstSemantics.md
titzer: On Thu, Jun 18, 2015 at 3:58 PM, Dan Gohman <notifications@github.com>
wrote:

> In AstSemantics.md
> <https://github.com/WebAssembly/design/pull/200#discussion_r32787666>:
>
> > @@ -198,12 +200,11 @@ is why it isn't the specified default.)
> >
> >  ### Out of bounds
> >
> > -The ideal semantics is for out-of-bounds accesses to trap. A module may
> > -optionally define that "out of bounds" includes low-memory accesses.
> > +The ideal semantics is for out-of-bounds accesses to trap, but the performance
> > +implications and ability to polyfill to JS are not yet fully clear.
>
> Polyfill considerations were specifically moved out to a separate document
> <http://Polyfill.md>.
>

That's fine. It just read funny previously. "The ideal semantics is for
XXX." And then a bulleted list of other potential things, leaving the
reader wondering why. Thus, a one sentence justification.

AstSemantics.md
titzer: On Thu, Jun 18, 2015 at 3:59 PM, Dan Gohman <notifications@github.com>
wrote:

> In AstSemantics.md
> <https://github.com/WebAssembly/design/pull/200#discussion_r32787706>:
>
> > @@ -350,7 +352,7 @@ The same operations are available on 64-bit integers as the ones available for
> >  Floating point arithmetic follows the IEEE-754 standard, except that:
> >   - The sign bit and significand bit pattern of any NaN value returned from a
> >     floating point arithmetic operation other than `neg`, `abs`, and `copysign`
> > -   are computed nondeterministically. In particular, the "NaN propagation"
> > +   are not specified. In particular, the "NaN propagation"
>
> "Not specified" makes it unclear whether implementations are allowed to
> change their behavior over time. We changed this text to say
> "nondeterministic" since there seems to be a consensus that running
> programs could migrate from one CPU to another with a different
> implementation, and there doesn't seem to be any clean way to handle that
> in the semantics.
>
Well, we do say "any NaN value returned from a floating point arithmetic
operation", which with the change to "not specified" should imply that
different operations could return different bit patterns.

The motivation for this change is that "are computed nondeterministically"
could be read as prescriptive, implying that the engine _must_ compute the
bits nondeterministically.

AstSemantics.md
sunfishcode: I agree. I think your rewording is fine if you take out "and ability to polyfill to JS".

titzer: On Thu, Jun 18, 2015 at 7:08 PM, Dan Gohman <notifications@github.com>
wrote:

> In AstSemantics.md
> <https://github.com/WebAssembly/design/pull/200#discussion_r32796589>:
>
> > @@ -198,12 +200,11 @@ is why it isn't the specified default.)
> >
> >  ### Out of bounds
> >
> > -The ideal semantics is for out-of-bounds accesses to trap. A module may
> > -optionally define that "out of bounds" includes low-memory accesses.
> > +The ideal semantics is for out-of-bounds accesses to trap, but the performance
> > +implications and ability to polyfill to JS are not yet fully clear.
>
> I agree. I think your rewording is fine if you take out "and ability to
> polyfill to JS".
>
Done.

AstSemantics.md
sunfishcode: To be clear, even the same operation, with the same operand values, can return different values each time it is executed. But I see what you mean about the existing text being prescriptive. I'll be ok with "not specified" for now.

sunfishcode:
lgtm

AstSemantics.md
```diff
@@ -1,28 +1,31 @@
 # Abstract Syntax Tree Semantics
 
-The Abstract Syntax Tree (AST) has a basic division between statements and
-expressions. Expressions are typed; validation consists of simple, bottom-up,
-`O(1)` type checking.
+WebAssembly code is represented as an abstract syntax tree
+that has a basic division between statements and
+expressions. Each function body consists of exactly one statement.
```
jfbastien: Could you explain why there's a division between statement/expression, but functions only have a single statement? Explaining that there's a division, but just on statement doesn't contribute much explanation IMO.

kg: Do we even need a division between statements and expressions in the first place? It might be a nice simplification for our only polymorphic-ish node type to be the expression, and in cases where an expression yields no meaningful value we can just define it to have a void type (since functions can already have a void type anyway.)

Most of the things I can think of where the statement/expression difference would matter are JS/source language things that are erased by the time you encode down to wasm.

AstSemantics.md
```diff
@@ -142,11 +143,12 @@ which leads to the following advantages:
...
 
-Each heap access is annotated with a *Memory type* and
-the presumed alignment of the incoming pointer. As discussed previously, loads may
-include explicit zero- or sign-extension and stores may include implicit wrapping.
-
-Indexes into the heap are always byte indexes.
+Programs address the heap by using integers that are interpreted as unsigned byte indexes
+starting at `0`.
+Accesses to the heap at indices larger than the size of the heap are considered out-of-bounds,
+and a module may optionally define that out-of-bounds includes small indices close to `0`.
```
jfbastien: I think it's worth discussing how this works. I like the idea, it would be good to define it further. Could you add a link to #204 so it's clear that this is still TBD?

AstSemantics.md
lukewagner: I was also initially concerned with "not specified", but reading the context it seems we're still limiting the nondeterminism.

AstSemantics.md
titzer: Done.

jfbastien:
lgtm

AstSemantics.md
titzer: On Fri, Jun 19, 2015 at 11:28 AM, Katelyn Gadd <notifications@github.com>
wrote:

> In AstSemantics.md
> <https://github.com/WebAssembly/design/pull/200#discussion_r32814070>:
>
> > @@ -1,28 +1,31 @@
> >  # Abstract Syntax Tree Semantics
> >
> > -The Abstract Syntax Tree (AST) has a basic division between statements and
> > -expressions. Expressions are typed; validation consists of simple, bottom-up,
> > -`O(1)` type checking.
> > +WebAssembly code is represented as an abstract syntax tree
> > +that has a basic division between statements and
> > +expressions. Each function body consists of exactly one statement.
>
> Do we even need a division between statements and expressions in the first
> place? It might be a nice simplification for our only polymorphic-ish node
> type to be the expression, and in cases where an expression yields no
> meaningful value we can just define it to have a void type (since functions
> can already have a void type anyway.)
>

Good question. In my decoder it's tempting to just allow an expression be
the entire body of a function, and with the comma operator, one can
basically express a list of imperative statements followed by a
value-producing expression. But early returns, breaks, continues, etc. kind
of mess up the works. It'd be weird to allow them to be expressions, and
forcing producers to rewrite early returns to a block + break may be
inconvenient.

qwertie:
Yes, if possible I think all executable code (if not more of the system) should be expression-based. But what about control flow? I haven't thought much about it but

- Each function is just a single expression.
- If void (unit, ()) is understood as a proper data type, then `if` and `do_while` can be seen as expressions that return void, and if-else can be seen as equivalent to the conditional operator ?:
- `forever` can be seen as an expression whose arguments are the statements inside the loop, and `break` could include a 'return value', which will be the value of an enclosing `forever` loop.
- Under this idea, `return` is not needed; return values mid-function can be simulated by enclosing the function in a `forever` that uses `break` to specify the return value.

titzer:
What about continue? If's with one branch? What about discarding the value of a side-effecting expression or call?

While there is an attractive elegance to making everything an expression, the analogy gets strained in imperative-style code which is prevalent in the Algol branch of languages (C, C++, Java, JavaScript, C#, etc). We could bridge the gap with "void" or "unit" type that is the result of expressions that don't have a normal return value, which would otherwise be statements, I think this will result in a lot of confusion and an "impedance mismatch" for the very popular languages while giving on a very marginal benefit to much less popular functional languages. In short, it doesn't serve either master well.

titzer:
Since this PR is mostly wordsmithly and not changing semantics, I'll merge this later today if there are no objections.

qwertie:
@titzer: `continue` simply returns to an earlier location, so I don't see any problem with it. `if`s with one branch would return `void`. Side effects can occur inside an expression, and the comma operator provides sequences. Everyone is already familiar with this effect in C, e.g. (x = y * y, (x += y) + 5)

I would stress that merging statements and expressions into a single concept called 'expressions' seems to produce a strict superset of the current plan. I think it will be a simpler and more capable model than what is currently proposed, and it won't substantially alter code generation for the Algol branch of languages.

I am already accustomed to thinking this way, since I designed Enhanced C# as an expression-based language which is a strict superset of C#.

titzer:
On Wed, Jun 24, 2015 at 5:54 PM, qwertie <notifications@github.com> wrote:

> @titzer <https://github.com/titzer>: continue simply returns to an
> earlier location, so I don't see any problem with it. ifs with one branch
> would return void. Side effects can occur inside an expression, and the
> comma operator provides sequences. Everyone is already familiar with this
> effect in C, e.g. (x = y * y, (x += y) + 5)
>
> I would stress that merging statements and expressions into a single
> concept called 'expressions' seems to produce a strict superset of the
> current plan. I think it will be a simpler and more capable model than what
> is currently proposed, and it won't substantially alter code generation for
> the Algol branch of languages.
>
> I am already accustomed to thinking this way, since I designed Enhanced C#
> as an expression-based language which is a strict superset of C#.
>
I've filed https://github.com/WebAssembly/design/issues/223 for the
expression/statement discussion, since this PR is tailored toward simply
revising text.

PR #206, opened 2015-06-19
Brotli for BinaryEncoding compression

jfbastien:
I received a suggestion to support Brotli, and the rationale is pretty solid:

> The WOFF 2.0 team dropped both gzip and lzma to power their latest system with brotli. Because of that, brotli has already found its way to chrome and android. http://www.w3.org/TR/WOFF20ER/
>
> Brotli can compress 5-25 % more than gzip, is a little bit more efficient for small files than LZMA, and decompresses 3-5x faster than LZMA. (Brotli typically compresses less than LZMA for large files -- above 1 MB or so.)
>
> https://github.com/google/brotli
> 
> https://datatracker.ietf.org/doc/draft-alakuijala-brotli/
> 
> Once the specifics for WebAssembly are clear, we could even specialize the context map or the static dictionary for WebAssembly (or you can design WebAssembly it in a way that it takes maximum benefit of the existing brotli context modes -- there are three binary modes and an utf-8 mode).

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -30,7 +30,8 @@ Reducing download size, is achieved through three layers:
        [can improve general compression](https://www.rfk.id.au/blog/entry/cromulate-improve-compressibility/).
  * **Generic** compression, such as gzip, already supported in browsers. Other
    compression algorithms being considered and which might be standardized
-   include: LZMA, [Brotli](https://datatracker.ietf.org/doc/draft-alakuijala-brotli/).
+   include: LZMA, [LZHAM](https://github.com/richgel999/lzham_codec),
+    [Brotli](https://datatracker.ietf.org/doc/draft-alakuijala-brotli/).
 
 Each of the three layers work to find compression opportunities to the best of
 its abilities, without encroaching upon the subsequent layer's compression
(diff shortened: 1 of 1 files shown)


lukewagner:
Brotli might make sense if the WebAssembly were going to specify a form of generic compression, but so far we've decided against that.  Perhaps one day we'll get a builtin stream transformer (from the Streams API) so Brotli will be an attractive option for that reason.  Otherwise, I'd at least include lzma and [lzham](https://github.com/richgel999/lzham_codec) in the list of alternatives.

jfbastien:
@lukewagner agreed, I added LZHAM.

MikeHolman:
lgtm

jedie:
I also look for a good compressor. The starting point is: https://github.com/pypyjs/pypyjs.github.io/issues/4

It's not easy to find a compressor. Because there are different requirements.
In this case, i think the format must be have:
* fast decompression speed
* less memory usage while decompression

The main question is: What's fast enough?!?
This depend on download speed.

There are two good resources for compression algorithm:
* https://quixdb.github.io/squash-benchmark/
* http://mattmahoney.net/dc/text.html



Here a chart from the first link with:
* data: **Tarred executables of Mozilla 1.0 (Tru64 UNIX edition)**
* machine: **beagleboard-xm**

unbenannt

You can see, in this case **brotli** is nice. But **lzham** is faster and has a better ratio...

I don't know if **lzham** will be fast enough.

There are some more interesting compressors. You will find some interesting at the second link: http://mattmahoney.net/dc/text.html

PR #211, opened 2015-06-22
Point to Streams Standard in active development

sideshowbarker:
This duplicates #195 from @annevk (which couldn’t be merged since the fork it came from was afterward deleted).


lukewagner:
Thanks!

Issue #228, opened 2015-06-25
Naming consistency for conversions: extensions, truncations, rounding and wrapping


titzer:
We are somewhat inconsistent in the naming of the AST operations at the moment. In particular, we use a lot of different terms but haven't defined exactly what we mean by them. Also, the naming w.r.t. signed/unsigned integer operations is slightly inconsistent.

I'll use the generic term "conversion" to refer to all operations that take a single value of one type and produce a single value of the same or different type where the result is an approximation or different representation of the same value. Conversion operations C generally have these properties:

1. Pure function. Neither reads nor writes mutable state, always returns the same result for the same input.
2. Unary. One input, one output.
3. Mostly-invertable. For a large number of inputs v, there exists an inverse conversion R where R(C(v)) == v.

Our operations could be broken down as follows. The ones that are fully invertable (i.e. an inverse function exists for all input values v) are noted, while the others are mostly invertable.

1. Integer extension. For example, converting a smaller integer to a larger integer by sign extending or zero extending. Fully invertable.
2. Integer wrapping. For example, converting a larger integer to a smaller integer by discarding the upper bits. Only invertable when the upper bits are the sign-extension or zero-extension of the lower bits appropriately.
3. Floating point rounding. For example, one of the IEEE 754 rounding modes. Inverse only exists when the output is equal to the input.
4. Floating point nearest integer. A special case of floating point rounding to integers.
5. Integer to floating point extension. For example, converting an integer to a floating point number where there can be no loss of precision, such as int32 to float64. Fully invertable.
6. Integer to floating point conversion. For example, converting an integer to a floating point number where there could be loss of precision, such as int32 to float32. Only invertable when the discarded lower bits are all 0.
7. Integer/floating point reinterpretation. For example, reinterpreting the bits of an integer as a floating point number or vice versa. Fully invertable if NaN bit patterns are not canonicalized.
8. Floating point to integer truncation with saturation. For example, converting a floating point to a number where values too large or too small to represent are clamped to -MIN_INT or MAX_INT, and NaN is interpreted as some integer value, e.g. 0. Only invertable if there are no fractional bits, there is no overflow, and the input is not a NaN.
9. Any of the variants of the above that trap when an inverse function does not exist. Invertable, by definition, if the input does not cause a trap.

titzer:
I'll take a first stab at naming some of the operations we have so far, somewhat along the lines we've been following, using a convention where the local type of the result comes first:

result_type.operation[input_type]

[...]

PR #229, opened 2015-06-25
Correct behavior of int to float conversions.

sunfishcode:
IEEE 754 specifies that these operations round rather than just overflow.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -458,8 +458,8 @@ the opcode, and the operand.
 
 Reinterpretations always succeed.
 
-Conversions from integer to floating point always succeed, though they may
-overflow to infinity or negative infinity as specified by IEEE-754.
+Conversions from integer to floating point always succeed, and use
+round-to-nearest ties-to-even rounding.
 
 Truncation from floating point to integer where IEEE-754 would specify an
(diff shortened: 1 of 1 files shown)


AstSemantics.md
```diff
@@ -458,8 +458,8 @@ the opcode, and the operand.
 
 Reinterpretations always succeed.
 
-Conversions from integer to floating point always succeed, though they may
```
cheehaur: JJ

AstSemantics.md
```diff
@@ -458,8 +458,8 @@ the opcode, and the operand.
 
 Reinterpretations always succeed.
```
chealili: www.chealili@gmail.com

chealili: lili model

chealili: @@ -458,8 +458,8 @@ the opcode, and the operand.

chealili:
01087232e

MikeHolman:
lgtm

titzer:
lgtm, though I think we should continue discussion on the conversion issue (https://github.com/WebAssembly/design/issues/228) to nail down what "succeed" and "fail" mean.

PR #237, opened 2015-06-29
add opcode definitions section

MikeHolman:


BinaryEncoding.md
```diff
@@ -67,18 +67,30 @@ Yes:
...
+    offset (within the module), followed by
   * a sequence of sections.
 * A section contains:
   * a header followed by
   * the section contents (specific to the section type)
+* A definitions section contains:
+  * the generic section header
+  * a table containing, for each opcode-space, a standardized string literal
+    type name (where index defines its type), offset (within the section),
+    sorted by offset, followed by
```
jfbastien: Could you make these sub-bullets?

What do you mean by "index defines its type"?

BinaryEncoding.md
```diff
@@ -67,18 +67,30 @@ Yes:
...
+  * a sequence of opcode tables
+  * An opcode table contains:
+    * a sequence of standardized string literal opcode names, where order
+      determines opcode index
```
jfbastien: I know what the standardized type names and opcode names are, but it isn't quite clear from this. Could you add a definition? Link to ASTSemantics?

BinaryEncoding.md
MikeHolman: I mean that when we have places where we need to reference a type (e.g. function definitions), we don't want to put "int32", would rather put 0.

BinaryEncoding.md
```diff
@@ -67,18 +67,30 @@ Yes:
...
-  * a table containing, for each function, it's signature, offset (within the section), sorted by offset, followed by
+  * a table containing, for each function, its signature, offset (within the
+    section), sorted by offset, followed by
   * a sequence of functions
-* A function contains:
-  * a table containing, for each type, how many locals are indexed by the function body of that type
-  * the serialized AST
+  * A function contains:
+    * a table containing, for each type, how many locals are indexed by the
+      function body of that type
```
jfbastien: What if a function doesn't use some types? Do we need a bunch of zero-sized entries for all SIMD types?

jfbastien:
What does it mean to have multiple sections with functions in each? Can one section's function call a function in another section?

Or do we just have one code section for now?

How do I access different data sections?

It looks like we're close to a container format... This is related to #74 about using ELF.


The definition is also beginning to look like BNF!

MikeHolman:
>What does it mean to have multiple sections with functions in each? Can one section's function call a function in another section? Or do we just have one code section for now?

I don't see a logical reason to have more than one code section right now, so I think limiting to one should be ok.

>How do I access different data sections?

I guess there are a couple ways we could go about it. If different data section types are all singletons (e.g. you can only have a single import sections), then you can directly access it and decoder will know where to look when you ask for import[0]. Otherwise you can indirect through the section list for the section you want and access your data from there (not quite as efficient though).

>It looks like we're close to a container format... This is related to #74 about using ELF.

I don't really know ELF, but the conversation made it sound like the format will not map well for us. So I'd like to move forward with something else hedging against it.

BinaryEncoding.md
MikeHolman: That's what is implied, but I think we can change it so encoding is something like:

[{0,29},{1,3}] //where int32==0, float64==1

sunfishcode:
ELF isn't yet ruled out. These various tables could just map to special sections/segments in ELF. But I don't think we need to worry about that now. Let's design what we want first, and figure out whether ELF makes sense once we have that.

BinaryEncoding.md
```diff
@@ -65,20 +65,36 @@ Yes:
 
 ## Global structure
 
-* A module contains:
-  * a header followed by
-  * a table (sorted by offset) containing, for each section, its type and offset (within the module), followed by
-  * a sequence of sections.
+* A module contains (in this order):
+  * A header
```
jfbastien: Define header.

BinaryEncoding.md
```diff
@@ -65,20 +65,36 @@ Yes:
 
 ## Global structure
 
-* A module contains:
-  * a header followed by
-  * a table (sorted by offset) containing, for each section, its type and offset (within the module), followed by
-  * a sequence of sections.
+* A module contains (in this order):
```
jfbastien: Can you use the different styles of bullet lists below, so it's easier to read the text. E.g.:
* A
  - Deeply
    + Nested
  - List
  - Is
* Easier
* To
  - Read
    + With
    + Different
    + Bullets

BinaryEncoding.md
```diff
@@ -65,20 +65,36 @@ Yes:
...
+  * A header
+  * A table (sorted by offset) containing, for each section:
+    * Section type
```
jfbastien: Define type.

BinaryEncoding.md
```diff
@@ -65,20 +65,36 @@ Yes:
...
+    * Offset within the module
```
jfbastien: 32-bit

BinaryEncoding.md
```diff
@@ -65,20 +65,36 @@ Yes:
...
-  * a header followed by
-  * the section contents (specific to the section type)
-* A code section contains:
-  * the generic section header
-  * a table containing, for each function, it's signature, offset (within the section), sorted by offset, followed by
-  * a sequence of functions
-* A function contains:
-  * a table containing, for each type, how many locals are indexed by the function body of that type
-  * the serialized AST
+  * A header followed by
```
jfbastien: Null-terminated UTF8?

jfbastien:
Maybe I'm going into too many details?

BinaryEncoding.md
MikeHolman: You caught me. I don't know what the headers contain. At the very least, the module header contains the magic number, but besides that I don't really have anything in particular decided. Maybe some things like whether heap is 32/64 bit, source language (for ABI), and entry point. This will need to be figured out, but based on the level of detail in the rest of our design docs I'm not sure how much detail to go in here.

Maybe I should just mention some things like this as "ideas" for what a header would contain?

jfbastien: Yeah that sounds good. I like what you're adding overall, so I'll step back, take this improvement, and we can iterate later :-)

MikeHolman:
@jfbastien Maybe a bit with null terminated UTF8 (which is what I had in mind, but thought I was already bordering on too verbose), but you are right that "header" and "type" deserved some clarifications.

jfbastien:
lgtm

BinaryEncoding.md
```diff
@@ -65,20 +65,40 @@ Yes:
...
 ## Global structure
 
-* A module contains:
-  * a header followed by
-  * a table (sorted by offset) containing, for each section, its type and offset (within the module), followed by
-  * a sequence of sections.
+* A module contains (in this order):
+  - A header, containing:
+    + The [magic number](https://en.wikipedia.org/wiki/Magic_number_%28programming%29)
+    + Other data TBD (possibly entrypoint, memory bitness, source language, etc.)
```
sunfishcode: What is memory bitness?

At an initial glance, source language seems like something we'd specifically try to avoid including in the main header, because it suggests special magical per-source-language semantics.

BinaryEncoding.md
```diff
@@ -65,20 +65,40 @@ Yes:
...
+  - A table (sorted by offset) containing, for each section:
+    + A string literal section type name
+    + 32-bit offset within the module
```
sunfishcode: Is 32 bits enough for everyone?

waywardmonkeys: That's an area where using LEB128 encoding might be helpful?

jfbastien: Yeah we can go with 64 for now. There won't be that many of them anyways.

MikeHolman: 64 bit is fine with me. that's why I didn't specify originally, I thought this might be decided based on experimental data. do we expect modules that are larger than 4GB? That feels like a pretty insane amount of code.

MikeHolman: Actually, I guess I can see that there might be some meta-data sections that could blow up size pretty quickly in certain scenarios.

BinaryEncoding.md
MikeHolman: >What is memory bitness?

I mean whether your linear memory has a 64 bit or 32 bit address space (i.e. whether ptr type is int32 or int64). Maybe not necessary here, but was just an idea for something we might want. I think we might need some version info for the module format as well. We may never need to break compat, but I can imagine a scenario where we eventually want to make format changes, and having a byte to allow for that would be useful.

MikeHolman: I'll just say other data TBD for now and remove the rest

BinaryEncoding.md
jfbastien: There are 2 other instances of 32 below.

BinaryEncoding.md
titzer: In the AstSemantics.md, we've just been assuming that one can index the
linear memory with either 32-bit or 64-bit offsets. In the v8 native
prototype, there are different bytecode numbers for whether the memory
offset operand is an Int32 or an Int64.

BinaryEncoding.md
MikeHolman: Yeah, I thought that those could still be 32-bit. I can see how you might have a bigger metadata section, but if you exhaust 32 bit address space in opcodes alone, then we have done something seriously wrong. And to a lesser extent, I have never seen a dll that was more than like 40MB, and if we are really a size efficient format our encoding shouldn't be orders of magnitude worse. A 4GB code section seems completely implausible.

titzer: At Google we've exceeded the 2gb JAR size limitation in the JVM and have
had to patch the JVM. So, it could happen. And, I'm crying saying that.

MikeHolman: @titzer That sounds extremely painful. I'd rather not put someone else through the same, so ok sure, I'll make code sections 64 bit as well.

jfbastien: @titzer is asking for more [@horse_wasm attention](https://twitter.com/horse_wasm/status/615985298955091968) with this "at Google..." :-)

Whoever runs the account: good job!

MikeHolman: @jfbastien Oh no, I've been quoted!

jfbastien:
I think this is good to go, we can iterate on top.

PR #240, opened 2015-07-01
Minor edits to binary encoding of global structure

jfbastien:


BinaryEncoding.md
```diff
@@ -81,7 +81,7 @@ Yes:
   - A table (sorted by offset) containing, for each type which has opcodes:
     + A standardized string literal [type name](AstSemantics.md#local-and-memory-types).
       The index of a type name in this table is referred to as a type ID
-    + 32-bit offset of its opcode table within the section
+    + 64-bit offset of its opcode table within the section
```
MikeHolman: I'm absolutely certain we don't need 4gb here. I'm not going to argue over what will be a small handful of bytes, but I just don't see the necessity

MikeHolman:
lgtm, and strongly agree with adding meta-data for function hotness

BinaryEncoding.md
jfbastien: Agreed, I just figure we'd go with consistency, and cut the fat if & where needed later :-)

waywardmonkeys: That's one reason that a LEB128 encoding for these sorts of values is convenient.

jfbastien: @waywardmonkeys it's still not clear that LEB128 or any other variable-length integer encoding pays off after applying generic encoding such as gzip, lzham or brotli. We're optimizing a multi-variable encoding approach, and I'm pretty wary of early optimization for now!

waywardmonkeys: I like it for some things because then I don't have to specify 32 or 64 bit sizes for something that might change in the future. (Like storage of data from a profiler where values may reasonably be either 32 or 64 bit depending on the target or the nature of the data being sampled.) The compression is a secondary concern in that sort of case.

kg: Fixed-size fields in structures are arguably a virtue compared to everything being variable-length, though. Decoding is simpler and you can seek to known offsets. In a scenario like this saving a couple bytes by using a variable-length integer instead of just a 64-bit one would be... hm.

MikeHolman: I'm sorry I even mentioned it. We're probably talking single digit number of bits for compressed binary. This is fine.

Issue #241, opened 2015-07-02
Adding greater-than operations


MikeHolman:
Our reasoning for not having greater-than operations is stated like this in ASTSemantics:

>Note that greater-than and greater-than-or-equal operations are not required,
since `a < b` is equivalent to `b > a` and `a <= b` is equivalent to `b >=
a`.

If `a` could have side effects which might impact `b`, then you must do something like `!(a <= b)`. I don't think we need to be so stingy as to require this.

creationix:
Good point.  In traditional bytecode where the operands are either simple literals or register lookups, there is no reason to have both sets unless you're dealing with floating point numbers.  From the luajit docs:

> Q: Why do we need four different ordered comparisons? Wouldn't < and <= suffice with appropriately swapped operands?

> A: No, because for floating-point comparisons (x < y) is not the same as not (x >= y) in the presence of NaNs.

Since we have both floating point number types and are operands can be deep AST trees containing nodes with side effects, the order of the operands matters a lot.

jfbastien:
NaN ordering isn't an issue for wasm at the moment, since NaN bits aren't specified.
The problem with side-effects isn't whether things are evaluated or not, it's rather about evaluation order. I think this is sufficient reason to add this feature.

MikeHolman:
@jfbastien Alright sounds good, I'll create a PR to add these.

titzer:
On Thu, Jul 2, 2015 at 7:11 PM, Tim Caswell <notifications@github.com>
wrote:

> Good point. In traditional bytecode where the operands are either simple
> literals or register lookups, there is no reason to have both sets unless
> you're dealing with floating point numbers. From the luajit docs:
>
> Q: Why do we need four different ordered comparisons? Wouldn't < and <=
> suffice with appropriately swapped operands?
>
> A: No, because for floating-point comparisons (x < y) is not the same as
> not (x >= y) in the presence of NaNs.
>
>
(x < y) == (y < x), even in the presence of NaNs.

sunfishcode:
@titzer That's true, however (f() > g()) is not equivalent to (g() < f()) because the operands are evaluated in a different order. We can always change (f() > g()) to (tmp = f(), g() < tmp), but that's less nice.

lskyum:
I suppose (f() > g()) would be the same as !(f() <= g()) but that would increase the size of the AST. Although not verry much...

titzer:
[...]

PR #242, opened 2015-07-02
add greater than ops

MikeHolman:


MikeHolman:
Addresses #241

jfbastien:
lgtm

sunfishcode:
While (x > y) is equivalent to (y < x), (f() > g()) is not equivalent to (g() < f()). So lgtm.

Issue #148, opened 2015-06-11
What about subnormals?


jfbastien:
Forking from: https://github.com/WebAssembly/design/pull/141#discussion_r32198186

Should denormal be:
* Unspecified (say so in Nondeterminism.md).
* Fully specified as IEEE-754 compliance.
* Fully specified as IEEE-754 compliance for scalars, and do something for SIMD because existing hardware (especially ARMv7 NEON) doesn't support denormals.
* Specified as DAZ/FTZ (not IEEE-754 compliant).

We should probably let ourselves change this from developer feedback, but I'd like to make some decision for MVP. I [suggested on esdiscuss](https://esdiscuss.org/topic/float-denormal-issue-in-javascript-processor-node-in-web-audio-api#content-3) that JavaScript go full unspecified, and just do DAZ/FTZ because it's often faster. Yes, x86 is better than it used to be but that's not universal, ignores current hardware, and doesn't look towards what new hardware will do. I like leaving the door open :-)

For @sunfishcode searchability, I'll use the word "subnormals" too :-)

angryoctopus:
I think the vast majority of developers that encounter denormals discover them while troubleshooting unexpected performance hits. In the DSP world it tends to be a very common performance gotcha so I would advocate defaulting FTZ with an optional global IEEE-compliant mode.

mikolalysenko:
My preference would be full IEEE 754 support, or barring that to use DAZ. Undefined behavior would be a very terrible decision in my opinion.  Inconsistent semantics make it very difficult to implement algorithms from computational geometry, which require exactly computing things like the sign of a determinant for example.  A common technique to speed up these calculations is to use a floating point filter as a quick check before falling back to a more expensive exact arithmetic test.  If the floating point behavior is not specified, then it becomes much more difficult (and in some cases impossible) to construct such a filter.

sunfishcode:
Here is a discussable proposal which I believe gives most people what they want, though it makes some tradeoffs (as any proposal must):

 - Default to full support for subnormals.
 - A function attribute sets either "standard" or "maybe_flush", with "maybe_flush" meaning that it's nondeterministic whether subnormals are flushed as input and/or output at each operation within the function body (and does not apply within called functions).
 - Float32x4 is not accelerated on 32-bit NEON in standard mode, but is in "maybe_flush" mode.

The following questions seem interesting:

[...]

PR #243, opened 2015-07-06
subnormal_mode

sunfishcode:
This PR proposes a subnormal_mode AST node, as a possible resolution to #148.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -500,8 +500,7 @@ Promotion and demotion of floating point values always succeed.
 Demotion of floating point values uses round-to-nearest ties-to-even rounding,
 and may overflow to infinity or negative infinity as specified by IEEE-754.
 If the operand of promotion or demotion is NaN, the sign bit and significand
-of the result are computed from an unspecified function of the implementation,
-the opcode, and the operand.
+of the result are not specified.
 
 Reinterpretations always succeed.
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -270,6 +270,21 @@ nested. This guarantees that all resulting control flow graphs are well-structur
     feature would allow efficient compilation of arbitrary irreducible control
     flow.
 
+## Floating Point Subnormal Handling
+
+  * `subnormal_mode`: Like [`block`](AstSemantics.md#control-flow-structures),
+     but with a mode attribute specifying special semantics for the handling of
+     subnormal values in all lexically contained [floating point operations][]
+     except `neg`, `abs`, and `copysign`, and [conversions][], except those in
+     contained `subnormal_mode` blocks. The mode may be one of:
+    * `standard`: standard semantics are followed
+    * `dont_care`: Subnormal values may or may not be interpreted as zero of the
+       same sign. Subnormal result values may or may not be replaced by any
+       subnormal value of the same sign, or zero of the same sign.
+
+  [floating point operations]: AstSemantics.md#floating-point-operations
+  [conversions]: AstSemantics.md#datatype-conversions-truncations-reinterpretations-promotions-and-demotions
+
 ## Calls
 
 Direct calls to a function specify the callee by index into a function table.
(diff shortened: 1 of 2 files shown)


AstSemantics.md
```diff
@@ -146,6 +146,21 @@ nested. This guarantees that all resulting control flow graphs are well-structur
...
+
+  * `subnormal_mode`: Like [`block`](AstSemantics.md#control-flow-structures),
+     but with a mode attribute specifying special semantics for the handling of
+     subnormal values in all lexically contained [floating point operations][]
+     except `neg`, `abs`, and `copysign`, and [conversions][], except those in
+     contained `subnormal_mode` blocks. The mode may be one of:
+    * `standard`: standard semantics are followed
+    * `dont_care`: Subnormal values may or may not be interpreted as zero of the
+       same sign. Subnormal result values may or may not be replaced by any
+       subnormal value of the same sign, or zero of the same sign.
```
jfbastien: What's "standard"? Wouldn't "IEEE754" be a better name?
I thought that we'd settled on "fastest" instead of "dont_care"?

jfbastien:
I'm still not sure that I like using a block as you propose, versus tagging each individual instruction.

We haven't really discussed what happens when there's a call inside of such a block. Do it affect the callee?

AstSemantics.md
sunfishcode: "Standard" means the IEEE-754 standard, and also the WebAssembly standard, since it's the behavior we'll otherwise specify. Is there another standard which is remotely as authoritative in this regard?

In my opinion, "fastest" is implied. We always want the fastest performance permitted under the chosen semantics. "dont_care" makes it explicit that the programmer is choosing not to care about what happens, rather than blindly hoping that something named "fastest" will make their program faster without understanding what it means.

sunfishcode:
The advantage of a scoped construct is that it eliminates the need for an implementation to run a code placement algorithm to emit reasonably-well optimized mode switching code.

By the phrase "lexical extent", I mean that it doesn't apply to callees. One of the main goals here is to make it possible (and easy) to statically determine the mode for every operation.

kg:
The checked/unchecked mechanism in C# is a relevant example here:

https://msdn.microsoft.com/en-us/library/74b4xzyw.aspx

they recognized that there are multiple levels at which you want to control checking, so they let you decorate a single expression ```checked(x)```, a block ```checked { a; b; }``` or an entire module (by changing compiler options). It's not inherited by callees, so it's pretty easy to compile and audit.

sunfishcode:
Thanks for the reference @kg. That covers integer overflow checking which has some similar usage patterns as subnormal handling, though a few differences as well.

Block-level: that's what I'm proposing :-).

Module-level: We can achieve similar functionality by just wrapping every function body in a subnormal_mode block. This is also easier for WebAssembly to do than it is for C# because WebAssembly is primarily a compiler target. The user interface can still be to just set a compiler option, and compilers can just make the right thing happen everywhere. Also, module-level flags complicate dynamic linking, so we'd like to avoid them when we can.

Per-operation: Unlike integer-overflow checking, subnormal flushing is typically implemented in hardware by setting a control register, and there is a cost each time the mode is changed. Since the point of the mode is to make code go faster, one typically doesn't wish to take on extra costs by changing modes frequently. Consequently, it's much more common and meaningful to just set it once for an entire block of code.

titzer:
I support block-level declaration, though IIUC a conforming implementation
can default to IEEE 754 everywhere, correct?

sunfishcode:
Yes. The attribute choices proposed here are "standard" and "dont_care", so using standard mode for everything is fully conforming.

jfbastien:
Please hold on this PR, @davidsehr and I are discussing a different / wider solution.

sunfishcode:
After discussing this with people more, I'm ok closing this for now. While a flag to disable subnormals is worth considering, the real-world experience of asm.js tells us that it ultimately isn't necessary for an MVP.

PR #244, opened 2015-07-06
Add a FutureFeatures section about integer overflow.

sunfishcode:
This PR just adds a FutureFeatures.md section about integer overflow detection ideas that have been present in several offline discussions.

### DIFF
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -336,3 +336,26 @@ appeared. Another option would be to offer another non-standard execution mode,
 enabled only from developer tools, that would enable traps on selected floating
 point exceptions, however care should be taken, since not all floating point
 exceptions indicate bugs.
+
+## Integer Overflow Detection
+
+There are two different use cases here, one where the application is prepared
+to handle overflow, and one where it isn't.
+
+When the application is prepared to handle overflow, it would be useful to have
+arithmetic operations which can indicate when overflow occured. An example of
+this is the checked arithmetic builtins available in some compilers:
+
+ * http://clang.llvm.org/docs/LanguageExtensions.html#checked-arithmetic-builtins
+ * https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html
+
+If WebAssembly is made to support nodes with multiple return values, that could
+be used instead of passing a pointer.
+
+When the application is not prepared to handle overflow, it would be useful to
+have arithmetic operations which trap on overflow. Following the rule that
+explicitly signed and unsigned operations trap whenever the result value can not
+be represented in the result type, it would be possible to add explicitly signed
+and unsigned versions of integer `add`, `sub`, and `mul`, which would trap on
+overflow. The main reason we haven't added these already is that they're not
+efficient for general-purpose use on today's popular hardware.
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -344,10 +344,10 @@ to handle overflow, and one where it isn't.
 
 When the application is prepared to handle overflow, it would be useful to have
 arithmetic operations which can indicate when overflow occured. An example of
-this is the checked arithmetic builtins available in some compilers:
-
- * http://clang.llvm.org/docs/LanguageExtensions.html#checked-arithmetic-builtins
- * https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html
+this is the checked arithmetic builtins available in compilers such as
+[clang](http://clang.llvm.org/docs/LanguageExtensions.html#checked-arithmetic-builtins)
+and
+[GCC](https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html).
 
 If WebAssembly is made to support nodes with multiple return values, that could
 be used instead of passing a pointer.
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -339,23 +339,30 @@ exceptions indicate bugs.
 
 ## Integer Overflow Detection
 
-There are two different use cases here, one where the application is prepared
-to handle overflow, and one where it isn't.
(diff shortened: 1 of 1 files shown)


FutureFeatures.md
```diff
@@ -336,3 +336,26 @@ appeared. Another option would be to offer another non-standard execution mode,
...
+
+There are two different use cases here, one where the application is prepared
+to handle overflow, and one where it isn't.
+
+When the application is prepared to handle overflow, it would be useful to have
+arithmetic operations which can indicate when overflow occured. An example of
+this is the checked arithmetic builtins available in some compilers:
+
+ * http://clang.llvm.org/docs/LanguageExtensions.html#checked-arithmetic-builtins
+ * https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html
```
jfbastien: Link to [clang](http://clang.llvm.org/docs/LanguageExtensions.html#checked-arithmetic-builtins) and [GCC](https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html) instead of long URLs.

FutureFeatures.md
```diff
@@ -336,3 +336,26 @@ appeared. Another option would be to offer another non-standard execution mode,
...
+If WebAssembly is made to support nodes with multiple return values, that could
+be used instead of passing a pointer.
+
+When the application is not prepared to handle overflow, it would be useful to
+have arithmetic operations which trap on overflow. Following the rule that
+explicitly signed and unsigned operations trap whenever the result value can not
+be represented in the result type, it would be possible to add explicitly signed
+and unsigned versions of integer `add`, `sub`, and `mul`, which would trap on
+overflow. The main reason we haven't added these already is that they're not
+efficient for general-purpose use on today's popular hardware.
```
jfbastien: I think this should be a toolchain-side sanitizer instead, which ends up using the intrinsics you discuss in the above paragraph.

sunfishcode: For C/C++ code today on today's architectures, this is true.

However, consider a language which wants to implement transparent bignums, or consider a JavaScript JIT running inside WebAssembly that wants to specialize doubles to integers. In both cases, it's desirable to avoid making a branch after every add. x86 and ARM presently give us no choice. But one could hope for better hardware which has a trapping add.

And even in the case of C/C++ code, if one had hardware support for trapping add, and the option of enabling it by default (in production, if one wishes), that would have advantages over an on-the-side sanitizer. (Of course, you'd still have to do something about those crazy compilers, but that's another matter.)

jfbastien: Optimizing for non-existent ISAs seems a bit premature. I agree the feature is useful, but it's exactly the same as using the builtins, and really easy for compilers to figure out were such an ISA in existence.

Added bonus: the operations will compress well because of parameterizable macros compression.

sunfishcode: This is FutureFeatures.md after all ;-). But also, it's not exactly the same thing as the builtins. The builtins require branching, and a branch target inside the same function, and no plausible possibility of doing something like the magical things we're hoping to do with traps some day.

And added bonus: this is a beautiful application of the rule that explicitly signed/unsigned operations trap whenever they can't represent their result.

FutureFeatures.md
sunfishcode: Ok.

FutureFeatures.md
kripken: I like the consistency of explicitly signed/unsigned always trapping. Not having that would mean one more thing to keep in mind when thinking about operations and trapping.

I also like the idea of security-sensitive code being able to just use explicitly signed/unsigned operations, and register a trap handler once for all of that, instead of each function needing a trampoline to a handler.

The part about VMs and CPUs possibly finding better ways to optimize this over time is an extra bonus, from my perspective.

jfbastien: This being "future features" doesn't mean we should mention things that aren't [extensible web manifesto](https://extensiblewebmanifesto.org) friendly. In general we should standardize the bare minimum features to allow tooling to do what it needs to. Adding optimizations up-front for theoretical cases is something we've avoided, and I would like us to avoid here too.

I think we agree that the basic checked arithmetic builtins allow tooling to do what the second paragraph suggests?

sunfishcode: MIPS and other actual relevant architectures have the ability to trap on integer overflow, so this isn't just theoretical.

And I believe we still disagree. Trapping arithmetic and arithmetic that returns a bool for overflow have similar functionality but disjoint usecases, as discussed above.

jfbastien: > And I believe we still disagree. Trapping arithmetic and arithmetic that returns a bool for overflow have similar functionality but disjoint usecases, as discussed above.

That's disagreeing on what one would do with the feature, not whether the first implementation can be used to implement the second. I believe we agree on this? If so, it means that we can implement the first as a standalone thing which enables the second. We can always add the second later.

I'm not opposed to mentioning the second, but I want the text to make the distinction in implementability. The current text puts them on the same footing, which is what I disagree on.

Incremental wins.

sunfishcode: I also disagree that the builtin operations could reasonably implement the functionality of the second paragraph. They're limited to handling overflow locally. The use cases @kripken and I mentioned above of really want something along [these lines](https://github.com/WebAssembly/design/blob/master/FutureFeatures.md#trapping-or-non-trapping-strategies).

Incremental is indeed great. But this is FutureFeatures.md.

jfbastien: Can you capture this discussion on the text, then? I'd at least want future features to reflect the open design space.

sunfishcode: Ok.

jfbastien:
lgtm

sunfishcode:
Merging based on lgtm above and no other objections. Of course, future objections are still welcome.

PR #245, opened 2015-07-06
Types and linear memory

sunfishcode:
The current AstSemantics sections on types and linear memory are a little disorganized at the moment. Linear memory access operations are described in the types section, and the linear memory section itself has an obsolete set of load and store opcodes.

This PR reorganizes and generally revises these sections, creating a simple section for just types, and putting all the linear memory content in the linear memory section. It doesn't make any semantic changes, though it does make a few terminology changes.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -126,9 +126,9 @@ their value operand, with no conversion applied.
 ### Addressing
 
 Each linear memory access operation also has an address operand and an immediate
-integer offset attribute. The infinite-precision sum of the address operand's
-value with the offset attribute's value is called the *effective address*, which
-is interpreted as an unsigned byte index.
+integer byte offset attribute. The infinite-precision sum of the address
+operand's value with the byte offset attribute's value is called the
+*effective address*, which is interpreted as an unsigned byte index.
 
 Linear memory accesses access the bytes starting at the location in the linear
 memory storage indexed by the effective address, and extending for the number
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -150,8 +150,8 @@ address modes in hardware, and to simplify bounds checking optimizations.
 In the MVP, address operands and offset attributes have type `int32`, and linear
 memory sizes are limited to 4 GiB (of course, actual sizes are further limited
 by [available resources](Nondeterminism.md)). In the future, to support
-[>4GiB linear memory](FutureFeatures.md#heaps-bigger-than-4gib), there will be
-a mode in which they have type `int64`.
+[>4GiB linear memory](FutureFeatures.md#heaps-bigger-than-4gib), support for
+indices with type `int64` will be added.
 
 ### Alignment
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -120,6 +120,9 @@ basic type to their memory type, and write the resulting value to linear memory:
 Wrapping of integers simply discards any upper bits; i.e. wrapping does not
 perform saturation, trap on overflow, etc.
 
+In addition to storing a value to linear memory, store instructions also return
+their value operand, with no conversion applied.
+
 ### Addressing
 
 Each linear memory access operation also has an address operand and an immediate
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -147,8 +147,9 @@ address for an access is out-of-bounds, the effective address will always also
 be out-of-bounds. This is intended to simplify folding of offsets into complex
 address modes in hardware, and to simplify bounds checking optimizations.
 
-In the MVP, address operands and offset attributes have type `int32`, and heap
-sizes are limited to 4 GiB. In the future, to support
+In the MVP, address operands and offset attributes have type `int32`, and linear
+memory sizes are limited to 4 GiB (of course, actual sizes are further limited
+by [available resources](Nondeterminism.md)). In the future, to support
 [>4GiB linear memory](FutureFeatures.md#heaps-bigger-than-4gib), there will be
 a mode in which they have type `int64`.
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -127,18 +127,19 @@ their value operand, with no conversion applied.
 
 Each linear memory access operation also has an address operand and an immediate
 integer offset attribute. The infinite-precision sum of the address operand's
-value with the offset attribute's value is called the *effective address*.
+value with the offset attribute's value is called the *effective address*, which
+is interpreted as an unsigned byte index.
 
-Effective addresses are interpreted as unsigned byte indices into the linear
-memory storage, starting at `0`. Linear memory accesses access memory starting
-at this byte and extending for the number of bytes implied by the memory
-attribute.
+Linear memory accesses access the bytes starting at the location in the linear
+memory storage indexed by the effective address, and extending for the number
+of bytes implied by the memory type attribute of the access.
 
-Accesses to any bytes of linear memory beyond the `memory_size` are considered
+If any of the accessed bytes are beyond `memory_size`, the access is considered
 *out-of-bounds*. A module may optionally define that out-of-bounds includes
 small effective addresses close to `0`
-(see [discussion] (https://github.com/WebAssembly/design/issues/204)).
-The semantics of out-of-bounds accesses are discussed below.
+(see [discussion](https://github.com/WebAssembly/design/issues/204)).
+The semantics of out-of-bounds accesses are discussed
+[below](AstSemantics.md#out-of-bounds).
 
 The use of infinite-precision in the effective address computation means that
 the addition of the offset to the address does is never wrapped, so if the
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -32,35 +32,62 @@ Callstack space is limited by unspecified and dynamically varying constraints.
 If program callstack usage exceeds the available callstack space at any time,
 a trap occurs.
 
-## Local and Memory types
+## Types
 
-Individual storage locations in WebAssembly are typed, including global
-variables, local variables, and parameters. The main storage of a wasm module, 
-called the *Linear memory*, is a contiguous, byte-addressable range of memory
-spanning from offset `0` to `memory_size`. The linear memory can be considered 
-to be a untyped array of bytes which is stored separately from the global variables.
-The linear memory is sandboxed; it does not alias the execution engine's internal
-data structures, the execution stack, or other process memory.
-All accesses to memory are annotated with a type. The legal types for
-global variables and memory accesses are called *Memory types*.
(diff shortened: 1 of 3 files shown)


AstSemantics.md
```diff
@@ -32,35 +32,62 @@ Callstack space is limited by unspecified and dynamically varying constraints.
...
+byte-addressable range of memory spanning from offset `0` and extending for
+`memory_size` bytes. The linear memory can be considered to be a untyped array
+of bytes. The linear memory is sandboxed; it does not alias the execution
+engine's internal data structures, the execution stack, local variables, global
+variables, or other process memory.
+
+In the MVP, linear memory is not shared between threads. When
+[threads](PostMVP.md#threads) are added as a feature, regular load and store
+nodes will have the most relaxed semantics specified in the memory model and new
+memory-access nodes will be added with atomic and ordering guarantees.
```
jfbastien: Regular loads and stores won't have relaxed semantics: they're instead bound by happens-before with atomics in the same thread of execution, which themselves synchronize-with atomics in other threads.

AstSemantics.md
```diff
@@ -96,107 +120,59 @@ following rules:
...
-others, etc.
+Each linear memory access operation also has an address operand and an immediate
+integer offset attribute. The infinite-precision sum of the address operand's
+value with the offset attribute's value is called the *effective address*.
 
-## Control flow structures
+Effective addresses are interpreted as unsigned byte indices into the linear
+memory storage, starting at `0`. Linear memory accesses access memory starting
+at this byte and extending for the number of bytes implied by the memory
+attribute.
```
jfbastien: I find this slightly confusing: is it `base_address_offset_from_0_in_byte + offset_in_bytes` or is the offset instead then number of elements to index from (so, multiply by `sizerof(type)` for number of bytes)?

AstSemantics.md
sunfishcode: This is text that was already present, and is just being moved in this patch. I can attempt to clarify this text though and avoid the ambiguous word "relaxed" in this context.

AstSemantics.md
```diff
@@ -96,107 +120,59 @@ following rules:
...
+address for an access is out-of-bounds, the effective address will always also
+be out-of-bounds. This is intended to simplify folding of offsets into complex
+address modes in hardware, and to simplify bounds checking optimizations.
 
-Break and continue statements can only target blocks or loops in which they are
-nested. This guarantees that all resulting control flow graphs are well-structured.
+In the MVP, address operands and offset attributes have type `int32`, and heap
+sizes are limited to 4 GiB. In the future, to support
+[>4GiB linear memory](FutureFeatures.md#heaps-bigger-than-4gib), there will be
+a mode in which they have type `int64`.
```
jfbastien: Limited to 4GiB: link to Nondeterminism.md's description of "Environment-dependent resource limits may be exhausted".

AstSemantics.md
```diff
@@ -96,107 +120,59 @@ following rules:
...
-correctly (as if the alignment operand was `1`). However, on some platforms,
-such misaligned accesses may incur a *massive* performance penalty (due to trap
-handling). Thus, it is highly recommend that every WebAssembly producer provide
-accurate alignment. Note: on platforms without unaligned accesses,
-smaller-than-natural alignment may result in slower code generation (due to the
-whole access being broken into smaller aligned accesses).
+Thus, it is recommend that WebAssembly producers align frequently-used data
+to permit the use of natural alignment access, and use loads and stores with
+the grestest alignment values practical, while always avoiding misaligned
+accesses.
```
jfbastien: Could you also mention that a misaligned access which is partly out-of-bounds traps?

AstSemantics.md
```diff
@@ -96,107 +120,59 @@ following rules:
...
+at this byte and extending for the number of bytes implied by the memory
+attribute.
 
-WebAssembly offers basic structured control flow. All control flow structures
-are statements.
+Accesses to any bytes of linear memory beyond the `memory_size` are considered
+*out-of-bounds*. A module may optionally define that out-of-bounds includes
+small effective addresses close to `0`
+(see [discussion] (https://github.com/WebAssembly/design/issues/204)).
+The semantics of out-of-bounds accesses are discussed below.
```
jfbastien: Add a link to the below section.

AstSemantics.md
MikeHolman: Offset in bytes makes more sense to me -- e.g. for accessing elements in a heap allocated struct

AstSemantics.md
```diff
@@ -241,14 +234,48 @@ Accesses to global variables specify the index as an integer literal.
...
+    (back to the top of the loop)
+  * `forever`: infinite loop statement (like `while (1)`), basically an
+    unconditional branch (back to the top of the loop)
+  * `continue`: continue to start of nested loop
+  * `break`: break to end from nested loop or block
+  * `return`: return zero or more values from this function
+  * `switch`: switch statement with fallthrough
+
+Break and continue statements can only target blocks or loops in which they are
+nested. This guarantees that all resulting control flow graphs are well-structured.
```
jfbastien: `switch` doesn't allow Duff's device, even though it supports fallthrough, because the AST structure prevents nesting a loop which straddles `switch` statement blocks.

AstSemantics.md
```diff
@@ -241,14 +234,48 @@ Accesses to global variables specify the index as an integer literal.
...
+    [algorithm](http://dl.acm.org/citation.cfm?id=2048224&CFID=670868333&CFTOKEN=46181900),
+    with guaranteed low code size overhead, and typically minimal throughput
+    overhead (except for pathological cases of irreducible control
+    flow). Alternative approaches can generate reducible control flow via node
+    splitting, which can reduce throughput overhead, at the cost of increasing
+    code size (potentially very significantly in pathological cases).
+  * The
+    [signature-restricted proper tail-call](PostMVP.md#signature-restricted-proper-tail-calls)
+    feature would allow efficient compilation of arbitrary irreducible control
+    flow.
```
jfbastien: This list is missing a paragraph that explains it.

AstSemantics.md
jfbastien: Ha, I didn't notice it the first time around. I can also do a separate PR following yours, if that's cleaner.

sunfishcode: Yeah, if you don't mind, that'd be simpler.

AstSemantics.md
sunfishcode: It's meant to be an offset in bytes. I'll clarify that.

AstSemantics.md
sunfishcode: This is also pre-existing text. I'll submit a followup PR for this.

AstSemantics.md
sunfishcode: This is more pre-existing text :-). I agree that there's a lot of clarification needed. Let's handle this in follow-up PRs.

AstSemantics.md
sunfishcode: I'm rewording the out-of-bounds discussion per your comments above, and that will hopefully clarify this.

AstSemantics.md
jfbastien: Will do. Filed #248 to do this once this PR is checked in.

AstSemantics.md
jfbastien: SGTM

AstSemantics.md
```diff
@@ -96,107 +120,61 @@ following rules:
...
+### Addressing
 
-  * `get_local`: read the current value of a local variable
-  * `set_local`: set the current value of a local variable
-
-The details of index space for local variables and their types will be further clarified,
-e.g. whether locals with type `int32` and `int64` must be contiguous and separate from 
-others, etc.
+Each linear memory access operation also has an address operand and an immediate
+integer offset attribute. The infinite-precision sum of the address operand's
```
jfbastien: I'd say "immediate integer offset attribute, in bytes".

sunfishcode: How about "immediate integer byte offset attribute"? I updated the pull request to say that.

jfbastien:
lgtm, assuming I do a follow-up PR and you do 2 :-)

sunfishcode:
Ok, I submitted #253 which includes updates for both issues I said I'd fix above. Note that that PR depends on this one.

AstSemantics.md
```diff
@@ -32,35 +32,62 @@ Callstack space is limited by unspecified and dynamically varying constraints.
...
+[threads](PostMVP.md#threads) are added as a feature, regular load and store
+nodes will have the most relaxed semantics specified in the memory model and new
+memory-access nodes will be added with atomic and ordering guarantees.
+
+### Linear Memory Operations
+
+Linear memory operations are annotated with a memory type and perform a
+conversion between that memory type and a basic type.
+
+Loads read data from linear memory, convert from their memory type to a basic
```
titzer: We've been using the term local type in other places.

AstSemantics.md
```diff
@@ -96,107 +120,61 @@ following rules:
 Wrapping of integers simply discards any upper bits; i.e. wrapping does not
 perform saturation, trap on overflow, etc.
 
-## Addressing local variables
+In addition to storing a value to linear memory, store instructions also return
+their value operand, with no conversion applied.
```
titzer: "Store expressions" and "produce a value" rather than return, since we want to avoid confusion about returning from the function.

sunfishcode: Ok, fixed.

AstSemantics.md
sunfishcode: I grepped the design repo and didn't see any other references. Do you prefer "local" over "basic" here? These types do get used outside of "local" contexts.

titzer: I recall using "local type" in AstSemantics. Maybe those references have
been refactored away. "Local" is nice because it relates to local variables.

sunfishcode:
I filed #255 to discuss whether 64-bit addressing is a mode or separate opcodes. And I've now updated the PR to avoid mentioning modes.

AstSemantics.md
sunfishcode: Yes, I had found and replaced them all :-). But I don't feel strongly about this right now, so I've now added another patch to revert back to "local types".

sunfishcode:
I believe I've addressed all outstanding comments as of this comment. The text mentioning modes has been changed to be more vague.

jfbastien:
Can this land, so subsequent PRs are unblocked?

sunfishcode:
As far as I know, all outstanding concerns have been addressed.

titzer:
lgtm

jfbastien:
Many LGTMs, so merge is yes.

Issue #249, opened 2015-07-06
Can we leave the door open for WebAssembly growing into a universal binary format for.. everything?


aardappel:
I appreciate that there's "Web" in WebAssembly, and that the MVP goals are all about making sure that WebAssembly performs well as a replacement for current asm.js & NaCl use cases.

But I see a bigger future. We've never had a cross platform, stable, portable and widely supported binary standard that supports C level semantics well. They've either not been very stable (LLVM), don't support C efficiently (JVM) or not widely supported (most everything else).

If in the future all browsers will have high quality WebAssembly implementation, the attraction to also use this for client (desktop/mobile) & server deployment will be huge. I know there's this:
https://github.com/WebAssembly/design/blob/master/NonWeb.md,
but that seems to mostly hint at the Node.js style use case. I'd hope that non-web client use cases (especially e.g. games) are not ignored.

Cross-platform C/C++ building, testing and deployment continues to be a huge pain till this day. The thought that I can make a C++ game on my favourite platform, spit out a wasm binary, test it locally, show it to others using the web, then deploy on other native platforms without further compilation (either by bundling it with a pre-made binary that can run wasm or because the platform supports it natively) makes me positively drool.

Emscripten already comes with ways to access OpenGL/SDL-alike functionality out of the box, so turning this into something that typically all WebAssembly client platforms provide would be amazing for all sorts of client development.

Additionally, I think WebAssembly growing into an even bigger standard would do wonderful things for future programming languages, which could then be on equal footing with C/C++ by simply outputting WebAssembly. Currently, that requires a level of toolchain integration on many platforms that's a huge barrier (see current mobile platforms, especially).

tl;dr: any design choices that tie WebAssembly too closely to JS semantics / JS APIs / DOM today may slow down or even prevent this very beneficial future direction.

jfbastien:
+1 that's exactly what we're aiming for.

You asking this question means it's not clear. Do you have suggestions of what parts of the design we can clarify? You can even send pull requests if you're willing to join the [Community Group](https://www.w3.org/community/webassembly/participants).

sunfishcode:
I believe there's already a desire to ensure that WebAssembly can be used independently of JS. I created #251 to add a paragraph about this.

Beyond that, I think the functionality you describe would mostly just need a person to do the work to develop, maintain, and package such an environment. That's not something the WebAssembly Community Group itself is likely to do (for the foreseeable future), but others could.

BrendanEich:
[...]

PR #251, opened 2015-07-06
Clarify that WebAssembly doesn't depend on JS

sunfishcode:
This replaces "browser" with "Web" in a few places, and adds a paragraph to NonWeb.md mentioning that WebAssembly is being designed to be independent of JS. This is relevant to #249.

PR #258, opened 2015-07-08
Clarified possible mobile/desktop use cases in NonWeb.md

aardappel:
As discussed in:
https://github.com/WebAssembly/design/issues/249
NonWeb.md didn't capture the full generality of possible use cases
for WebAssembly as a universal binary standard.

### DIFF
--- a/NonWeb.md
+++ b/NonWeb.md
@@ -3,15 +3,16 @@
 While WebAssembly is designed to run [within a browser](Web.md), it is
 also desirable for it to be able to execute well in other environments,
 including everything from minimal shells for testing to full-blown
-application environments e.g. on servers in datacenters. It may even be
-desirable to execute WebAssembly embedded within larger programs.
+application environments e.g. on servers in datacenters, on IoT devices,
+or mobile/desktop apps. It may even be desirable to execute WebAssembly
+embedded within larger programs.
 
 Non-browser environments may provide different APIs than Web
 environments, which
 [feature testing](FeatureTest.md) and
 [dynamic linking](FutureFeatures.md#dynamic-linking) will make discoverable and
 usable.
-  
+
 Where there is overlap between the browser and popular non-browser environments,
 shared specs could be proposed, but this would be separate from the WebAssembly
 spec. A symmetric example in JavaScript would be the
@@ -20,8 +21,17 @@ both browsers and node.js. This situation is expected to be first encountered
 with POSIX features such as file I/O. In that respect, WebAssembly would err
 towards standardizing existing practice through libraries, and let developers
 choose which libraries to use.
+Similarly, the SDL interface (exposed by emscripten) could be made into a
+shared spec with non-browser mobile/desktop environments, which combined with
+the existing WebGL spec would for example allow for cross platform browser
+and non-browser games (and other things).
 
 The WebAssembly spec will not try to define any large portable libc-like
 library. However, certain features that are core to WebAssembly semantics that
 are found in native libc *would* be part of the core WebAssembly spec as either
 primitive opcodes or a special builtin module (e.g., `sbrk`, `mmap`).
+
+In general, by keeping the non-browser path such that it doesn't require
+browser APIs, WebAssembly could be used as a portable binary format on many
+platforms, bringing great benefits in portability, tooling and
+language-agnosticity (since it supports C/C++ level semantics).


NonWeb.md
```diff
@@ -3,15 +3,16 @@
 While WebAssembly is designed to run [within a browser](Web.md), it is
 also desirable for it to be able to execute well in other environments,
 including everything from minimal shells for testing to full-blown
-application environments e.g. on servers in datacenters. It may even be
-desirable to execute WebAssembly embedded within larger programs.
+application environments e.g. on servers in datacenters, or mobile/desktop
```
jfbastien: Could you add "IoT" too? It's a wonderful buzzword, which the high-level goals also mention.

NonWeb.md
```diff
@@ -20,8 +21,19 @@ both browsers and node.js. This situation is expected to be first encountered
 with POSIX features such as file I/O. In that respect, WebAssembly would err
 towards standardizing existing practice through libraries, and let developers
 choose which libraries to use.
+Similarly, the WebGL and SDL interface (
+[exposed by emscripten](http://kripken.github.io/emscripten-site/docs/porting/multimedia_and_graphics/OpenGL-support.html)
+in C/C++) could be made into a shared spec with non-browser mobile/desktop
+environments, allowing for cross platform browser & non-browser games and other
```
jfbastien: "and" Instead of "&"

NonWeb.md
```diff
@@ -20,8 +21,19 @@ both browsers and node.js. This situation is expected to be first encountered
...
+things.
```
sunfishcode: WebGL already has a spec, and WebAssembly is planned to be extended with GC handles and other things to allow it to talk to WebIDL bindings directly. For cross-platform browser & non-browser games, that seems to be the obvious way forward. Consequently, it's not clear what kind of new spec this paragraph is proposing.

NonWeb.md
```diff
@@ -20,8 +21,19 @@ both browsers and node.js. This situation is expected to be first encountered
...
 
 The WebAssembly spec will not try to define any large portable libc-like
 library. However, certain features that are core to WebAssembly semantics that
 are found in native libc *would* be part of the core WebAssembly spec as either
 primitive opcodes or a special builtin module (e.g., `sbrk`, `mmap`).
+
+In general, by keeping the non-browser path such that it doesn't require too
+much "browser emulation" (JS/DOM dependencies), WebAssembly could be used as
```
trevnorris: I read "JS" to imply ECMA, which doesn't have any special tie to the browser for cases like node.js. The "browser emulation" would be more, for example, what's defined by WhatWG. May be insignificant at this level of detail but wanted to bring it up.

May possibly want to separate how JS impacts wasm (e.g. need for GC) and how the browser APIs do the same (e.g. WebWorkers, event loop, I/O).

NonWeb.md
```diff
@@ -20,8 +21,19 @@ both browsers and node.js. This situation is expected to be first encountered
...
+a universal binary standard for many platforms, with huge benefits in
+portability, tooling and language-agnosticity (this being the first widely
+supported "bytecode" that supports C/C++ level semantics).
```
sunfishcode: > "browser emulation" (JS/DOM dependencies)

I'd like to keep this section specific and not handwave about "browsers" and some need for "emulation". I proposed #251, and it make sense to mention the DOM specifically as an API that NonWeb use cases need not provide, or other specific things people might be concerned about.

I'm also not really comfortable with the rest of the language in this paragraph. "Universal", "binary standard", "first", "widely supported" and "bytecode" are all phrases which may lead to widely varying interpretations. It seems sufficient to me to add mobile and desktop (and IoT, rock on, @jfbastien!) to the use cases at the top.

NonWeb.md
aardappel: Done.

NonWeb.md
aardappel: Ok, will reword that. Besides WebGL, there's the SDL (used for input, timing, audio, etc. in emscripten).

sunfishcode: For SDL, should there be an SDL-level spec, or lower-level input/timing/audio specs with SDL being just a library on top of them? I personally don't have a sense of what's best here. But if you want portability to browsers, we have Web APIs with specs for the low-level things today.

NonWeb.md
aardappel: will reword

NonWeb.md
aardappel: Ok, removed all those words and reworded.

NonWeb.md
aardappel: SDL is a fairly clean and simple C style API, so that would be the preferred way, but the low level approach would work too (which presumably is what emscripten uses today).

aardappel:
Updated, with all of your suggestions.

jfbastien:
lgtm

Note: overall this depends heavily on dynamic linking (which we mention towards the top) to avoid defining specific APIs in the WebAssembly spec. I think it's clear when reading the entire document, but not necessarily clear when reading this PR on its own.

kripken:
lgtm as well

PR #259, opened 2015-07-09
Clarify regular load/store semantics when threads are added

### LINKED ISSUE #248: Fix memory ordering of regular load / store
Self-issue to track [discussion in #245](https://github.com/WebAssembly/design/pull/245#discussion_r33957301). Once #245 is in I'll submit a separate PR to fix the wording.

jfbastien:
Resolves #248.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -76,8 +76,11 @@ of bytes. The linear memory is sandboxed; it does not alias the execution
 engine's internal data structures, the execution stack, local variables, global
 variables, or other process memory.
 
-In the MVP, linear memory is not shared between threads of execution. This will
-change when [threads](PostMVP.md#threads) are added.
+In the MVP, linear memory is not shared between threads of execution. Separate
+modules can execute in separate threads but have their own liinear memory and
+can only communicate through messaging, e.g. in browsers using `postMessage`. It
+will be possible to share linear memory between threads of execution when
+[threads](PostMVP.md#threads) are added.
 
 ### Linear Memory Operations
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -78,8 +78,9 @@ variables, or other process memory.
 
 In the MVP, linear memory is not shared between threads. When
 [threads](PostMVP.md#threads) are added as a feature, regular load and store
-nodes will have the most relaxed semantics specified in the memory model and new
-memory-access nodes will be added with atomic and ordering guarantees.
+nodes will be bound by a happens-before relationship to atomic operations in
+the same thread of execution, which themselves synchronize-with atomics in other
+threads. This follows the [C++11 memory model](http://www.hboehm.info/c++mm/).
 
 ### Linear Memory Operations
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -76,11 +76,14 @@ of bytes. The linear memory is sandboxed; it does not alias the execution
 engine's internal data structures, the execution stack, local variables, global
 variables, or other process memory.
 
-In the MVP, linear memory is not shared between threads. When
-[threads](PostMVP.md#threads) are added as a feature, regular load and store
-nodes will be bound by a happens-before relationship to atomic operations in
-the same thread of execution, which themselves synchronize-with atomics in other
-threads. This follows the [C++11 memory model](http://www.hboehm.info/c++mm/).
+In the MVP, linear memory is not shared between threads. New atomic memory
+operations will be added when [threads](PostMVP.md#threads) are added as a
+feature, including loads/stores annotated with their atomic ordering property,
+following the [C++11 memory model](http://www.hboehm.info/c++mm/). Regular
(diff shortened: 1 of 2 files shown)


AstSemantics.md
```diff
@@ -78,8 +78,9 @@ variables, or other process memory.
 
 In the MVP, linear memory is not shared between threads. When
 [threads](PostMVP.md#threads) are added as a feature, regular load and store
-nodes will have the most relaxed semantics specified in the memory model and new
-memory-access nodes will be added with atomic and ordering guarantees.
+nodes will be bound by a happens-before relationship to atomic operations in
+the same thread of execution, which themselves synchronize-with atomics in other
+threads. This follows the [C++11 memory model](http://www.hboehm.info/c++mm/).
```
titzer: Maybe clarify that we're envisioning an atomicity annotation on loads and stores, with the default being the most relaxed.

jfbastien: Adding mention of atomicity annotation.
"relaxed" isn't quite appropriate because it has a meaning in the C++ memory model. Non-annotated accesses have different semantics than relaxed accesses: they can be elided, duplicated, split up.

AstSemantics.md
```diff
@@ -76,10 +76,14 @@ of bytes. The linear memory is sandboxed; it does not alias the execution
...
-nodes will have the most relaxed semantics specified in the memory model and new
-memory-access nodes will be added with atomic and ordering guarantees.
+In the MVP, linear memory is not shared between threads. New atomic memory
+operations will be added when [threads](PostMVP.md#threads) are added as a
+feature, including loads/stores annotated with their atomic ordering property,
+following the [C++11 memory model](http://www.hboehm.info/c++mm/). Regular
+loads and stores will be bound by a happens-before relationship to atomic
+operations in the same thread of execution, which themselves synchronize-with
+atomics in other threads. Following these rules, regular load/store operations
+can still be elided, duplicated, and split up.
```
sunfishcode: This is also good content, but it seems like it belongs in PostMVP.md#threads too, so that all the interesting information about threads is in one place.

AstSemantics.md
```diff
@@ -76,10 +76,14 @@ of bytes. The linear memory is sandboxed; it does not alias the execution
...
 variables, or other process memory.
 
-In the MVP, linear memory is not shared between threads. When
-[threads](PostMVP.md#threads) are added as a feature, regular load and store
-nodes will have the most relaxed semantics specified in the memory model and new
-memory-access nodes will be added with atomic and ordering guarantees.
+In the MVP, linear memory is not shared between threads. New atomic memory
+operations will be added when [threads](PostMVP.md#threads) are added as a
+feature, including loads/stores annotated with their atomic ordering property,
+following the [C++11 memory model](http://www.hboehm.info/c++mm/). Regular
```
sunfishcode: I think a better place for this link would be in PostMVP.md#threads itself.

jfbastien:
Updated per @sunfishcode's suggestions.

AstSemantics.md
```diff
@@ -76,10 +76,8 @@ of bytes. The linear memory is sandboxed; it does not alias the execution
 engine's internal data structures, the execution stack, local variables, global
 variables, or other process memory.
 
-In the MVP, linear memory is not shared between threads. When
-[threads](PostMVP.md#threads) are added as a feature, regular load and store
-nodes will have the most relaxed semantics specified in the memory model and new
-memory-access nodes will be added with atomic and ordering guarantees.
+In the MVP, linear memory is not shared between threads of execution. This will
+change when [threads](PostMVP.md#threads) are added.
```
lukewagner: I think this is technically correct, but only in the way that `false implies P` is vacuously true so I think it'd be more clear to reword.  The fact that you can have separate wasm modules executing in separate workers (and thus on separate threads) is irrelevant since the wasm spec only considers the execution of a single module (much like the C spec is only talking about what's happening in a single process).  I guess the goal is to provide a link from here to PostMVP.md#threads; perhaps clearer to just say "When [threads] are added, the memory model will be augmented to define the semantics of concurrent linear heap accesses." or something.

jfbastien: Good point, updated.

jfbastien:
Ping?

AstSemantics.md
```diff
@@ -76,10 +76,11 @@ of bytes. The linear memory is sandboxed; it does not alias the execution
 engine's internal data structures, the execution stack, local variables, global
 variables, or other process memory.
 
-In the MVP, linear memory is not shared between threads. When
-[threads](PostMVP.md#threads) are added as a feature, regular load and store
-nodes will have the most relaxed semantics specified in the memory model and new
-memory-access nodes will be added with atomic and ordering guarantees.
+In the MVP, linear memory is not shared between threads of execution. Separate
+modules can execute in separate threads but have their own liinear memory and
```
sunfishcode: typo: liinear

sunfishcode:
lgtm, with typo fixed.

PR #263, opened 2015-07-13
Binary encoding format proposal

kg:
I've been doing some binary encoding/file format research in the js-astcompressor-prototype repo since late April. Based on my results, I'd like to propose an approach for the on-disk encoding side of the format.

My proposal is a simple table-based format that can encode arbitrary graphs given a schema that describes the node types. This will be able to efficiently represent whatever set of opcodes and tree nodes we decide to ship with in the MVP. The basic format ('layer 0', as described in the PR) is extremely simple to encode and decode, which makes it ideal for experimental implementations, JIT, debugging tools, and so on. A handful of optimization techniques are layered on top of this to produce a format that is designed to be efficient to decode and compress extremely well with GZIP, LZHAM and other codecs.

As we don't have anything resembling a finalized AST structure or opcode set (though we're approaching that point), I don't have measurements for that. However, I've been doing extensive experiments using representative data - massaged JS parse trees for compiled JS applications like BananaBread and AngryBots. I've also tested this same representation on more interesting code and found good results there. Some preliminary numbers for BananaBread:

-

Applying all the optimizations described in the proposal reduces the size of a representative Layer 0 file on-disk from **23,455KiB** down to **3,389KiB**.
Stream-compressed results are even more significant, with the size post-GZIP decreasing from **6,716KiB** to **1,072KiB**, and the size post-LZHAM going from **4,499KiB** to **986KiB**.

The optimizations do not just reduce file size - they also increase decoding speed **and reduce the amount of heap necessary to decode the file**. For the graph used in the above size tests, the amount of heap used to decode it from JSON is **262,717KiB**, the amount of heap used to decode it from Layer 0 is **177,741KiB**, and the amount of heap used to decode the fully-optimized Layer 2 file is **47,219KiB**, a reduction of **nearly 4x**. Decoding speed likewise increases.

-

I believe this proposal strikes the right balance for the current state of the project:
* We want to make it easy for compiler authors and tools developers to interact with WebAssembly.
  * These developers can interact with raw Layer 0 files if they want, leveraging existing tools to convert to/from Layer 2.
* We want to hit the ground running and put working decoders in browsers as soon as possible.
  * Likewise, we can start with just a simple Layer 0 implementation and rely on a polyfill or offline converter to lower Layer 1/2 executables down to what the native decoder implements.
* We want to ship a good interpreter and a good polyfill.
  * Decoding layer 0 is extremely simple and decoding layer 2 is not much harder. My prototype is written in pure javascript and manages to be relatively small despite the fact that it supports enabling/disabling every proposed optimization with a flag.
* We want to be able to adapt the format in response to user feedback once we ship the MVP.
  * There's lots of room to do additional encoder-side optimization with this format, and it will be easy to introduce a Layer 3 (compatibility-breaking file format changes) with polyfill fallback to convert it to what native decoders currently support.

This approach will also make it easier for us to experiment going forward, since we will be able to generate and consume our ASTs in JSON for prototyping purposes.


kg:
For some reason the PR UI doesn't make it easy to view documents like this. Here's a link:
https://github.com/kg/design/blob/58196d8fddc38db2483e32c03138f26a87383b25/LayeredBinaryEncoding.md

svick:
What confuses me about that description is the schema.

How does the schema look like? How is it encoded? What is the relationship between the schema and the type tags table?

Or it that just not in the scope of this document (yet?)?

kg:
The schema is a separate concern; it essentially represents the opcodes/operands stuff that we're still sorting out. What opcodes are there, etc.

I don't have any firm opinion on what the schema needs to look like, and it might live outside of the files instead of inside. Right now the schema is a tiny JSON file that describes the shape of the AST I'm using for testing, see here: https://github.com/WebAssembly/js-astcompressor-prototype/blob/master/shapes-jsontree.json

The type tags table is a set of names that map to names in the schema. The idea there is that you may not end up using every opcode from the schema in your executable; this also provides a way to notify the runtime that you need particular opcodes (for feature detection, fallback, etc).

The type tag table and string table are two things that will definitely end up changing in some fashion if this proposal is adopted.

sunfishcode:
This is a very interesting proposal! Thanks for preparing it, and making this detailed design document. I've just finished an initial pass through it; here are my notes and questions:

The idea of splitting decoding into layers, making possible a "layer 0" encoding that could be used be simple low-level tools, is nice.

"type tags" initially confused me because of the ambiguity around the word "type". Would "opcode tags" make sense?

The paragraph on deduplication says "70% or more of of all nodes are deduplicated". Does this include literal constant nodes? Also, is this on JS ASTs of asm.js code?

There has been a fair amount of discussion elsewhere about a (non-nullary) macro layer. Would that be something we might call a new layer, between layers 1 and 2, or do you have a different plan?

Would we consider the final generic compression (gzip, lzham, etc) a separate "layer"?

Deduplication is similar to the "local variable" construct in the current AST design, although it looks more powerful because it just cares about tree structure, so it can work even if the values of variables change, and it can work across a whole module too. And, I see that it decreases the post-stream-compression memory usage. That makes it interesting. However, if we have a stream compression layer, does deduplication (the nullary macros in layer 0) give us a final compression advantage?

Layer 1 claims
> the ability to store entire graphs of objects inside a single table entry, by writing them directly using a pre-order tree representation.

From my reading, it just supports trees, rather than arbitrary graphs, right?

> Preorder Encoded Object Tag Splitting

How will the decoder know when to look at the Inline Flags table? Without having those bytes inline, it's not clear how inline objects are indicated.

> Clustering all integer literals, double literals, string references &c into their own streams dramatically increases compressibility.

Note that WebAssembly programs won't have string literals embedded in their ASTs (string data will be just bytes in a constant data section somewhere). And, in many ASTs, double literals tend to be fairly localized just because of the nature of how they're used.

> However, once all Layer 2 optimizations have been applied, it is roughly the same size as the input JavaScript (both before and after stream compression), if not smaller.

JavaScript is a pretty inefficient encoding of asm.js's abstract structure, so it's unclear how to interpret this.

> This format accomplishes many of the same goals as JavaScript minification, so compression gains are more significant when used on an unminified JavaScript file."

Can you clarify what this signifies?

> Compression results are similar (if not better) for the AngryBots demo.

What does "similar (if not better)" mean?

Thinking about decode efficiency, I recognize the advantage in reducing memory usage post-stream-decompression of an entire module. Although, since it looks like it requires random access over the entire object table, will this limit our options for streamed decoding? One of the ideas for the pure AST serialization approach was that a very fast decoder could stream its input in and avoid the need to ever store the entire decompressed AST in memory all at once.

Since objects are variable-length, will decoders need to build an index for looking up object offsets?

kg:
> "type tags" initially confused me because of the ambiguity around the word "type". Would "opcode tags" make sense?

Opcode tags is probably the right term here, yes. Or just 'opcodes'.

> The paragraph on deduplication says "70% or more of of all nodes are deduplicated". Does this include literal constant nodes? Also, is this on JS ASTs of asm.js code?

The current test cases are JS ASTs of asm.js code. I've also tested it with large non-asm JS code, but my two main test cases are bananabread and angrybots. This includes all parts of the tree except string literals (which are interned into a string table). It sounds like string literals would go into the heap so the string table would only be used for things like the names of exported functions, so that table becomes less important.

> There has been a fair amount of discussion elsewhere about a (non-nullary) macro layer. Would that be something we might call a new layer, between layers 1 and 2, or do you have a different plan?

Yes, I think a non-nullary macro mechanism would probably be a new layer. You could slot it in anywhere. I don't think you could introduce that without compatibility-breaking format changes.

> Would we consider the final generic compression (gzip, lzham, etc) a separate "layer"?

Sure. It's definitely optional. I think debug builds and JITs definitely would omit generic compression.

> Deduplication is similar to the "local variable" construct in the current AST design, although it looks more powerful because it just cares about tree structure, so it can work even if the values of variables change, and it can work across a whole module too. And, I see that it decreases the post-stream-compression memory usage. That makes it interesting. However, if we have a stream compression layer, does deduplication (the nullary macros in layer 0) give us a final compression advantage?

When everything is applied, turning off deduplication isn't much of a post-compression size penalty - lzham and brotli are especially good here. However, deduplication dramatically shrinks the uncompressed size along with the heap size reduction, so it pays for itself pretty well. I think it can probably produce better gains with a more intelligent implementation.

> > the ability to store entire graphs of objects inside a single table entry, by writing them directly using a pre-order tree representation.

> From my reading, it just supports trees, rather than arbitrary graphs, right?

You're right, this is an error. While the table approach can encode arbitrary graphs, the pre-order encoding can only represent trees. A mitigation for this is that you can pre-order encode a few levels and then encode a reference instead of continuing to recurse, which would allow you to represent a cycle or other structure. However, I am assuming we only care about trees - the generality only exists since for a while people were talking about CFGs and other representations.

> How will the decoder know when to look at the Inline Flags table? Without having those bytes inline, it's not clear how inline objects are indicated.

I think I was too abstract here. The current implementation is very simple: If preorder is turned on, an inline flag byte precedes *every object reference*. Depending on the value of the byte, the following data is either a preorder tree or an index pointing into the object table. You'd probably want to bit-pack this or use a sentinel value, but I ended up just letting stream splitting solve the problem as a starting point. This is similar to some of the discussions we were having about reserving a couple bits in our opcode space for certain purposes.

> Note that WebAssembly programs won't have string literals embedded in their ASTs (string data will be just bytes in a constant data section somewhere). And, in many ASTs, double literals tend to be fairly localized just because of the nature of how they're used.

This is a good point - in my tests those literal streams are not particularly large. It's a compression boost regardless, in part because it gets that noise out of the other streams - in general the win is from having every data type in its own stream so the compressor is able to do a better job as a whole. asm.js applications do have a great number of integer literals (heap offsets etc) so some gains there add up.

> JavaScript is a pretty inefficient encoding of asm.js's abstract structure, so it's unclear how to interpret this.

I was reluctant to mention this at all, but I figured I would get questions about roughly how large these test files are compared to the input JavaScript I used to generate the AST.

> > This format accomplishes many of the same goals as JavaScript minification, so compression gains are more significant when used on an unminified JavaScript file."

> Can you clarify what this signifies?

This is mostly referring to cases where you have a significant number of exported symbols, identifiers or string literals, since minification condenses those. If string literals live in the user space heap, this isn't really a factor, though. I think we had been discussing having debug information stored externally (source maps etc) so that also removes identifiers from the equation.

> What does "similar (if not better)" mean?

Most of my exhaustive tests are against BananaBread because it's reasonably-sized. I run less-frequent tests against AngryBots and the compression results are somewhat better there. I mostly wanted to call this out since AngryBots was the real-world scenario Luke was testing against, and I made sure that this approach is at least feasible for encoding something that big.

> Thinking about decode efficiency, I recognize the advantage in reducing memory usage post-stream-decompression of an entire module. Although, since it looks like it requires random access over the entire object table, will this limit our options for streamed decoding? One of the ideas for the pure AST serialization approach was that a very fast decoder could stream its input in and avoid the need to ever store the entire decompressed AST in memory all at once.

If the encoder orders entries in the object table correctly, it will be possible for a decoder to stream over the file and generate code from it, as far as I know. There might be some corner cases here. This is definitely a single-pass decoding process, even with pre-order encoding introduced. You're correct that it requires the decoder to retain objects in order to be able to construct cross-references. The number of objects retained for this purpose is much smaller if pre-order encoding is enabled, though, so that mitigates the problem considerably.

If memory serves, when you encode bananabread with all the opts turned on, it deduplicates around ~80% of the nodes and then encodes around 97% of the remaining nodes as pre-order trees. So the number of entries in the object table that need to be retained is actually quite small. I can get you more exact numbers if that helps.

If streaming decoding is important there are probably things you could do to this format to enable it. One option would be to disable deduplication and force the use of pre-order trees at some level of granularity (functions or basic blocks, probably?)

> Since objects are variable-length, will decoders need to build an index for looking up object offsets?

Objects of a given type are fixed length, unless pre-order encoding or variable length ints are enabled. However, since the object table contains an arbitrary mix of types, objects are effectively variable-length. The assumption here is that a decoder will at least do a pass over the tags (headers, if you will) in the object table. During that step you can build a table of offsets if necessary, though the ideal approach is to just construct the objects right then and there. Stream splitting moves all those tags into a single buffer, so building a seek/lookup table becomes easier.

lukewagner:
It's not clear what the numbers quoted in the OP are measuring; I see big reductions, but if the starting point is JSON representing asm.js source then it's harder to gauge.  My main question is which of these techniques are generally beneficial and which are addressing the inefficiencies of starting with JSON. Could you perhaps post a table allowing apples-to-apples comparison against the size tables in the [FAQ](https://github.com/WebAssembly/design/blob/master/FAQ.md#can-the-polyfill-really-be-efficient)?  (Instead of just "binary", I realize there are several layers and it'd be nice to list them out.)  Note, though, that polyfill-prototype-1 doesn't do any more deduplication than int/double constant pools, so it won't quite be apples-to-apples until polyfill-prototype-1 applies an analogous deduplication layer.

kg:
> Could you perhaps post a table allowing apples-to-apples comparison against the size tables in the FAQ?

No, because I have no way to get the graph out of your polyfill. :(

lukewagner:
We talked about this on IRC, but for posterity: getting the "graph" out of the polyfill isn't necessary, but rather filtering out all the extra stuff kept by js-astcompressor that is implicitly discarded by polyfill-prototype-1 (since it's not necessary for asm.js-generation).

kg:
I'm going to aggressively revise this proposal to bring it closer to something we can merge in. 

Two things I've tried so far based on feedback:

Eliminating the string table entirely is a small size increase because it inlines the strings directly into the objects. However, string literal objects can get deduplicated like any other object, which effectively produces a string table. Identifiers won't be strings, so that mitigates this as well. The deduplication heuristic could always be adjusted for string literals specifically to maximize the efficiency here.

Name interning (tree builder maintains a scope chain, replaces names with indices into the current scope) brings the post-compression size down about as much as removing the string table raised it. This makes the actual post-decode AST smaller too. This matches expectations for what we are planning to do with the real AST (numbered locals, etc.)

titzer:
Reasoning backwards from the AST that a level 0 decoder can process, it seems like a preorder encoding (i.e. inline) is more fundamental and denser than a graph. As such I think it belongs in layer 0 rather than layer 1. If the most fundamental level 0 is a graph model where we need to have pointers between different objects, then there will be a lot of pointer chasing in the decoder and I think performance will suffer.

In my mental model the layer 0 decoder will build compiler IR directly from some tree, and so far the v8-native-prototype works directly on the preorder encoded bytes (i.e. without an actual C++-level AST data structure). This is an important win in decoding speed and memory usage, IMO. For example I was able to achieve 65mb/s decoding and validation speed on a single thread by operating directly on the bytes. 

Would it make sense to think of the tree being the more fundamental representation, and the layer 1 decoding uses deduplication to save space, and that we could "reduplicate" the layer 1 graph back to a tree?

kg:
You still end up with an object table eventually, it's just a question of whether the objects are some arbitrary unit (functions, for example) or whether they are objects at a basic level.

If you're saying that you want preorder encoding as a baseline guarantee, that seems sensible? But you can't avoid the table representation, and a naive encoder may opt to never do preorder encoding.

titzer:
On Thu, Jul 23, 2015 at 2:01 PM, Katelyn Gadd <notifications@github.com>
wrote:

> You still end up with an object table eventually, it's just a question of
> whether the objects are some arbitrary unit (functions, for example) or
> whether they are objects at a basic level.
>
> If you're saying that you want preorder encoding as a baseline guarantee,
> that seems sensible? But you can't avoid the table representation, and a
> naive encoder may opt to never do preorder encoding.
>
Tables for functions, strings, globals, etc are totally fine. At the level
of individual function bodies, a preorder-encoded tree representing the AST
for the function body seems like the right unit for fast decoding. We could
probably share lots of internal trees across function bodies at layer 1,
especially if the wasm producer did very heavy inlining and thus had very
redundant code to start with.

kg:
I'd be happy to merge layer 0 into layer 1 and have the baseline be support for encoding things as pre-order trees or graphs of object references. I'm not sure there's any way to bake requirements about which specific object types can be pre-order encoded into the file format, since putting deduplication and stream splitting back in means those requirements have to be undone anyway.

The closest thing I can think of would be to just state that entries in the object table can't be subtrees. That turns the following layers into a more significant file format change, but maybe that's okay if it makes native decoders considerably simpler? (I'm not sure it actually would)

kg:
@titzer @lukewagner @jfbastien
So I'd like to try and get a rough consensus on what layer 0 should look like that everyone's okay with, and move forward on that. I'll push all the optional compression stuff off indefinitely so we can figure it out later in the design process.

My rough proposal here would be that a module is an opcode table plus an object table, as specified in the old draft in the PR. By definition, the object table will be where you store your functions, so most (or all) of the objects in that table will be functions.

An encoder will have the ability to store non-function-objects into the table, which covers both deduplication, non-preorder scenarios, and lazy encoders.

So from the decoder side the table is basically a function table with some extra entries that you don't care about*. Most entries in the table will be a fully preorder encoded function body. The inclusion of preorder encoding in layer 0 means that you will have an inline/not inline flag on every object, but that can be packed into the type (opcode) tag.

* If you find a non-inline object while decoding a function, then that's a reference to one of those non-function entries in the table. But you don't have to care about those until you find a reference to them.

How would you feel about this as a baseline? It is roughly equivalent to 'layer 1' from the original draft, but with some complication removed.

lukewagner:
I talked to @kg on IRC just now about how and when we should go about defining our "layer 0" (the only layer that's in the spec).  Based on the plan in #287, we were both agreeing that our next steps should be to focus on the AST and semantics and write all our tests/prototypes as S-Expressions that mirror the AST.  That way, we can get to a state where we know exactly what the structure of a wasm module is and we'll able to generate real wasm modules (as S-Expressions) from Emscripten (by having the `pack-asmjs` tool in polyfill-prototype-1 generate S-Expressions) and, if it's ready, llvm-wasm.  At *that* point, we'll be in a really good place to focus all our efforts on defining a good layer 0 format since we'll have totally real-world test corpuses (corpii?) in a simple input format that will allow direct apples-to-apples comparisons.

PR #264, opened 2015-07-13
FutureFeatures: clarify trigonometric functions

jfbastien:
The text was missing some of the background discussions that led to moving these intrinsics to future features, especially dynamic linking.

### DIFF
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -275,17 +275,25 @@ quadruple precision.
 
 These operations aren't needed for the MVP because they can be implemented in
 WebAssembly code and linked into WebAssembly modules at small size cost (as is
-traditionally done through C libraries such as libm). This avoids a non-trivial
-specification burden for their semantics, precision, and performance.
+traditionally done through C libraries such as libm). This approach:
 
-[Dynamic linking](FutureFeatures.md#dynamic-linking) will allow caching of libm,
-making this sharing cost trivial.
+* Avoids a non-trivial specification burden for their semantics, precision, and
+  performance.
+* Allows developers to make different application-specific tradeoffs of
+  precision/robustness versus performance, while still getting deterministic
+  results across implementations.
+* Reduces the implementation burden for WebAssembly, since more than the few
+  math functions listed below may be needed by different developers.
+
+[Dynamic linking](FutureFeatures.md#dynamic-linking) will also allow caching of
+libm, making this already low-cost sharing trivial.
 
 Adding these intrinsics would potentially allow for better high-level backend
 optimization of these intrinsics that require builtin knowledge of their
 semantics. WebAssembly may support these operations in the future if data shows
 it would be useful. The rounding behavior of these operations would need
 clarification.
+
 
   * `float64.sin`: trigonometric sine
   * `float64.cos`: trigonometric cosine
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -273,14 +273,19 @@ quadruple precision.
 
 ## Floating point library intrinsics
 
-These operations aren't needed because they can be implemented in WebAssembly
-code and linked into WebAssembly modules as at small size cost, and this avoids
-a non-trivial specification burden of their semantics/precision. Adding these
-intrinsics would allow for better high-level backend optimization of these
-intrinsics that require builtin knowledge of their semantics. On the other
-hand, a code generator may continue to statically link in its own
-implementation since this provides greater control over precision/performance
-tradeoffs.
+These operations aren't needed for the MVP because they can be implemented in
+WebAssembly code and linked into WebAssembly modules at small size cost (as is
+traditionally done through C libraries such as libm). This avoids a non-trivial
+specification burden for their semantics, precision, and performance.
+
+[Dynamic linking](FutureFeatures.md#dynamic-linking) will allow caching of libm,
+making this sharing cost trivial.
+
+Adding these intrinsics would potentially allow for better high-level backend
+optimization of these intrinsics that require builtin knowledge of their
+semantics. WebAssembly may support these operations in the future if data shows
+it would be useful. The rounding behavior of these operations would need
+clarification.
 
   * `float64.sin`: trigonometric sine
   * `float64.cos`: trigonometric cosine
@@ -302,8 +307,6 @@ tradeoffs.
   * `float32.exp`: exponentiate e
   * `float32.ln`: natural logarithm
   * `float32.pow`: exponentiate
-
-The rounding behavior of these operations would need clarification.
 
 ## Full IEEE-754 conformance
 


FutureFeatures.md
```diff
@@ -273,14 +273,19 @@ quadruple precision.
...
-hand, a code generator may continue to statically link in its own
-implementation since this provides greater control over precision/performance
-tradeoffs.
+These operations aren't needed for the MVP because they can be implemented in
+WebAssembly code and linked into WebAssembly modules at small size cost (as is
+traditionally done through C libraries such as libm). This avoids a non-trivial
+specification burden for their semantics, precision, and performance.
+
+[Dynamic linking](FutureFeatures.md#dynamic-linking) will allow caching of libm,
+making this sharing cost trivial.
```
sunfishcode: Even without dynamic linking, the code for most of these functions isn't actually very big (often just a few k).

jfbastien: Are you suggesting that I change the top of the paragraph "at small size cost" or the bottom "making this sharing cost trivial"? Agreed it's very small and will become even smaller! I'm keeping these mentions here because they were part of the discussions we had originally, where folks were worried of the size overhead.

FutureFeatures.md
```diff
@@ -273,14 +273,19 @@ quadruple precision.
...
-a non-trivial specification burden of their semantics/precision. Adding these
-intrinsics would allow for better high-level backend optimization of these
-intrinsics that require builtin knowledge of their semantics. On the other
-hand, a code generator may continue to statically link in its own
-implementation since this provides greater control over precision/performance
-tradeoffs.
+These operations aren't needed for the MVP because they can be implemented in
+WebAssembly code and linked into WebAssembly modules at small size cost (as is
+traditionally done through C libraries such as libm). This avoids a non-trivial
+specification burden for their semantics, precision, and performance.
```
sunfishcode: Additional advantages include:
 - developers (or tools/libraries/etc. made for developers) can make different tradeoffs of precision/robustness versus performance, while still getting deterministic results across implementations.
 - there's no temptation among implementations to "race to the bottom" in precision or robustness to compete for performance.
 - There are a *lot* of math functions, more than just the handful of trig and exponentiation variations listed here, and right now we have a very clear line of what belongs in the platform and what belongs in a library. Adding all these new functions would make it a lot less clear and open up the door to people wanting us to implement more functions in the future.

sunfishcode: Another advantage:
 - This gives us the best chance of getting consistent performance across implementations.

FutureFeatures.md
sunfishcode: It just felt like talking about code size at all makes it feel like it's something people should be concerned about in the MVP. Perhaps this sentence could say "Dynamic linking will even allow caching of libm, making this already-low sharing cost trivial."

FutureFeatures.md
```diff
@@ -273,14 +273,19 @@ quadruple precision.
...
+
+[Dynamic linking](FutureFeatures.md#dynamic-linking) will allow caching of libm,
+making this sharing cost trivial.
+
+Adding these intrinsics would potentially allow for better high-level backend
+optimization of these intrinsics that require builtin knowledge of their
+semantics. WebAssembly may support these operations in the future if data shows
+it would be useful. The rounding behavior of these operations would need
+clarification.
```
sunfishcode: If we're going to discuss this topic in detail, we should also mention an alternative way to achieve the same benefits, and much more: develop an optimization/JIT library, which could run as a library inside WebAssembly, and would be useful for general-purpose JITs of all kinds. Such a library could much more freely offer a much greater degree of nondeterminism because it wouldn't be baked into the platform, so it could theoretically offer services like:
 - math functions that aren't always perfectly rounded
 - lots of fast-math optimizations
 - evaluation of constant expressions at greater intermediate precision
 - optimizations that depend on knowing that integer arithmetic doesn't overflow
 - optimizations that depend on knowing that all loops either have side effects or terminate
 - higher-level semantics
 - many others!

One possibility here would be to compile and use LLVM inside WebAssembly, and I expect some people will do exactly this. There's also room for simpler and lighter-weight solutions as well; code for a simple IR, with basic inlining and constant folding (including of trigonometic and exponential functions!) would be quite simple.

And the great thing about a library approach is that one can depend on getting the same optimizations in all implementations. There's no need to make sure that every browser does the same inlining and thus achieves the same math function constant folding. One's optimization library either does it or it doesn't, everywhere.

FutureFeatures.md
jfbastien: Yes to the "race to the bottom" argument, but it's encompassed by having a specification for precision so I don't think it's worth mentioning (not everyone knows about transcendental cache sins, let's not dive into too many details).

I'm not sure getting consistent performance across implementations is really a goal. Code is code! If implementations lag behind then they need to work harder.

FutureFeatures.md
jfbastien: I don't think this specific feature is the right place to discuss what you suggest: it's just one thing that would benefit from this suggestion! Could you do a separate PR? I agree with the high-level, but I think it would be useful to discuss this in general, with a few other folks.

jfbastien:
Addressed @sunfishcode's comments.

FutureFeatures.md
sunfishcode: Yes indeed, let's talk about the transcendental math caches, a topic many of us are familiar with. This is an example where some implementations were faster than others on some workloads, and it isn't a simple matter of the others "lagging behind" and needing to "work harder". And while Sunspider may have started it, it turns out that some real-world applications benefit from the caches as well. So I'll add another bullet point to my list:
 - there's no temptation for implementations to optimize for some input patterns at the expense of others.

(It's true that we face this temptation in other areas, but that doesn't mean we shouldn't wish to avoid it here.)

FutureFeatures.md
sunfishcode: I don't think this specific feature belongs in WebAssembly in the first place, and I've given numerous reasons for it here, and I've also proposed an alternative approach which would solve the problem that this feature solves, while having many advantages besides. If you want to have this feature in the documentation, I'd like you to acknowledge and discuss the alternative approach as well.

sunfishcode: Ok, I've now created PR #265 to propose my idea.

jfbastien: Which "specific feature" are you referring to?

sunfishcode: The feature of having operators for libm functions.

jfbastien: I agree with you :-)

@titzer was pretty adamant on considering them in the future, I therefore don't want to outright delete this section. I do want to capture the discussion that led to this, which is why I'm doing this PR.

sunfishcode:
ok, lgtm. I'll rebase #265 on top of this.

Issue #256, opened 2015-07-07
Incorrect example of browser/node.js library compatibility


trevnorris:
There's a problematic example from the following in `NonWeb.md`:

> A symmetric example in JavaScript would be the Loader spec, intended to be implemented by both browsers and node.js.

Using the Loader spec in node.js has practical issues. Mainly because it specifies that all loading of imports must be done asynchronously. I've outlined a few issues and limitations of using asynchronous module loading in node.js:

- All modules exist locally (i.e. the hard drive) and their retrieval does not suffer from latency, round-trip or other browser related I/O issues.

- All `import`s will be resolved and loaded at the beginning of the application. Preventing the application from starting until all modules have been loaded. Basically even asynchronous loading will still block the application.

- The number of modules that can be asynchronously retrieved is limited by both the number of leafs in the dependency graph and the number of threads in libuv's thread pool. Further inhibiting possible performance gains from simultaneous retrieval.

- The asynchronous nature of Loader cannot be used safely with the existing synchronous `require()`. For example, a synchronously loaded module may cause side affects that impact the asynchronously loaded module.

- The `import` syntax in general is limiting to node.js module loading. Specifically pertaining to the loading of native modules that decide at run-time what build to use (e.g. debug/release).

Basically, implementing Loader in node.js has little to zero likelihood of a performance advantage and has high potential to cause non-trivial breakage in the ecosystem.

I've been considering what similar implementations would work well on browser/node.js and to be honest I can't think of one. Starting in v3 we use a bastardized `Uint8Array()` as the backing store for `Buffer` to get around performance issues. All of our I/O bindings sit very close to the syscalls underneath and support almost anything you could do at that level.

In short, any browser/node.js compatibility I've been able to think of may be useful for module authors who want to use the same code on both, but it's not useful for the core code.

bmeck:
Might want to note as well: standard feature detection of polyfills uses variables / conditional branching to figure out if a polyfill is requires to be loaded, current ecmascript loader syntax does not support this.

lukewagner:
@trevnorris IIUC your comment, you're saying that there may actually *not* be an intention of the Loader spec to be implemented in node and so the example should be removed.  If so, feel free to file a PR to remove.

[...]

PR #266, opened 2015-07-15
Clarify wording concerning portable libraries in NonWeb.md

lukewagner:
This makes two clarifications suggested in #256 concerning NonWeb.md.  The first is to dial down the wording about the Loader spec from "intended to be implemented" to "proposed".  The second is to be clear that the wasm spec isn't trying to spec any large portability libraries but rather expose the host environment's builtin libraries and allow community-developed libraries to provide portable abstractions.

### DIFF
--- a/NonWeb.md
+++ b/NonWeb.md
@@ -13,23 +13,23 @@ environments, which
 [dynamic linking](FutureFeatures.md#dynamic-linking) will make discoverable and
 usable.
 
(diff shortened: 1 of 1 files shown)


trevnorris:
Excellent. Thanks. Could never get quite the right wording myself. This reads well.

jfbastien:
lgtm

Issue #84, opened 2015-05-27
Feature detection and running Web Assembly


jfbastien:
The initial polyfill will unconditionally load JavaScript code, then load the Web Assembly binary blob, and somehow translate it to a form that'll execute using the JavaScript VM.

Once we have native support this polyfill will be a fallback: developer's code should only load the polyfill and use it if Web Assembly isn't natively supported.

How will we detect native Web Assembly support? Something like `navigator.mimeTypes['application/x-wasm'] !== undefined`?

How will we trigger the browser's native support? Use an `<embed>` or `<script>` tag?

jfbastien:
WebAssembly will be an object in JS, so this is not relevant anymore.

PR #270, opened 2015-07-17
Consolidate explanation of modules into a new Modules.md and improve explanation

lukewagner:
This PR is an attempt to consolidate the discussion of modules and their integration with ES6 modules into one place and also provide a better explanation.

Independently, there's been some agreement to work on a new "level 0" version of the [loader spec](http://whatwg.github.io/loader/) with the goal of being "just enough to be able to ship ES6 modules and WebAssembly modules".  When there is a URL to point to, I'll update the TODO link in Modules.md.

Especially interested for feedback from @ajklein.

(On a side note, with this PR, MVP.md is now pretty empty and ad hoc, it'd be good to tidy it up.  I might do that if noone beats me to it.)

### DIFF
--- a/MVP.md
+++ b/MVP.md
@@ -15,7 +15,10 @@ separate docs with more precise descriptions of:
  * [Polyfill to JavaScript](Polyfill.md);
  * [AST semantics](AstSemantics.md);
  * [Binary encoding](BinaryEncoding.md);
+ * [Text format](TextFormat.md);
  * Implementation [in the browser](Web.md) and [outside the browser](NonWeb.md).
+
+**Note**: This content is still in flux and open for discussion.
 
 ## Linear Memory
 
@@ -33,6 +36,21 @@ separate docs with more precise descriptions of:
     detaches any existent `ArrayBuffer`.
 * See the [AST Semantics linear memory section](AstSemantics.md#linear-memory)
   for more details.
+
+## Binary format
+
+* A [binary format](BinaryEncoding.md) provides efficiency: it reduces download
+  size and accelerates decoding, thus enabling even very large codebases to have
+  quick startup times. Towards that goal, the binary format will be natively
+  decoded by browsers.
+* The binary format has an equivalent and isomorphic
+  [text format](MVP.md#text-format).  Conversion from one format to the other is
+  both straightforward and causes no loss of information in either direction.
+
+## Text format
+
+The [text format](TextFormat.md) provides readability to developers, and is
+isomorphic to the [binary format](BinaryEncoding.md).
  
 ## Security
 
--- a/MVP.md
+++ b/MVP.md
@@ -11,83 +11,11 @@ even on mobile devices, which leads to roughly the same functionality as
 This document explains the contents of the MVP at a high-level. There are also
 separate docs with more precise descriptions of:
 
+ * [Modules](Modules.md)
  * [Polyfill to JavaScript](Polyfill.md);
  * [AST semantics](AstSemantics.md);
  * [Binary encoding](BinaryEncoding.md);
  * Implementation [in the browser](Web.md) and [outside the browser](NonWeb.md).
-
-**Note**: This content is still in flux and open for discussion.
-
-## Modules
-
-* The primary unit of loadable, executable code is a **module**.
-* A module can declare a subset of its functions and global variables to be
-  **exports**. The meaning of exports (how and when they are called) is defined
-  by the host environment. For example, `_start` and `init` can be the only
-  meaningful exports.
-* A module can declare a set of **imports**. An import is a tuple containing a
-  module name, export name, and the type to use for the import within the
-  module. The host environment controls the mapping from module name to which
-  module is loaded.
-* The spec defines the semantics of loading and calling exports of a *single*
-  module. The meaning of a call to an import is defined by the host environment.
-  * In a minimal shell environment, imports could be limited to builtin modules
-    (implemented by the shell) and/or shell scripts.
-  * The [dynamic linking](FutureFeatures.md#dynamic-linking) post-MVP feature
-    would extend the semantics to include multiple modules and thus allow sharing 
-linear memory and pointers. Dynamic linking would be semantically distinct from
-    importing, however.
-* When compiling from C++, imports would be generated for unresolved `extern`
-  functions and calls to those `extern` functions would call the import.
-* Host environments can define builtin modules that are implemented natively but
-  can otherwise be imported like [other modules](MVP.md#modules). As examples:
-  * A WebAssembly shell might define a builtin `stdio` library with an export
-    `puts`.
-  * In the browser, the WebIDL support mentioned in
-    [future features](FutureFeatures.md).
-* Any [ABI](https://en.wikipedia.org/wiki/Application_binary_interface) for
-  statically linked libraries will be specific to your source language compiler.
-  In the future, [standard ABIs may be defined](FutureFeatures.md#dynamic-linking)
-  to allow for compatibility between compilers and versions of compilers.
-* **TODO**: there is more to discuss here concerning APIs.
-
-## Module structure
-
-* At the top level, a module is ELF-like: a sequence of sections which declare
-  their type and byte-length.
- * Sections with unknown types would be skipped without error. 
- * Standardized section types:
-  * module import section;
-  * globals section (constants, signatures, variables);
-  * code section;
-  * memory initialization section.
-
-## Code section
-
-* The code section begins with a table of functions containing the signatures
-   and offsets of each function followed by the list of function bodies. This
-   allows parallel and streaming decoding, validation and compilation.
- * A function body consists of a set of typed variable bindings and an AST
-   closed under these bindings.
-  * The [Abstract Syntax Tree](AstSemantics.md) is composed of two primary kinds
-    of nodes: statements and expressions.
- * [Control flow](AstSemantics.md#control-flow-structures) is structured (no
-   `goto`).
-
-## Binary format
-
-* A [binary format](BinaryEncoding.md) provides efficiency: it reduces download
-  size and accelerates decoding, thus enabling even very large codebases to have
-  quick startup times. Towards that goal, the binary format will be natively
-  decoded by browsers.
-* The binary format has an equivalent and isomorphic
-  [text format](MVP.md#text-format).  Conversion from one format to the other is
-  both straightforward and causes no loss of information in either direction.
-
-## Text format
-
-The [text format](TextFormat.md) provides readability to developers, and is
-isomorphic to the [binary format](BinaryEncoding.md).
 
 ## Linear Memory
 
--- a/Modules.md
+++ b/Modules.md
@@ -1,0 +1,112 @@
+# Modules
+
+The distributable, loadable, and executable unit of code in WebAssembly
+is called a **module**. A module contains:
+* a set of [imports and exports](Modules.md#imports-and-exports);
+* a section defining the [initial state of linear memory](Modules.md#initial-state-of-linear-memory);
+* a section containing [code](Modules.md#code-section);
+* after the MVP, sections containing [debugging/symbol information](Tooling.md); and
+* possibly other sections in the future.
+
+While WebAssembly modules are designed to interoperate with ES6 modules
+in a Web environment (more details [below](Modules.md#integration-with-es6-modules)),
+WebAssembly modules are defined independently of JavaScript and do not require
+the host environment to include a JavaScript VM.
+
+## Imports and Exports
+
+A module defines a set of functions in its
+[code section](Modules.md#code-section) and can declare and name a subset of
+these functions to be **exports**. The meaning of exports (how and when they are
+called) is defined by the host environment. For example, a minimal shell
+environment might only probe for and call a `_start` export when given a module
+to execute.
+
+A module can declare a set of **imports**. An import is a tuple containing a
+module name, the name of an exported function to import from the named module,
+and the signature to use for that import within the importing module. Within a
+module, the import can be [directly called](AstSemantics.md#calls) like a
+function (according to its locally-declared signature).
+
+The WebAssembly spec does not define how imports are interpreted:
+* the host environment can interpret the module name as a file path, a URL,
+  a key in a fixed set of builtin modules or the host environment may invoke a
+  user-defined hook to resolve the module name to one of these;
+* the module name does not need to resolve to a WebAssembly module; it
+  could resolve to a builtin module (implemented by the host environment) or a
+  module written in another, compatible language; and
+* the meaning of calling an imported function is host-defined.
+
+The open-ended nature of module imports allow them to be used to expose
+arbitrary host environment functionality to WebAssembly code, similar to a
+native `syscall`. For example, a shell environment could define a builtin
+`stdio` module with an export `puts`.
+
+In C/C++, an undefined `extern` declaration could be compiled to an import and
+C/C++ calls to this `extern` would then be compiled to calls to this import. This
+is one way low-level C/C++ libraries could call out of WebAssembly in order to
+implement portable source-level interfaces (e.g., POSIX, OpenGL or SDL) in
+terms of host-specific functionality.
+
+### Integration with ES6 modules
+
+While ES6 defines how to parse, link and execute a module, ES6 does not
+define when this parsing/linking/execution occurs. An additional extension
+to the HTML spec is required to say when a script is parsed as a module instead
+of normal global code. This work is [ongoing](TODO). Currently, the following
+entry points for modules are being considered:
+* `<script type="module">`;
+* an overload to the `Worker` constructor;
+* an overload to the `importScripts` Worker API;
+
+Additionally, an ES6 module can recursively import other modules via `import`
+statements.
+
+For WebAssembly/ES6 module integration, the idea is that all the above module
+entry points could also load WebAssembly modules simply by passing the URL of a
+WebAssembly module. The distinction of whether the module was WebAssembly or ES6
+code could be made by namespacing or by content sniffing the first bytes of the
+fetched resource (which, for WebAssembly, would be a non-ASCII&mdash;and thus
+illegal as JavaScript&mdash;[magic number](https://en.wikipedia.org/wiki/Magic_number_%28programming%29)).
+Thus, the whole module-loading pipeline (resolving the name to a URL, fetching
+the URL, any other [loader hooks](http://whatwg.github.io/loader/)) would be
+shared and only the final stage would fork into either the JavaScript parser or
+the WebAssembly decoder.
+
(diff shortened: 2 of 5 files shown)


MVP.md
```diff
@@ -11,84 +11,12 @@ even on mobile devices, which leads to roughly the same functionality as
 This document explains the contents of the MVP at a high-level. There are also
 separate docs with more precise descriptions of:
 
+ * [Modules](Modules.md)
  * [Polyfill to JavaScript](Polyfill.md);
  * [AST semantics](AstSemantics.md);
  * [Binary encoding](BinaryEncoding.md);
  * Implementation [in the browser](Web.md) and [outside the browser](NonWeb.md).
 
-**Note**: This content is still in flux and open for discussion.
```
jfbastien: I think we should keep this not for now.

Modules.md
```diff
@@ -0,0 +1,112 @@
+# Modules
+
+The distributable, loadable, and executable unit of code in WebAssembly
+is called a **module**. A module contains:
+* a set of [imports and exports](Modules.md#imports-and-exports);
+* a section defining the [initial state of linear memory](Modules.md#initial-state-of-linear-memory);
+* a section containing [code](Modules.md#code-section);
+* after the MVP, sections containing [debugging/symbol information](Tooling.md); and
```
jfbastien: I was discussing this with @kg yesterday: debug info would be nice as a separate file too. Not to design a future feature here, but maybe "sections contained debugging/symbol information, or a reference to separate files containing them"?

Modules.md
```diff
@@ -0,0 +1,112 @@
...
+these functions to be **exports**. The meaning of exports (how and when they are
+called) is defined by the host environment. For example, a minimal shell
+environment might only probe for and call a `_start` export when given a module
+to execute.
+
+A module can declare a set of **imports**. An import is a tuple containing a
+module name, the name of an exported function to import from the named module,
+and the signature to use for that import within the importing module. Within a
+module, the import can be [directly called](AstSemantics.md#calls) like a
+function (according to its locally-declared signature).
```
jfbastien: Could you clarify "according to its locally-declared signature"? Are the exported and imported signatures allowed to be different?

Modules.md
```diff
@@ -0,0 +1,112 @@
...
+is one way low-level C/C++ libraries could call out of WebAssembly in order to
+implement portable source-level interfaces (e.g., POSIX, OpenGL or SDL) in
+terms of host-specific functionality.
+
+### Integration with ES6 modules
+
+While ES6 defines how to parse, link and execute a module, ES6 does not
+define when this parsing/linking/execution occurs. An additional extension
+to the HTML spec is required to say when a script is parsed as a module instead
+of normal global code. This work is [ongoing](TODO). Currently, the following
```
jfbastien: Ongoing outside of wasm, or is that something we need to tackle?

Modules.md
```diff
@@ -0,0 +1,112 @@
...
+entry points for modules are being considered:
+* `<script type="module">`;
+* an overload to the `Worker` constructor;
+* an overload to the `importScripts` Worker API;
```
jfbastien: Are these what wasm would want?

We kind of discussed this in #84, it would be good to update that issue and close it if we have a believable solution.

Modules.md
```diff
@@ -0,0 +1,112 @@
...
+
+Additionally, an ES6 module can recursively import other modules via `import`
+statements.
+
+For WebAssembly/ES6 module integration, the idea is that all the above module
+entry points could also load WebAssembly modules simply by passing the URL of a
+WebAssembly module. The distinction of whether the module was WebAssembly or ES6
+code could be made by namespacing or by content sniffing the first bytes of the
+fetched resource (which, for WebAssembly, would be a non-ASCII&mdash;and thus
+illegal as JavaScript&mdash;[magic number](https://en.wikipedia.org/wiki/Magic_number_%28programming%29)).
```
jfbastien: That's beautiful and hacky at the same time :-)
```
💩wasm
```
:-)

Modules.md
```diff
@@ -0,0 +1,112 @@
...
+transitively used by those libraries "played well" together (e.g., explicitly
+shared `malloc` and coordinated global address ranges). Instead, the
+[dynamic linking future feature](FutureFeatures.md#dynamic-linking) is intended
+to allow *explicitly* sharing linear memory between multiple modules.
+
+## Initial state of linear memory
+
+A module will contain a section declaring the heap size (initial and maximum
+size allowed by `sbrk`) and the initial contents of memory (analogous to
+`.data`, `.rodata`, `.bss` sections in native executables).
```
jfbastien: I think we may need to be more configurable, otherwise folks may just start with "tiny initial size, infinite maximum". That's a problem because wasm is in-process and fragmentation will be an issue.

Could this value also be configured at load time (e.g. "load a module, and give it X memory because I know that in this embedding I'll use more than usual"). That's interesting for reusable modules which sometimes use little memory, and sometimes a lot of it.

Modules.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Code section
+
+The WebAssembly spec defines the code section of a module in terms of an
+[Abstract Syntax Tree](AstSemantics.md) (AST). Additionally, the spec defines
+two concrete representations of the AST: a [binary format](BinaryEncoding.md)
+which is natively decoded by the browser and a [text format](TextFormat.md)
+which is intended to be read and written by humans. This design separates
+the concerns of specifying and reasoning about behavior, over-the-wire size
+and compilation speed, and ergonomic syntax.
```
jfbastien: This is a bit misleading: the module's section can't contain the textual representation of the AST, it can only contain the binary one.

MVP.md
```diff
@@ -11,84 +11,12 @@ even on mobile devices, which leads to roughly the same functionality as
...
-  statically linked libraries will be specific to your source language compiler.
-  In the future, [standard ABIs may be defined](FutureFeatures.md#dynamic-linking)
-  to allow for compatibility between compilers and versions of compilers.
-* **TODO**: there is more to discuss here concerning APIs.
-
-## Module structure
-
-* At the top level, a module is ELF-like: a sequence of sections which declare
-  their type and byte-length.
- * Sections with unknown types would be skipped without error.
```
jfbastien: We should keep this property.

MVP.md
```diff
@@ -11,84 +11,12 @@ even on mobile devices, which leads to roughly the same functionality as
...
-  quick startup times. Towards that goal, the binary format will be natively
-  decoded by browsers.
-* The binary format has an equivalent and isomorphic
-  [text format](MVP.md#text-format).  Conversion from one format to the other is
-  both straightforward and causes no loss of information in either direction.
-
-## Text format
-
-The [text format](TextFormat.md) provides readability to developers, and is
-isomorphic to the [binary format](BinaryEncoding.md).
```
jfbastien: I put a note below (in Modules.md): I think text format and binary format should stay in MVP.md.

Modules.md
ajklein: As previously discussed offline, this needs some work on the ES Loader side to ensure that it can be specified in terms of whatever hooks end up in Loader "level 1", rather than requiring a dependency from the ES Loader on WebAssembly.

Modules.md
ajklein: As I mention below, using an ES module loading mechanism to load wasm is something I don't see fitting into a level 0 Loader spec, so I think it still might be prudent for wasm to offer some more explicit hooks for loading from HTML (such as those discussed in #84). That really depends on the timelines for various parts of this: loading wasm code, importing ES from wasm, importing wasm from ES.

MVP.md
lukewagner: I'm happy to, I was just thinking that this was originally added when V1.md had a bunch of text and so now it seems a bit ad hoc and asymmetric since most files don't have it.  The important thing is the Readme.md has it bold.  Still want to keep?

MVP.md
lukewagner: Oops, yes, good catch.

MVP.md
lukewagner: Oops, I had meant to add a link to TextFormat.md to the list of links above (below BinaryEncoding.md).  It seems like these sections weren't adding much given their restatement in HighLevelGoals.md and Modules.md.  However MVP.md needs to be redone anyway (now that it's almost empty) to have a purpose, so I can leave it until that PR if you want.

Modules.md
lukewagner: Sounds good.

Modules.md
lukewagner: So in a module you can effectively have `import foo from bar with signature int (double)`.  This gives `foo` a local signature which is used for type checking calls to `foo`.  From previous discussions, when the imported module is wasm, I think we want the signatures to have to match exactly (no magic happens at the boundary).  But if the imported module is ES6, anything goes.

Modules.md
lukewagner: Outside of wasm (the new "loader level 0" we've been talking about).  When there is a URL, we can patch it in here.

Modules.md
kg: I think for anything beyond simple scenarios, there will end up being something along the lines of embind or p/invoke that acts as a shim between wasm modules and ES6 modules and does marshaling, type checking, etc. I assume loaders can cover these cases.

Modules.md
lukewagner: IIUC, this is a spec-factoring problem, not a fundamental implementation problem, is that right?  That is, if these were all the same spec, I don't think there would be an issue.  Assuming that's right, then I expect we could define some host hooks where the ES6 spec says "if the first byte isn't ASCII, then an error is raised unless the host environment has a separate way to decode this file".

Modules.md
lukewagner: @jfbastien Ideally yes, in my mind.  That would mean, e.g., it should be possible to swap out a JS module with a wasm module (or vice versa) by simply changing the contents of the file.  If we agree on this, happy to comment on #84.

@ajklein If we know what we want in "level 1", then it doesn't seem great to add a separate way to load a wasm module (unless we also wanted that separate mechanism to load ES6).  My expectation here is that we can do "level 0" first (purely in terms of ES6 modules), get that shipped, and follow on fast with wasm.

Modules.md
```diff
@@ -0,0 +1,112 @@
...
+scope of "all" is a nuanced question), a single app using multiple
+independent libraries would have to hope that all the WebAssembly modules
+transitively used by those libraries "played well" together (e.g., explicitly
+shared `malloc` and coordinated global address ranges). Instead, the
+[dynamic linking future feature](FutureFeatures.md#dynamic-linking) is intended
+to allow *explicitly* sharing linear memory between multiple modules.
+
+## Initial state of linear memory
+
+A module will contain a section declaring the heap size (initial and maximum
```
sunfishcode: Say "linear memory" instead of "heap", here and one spot below too.

Modules.md
lukewagner: I'm not quite following your proposal.  In particular, how would that be different than having the module heap-init declaration specify the smaller heap size and then having the `_start` hook dynamically decide whether to `sbrk` up to a bigger size?  More generally, I would wonder why such up-front heap sizing would be necessary (when you can always just `sbrk` on demand); the only reason I mention it here is because it's necessary to initialize the state of the heap.

Modules.md
lukewagner: Ah, good point, I'll try to clear that up.

Modules.md
ajklein: The point is that none of these are the same spec, and we want to keep it that way (so saying it's just a spec factoring problem seems overly dismissive). I'd like to make sure that ES proper need know nothing about wasm, and I'd also hope that the ES Loader wouldn't need to know about it either: whatever sits right above the ES Loader should be able to add the right hooks to the loader (conceptually, anyway).

Modules.md
lukewagner: Good catch.

Modules.md
lukewagner: I didn't mean to be dismissive but if we can agree on an intended behavior, then it seems like it's then becomes a matter of spec engineering to decide how to cut up the spec so that we don't have unintended dependencies.  E.g., there is *definitely* going to be an HTML5 portion of the whole specified loader pipeline so it seems like that is where we could mention both JS and wasm.

Modules.md
jfbastien: Could you clarify the TODO, so we know what link to add once there's something to link to?

lukewagner:
@jfbastien Which line are you referring to?

jfbastien:
Hmm, it looks like commenting on an incremental diff doesn't link well :-s
Chasing links:
https://github.com/WebAssembly/design/commit/17d32b8fc2cf12fa16a6d2dd8db4be07a832d1b3#commitcomment-12235834

lukewagner:
Updated to address feedback.

jfbastien:
This mostly lgtm, but I'd like @ncbray's concerns to be captured somewhere: I think there are still open issues as @ncbray pointed out. Overall this clarifies things, so I like it, but I don't want to paper over issues!

lukewagner:
I thought that I had; are there any outstanding issues not addressed in the current version?

lukewagner:
@ncbray Rebased to trunk

ncbray:
This is the only way I found to view a complete diff:
https://github.com/WebAssembly/design/compare/better-explain-modules
but it isn’t linked to the PR.  Is there a better way?  Out-of-line comments, because of that.

In general I think the consolidation, editing, and clarification is good.  The new ideas should probably be discussed with a wider audience?

At first glance, WASM importing WASM being type checked is potentially contentious/complicated.  It is a weird corner case, normally imports are “whatever the embedder wants” and this burns a hole that seems tricky to define.  There are also crazier things I’ve been thinking about, such as allowing a signed/unsigned distinction in imports so values could be fixed up before passing them to JS.  How does that interact?  It does seem reasonable that the implementation could type check imports to some degree (WASM or otherwise) but ironing out the details... hmmm. Prototypes would be nice.  If checking imports is complicated we could always drop it on the floor.

I have gotten the impression that content sniffing is considered harmful by a lot of web folks.  (Implementation complexity and security issues arising from content confusion.)  Polymorphic behavior based on URL content (JS or WASM) may be problematic unless we can require correct MIME types from the server 100% of the time.  I’ll talk to the web folks on this side of the fence who have voiced concerns before.  Content sniffing may also interact strangely with the “format layering” idea we’re kicking around for file format.

It will always possible to wrap a WASM instance in a ES6 module (manually or autogen) to prototype all of this without needed to change the loader.  Does ES6 module integration need to be part of MVP?  (I don't have strong feelings, it just noticed it seems separable.)

Nits / related comments:
* Modules may include a relocation table, depending on how function pointers and/or dynamic linking are implemented.  Not critical, but if we’re listing what’s contained in a module...
* We may not want to support sbrk.  I am considering scenarios where the system may want to steal memory from the program (automatically marshaling calls to the main thread, for example) and the only way to do this is forgetting sbrk and using a mmap-like interface.  Currently I’d rate this idea as “problematic” and it may not see the light of day, but FYI.
* General comment for everyone, no need for you to address it in this PR: there’s a lot of abstract thinking going on, concrete examples / psedocode would help make things clearer.

jfbastien:
@ncbray: for `sbrk` see #285.

lukewagner:
> This is the only way I found to view a complete diff:
> https://github.com/WebAssembly/design/compare/better-explain-modules
> but it isn’t linked to the PR

I've been naively assuming that this is what "Files Changed" is showing.  Anyhow, for reasons given in #282, out-of-line comments are preferable for non-nit comments.

> At first glance, WASM importing WASM being type checked is potentially contentious/complicated.

Thus far, people have been relieved to hear this proposed (e.g., @flagxor, iirc) and while it may seem asymmetric at first glance, if you consider that an import is probing the host environment not just for a name, but for a (name, signature) pair (and faulting when not found), then all this amounts to is that, for a JS import, the signature is ignored in the probe (b/c there is no concept of signature in JS) while for a wasm import, there is a signature and so a "type mismatch" is really just a failed lookup.  I would expect builtin imports to work the same way (less special cases to implement), so that actually makes JS the odd one out.  Do you think it'd be more clear to describe this design this way?

> There are also crazier things I’ve been thinking about, such as allowing a signed/unsigned distinction 
> in imports so values could be fixed up before passing them to JS. How does that interact?

Great idea!  I think it could be done independent of JS by saying that the set of types available to import/export signature declarations additionally include `sint32`/`uint32`.  From wasm's pov, these are both just `int32`s, but the extra sign information could be used by whatever language was on the other side of the interface.  Think it's worth mentioning in this PR?

> I have gotten the impression that content sniffing is considered harmful by a lot of web folks.

I have too, but I think a lot of this is from situations that are 100x harrier (e.g., character encoding, quirks mode and images) because the sniffing only came about after the fact.  However, in our case, we're starting with a pretty strict input requirement (invalid characters trigger hard syntax errors in JS) and a pretty clear a priori break which we can test for web compat extensively before deploying.  So while I'm definitely interested to hear specific concerns, I'm optimistic that our case is easier than those before.

>  Does ES6 module integration need to be part of MVP? (I don't have strong feelings, it just noticed it 
> seems separable.)

One way or another, we have to define how a wasm module gets loaded and how it looks to JS.  If we don't integrate with ES6 modules, we'll have to design something separate which I think will have worse ergonomics and lead to duplication (spec and impl) and accidental incongruencies.  If ES6 modules were imposing significant costs on the wasm design, then we'd have to weigh the pros/cons, but so far, and as I've tried to demonstrate in this PR, I'm not seeing any warts imposed on wasm due to ES6 modules atm.

> Modules may include a relocation table, depending on how function pointers and/or dynamic linking
> are implemented. Not critical, but if we’re listing what’s contained in a module...

By representing references symbolically (via an index into a module-local table of functions/globals/imports) instead of with hard-coded addresses, I'm not sure what extra information is needed to do relocations when dynamic linking.  Do you have anything in mind?

> General comment for everyone, no need for you to address it in this PR: there’s a lot of abstract 
> thinking going on, concrete examples / psedocode would help make things clearer.

Agreed; definitely looking forward to the next steps of giving a more formal/executable definition to all of this.  For now, though, I think it's valuable to have this in place to address a lot of the immediate confusion.

qwertie:
@ncbray What about generalizing that idea of signed/unsigned distinction to a general type, like ... "pointer to tuple (int32, sint)" or any little piece of metadata? Wasm doesn't necessarily have to define a metadata format, and could just offer a place to put it, with optional binary matching on import - e.g. if the export says `Square("sint32" int32 x)` where `"sint32"` is the metadata, an import could require `Square("sint32" int32)` or just `Square(int32)` if matching the metadata is not needed / supported. Could be stored as an index into a string table.

> there’s a lot of abstract thinking going on, concrete examples / psedocode would help make things clearer.

Agreed!

lukewagner:
Alright, have lgtm from @jfbastien and I don't see any specific outstanding objections to merging.  Happy to iterate further in future PRs.

PR #271, opened 2015-07-17
Move handling of subnormals into FutureFeatures.

sunfishcode:
With #243 (the subnormal_mode proposal) closed, this PR instead addresses #148 by moving subnormal flushing into FutureFeatures.md.

In the public discussion in #148, there is consensus for supporting subnormals by default in the MVP. And, we have abundant of data confirming that this is practical, since asm.js has subnormals enabled and does not even have any way to disable them.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -401,8 +401,6 @@ Floating point arithmetic follows the IEEE-754 standard, except that:
  - WebAssembly uses the round-to-nearest ties-to-even rounding attribute, except
    where otherwise specified. Non-default directed rounding attributes are not
    supported.
- - The strategy for gradual underflow (subnormals) is
-   [under discussion](https://github.com/WebAssembly/design/issues/148).
 
 In the future, these limitations may be lifted, enabling
 [full IEEE-754 support](FutureFeatures.md#full-ieee-754-conformance).
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -337,6 +337,14 @@ enabled only from developer tools, that would enable traps on selected floating
 point exceptions, however care should be taken, since not all floating point
 exceptions indicate bugs.
 
+## Flushing Subnormal Values to Zero
+
+Many popular CPUs have significant stalls when processing subnormal values,
+and support modes where subnormal values are flushed to zero which avoid
+these stalls. And, ARMv7 NEON has no support for subnormal values and always
+flushes them. A mode where floating point computations have subnormals flushed
+to zero in WebAssembly would address these two issues.
+
 ## Integer Overflow Detection
 
 There are two different use cases here, one where the application wishes to


jfbastien:
I'd still like to leave this as "under discussion", since making up our mind on what the default is will be pretty easy.

sunfishcode:
What are we waiting for?

jfbastien:
Data from real implementations. It's easy to figure out in the future, and harder / misleading to put in the FAQ as-is and then change later. I'm OK with the FAQ explaining this, but right now it's phrased as pretty definitive (FTZ is a future feature, denormals are in MVP) and I've had a few folks tell me they didn't like that conclusion (and would rather have the opposite). I think #148 has pretty good points but not consensus, so IMO the FAQ should reflect that.

titzer:
I think we should move the FTZ feature to FutureFeatures, reasoning thusly:

1.) it's a performance feature
2.) it's easy to add along the lines discussed, provided data appears that justifies it, but can't really be taken away without leaving a wart
3.) MVP should strive for minimal completeness

titzer:
Also 

4.) Most implementations will probably ignore FTZ at first

jfbastien:
To be clear: FTZ isn't necessarily a "feature" that one turns on or off. It can be the default. Denormal support can be the future feature that one turns on or off, than can get ignored, that breaks minimal completeness, and that can't be removed without leaving a wart.

This is why I want to choose what the default is later, with data. It's easy to set this default, and deciding now seems premature.

titzer:
There are lot of good reasons to make IEEE 754 (which requires subnormal
support) the default, as been discussed previously. All hardware
implementations that I am aware of that initially cut corners eventually
tend toward IEEE 754 compliance. That makes sense given that it's a widely
accepted standard. Under that assumption, FTZ or other deviations only make
sense if they improve performance above some threshold, and that needs to
be justified with data.

lukewagner:
+1

sunfishcode:
I agree with @jfbastien that data is important. We have just performed an extensive multi-year study into the performance of real-world apps running in the real world on real-world browsers with IEEE-conforming subnormal handling enabled. The data shows that:
 - The vast majority of applications see no significant performance impact due to subnormal handling.
 - Widely-used benchmark suites don't register any performance difference whether subnormal handling is enabled or disabled.
 - Participating game developers (reputed to be among the most demanding on this topic) have not reported subnormal handling as a significant concern in practice.
 - A few specific application domains, such as audio processing, regularly see an impact due to subnormal handling.

With respect to the concerns raised, it is possible that subnormal handling is merely being overshadowed by the lack of threading and SIMD in our experiment. For WebAssembly, since threads and SIMD are already planned to be postponed until after the MVP, we can expect to see the same overshadowing if that is the case.

With respect to those applications domains that are regularly affected, it is unfortunate that we can't provide all features desired by all application domains in the MVP, but this is consistent with the nature of the MVP. We're purposefully keeping the scope restricted in many areas, including many that are far more well known than subnormal handling, to keep the MVP focused.

Implementors opinions are also important. People from [several](https://github.com/WebAssembly/design/pull/271#issuecomment-124192930) [major](https://github.com/WebAssembly/design/pull/271#issuecomment-124390034) [browser](https://github.com/WebAssembly/design/issues/148#issuecomment-121444727) [engines](https://github.com/WebAssembly/design/issues/148#issuecomment-120761040) have spoken in favor making IEEE-754 behavior the default and addressing FTZ+DAZ in a future feature.

It may also be noted that the [proposal to disable subnormal handling on esdiscuss](https://esdiscuss.org/topic/float-denormal-issue-in-javascript-processor-node-in-web-audio-api#content-3) failed to gain traction. This doesn't mean that it was a bad idea, but it does help confirm that this feature isn't actually a major concern in the big picture, so deferring addressing it until after the MVP is reasonable.

The choice for the MVP is clear.

titzer:
Are we OK merging this?

lukewagner:
lgtm

Issue #104, opened 2015-06-02
What are Web Assembly threads?


jfbastien:
Are Web Assembly specified in terms of [WebWorkers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers), or are they different? What are the differences, or are WebWorkers just an implementation detail?

Maybe the polyfill is OK with WebWorkers, and Web Assembly does its own thing that's closer to pthreads.

We need to ensure that Web Assembly can work outside of a browser.

trevnorris:
If I may offer a few thoughts from the server-side of things (specifically node). There are three main use cases for threads (there are probably "proper" names for these, but you'll get the idea):

- Computationally heavy native work. These have no JS stack. Values are passed, a callback is set and when the numbers have been crunched the callback receives the return value.
- Immediately exiting thread. A single file is passed, along with optional arguments, which runs to completion, during which time return values can be messaged back to the parent thread. This one has two variations.
  - No I/O access thread. The spawned thread is essentially synchronous, but to further clamp down on the scope of work these threads can do they are not allowed to do any sync I/O operations either.
  - I/O thread. Allowed to require other modules and perform sync I/O.
- Event loop thread. These are essentially just like a running process. Except in the way they communicate, and have the ability to share memory.

These have variations on how developers want them implemented. Like does the immediately exiting thread create a new JS stack every time, or use a new one. Is the thread closed when the operation is complete or reused. And also is a thread, sync or async, joined causing the main thread to hault until the spawned thread is complete.

Sorry if this was too many examples. What I'm getting at is it seems like wasm's ability to work with threads will either not be low level enough or extensive enough to fit these needs (I don't expect it could concerning the first example). Leaving server applications needing to use their own bindings. This correct?

sunfishcode:
I don't have all the answers, but one thing I can comment on is that WebAssembly threads are going to have access to a pthreads-level API, so applications will have quite a lot of control. Decisions like when and how to use pthread_join are largely determined by the application (or by a library linked into the application).

Also, WebAssembly itself is being designed to be independent of JS, and it will be possible to have WebAssembly threads with no JS code on the stack.

trevnorris:
Thank you. That's excellent to hear.

sunfishcode:
What things here do we need to decide on for the MVP, and what things can wait until we actually introduce shared memory?

lukewagner:
[...]

PR #275, opened 2015-07-21
Fill in a more detailed sketch for GC/DOM/Web API integration

lukewagner:
Over the last few weeks, I had several chances to talk to Mozilla DOM/WebIDL folks about how WebAssembly/Web API integration could work.  This PR is a sketch of the basic plan we discussed which breaks down the overall goal of "GC/DOM integration" into two WebAssembly features (independent of JS/Web) and then how these features would integrate with the Web platform.  Definitely interested in feedback, esp. from more people with Web API experience.


FutureFeatures.md
```diff
@@ -66,16 +66,160 @@ Options under consideration:
...
+There is a lot of the design left to
+consider for this feature, but a few points of tentative agreement are:
+* To avoid baking in a single language's object model, define low-level GC
+  primitives (viz., structs and arrays) and allow the source language compiler
+  to build up features like virtual dispatch and access control.
+* GC struct and array types would have associated *struct/array reference
+  types* that were similar to and symmetric with 
+  [opaque reference types](FutureFeatures.md#opaque-reference-types)
+  (just not opaque).
+* The GC heap would be semantically distinct from linear memory and thus
```
kripken: I'm curious if there is precedent for this model, of having both a linear memory and a GC heap, in the same VM. I can't think of one offhand.

lukewagner: C++/CLI :)

kripken: I see, yes. IIUC, in C++/CLI, one can allocate e.g. C# structs on the regular C++ stack, which seems quite different than for us. Although I suppose that might have been mostly a syntactic sugar issue, and the compiler did actually manage two separate stacks, and so we are closer to the output of the C++/CLI compiler than its input.

kg: IIRC .NET maintains a single native stack, which is safe-enough because the only way to get the address of something on the stack (or read/write directly from addresses) is to drop into ```unsafe``` mode. Anyone using pointers in C# is interacting with this too, even though C++/CLI's access is more powerful.

.NET structs allocated on the GC heap either live in a box or inside a reference type, so they're never actually allocated *directly* on the GC heap. They are effectively native values, just like in C. They do have the ability to hold object references, though, so in my experience the runtime rejects any attempt to store any struct containing GC references into the native heap - only allowed to put it on the stack or in the GC heap.

I think their model works pretty well, but we have more stringent safety/robustness requirements, and we can't protect the stack as easily as they can since raw memory access is a baseline for us.

FutureFeatures.md
```diff
@@ -66,16 +66,160 @@ Options under consideration:
...
+  the fields of GC objects could safely hold reference types (unlike linear
+  memory). GC object fields could also hold 
+  [function pointers](AstSemantics.md#calls) directly (again, unlike
```
kg: To be clear: When you say they can hold function pointers directly... how would this work from wasm and JS? Does wasm store a safe wasm function into a GC object field, and the runtime converts it to a raw function pointer? Does JS get a JS Function instance when it reads from that field?

I'm guessing maybe this isn't thought through yet and we just want to call out that this can allow us to build efficient vtables, but it's not clear to me whether this is a fully speculative bit of prose or not.

FutureFeatures.md
```diff
@@ -66,16 +66,160 @@ Options under consideration:
...
+`js` module providing:
+* an exported `string` opaque reference type and exported functions
+  to allocate, query length, and index `string` values;
+* an exported `object` opaque reference type and exported functions
+  that correspond with the ES5 meta-object protocol;
+* an exported `value` opaque reference type with exported functions for
+  constructing `value`s from integers, floats, `object`s, and `string`s and
+  with exported functions for querying the type of a `value` and extracting the
+  abovementioned payload types.
+
```
kg: I suspect JS ```Function``` will need to be included in the options here, at least at a basic level. People will need it for interacting with many DOM APIs and for interacting with existing third-party libraries.

FutureFeatures.md
```diff
@@ -66,16 +66,160 @@ Options under consideration:
...
+The first sub-feature is to extend [module imports](Modules.md#imports-and-exports)
+to allow modules to import *opaque reference types*. "Opaque" means that the
+reference type itself has no structural content and does not, e.g., define any
+methods or fields. Once imported, an opaque reference type can be used in the
+signature of other imported functions. Thus, the point of an opaque reference
+type is to be passed to and returned from exported functions.
+
+Reference types are allowed to be used as the types of locals, parameters
+and return types. Additionally, references would be allowed as operands to
+operators that treat their values as black boxes (`conditional`, `comma`, maybe
```
kg: ```eq``` will be an absolute necessity, I think. So we can probably remove the 'maybe' here.

FutureFeatures.md
lukewagner: IIUC (not a C# expert), C# structs are value types and so you cannot point at them with a reference.  Thus, to compile a C# struct to wasm, we could decompose the struct into scalar locals such that all field references were just local accesses.  (I guess array-typed fields of structs would require the heap-stack unless we extended wasm with fixed-size array local types... maybe not a bad idea :)

lukewagner: Oops, @kg beat me to it.

FutureFeatures.md
```diff
@@ -66,16 +66,160 @@ Options under consideration:
...
+Whether the cast succeeds is up to the host environment; WebAssembly itself
+will define no a priori subtyping relationship.
+
+For reasons of safety and limiting nondeterminism, imported opaque reference
+types would not be able to be loaded from or stored to linear memory where they
+could otherwise be arbitrarily aliased as integers. Instead, a new set of
+operations would be added for allocating, deallocating, loading and storing
+from integer-indexed cells that could hold references and were not aliasable by
+linear memory. There are several important alternatives to consider:
+* LIFO allocation of cells versus (or in addition to) ad hoc explicit
```
kg: Are cells more like heap or more like stack slots for GC objects? It's unclear to me whether they're one or both at once. I can imagine use cases where a function just wants to have N reserved GC cells like it has M reserved int32 locals, but also cases where you want to allocate some GC cells and hold on to them indefinitely. Are those both considered cells here, and we just have some lifetime distinction?

I think we need both stack cells and long-lived cells, and they need to have distinct access opcodes. Without that it'll be pretty easy to end up accessing a stack cell after its lifetime has expired. That kinda sucks.

For simplicity's sake, stack cells could be ruled out and they could always be long-lived and manually allocated/deallocated. That's probably the right place to go with the MVP.

FutureFeatures.md
lukewagner: It's all opaque, so it's up to the impl, but the idea is that an impl would be able to just store the function pointer and, when accessed from JS, that field type would produce a JS function that represented that wasm function (similar to wasm module exports viewed from JS).  There probably are some more details to work out here, though, and for function pointers in general.

FutureFeatures.md
lukewagner: Well, technically JS functions are just objects :), but your basic point stands that I should explicitly call out functions too.

FutureFeatures.md
kg: I think we need a better way to phrase this, but I'm having trouble coming up with something...

I don't think you ever want the canonical representation in the heap to be a function pointer - you need type information, etc. But you could definitely have a representation that is (function type info, function pointer) and mechanisms for calling into a stored fp with an expected signature. This would mean a type check, but it would predict well.

If the GC object fields store a function of a *specific signature*, we can enforce that signature all the way through the system and that should reduce an invocation into a raw function pointer call because we don't need to type-check at call time, just at write time.

lukewagner: If this is more confusing than it's worth, though, I can remove.

FutureFeatures.md
lukewagner: You're right that there are use cases for stack-y and ad hoc cells which is what I wanted to express in this bullet: there is the core feature: a cell that can hold a reference and be indexed by an integer, and a separate issue is their lifetime and how we allocate them.  Do you think this could be clarified some more?

kg: I think having the allocation stuff mentioned here implies some things about cell lifetime and makes one assume that they are heap allocated. Maybe introduce the concept of a 'cell' but don't talk about its lifetime or reservation policy.

FutureFeatures.md
kg: Well, according to ```typeof``` JS functions are not objects. :-) They do behave like objects other than being callable, though. When I see 'string' and 'object' listed but not 'function' I assume functions can't be stored as opaque references, because it's almost-mirroring-but-not-quite the JS type system. There will be more typeof enumerants once value types and typed objects land, I think?

It's probably sufficient to clarify this by stating what the three opaque reference types are, somehow:
```string``` stores JS string values (not ```String``` instances, which are objects)
```object``` stores JS objects (including the ```String``` wrapper), but not values
```value``` stores anything that can be stored in a JS variable (```string```, ```number```, ```object```...)

FutureFeatures.md
qwertie: FYI about .NET value types (C# structs):

- Normally C# structs are located on the stack, or 'inline' inside other objects (reference types). But you can also 'box' a value type and then it will be located "directly" on the heap (with a 2-word header, like a reference type). IIRC C++/CLI (but not C#) allows you to interact with boxed value types.
- You can make references directly to C# structs (located either on the stack or on the heap). Such references can only exist on the stack. Such a "managed pointer" is used as the 'this' pointer when you call a method on a value type, as well as when you pass a value type by reference. C# only allows such references to be created as part of a method call, but CIL bytecode lets you create them at any time.

FutureFeatures.md
kg: It just occurred to me that 'JS objects' is still ambiguous re: ```Function``` instances, so you may need something like 'object-like' or just 'objects or functions'.

Then to complicate things there's ```Proxy```, which according to ```typeof``` is either an ```object``` or a ```function``` depending on how you construct it...

FutureFeatures.md
lukewagner: Ah, so in the case of struct methods or by-reference passing, C# structs would need to go on the heap-stack unless wasm provided a similar sort of feature.

FutureFeatures.md
lukewagner: Yes, I was thinking that function pointer types would include the signature (so they could just be a raw pointer value).  I'm thinking this opens up a separate discussion and so I'll just strip this bit of text since it's not necessary to make the overall point.

FutureFeatures.md
lukewagner: FWIW, functions are literally defined as "ordinary objects" [[1](http://people.mozilla.org/~jorendorff/es6-draft.html#sec-ecmascript-function-objects)]; `typeof` is just overloaded for objects that define the `[[Call]]` internal property [[2](http://people.mozilla.org/~jorendorff/es6-draft.html#sec-typeof-operator-runtime-semantics-evaluation)].

Anyhow, lmk if updates make sense.

lukewagner:
@davemichael I think weak handles should definitely be considered for both WebAssembly and JS and there is long-standing demand from JS.  It's easy to miss, but this is explicitly mentioned as an alternative to consider in the "Opaque reference types" section by the "Cells that hold a weak reference." bullet.  But this could be generalized so that not just cells, but also fields of GC objects could be weak.

It's great to see your GDoc, iiuc, coming to a similar conclusion as what is proposed here.

For the last question, I have a [proposal](https://github.com/WebAssembly/design/issues/104#issuecomment-108631787) buried in #104 for a hybrid model that, I believe, offers the best of both worlds.  From the perspective of WebIDL bindings, though, you can identify threads with workers which really simplifies the story for the rest of the browser.

davemichael:
Thanks! Overall, I really like this proposal, thanks for writing it and working on it.

Re: the proposal about the thread model... (i.e., what threads can use what opaque handles and what APIs). In particular...
"What happens when a non-main thread calls an import that (in a browser) is JS? (This whole idea from @ncbray) The calling thread would block and the call would be executed on the main thread of the module"
This is interesting. I'm assuming the developer has access to this main thread of the module (it's not hidden; they can schedule timeouts there, etc), and can make calls from there? This bears some similarity to how Pepper handled threads. Virtually all APIs could be used on all threads. The main thread was "special" in that it needed to not hang so that calls could come and go appropriately. For that reason, we had this model of completion callbacks:
https://code.google.com/p/chromium/codesearch#chromium/src/ppapi/c/pp_completion_callback.h&l=93
...that allowed you to make calls in a "blocking" way, but _only_ on non-main threads. Another useful option was "OPTIONAL", which matters for things like IO. If there's data in the buffer, you can deal with it immediately instead of having to pop the stack and handle a callback later. Not sure if a similar concept would be useful in the bindings for WebAssembly.

Is the main thread of the module (for cases where there is a JS context) actually going to run on the same thread that the JS context lives on? It seems like it should, to avoid any extra overhead when interacting with JS. It would obviously always be an "event loop" type of wasm thread.

Sorry if I'm asking stupid questions addressed elsewhere ;-). If there's any chance our battle scars with Pepper matter here, I wanted to offer some info on what worked for us and what didn't. I think ncbray knows enough about Pepper, though, that he's probably largely filled you in.

lukewagner:
The completion callback idea makes sense and seems similar to what we'll need to do for [callback function types](http://www.w3.org/TR/WebIDL/#idl-callback-function).  The "OPTIONAL" strategy sounds useful; I think to add that we'd need to add it to the Web API itself, not at the wasm/WebIDL-integration level.

> Is the main thread of the module (for cases where there is a JS context) actually going to run on the
> same thread that the JS context lives on?

Yes, and synchronous calls to/from JS/Web APIs would share the same (trusted) stack.

> Sorry if I'm asking stupid questions addressed elsewhere ;-). If there's any chance our battle 
> scars with Pepper matter here, I wanted to offer some info on what worked for us and what didn't.

Happy to have all these constructive questions and learn from Pepper experience.  I realize that there are some things still lacking, viz., being able to directly share a resource between threads without doing this blocking-cross-thread communication.  My expectation here is that we should have a v.1 which makes minimal asks of the rest of the browser and then build on that by proposing new Web APIs (perhaps a MutableFile that was directly shareable between threads).

lukewagner:
(rebased) Any more comments to address before merging or objections?

jfbastien:
I defer to @ncbray and @davemichael's wisdom in this area.

titzer:
I think the discussion of cells for holding opaque reference types is too
detailed and that we should preserve the mystery about how exactly opaque
reference types are mapped to integers and back for storage into the linear
memory. Therefore I would advocate dropping that language about cells.

Let's avoid being too detailed in Future Features itself. Maybe split off a
separate document and file some issues for discussion?

lukewagner:
@titzer Done

titzer:
lgtm

lukewagner:
Merging, based on lgtm from @titzer and positive comments from @davemichael.  Happy to iterate further in future PRs, of course.

PR #276, opened 2015-07-21
Iterate on the integer operators.

sunfishcode:
Clarify that signed integer operations don't have silent overflow,
clarify the nature of signed division and remainder, mention that
clz/ctz/popcnt are sign-agnostic, and reorganize the text a little.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -335,18 +335,22 @@ expression-tree-building opportunities.
 
 ## 32-bit Integer operations
 
-Many operations available on 32-bit integers are sign-agnostic. Signed
-integers are always represented as two's complement and arithmetic that
-overflows conforms to the standard wrap-around semantics. All comparison
-operations yield 32-bit integer results with `1` representing `true` and `0`
-representing `false`.
+Integer operations are signed, unsigned, or sign-agnostic. Signed operations
+use two's complement signed integer representation.
+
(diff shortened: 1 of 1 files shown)


AstSemantics.md
```diff
@@ -361,21 +365,18 @@ representing `false`.
...
+  * `int32.ctz`: sign-agnostic count trailing zeroes (defined for all values, including zero)
+  * `int32.popcnt`: sign-agnostic count number of ones
 
 Shifts interpret their shift count operand as an unsigned value. When the shift
 count is at least the bitwidth of the shift, `shl` and `shr` produce `0`, and
 `sar` produces `0` if the value being shifted is non-negative, and `-1`
 otherwise.
 
+All comparison operations yield 32-bit integer results with `1` representing
+`true` and `0` representing `false`.
```
jfbastien: Do 64-bit comparisons return 64-bit integer results?
What about FP?

I kind of like having non-storable i1 types instead. The only thing you can do with them is bitwise operations, and extend/truncate from/to them.

sunfishcode: This is pre-existing text that's just getting moved here. Incidentally, I am also contemplating proposing a non-storable bool type :-).

sunfishcode:
Merging based on lgtm and because this is just refining and reformatting existing ideas.

PR #277, opened 2015-07-22
Conversions have nondeterministic NaN bits too.

sunfishcode:
An earlier rule tried to limit the nondeterminism of NaN bits. This was found to be untenable and was loosened for floating point operations. This pull request does for for conversion operations as well.


jfbastien:
lgtm

sunfishcode:
Merging, based on lgtm here, and because I don't think this patch is particularly controversial. This patch just makes conversions consistent with other floating point operators, so if anyone doesn't like how they work in general, we should open a new issue.

Issue #89, opened 2015-05-29
Indirect calls: how do they work?


jfbastien:
The current [AST semantics document](https://github.com/WebAssembly/spec/blob/master/AstSemantics.md) states:

> Indirect calls may be made to a value of function-pointer type. A function- pointer value may be obtained for a given function as specified by its index in the function table.
>
> * CallIndirect - call function indirectly
> * AddressOf - obtain a function pointer value for a given function
> Function-pointer values are comparable for equality and the AddressOf operator is monomorphic. Function-pointer values can be explicitly coerced to and from integers (which, in particular, is necessary when loading/storing to the heap since the heap only provides integer types). For security and safety reasons, the integer value of a coerced function-pointer value is an abstract index and does not reveal the actual machine code address of the target function.
>
> In v.1 function pointer values are local to a single module. The dynamic linking feature is necessary for two modules to pass function pointers back and forth.

IIUC this basically is what Emscripten does.

I'd like us to discuss this a bit more to make sure we consider alternatives before choosing a specific approach:
* Are different implementations of Web Assembly allowed to return different abstract integers for a function pointer?
* What's the performance cost?
* Does this have caveats with C++ UB on function pointers that happen to work on most implementations?
* Does this support C++ pointer to member function sufficiently?
* How does this interact with dynamic linking, late binding, and relocations?
* Is this sufficient for non-C++ languages?
  * Does ObjectiveC work properly / efficiently?
  * Multimethods?
* Can sanitizers (such as [control-flow sanitizer][]) be implemented efficiently (without Web Assembly runtime involvement)?
* Can the Web Assembly implementation use a sandboxing approach that doesn't rely on a language VM for security?
  * Can this target [NaCl][] efficiently?
  * can this target [MinSFI][] efficiently?
  * Are we hindering future sandboxing research?

Anything else?

  [control-flow sanitizer]: http://clang.llvm.org/docs/ControlFlowIntegrity.html
  [NaCl]: gonacl.com
  [MinSFI]: https://groups.google.com/forum/#!msg/native-client-dev/xs4khSwhAso/hk1hRgGREQ4J

sunfishcode:
#278 is a pull request which is my attempt to capture the present conclusions of this conversation.

binji:
This has been resolved by #392.

PR #278, opened 2015-07-22
Flesh out calls, indirect calls, and function pointers.

sunfishcode:
The function pointer type and its operations (conversions, comparisons)
were previously only implied. This patch documents them like regular
types and operations.

It also lays the groundwork for a future with multiple-return-value
functions. This is done by eliminating the `void` type and defining
return types as a sequence. In the MVP, the sequence can have at most
one value.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -34,9 +34,9 @@ a trap occurs.
 
 ## Types
 
-### Local Types
+### Basic Types
 
-The following types are called the *local types*:
+The following types are called the *basic types*:
 
   * `int32`: 32-bit integer
   * `int64`: 64-bit integer
@@ -47,15 +47,21 @@ Note that the local types `int32` and `int64` are not inherently signed or
 unsigned. The interpretation of these types is determined by individual
 operations.
 
+Also note that there is no need for a `void` type; function signatures use
+[sequences of types](Calls.md) to describe their return values, so a `void`
+return type is represented as an empty sequence.
+
+### Local Types
+
+*Local types* are a superset of the basic types, adding the following:
+
+  * `funcid`: a function identifier for use in `call_indirect`
+
+The zero value of `funcid` is the identifier for the first function in the
+function table. (C/C++ compilers may wish to put a placeholder function at
+this point in the table to implement a null pointer concept.)
+
 Parameters and local variables use local types.
-
-### Expression Types
-
-*Expression types* include all the local types, and also:
-
-  * `void`: no value
-
-AST expression nodes use expression types.
 
 ### Memory Types
 
@@ -293,39 +299,53 @@ may be added in the future.
 
 ## Calls
 
-Direct calls to a function specify the callee by index into a function table.
+Each function has a *signature*, which consists of:
+
+  * Return types, which are a sequence of local types
+  * Argument types, which are a sequence of local types
+
+Note that WebAssembly itself does not support variable-length argument lists
+(aka varargs). C and C++ compilers are expected to implement this functionality
+by storing arguments in a buffer in linear memory and passing a pointer to the
+buffer.
+
+In the MVP, the length of the return types vector may only be 0 or 1. This
+restriction may be lifted in the future with the addition of support for
+[multiple return values](FutureFeatures.md#multiple-return-values).
+
+There are two forms of calls:
 
   * `call_direct`: call function directly
-
-Each function has a signature in terms of expression types, and calls must match
-the function signature
-exactly. [Imported functions](MVP.md#code-loading-and-imports) also have
-signatures and are added to the same function table and are thus also callable
-via `call_direct`.
-
-Indirect calls may be made to a value of function-pointer type. A 
-function-pointer value may be obtained for a given function as specified by its index
-in the function table.
-
   * `call_indirect`: call function indirectly
-  * `addressof`: obtain a function pointer value for a given function
 
-Function-pointer values are comparable for equality and the `addressof` operator
-is monomorphic. Function-pointer values can be explicitly coerced to and from
-integers (which, in particular, is necessary when loading/storing to memory
-since memory only provides integer types). For security and safety reasons,
-the integer value of a coerced function-pointer value is an abstract index and
-does not reveal the actual machine code address of the target function.
+Direct calls identify their function statically. Indirect calls have a
+`funcid` operand which identifies the function at runtime.
 
-In the MVP, function pointer values are local to a single module. The
+Calls have a signature, which is the expected return types and argument types
+(ignoring the `funcid` operand, in the case of `call_indirect`) of the
+AST node. Call operations trap if the signature of the call differs from the
+signature of the called function.
+
+### Function pointers
+
+Function pointer values are obtained through the use of a special operator:
+
+  * `addressof`: obtain a `funcid` value for a given statically-identified function
+
+and are comparable for equality:
+
+  * `funcid.eq`: function identifier compare equal
+
+Note that it is not possible to directly observe the bits of a `funcid`
+value. They may be [converted into integers][], but the integers only hold an
+index into the *function table*, a table with an entry for each function
+appended to the table in the order that they are loaded into the program.
+
+In the MVP, `funcid` values are local to a single module. The
 [dynamic linking](FutureFeatures.md#dynamic-linking) feature is necessary for
-two modules to pass function pointers back and forth.
+two modules to pass `funcid` values back and forth.
 
-Multiple return value calls will be possible, though possibly not in the
-MVP. The details of multiple-return-value calls needs clarification. Calling a
-function that returns multiple values will likely have to be a statement that
-specifies multiple local variables to which to assign the corresponding return
-values.
+  [converted into integers]: AstSemantics.md#datatype-conversions-truncations-reinterpretations-promotions-and-demotions
 
 ## Literals
 
@@ -507,6 +527,8 @@ is NaN, and *ordered* otherwise.
   * `float64.cvt_unsigned[int32]`: convert an unsigned 32-bit integer to a 64-bit float
   * `float64.cvt_unsigned[int64]`: convert an unsigned 64-bit integer to a 64-bit float
   * `float64.reinterpret[int64]`: reinterpret the bits of a 64-bit integer as a 64-bit float
+  * `funcid.decode[int32]` : convert an unsigned 32-bit integer to a function identifier
+  * `int32.encode` : convert a function identifier to an unsigned 32-bit integer
 
 Wrapping and extension of integer values always succeed.
 Promotion and demotion of floating point values always succeed.
@@ -523,3 +545,10 @@ round-to-nearest ties-to-even rounding.
 Truncation from floating point to integer where IEEE-754 would specify an
 invalid operation exception (e.g. when the floating point value is NaN or
 outside the range which rounds to an integer in range) traps.
+
+Encoding a `funcid` returns the index into the function table. If the index of
+the function is too great to fit in the result type, encoding traps. Decoding
+returns the `funcid` from an encoded function index. If the index is out of
+bounds in the function table, decoding traps. In the MVP, `funcid` values may
+only be converted to and from 32-bit integers. Support for 64-bit funcid may be
+added in the future.
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -393,3 +393,37 @@ reason we haven't added these already is that they're not efficient for
 general-purpose use on several of today's popular hardware architectures.
 
   [handle trap specially]: FutureFeatures.md#trapping-or-non-trapping-strategies
+
+## Multiple Return Values
+
+In the MVP, functions and operators are limited to having at most one return
+value. It is desirable to lift this restriction in the future.
+
+Currently, when the result of an operator is used more than once, it must be
+stored in a local variable. Multiple return values could generalize this rule
+to require when an operator has multiple result values, the result values must
+be all stored in local variables too. This suggests a new form of `set_local`
+which can directly assign the results of a multiple-result operator into
+multiple local variables, and a new form of `return` which can return multiple
+values from a function. An advantage of this approach is avoiding the need
+for tuple values to be live anywhere.
+
+One interesting question is whether it should be possible to have a function
+call another function which returns multiple result values, and then forward
+all of those result values to its own result, without copying them all into
+local variables first. It gets even more interesting with the discussion of
+[statements versus expressions](https://github.com/WebAssembly/design/issues/223).
+
+## C++-style vtable optimizations
+
+Storing function pointers in memory requires converting them into function table
+indices. Since the traditional implementation of C++ virtual functions is to
+have vtables stored as arrays in the address space, a traditional implementation
+of C++ virtual functions in WebAssembly would work, but would require an extra
+table lookup.
+
+However, because vtable objects are not exposed at the C++ level, it isn't
+actually necessary to represent them inside the address space of the program. If
+WebAssembly allowed modules to define alternate function tables, C++ compilers
+could lower vtables into these alternate tables, so their contents could be
+actual function pointers instead of function table indices.
--- a/PostMVP.md
+++ b/PostMVP.md
@@ -38,7 +38,7 @@ variables are not aliasable and thus allow more aggressive optimization.
 Support fixed-width SIMD vectors, initially only for 128-bit wide vectors as
 demonstrated in [PNaCl's SIMD][] and [SIMD.js][].
 
-SIMD adds new primitive variable and expression types (e.g., `float32x4`) so it
+SIMD adds new local types (e.g., `float32x4`) so it
 has to be part of the core semantics. SIMD operations (e.g., `float32x4.add`)
 could be either builtin operations (no different from `int32.add`) or exports of
 a builtin SIMD module.


kg:
This looks reasonable to me from a first glance. I'll try to review it again in depth later.

FWIW, this does seem to conflict with multiple past assumptions and ideas about our expression trees.

In the past we assumed assignments have a single target with an expression on the RHS. This introduces the idea of assignments to N targets, where right now we're constraining N to 0 or 1, but eventually it will be more.
 * This implies some sort of generalization, I think? I don't really know how we want to express this. I think maybe function calls become a special case and they can only assign their result to a local? Otherwise we need a generalized 'multiple assignment' concept and that's really confusing.
 * This also implies that the previously proposed idea of inferring subexpression types based on the expression of the outer type is no longer viable.

We probably need a way to indicate that one or more of a function's return values are unused. This will be important for optimization and general codegen, I think. With return values as out-pointer parameters, this is relatively simple because you just pass a nullptr and that turns into an easily eliminated ```if (p)``` branch on a constant.

sunfishcode:
The idea with multiple return values (in FutureFeatures.md) is that they'd only be able to participate in expression trees if they have exactly one result (a mild extension of the existing rule that they can only participate in expression trees if they have exactly one user). Consequently, the only things that would have to accept multiple results would be special forms of `set_local` and `return`, and in terms of the binary encoding, these could just reference the input node as if it were a regular operand and then just implicitly wire up all the operand's values to the appropriate sinks. From the binary encoding proposals I've seen, I believe this should fit in pretty naturally.

kg:
Maybe we narrow this down and have a concept of 'most recent function return values'? So instead of (strawman):

```
local @a, @b, ...
setlocals [@a, @b] (call @fn (const 1), const (2))
call @print_two (getlocal @a), (getlocal @b)
```

something like:

```
call @fn (const 1), (const 2)
call @print_two (retval 0), (retval 1)
```

Maybe we layer some validation rules onto this, like that retvals can only be used once, and accessing a retval that isn't currently valid traps. I think those can be trivially verified because they're sequential, as long as they don't cross a branch.

Given a model like this, for a call with a single return value, you can easily decompose it:

```
call @print_one (call @sin (const 0.5))
```

->

```
call @sin (const 0.5)
call @print_one (retval 0)
```

This would potentially eliminate a lot of mostly-wasted local slots and eliminate the need for some liveness analysis on locals that are used once. That's probably good. It also makes it easier to tell whether a given return value is ever used.

sunfishcode:
As we discussed on IRC, let's move the discussion of how to efficiently encode multiple return values into a different PR, because it's an interesting but separate topic.

lukewagner:
Nice work capturing all of this.

AstSemantics.md
```diff
@@ -47,15 +47,17 @@ Note that the local types `int32` and `int64` are not inherently signed or
...
-Parameters and local variables use local types.
+### Local Types
 
-### Expression Types
+*Local types* are a superset of the basic types, adding the following:
 
-*Expression types* include all the local types, and also:
+  * `funcptr`: a function identifier for use in `call_indirect`
 
-  * `void`: no value
```
jfbastien: I'd like this document to explain why there's no void type. I'd also like to understand what a function does return if it doesn't return anything in source form.

AstSemantics.md
```diff
@@ -47,15 +47,17 @@ Note that the local types `int32` and `int64` are not inherently signed or
 unsigned. The interpretation of these types is determined by individual
 operations.
 
-Parameters and local variables use local types.
+### Local Types
 
-### Expression Types
+*Local types* are a superset of the basic types, adding the following:
```
jfbastien: Can you explain where basic and local types can be used?

jfbastien: Ah I see that you do later... Ignore :)

AstSemantics.md
```diff
@@ -273,39 +275,53 @@ nested. This guarantees that all resulting control flow graphs are well-structur
...
+Each function has a *signature*, which consists of:
 
-  * `call_direct`: call function directly
+  * Return types, which are a sequence of local types
+  * Argument types, which are a sequence of local types
+
+Note that WebAssembly itself does not support variable-length argument lists.
+C and C++ compilers are expected to implement this functionality by storing
+an array of arguments in a buffer in linear memory and passing a pointer to
+the buffer.
```
jfbastien: I'd clarify: store a struct of naturally-aligned elements to the stack, where the struct's elements are the variable argument, ordered as they were from the callee, and pass an opaque pointer to the start of this struct. What's the type of this pointer? It's technically `void*` from a C perspective, but `byte*` or `uint8*` is more sensible for WebAssembly.

Link to: https://chromium.googlesource.com/native_client/pnacl-llvm/+/master/lib/Transforms/NaCl/ExpandVarArgs.cpp
?
I think it does exactly what we want :-)

AstSemantics.md
```diff
@@ -273,39 +275,53 @@ nested. This guarantees that all resulting control flow graphs are well-structur
...
+Direct calls identify their function statically. Indirect calls have a
+funcptr operand which identifies the function at runtime.
+
+Calls have a signature, which is the expected return types and argument types
+(ignoring the funcptr operand, in the case of `call_indirect`) of the
+AST node. Call operations trap if the signature of the call differs from the
+signature of the called function.
 
-In the MVP, function pointer values are local to a single module. The
+### Function pointers
```
jfbastien: Should we also discuss pointer-to-member-function? Itanium has done weird things that I think we want to avoid?

AstSemantics.md
```diff
@@ -47,15 +47,17 @@ Note that the local types `int32` and `int64` are not inherently signed or
...
+### Local Types
 
-### Expression Types
+*Local types* are a superset of the basic types, adding the following:
 
-*Expression types* include all the local types, and also:
+  * `funcptr`: a function identifier for use in `call_indirect`
 
-  * `void`: no value
+The zero value of funcptr is the identifier for the first function in the
```
jfbastien: This implies that `funcptr` is an integer? Could you expand on this?

AstSemantics.md
```diff
@@ -273,39 +275,53 @@ nested. This guarantees that all resulting control flow graphs are well-structur
...
+(ignoring the funcptr operand, in the case of `call_indirect`) of the
+AST node. Call operations trap if the signature of the call differs from the
+signature of the called function.
 
-In the MVP, function pointer values are local to a single module. The
+### Function pointers
+
+Function pointer values are obtained through the use of a special operator:
+
+  * `addressof`: obtain a funcptr value for a given statically-identified function
```
jfbastien: Code-quote `funcptr` here and below?

AstSemantics.md
```diff
@@ -273,39 +275,53 @@ nested. This guarantees that all resulting control flow graphs are well-structur
...
+
+and are comparable for equality:
+
+  * `funcptr.eq`: funcptr compare equal
+
+Note that it is not possible to directly observe the bits of a funcptr
+value. They may be [converted into integers][], but the integers only hold an
+index into the *function table*, a table with an entry for each function
+appended to the table in the order that they are loaded into the program.
```
jfbastien: Are these deterministic?

Does this have any bearings on dynamic linking (cc @dschuff)?

AstSemantics.md
```diff
@@ -497,3 +515,10 @@ round-to-nearest ties-to-even rounding.
 Truncation from floating point to integer where IEEE-754 would specify an
 invalid operation exception (e.g. when the floating point value is NaN or
 outside the range which rounds to an integer in range) traps.
+
+Encoding a funcptr returns the index into the function table. If the index of
+the function is too great to fit in the result type, encoding traps. Decoding
+returns the funcptr from an encoded function index. If the index is out of
+bounds in the function table, decoding traps. In the MVP, funcptrs may only be
+convered to and from 32-bit integers. Support for 64-bit funcptrs may be added
+in the future.
```
jfbastien: I'd be OK with just 32. 32-bit's worth of functions should be good enough for everyone?

AstSemantics.md
kg: I think we should avoid using the name ```funcptr``` because it implies that it's a pointer into the heap, when it's actually an index into a table somewhere (and there are probably separate tables per signature type). ```funcid```? Not really interested in bikeshedding beyond that implying pointer-ness is probably bad here, even if compilers happen to use this type to *implement* function pointers.

AstSemantics.md
sunfishcode: Those are ABI concerns that don't belong in AstSemantics.md. The main goal here is just to let people know that we haven't forgotten about varargs :-). But I agree that "array" isn't quite right, so I'll fix that.

AstSemantics.md
sunfishcode: That's an ABI concern too.

AstSemantics.md
titzer: Nick and I had a discussion about function pointers today over VC. There
are several concerns here.

Add function types as local types means that verification of the bytecode
is more costly, because there are now complex function types, and
presumably function types can contain function types. That could be avoided
by using indexes and multiple function tables, where each function table
has a single signature.

Another option is to accept the cost of a dynamic type check at indirect
call sites or "decode" sites. That would mean just one function pointer
type that is dynamically checked at calls. The check could be cheap (e.g.
check a canonically-encoded signature descriptor matches the expectation).

In some sense I think we need to think about what other complex types are
going to arise in the future before we cross the barrier to function
pointer types.

sunfishcode: That's a good idea. I like funcid. I'll update the PR.

AstSemantics.md
sunfishcode: Per the discussions in #89, there's one table, not one table per signature, and:
> Call operations trap if the signature of the call differs from the signature of the called function

The order is deterministic if the application loads functions into the program in a deterministic fashion:
> appended to the table in the order that they are loaded into the program

Also, I don't expect this PR is the last word on function pointers. This is just trying to update the text to where the current discussions have lead and clean up the text to facilitate the next rounds of discussion.

AstSemantics.md
sunfishcode: There are some scenarios where the table may not be densely packed, and one could imagine running out of 32-bit indices.

AstSemantics.md
sunfishcode: @titzer @ncbray The PR here has a single non-paremterized function pointer type, and implements the "Another option" approach described in your comment here.

There are tradeoffs here, and other alternatives are possible, and further discussion is useful. This PR is meant to capture what seems like the current state of the discussion, to provide a basis for discussions going forward.

AstSemantics.md
```diff
@@ -273,39 +279,53 @@ nested. This guarantees that all resulting control flow graphs are well-structur
...
+(aka varargs). C and C++ compilers are expected to implement this functionality
+by storing arguments in a buffer in linear memory and passing a pointer to the
+buffer.
 
-Indirect calls may be made to a value of function-pointer type. A function-
-pointer value may be obtained for a given function as specified by its index
-in the function table.
+In the MVP, the length of the return types vector may only be 0 or 1. This
+restriction may be lifted in the future with the addition of support for
+[multiple return values](FutureFeatures.md#multiple-return-values").
```
jfbastien: Stray quote?

sunfishcode: Yep, fixed.

titzer: I don't think the PR represents consensus yet, so maybe it's more
appropriate to make an issue for discussion first?

AstSemantics.md
titzer: On Thu, Jul 23, 2015 at 8:43 PM, Dan Gohman <notifications@github.com>
wrote:

> In AstSemantics.md
> <https://github.com/WebAssembly/design/pull/278#discussion_r35356329>:
>
> >
> > -In the MVP, function pointer values are local to a single module. The
> > +### Function pointers
> > +
> > +Function pointer values are obtained through the use of a special operator:
> > +
> > +  * `addressof`: obtain a funcptr value for a given statically-identified function
> > +
> > +and are comparable for equality:
> > +
> > +  * `funcptr.eq`: funcptr compare equal
> > +
> > +Note that it is not possible to directly observe the bits of a funcptr
> > +value. They may be [converted into integers][], but the integers only hold an
> > +index into the *function table*, a table with an entry for each function
> > +appended to the table in the order that they are loaded into the program.
>
> There's a basic design decision to be made somewhere about whether we
accept a dynamic signature check at calls/decodes or whether we try to
design a static mechanism. The addition of local types (that include
signatures) for function pointers only makes sense if we can remove the
need for most, if not all dynamic checks.


> Per the discussions in #89
> <https://github.com/WebAssembly/design/issues/89>, there's one table, not
> one table per signature, and:
>
> Call operations trap if the signature of the call differs from the
> signature of the called function
>
>
That implies a dynamic signature check, which I thought this proposal was
trying to avoid.

>  The order is deterministic if the application loads functions into the
> program in a deterministic fashion:
>
> appended to the table in the order that they are loaded into the program
>
> Also, I don't expect this PR is the last word on function pointers. This
> is just trying to update the text to where the current discussions have
> lead and clean up the text to facilitate the next rounds of discussion.
>
The idea of using multiple function tables, with one per signature, is to
avoid the need for a dynamic signature check and also the need for function
types to be local types, since the program only ever manipulates indices.
To be clear, I am not sure if that's a better solution, since it will
complicate dynamic linking, but it does have those advantages.

A key issue really is how C++ will encode vtables into wasm. Per-signature
function tables can work in an mtable scenario, where the vtable is
represented as a class number and the class number is used to index into
per-virtual-method tables. That requires numbering the classes in a
specific manner, but is dense and single-load efficient.

sunfishcode:
Because Github hides it, this PR is blocked as of @titzer's remark above:

> I don't think the PR represents consensus yet, so maybe it's more appropriate to make an issue for discussion first?

#89 already exists as the issue for discussion.

lukewagner:
I'd suggest:
1. Splitting out multiple-return values into a separate PR (although I think in the other issue we had some rough consensus about how to do that).
2. Having type-specific function-pointer types since for some impl strategies I can see it avoiding dynamic signature-match checks.
3. Instead of having the index order defined implicitly, having an explicit func-ptr-table declaration (effectively moving the "vtable optimization" from future features into the MVP, since I see little cost and immediate applicability).

@titzer What are the open issues you're thinking about?

rossberg-chromium:
I agree with Luke's suggestion that function ids ought to be typed. What is the motivation for having runtime signature checks instead?

jfbastien:
Can we move discussions to separate issues instead of this PR, as discussed in #282? I think we have a few forks to make :-)

lukewagner:
@jfbastien This isn't a diff; it's the main conversation of a PR and thus won't disappear if the branch is updated which was the problem.

lukewagner:
On further thought, I'm realizing that there is a lot more nuance to func-ptr types:

The big question is: what happens if you `funcid.decode` some bogus integer (out of range or, if we have typed `funcid`s, type mismatch)?  An obvious option would be to fault.  However, this would inhibit the C compiler from aggressively representing C function pointers as `funcid`s since sometimes (questionably legal) C code puts random bits into a func ptr, never calls that func ptr, and expects to get those bits back later.  With faulting semantics, a C compiler would only be able to generate a `funcid.decode` when it dominated a `call_indirect` in which case there's little benefit from having `funcid`s at all over simply having `call_indirect` take an int32 directly (and relying on the wasm backend to do any hoisting it can).

The other option is to have a non-faulting `funcid.decode` which has the property that `int32.encode` of `funcid.decode` of an int32 `x` equals `x` for all `x`, valid or not.  Thus `funcid`s would effectively be opaque int32s so the question is, again, what's the benefit over just having a `call_indirect` that took an int32.  I can imagine a tricky impl strategy that represented `funcid`s as function addresses by generating thunks on demand for invalid integer values (assuming this is a rare case), but I'm not sure if I'd want to do this over simply representing `funcid` as an int32.

Perhaps this is the same "need more impl experience" situation you were thinking about @titzer?

titzer:
On Fri, Aug 14, 2015 at 12:03 PM, Luke Wagner <notifications@github.com>
wrote:

> On further thought, I'm realizing that there is a lot more nuance to
> func-ptr types:
>
> The big question is: what happens if you funcid.decode some bogus integer
> (out of range or, if we have typed funcids, type mismatch)? An obvious
> option would be to fault. However, this would inhibit the C compiler from
> aggressively representing C function pointers as funcids since sometimes
> (questionably legal) C code puts random bits into a func ptr, never calls
> that func ptr, and expects to get those bits back later. With faulting
> semantics, a C compiler would only be able to generate a funcid.decode
> when it dominated a call_indirect in which case there's little benefit
> from having funcids at all over simply having call_indirect take an int32
> directly (and relying on the wasm backend to do any hoisting it can).
>
> The other option is to have a non-faulting funcid.decode which has the
> property that int32.encode of funcid.decode of an int32 x equals x for
> all x, valid or not. Thus funcids would effectively be opaque int32s so
> the question is, again, what's the benefit over just having a
> call_indirect that took an int32. I can imagine a tricky impl strategy
> that represented funcids as function addresses by generating thunks on
> demand for invalid integer values (assuming this is a rare case), but I'm
> not sure if I'd want to do this over simply representing funcid as an
> int32.
>
> Perhaps this is the same "need more impl experience" situation you were
> thinking about @titzer <https://github.com/titzer>?
>

Yeah, I think we can move quickly to get some experience if we take one of
a number of shortcuts here. E.g. we could:

a.) punt: have the producer generate their own dispatcher routines, just to
make progress. we'll probably try this first.
b.) per-signature function tables: something similar to asm.js, but since
it would be part of wasm, could be compiled into a single
boundscheck+indexedload+indirect branch (i.e. about 4-6 instructions).
c.) dynamic check: add a single funcptr type and try to make signature
checks cheap.

Those are in rough order of personal palatability. The check in c.) could
be made efficient if we require exact signature matching. That check could
then just be to load a signature id from the word immediately before the
machine code of the funcptr, compare against the callsite's expected id,
branching far away to trapping land if it fails (i.e. about 3-4
instructions).

The advantage of fully typed funcptrs I see here is that dispatch is 1-2
instructions.

Obviously, any of the strategies above could be sped up with an IC
mechanism that caches the last target and does a direct call with code
patching.

titzer:
On Fri, Aug 14, 2015 at 12:41 PM, Ben L. Titzer <titzer@google.com> wrote:

>
> On Fri, Aug 14, 2015 at 12:03 PM, Luke Wagner <notifications@github.com>
> wrote:
>
>> On further thought, I'm realizing that there is a lot more nuance to
>> func-ptr types:
>>
>> The big question is: what happens if you funcid.decode some bogus
>> integer (out of range or, if we have typed funcids, type mismatch)? An
>> obvious option would be to fault. However, this would inhibit the C
>> compiler from aggressively representing C function pointers as funcids
>> since sometimes (questionably legal) C code puts random bits into a func
>> ptr, never calls that func ptr, and expects to get those bits back later.
>> With faulting semantics, a C compiler would only be able to generate a
>> funcid.decode when it dominated a call_indirect in which case there's
>> little benefit from having funcids at all over simply having
>> call_indirect take an int32 directly (and relying on the wasm backend to
>> do any hoisting it can).
>>
>> The other option is to have a non-faulting funcid.decode which has the
>> property that int32.encode of funcid.decode of an int32 x equals x for
>> all x, valid or not. Thus funcids would effectively be opaque int32s so
>> the question is, again, what's the benefit over just having a
>> call_indirect that took an int32. I can imagine a tricky impl strategy
>> that represented funcids as function addresses by generating thunks on
>> demand for invalid integer values (assuming this is a rare case), but I'm
>> not sure if I'd want to do this over simply representing funcid as an
>> int32.
>>
>> Perhaps this is the same "need more impl experience" situation you were
>> thinking about @titzer <https://github.com/titzer>?
>>
>
> Yeah, I think we can move quickly to get some experience if we take one of
> a number of shortcuts here. E.g. we could:
>
> a.) punt: have the producer generate their own dispatcher routines, just
> to make progress. we'll probably try this first.
> b.) per-signature function tables: something similar to asm.js, but since
> it would be part of wasm, could be compiled into a single
> boundscheck+indexedload+indirect branch (i.e. about 4-6 instructions).
> c.) dynamic check: add a single funcptr type and try to make signature
> checks cheap.
>
> Those are in rough order of personal palatability. The check in c.) could
> be made efficient if we require exact signature matching. That check could
> then just be to load a signature id from the word immediately before the
> machine code of the funcptr, compare against the callsite's expected id,
> branching far away to trapping land if it fails (i.e. about 3-4
> instructions).
>
> The advantage of fully typed funcptrs I see here is that dispatch is 1-2
> instructions.
>
> Obviously, any of the strategies above could be sped up with an IC
> mechanism that caches the last target and does a direct call with code
> patching.
>

Also, long-term I think we will want to have typed function pointers that
fit in with a GC'd object model so that users can build vtables and such.
The question right now is if we need to solve this for MVP.

qwertie:
> this would inhibit the C compiler from aggressively representing C function pointers as `funcid`s

True, but couldn't C++ vtables still benefit, as well as dispatch mechanisms in many other languages? (the overall workflow of how wasm code will use `funcptr`s isn't clear to me.)

The C compiler might be able to help too. At least as a global optimizaton, it could use knowledge about whether "garbage" was casted to a function pointer type or not, and generate slower code if yes.

sunfishcode:
@lukewagner Imho, that's not questionably legal, that's seriously illegal. I don't want to penalize everyone just to optimize for that case. C compilers wishing to support that can always just be conservative.

If we agree that signature-parameterized types are something we'll be ok with long-term anyway, what are the downsides to just doing that in the MVP? We can figure out the vtable/etc. optimization parts after the MVP.

lukewagner:
@sunfishcode It doesn't matter if it's illegal if many codebases do it and expects it to work.  (E.g., SpiderMonkey does this all over the place via `JS_DATA_TO_FUNC_PTR`.)  Also, I was trying to read more about this and it looks like it is valid in C99 and conditionally legal in C++11.

@titzer 
I think it's possible to specify (c) but have an impl that does (b).  That is, even if there is a single specified index space, the impl could generate per-signature tables (selected statically by callsite signature) that contain throw stubs for all but the valid callees.  The downside is potentially a lot more memory; I had been thinking this would be an optional optimization that fell back to signature checks if the memory overhead would be prohibitive.

Interestingly, even if (b) is spec'd, real code depends on functions having unique addresses across different signatures.  This was enough of a problem in practice that Emscripten has a [workaround flag](https://github.com/kripken/emscripten/blob/master/src/settings.js#L170) that manually pads the tables in the same manner described above.  The difference is that this padding (and memory usage) gets baked into the semantics, instead of being an internal optimization.

Agreed on eventually wanting func-ptr types as GC object field types for implementing vtables; maybe that's reason enough to add func-ptr types now?

sunfishcode:
@lukewagner `JS_DATA_TO_FUNC_PTR` etc. for function pointers in `object*`, not the other way around. You're right though that while C/C++ present several obstacles to storing arbitrary bits in a function pointer, reinterpret_cast semantics ultimately make it possible. But in any case, this is just a minor variant of the broader problem of function pointers being storable in linear memory. If wasm has func_id types, producers from languages like C will need to do escape analysis optimize language function pointers into wasm func_ids anyway, and that'll also handle the various type conversion cases.

lukewagner:
There is also `JS_FUNC_TO_DATA_PTR` which is often used in conjunction to go the other way.

The alias analysis necessary to use `funcid` instead of an int in the heap would be the same intra-procedural analysis necessary to put anything into a local instead of the heap stack.  However, to use `funcid`, e.g., as the parameter type when the C++ code has a function-pointer parameter type, you need interprocedural analysis to see all the sources of values and/or uses.  I think the ideal here is that translation to `funcid`s could be trivial and predictable: any time you have an unaliased C++ function-pointer in a stack variable, you get a `funcid`.

sunfishcode:
(JS_FUNC_TO_DATA_PTR is for putting a function pointer in an Object*, and JS_DATA_TO_FUNC_PTR is for getting it out again. From a quick scan through SpiderMonkey, I don't think they're ever used in the reverse to put arbitrary bits into a function pointer type.)

Making placement of `funcptr` decode operations predictable from the C++ level where translation to `funcid` happens will be tricky; optimizing compilers aggressively discard variable lifetimes in the original code. Also, since translation from an index into a `funcid` will incur a cost, some cleverness may be desirable to place those operations to minimize the costs. I think we can do a reasonable job with this though.

Overall, what is the overall feeling here? Are people interested in having signature-parameterized `funcid` types in wasm, to allow type checks to be factored out in addition to range checks? Or are people leaning towards just having bare integer indices everywhere to start with?

lukewagner:
@sunfishcode Assuming the heap is much larger than the func-ptr table, casting an Object* to a func-ptr might as well be an arbitrary integer: both are invalid func-ptr indices and will spuriously fault if we have faulting `funcid.decode` semantics.

> Overall, what is the overall feeling here? Are people interested in having signature-parameterized > funcid types in wasm, to allow type checks to be factored out in addition to range checks? Or are 
> people leaning towards just having bare integer indices everywhere to start with?

This question seems to boil down to: is there any perf win from `funcid` over bare integers.  Once we have working prototypes, we should be able to do a quick-and-dirty experiment to test this.  As @titzer said above, we do *eventually* want fully-typed `funcid`s for use in GC types, but that use care may be sufficiently different from C++ function pointers (in particular, it might not even need `int32.encode`/`funcid.decode`) that we could get away with something much simpler.  Thus, we shouldn't preemptively add `funcid` now just based on the expected need in the future, but only with some clear perf win.

sunfishcode:
Closing this PR; the parts unrelated to function pointers are now merged in #307. The original issue #89 is still open.

Issue #255, opened 2015-07-07
Are 64-bit addresses a mode or separate opcodes?


sunfishcode:
In #245, @titzer asked whether 64-bit addressing (for >4GiB linear memory) should be a mode or just separate opcodes.

In practice, C/C++ and similar code is going to require one form or the other. Several people have looked at address-size abstractions, but ultimately decided to just accept it as a mode at the C/C++ ABI level, so we're currently expecting to have two ABIs anyway.

And, the utility of having a >4GiB heap with some code in the process that can only access the low 4 GiB of it seems marginal, and the potential for confusion and mistakes under such circumstances is significant.

Having two modes at the platform level seems indicated. Does anyone have other opinions?

titzer:
I think we should have separate bytecodes, since loads/stores already have
tons of bells and whistles, and adding another mode to configure them for a
module or a function or a scope seems superfluous. In the
v8-native-prototype the load/store bytecodes break down as follows (forgive
the macro-fu, it's a cut-and-paste):

#define FOREACH_LOAD_MEM_EXPR_OPCODE(V) \

  V(Int32LoadMemL, 0x20, i_i)           \

  V(Int64LoadMemL, 0x21, l_i)           \

  V(Float32LoadMemL, 0x22, f_i)         \

  V(Float64LoadMemL, 0x23, d_i)         \

  V(Int32LoadMemH, 0x24, i_l)           \

  V(Int64LoadMemH, 0x25, l_i)           \

  V(Float32LoadMemH, 0x26, f_l)         \

  V(Float64LoadMemH, 0x27, d_l)


// Store memory expressions.




#define FOREACH_STORE_MEM_EXPR_OPCODE(V) \

  V(Int32StoreMemL, 0x30, i_ii)          \

  V(Int64StoreMemL, 0x31, l_il)          \

  V(Float32StoreMemL, 0x32, f_if)        \

  V(Float64StoreMemL, 0x33, d_id)        \

  V(Int32StoreMemH, 0x34, i_li)          \

  V(Int64StoreMemH, 0x35, l_ll)          \

  V(Float32StoreMemH, 0x36, f_lf)        \

  V(Float64StoreMemH, 0x37, d_ld)


Having separate bytecodes has the nice advantage that the "signatures" i.e.
the input/output types of the bytecodes are fully determined without taking
into account a mode. Above, the -H suffix indicates a 64-bit memory index.

titzer:
[...]

PR #283, opened 2015-07-28
Elaborate on how 64-bit indices might work.

sunfishcode:
This is meant to capture the consensus in #255 about 64-bit indices:
 - 64-bit indices are a FutureFeature
 - initially, mixing 32-bit indices and 64-bit indices in the same program may be prohibited
 - separate opcode names will be used to leave the possibility of mixed modes in the future


jfbastien:
lgtm

Issue #236, opened 2015-06-29
Feature suggestion (memcpy/memset)


lskyum:
First of all - sorry if I missed it in the spec!

The suggestion is for simple memcpy/memset instructions, so copying/clearing can be done at maximum speed for any given platform/browser. For example bounds-checking doesn't have to be done pr. read/write, but only once. (Could be an issue on some implementations I think. Maybe also the polyfill...)

The memcpy could be provided in two versions, one that checks for overlaps and one that doesn't.

janesconference:
Aren't memcpy / memset / memmove usually implemented at an higher level, ie in the C stdlib? Example implememtation in XNU: http://www.opensource.apple.com/source/xnu/xnu-2050.18.24/libsyscall/wrappers/memcpy.c

waywardmonkeys:
In emscripten, ``memset`` and ``memcpy`` can result in pretty bloaty code, especially when compared to native code and what a loop expands into. (We suffer a good bit from things related to this in our code base that is compiled with emscripten.) I don't know how that would be in WebAssembly or if it would be less bloaty.

MikeHolman:
Interesting thought.

You might be able to rely on the wasm engine recognizing this pattern and optimizing the loop/removing bound checks (our engine is starting to do some optimizations when it recognizes the "memcpy" pattern in js). And when we have SIMD support we can implement memcpy pretty efficiently in wasm without as much reliance on optimizations from the wasm engine. To this end, it would be interesting to see how much SIMD helps memcpy in asm.js.

I think this should be less bloated in wasm (even in asm.js I think we could probably just not inline memcpy and clear up the code bloat -- I'd think in most cases inlining them won't really improve perf because call overhead would be dwarfed by the actual copy).

But it might be worth considering because it could help perf for baseline compiler (which might not have BCE), and it could reduce AST size (obviously important).

waywardmonkeys:
@MikeHolman The issue with not inlining in emscripten is that various patterns of code get converted to ``llvm.memcpy*`` and ``llvm.memset*`` calls ... this happens in constructors and other locations, as well as (sometimes) LLVM looking at a loop and deciding it can be converted to one of these calls.

sunfishcode:
WebAssembly does have fewer pain points than asm.js in this area, but there would still be advantages to adding these.

For memcpy, as opposed to memmove, it's not clear that introducing nondeterminism for the operlapping case is worth the minor performance difference, and we can always add it in the future if we wish.

[...]

PR #284, opened 2015-07-28
Define `memmove` and `memzero` operators.

sunfishcode:
This proposes what I sketched out in #236.

I'm quite open to discussing extensions to these operators to accept e.g. size and alignment hints (size hints may include things like known-multiples or known-ranges when the size isn't constant), but let's do so in a new issue.

Why `memzero` and not `memset`? It's simpler while still covering the main use case. On one hand, I'm already a nervous about adding what could be seen as macro-operators to WebAssembly in the first place. And on the other, the ability to paint memory regions with just a single-byte pattern is oddly restrictive. If we're going to add functionality like that, perhaps we should add forms for 32-bit and 64-bit patterns too.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -236,6 +236,8 @@ source overlap.
 `memzero`'s operands are a destination address, and a size in bytes. It writes
 zeros to memory of the given size at the destination address.
 
+`memmove` and `memzero` return the value of their `destination` address operand.
+
 If any of the accessed bytes are beyond `memory_size`, the access is considered
 [out-of-bounds](AstSemantics.md#out-of-bounds).
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -223,6 +223,26 @@ MVP, there are [future features](FutureFeatures.md#finer-grained-control-over-me
 proposed to allow setting protection and creating mappings within the
 contiguous linear memory.
 
+### Bulk-Memory Operations
+
+  * `memmove`: bulk-copy memory
+  * `memzero`: bulk-zero memory
+
+`memmove`'s operands are a destination address, a source address, and a size
+in bytes. It copies memory of the given size from the source address to the
+destination address, correctly handling the case where the destination and
+source overlap.
+
+`memzero`'s operands are a destination address, and a size in bytes. It writes
+zeros to memory of the given size at the destination address.
+
+If any of the accessed bytes are beyond `memory_size`, the access is considered
+[out-of-bounds](AstSemantics.md#out-of-bounds).
+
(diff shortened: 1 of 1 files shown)


jfbastien:
* It would be useful to add this under an "intrinsics" or "builtins" section.
* I'd have type signatures.
* I prefer `memset` over `memzero`. It's not really more complicated, is as familiar, and is useful.

sunfishcode:
>    It would be useful to add this under an "intrinsics" or "builtins" section.

This is a path which I'm specifically nervous about, and am liable to abandon and subsequently oppose the addition of even `memmove` if I suspect this is where it'll lead. We can always do `memmove` in a library reasonably well, especially once we get SIMD.

> I'd have type signatures.

"Everything is int32", and the future "Everything is int64", are pretty simple (and note that this is nicer with `memzero` over `memset`, will the value operand be int64 in the 64-bit pointer version?). Avoiding an explicit type signature declaration encourages us to keep things simple :-).

> I prefer memset over memzero. It's not really more complicated, is as familiar, and is useful.

`memset` might be confusing to some when they pass it an int32 and it ignores part of the value. Or when they figure out that there isn't a version that does use the full 32-bit value. But I'm content to see what the consensus is here.

jfbastien:
Type signatures are easier to read than English IMO :-)

sunfishcode:
Important to have in a section of intrinsics or builtins, surely :-).

titzer:
Is there a strong need to have a distinction between intrinsics and the
more basic AST operations? Engines are required to implement both, and both
have well-defined semantics. One view is that it is merely an
implementation detail if a given AST operation or intrinsic function is
implemented as inline code or a subroutine.

qwertie:
@titzer I've been thinking the same thing. In terms of file format, even control flow could be stored the same way as intrinsics (except maybe ... blocks? Blocks are the only thing coming to mind that really seem to need a variable argument list. There could be a binary operator `then`, but it's inefficient in case of many statements in a row.)

sunfishcode:
@titzer I agree, I don't think we need such a distinction.

jfbastien:
No distinction sounds fine. 

I'd still like explicit signatures instead of English. We do provide signatures for AST nodes.

I'd also like `memset` instead of `memzero`.

lukewagner:
For core semantics (`sbrk`, `memset`), I've been equating the two in my mind as well.  I think the only distinction comes when we get to host-dependent builtins.

sunfishcode:
Closing for now; we can reconsider these operators in the future when we look at other bulk memory operations too.

PR #288, opened 2015-08-05
Clarify, rename, and FAQ memory allocation

lukewagner:
Based on conversations in #285 which seemed to have settled down, this PR clarifies how WebAssembly could (in MVP + future features) implement most of mmap, renames sbrk to something without a connotation, and adds an FAQ that explains why no noncontiguous allocation.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -256,9 +256,6 @@ omission is:
   * This interleaving with unrelated allocations does not currently admit
     efficient security checks to prevent one module from corrupting data outside
     its heap (see discussion in [#285](https://github.com/WebAssembly/design/pull/285)).
-  * This interleaving would require making allocation nondeterministic.
-    Nondeterminism is something that WebAssemgly generally 
-    [tries to avoid](Nondeterminism.md) and history has clear examples of
-    memory allocator almost-determinism leading to real-world bustage
-    ([[1](https://technet.microsoft.com/en-us/magazine/ff625273.aspx)],
-    [[2](http://lxr.free-electrons.com/source/include/linux/personality.h?v=3.2#L31)]).
+  * This interleaving would require making allocation nondeterministic and
+    nondeterminism is something that WebAssemgly generally 
+    [tries to avoid](Nondeterminism.md).
--- a/FAQ.md
+++ b/FAQ.md
@@ -255,7 +255,7 @@ omission is:
   address space fragmentation). There are two problems with this:
   * This interleaving with unrelated allocations does not currently admit
     efficient security checks to prevent one module from corrupting data outside
-    its heap (see discussion in #285).
+    its heap (see discussion in [#285](https://github.com/WebAssembly/design/pull/285)).
   * This interleaving would require making allocation nondeterministic.
     Nondeterminism is something that WebAssemgly generally 
     [tries to avoid](Nondeterminism.md) and in this particular case, history
--- a/FAQ.md
+++ b/FAQ.md
@@ -229,3 +229,36 @@ WebAssembly implementations run on the user side, so there is no opportunity for
  * Most of the individual floating point operations that WebAssembly does have already map to individual fast instructions in hardware. Telling `add`, `sub`, or `mul` they don't have to worry about NaN for example doesn't make them any faster, because NaN is handled quickly and transparently in hardware on all modern platforms.
 
  * WebAssembly has no floating point traps, status register, dynamic rounding modes, or signalling NaNs, so optimizations that depend on the absence of these features are all safe.
+
+## What about `mmap`?
+
+The [`mmap`](http://pubs.opengroup.org/onlinepubs/009695399/functions/mmap.html)
+syscall has many useful features. While these are all packed into one
+overloaded syscall in POSIX, WebAssembly unpacks this functionality into
+multiple builtins:
+* the MVP starts with the ability to resize linear memory via a
+  [`resize_memory`](AstSemantics.md#resizing) builtin operation;
+* proposed [future features](FutureFeatures.md#finer-grained-control-over-memory)
+  would allow the application to change the protection and mappings for pages
+  in the contiguous range set by `resize_memory`.
+
+A significant feature of `mmap` that is missing from the above list is the
+ability to allocate disjoint virtual address ranges. The reasoning for this
+omission is:
+* The above functionality is sufficient to allow a user-level libc to
+  implement full, compatible `mmap` with what appears to be noncontiguous
+  memory allocation (but, under the hood is just coordinated use of
+  `memory_resize` and `mprotect`/`map_file`/`map_shmem`/`madvise`).
+* The benefit of allowing noncontiguous virtual address allocation would be if
+  it allowed the engine to interleave a WebAssembly module's linear memory with
+  other memory allocations in the same process (in order to mitigate virtual
+  address space fragmentation). There are two problems with this:
+  * This interleaving with unrelated allocations does not currently admit
+    efficient security checks to prevent one module from corrupting data outside
+    its heap (see discussion in #285).
+  * This interleaving would require making allocation nondeterministic.
+    Nondeterminism is something that WebAssemgly generally 
+    [tries to avoid](Nondeterminism.md) and in this particular case, history
+    has clear examples of memory allocator nondeterminism leading to real-world
+    bustage ([[1](https://technet.microsoft.com/en-us/magazine/ff625273.aspx)],
+    [[2](http://lxr.free-electrons.com/source/include/linux/personality.h?v=3.2#L31)]).
--- a/FAQ.md
+++ b/FAQ.md
@@ -258,7 +258,7 @@ omission is:
     its heap (see discussion in [#285](https://github.com/WebAssembly/design/pull/285)).
   * This interleaving would require making allocation nondeterministic.
     Nondeterminism is something that WebAssemgly generally 
-    [tries to avoid](Nondeterminism.md) and in this particular case, history
-    has clear examples of memory allocator nondeterminism leading to real-world
-    bustage ([[1](https://technet.microsoft.com/en-us/magazine/ff625273.aspx)],
+    [tries to avoid](Nondeterminism.md) and history has clear examples of
+    memory allocator almost-determinism leading to real-world bustage
+    ([[1](https://technet.microsoft.com/en-us/magazine/ff625273.aspx)],
     [[2](http://lxr.free-electrons.com/source/include/linux/personality.h?v=3.2#L31)]).
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -40,10 +40,27 @@ possible to use a non-standard ABI for specialized purposes.
 
 ## Finer-grained control over memory
 
-* `mmap` of files.
-* `madvise(MADV_DONTNEED)`.
-* Shared memory, where a physical address range is mapped to multiple physical
-  pages in a single WebAssembly module as well as across modules.
+Provide access to safe OS-provided functionality including:
+* `map_file(addr, length, Blob, file-offset)`: semantically, this operation
+   copies the specified range from `Blob` into the range `[addr, addr+length)`
+   (where `addr+length <= memory_size`) but implementations are encouraged
+   to `mmap(addr, length, MAP_FIXED | MAP_PRIVATE, fd)`
+* `dont_need(addr, length)`: semantically, this operation zeroes the given range
+   but the implementation is encouraged to `madvise(addr, length, MADV_DONTNEED)`
+* `shmem_create(length)`: create a memory object that can be simultaneously
+  shared between multiple linear memories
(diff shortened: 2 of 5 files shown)


FAQ.md
```diff
@@ -229,3 +229,36 @@ WebAssembly implementations run on the user side, so there is no opportunity for
...
+  implement full, compatible `mmap` with what appears to be noncontiguous
+  memory allocation (but, under the hood is just coordinated use of
+  `memory_resize` and `mprotect`/`map_file`/`map_shmem`/`madvise`).
+* The benefit of allowing noncontiguous virtual address allocation would be if
+  it allowed the engine to interleave a WebAssembly module's linear memory with
+  other memory allocations in the same process (in order to mitigate virtual
+  address space fragmentation). There are two problems with this:
+  * This interleaving with unrelated allocations does not currently admit
+    efficient security checks to prevent one module from corrupting data outside
+    its heap (see discussion in #285).
```
jfbastien: Fix link to #285.

FAQ.md
```diff
@@ -229,3 +229,36 @@ WebAssembly implementations run on the user side, so there is no opportunity for
...
+  * This interleaving would require making allocation nondeterministic.
+    Nondeterminism is something that WebAssemgly generally 
+    [tries to avoid](Nondeterminism.md) and in this particular case, history
+    has clear examples of memory allocator nondeterminism leading to real-world
+    bustage ([[1](https://technet.microsoft.com/en-us/magazine/ff625273.aspx)],
+    [[2](http://lxr.free-electrons.com/source/include/linux/personality.h?v=3.2#L31)]).
```
jfbastien: Eh, that's more about starting deterministic and and going non-deterministic, or even then just differently deterministic... It's more anecdotal than useful, I'd just drop it.

FutureFeatures.md
```diff
@@ -40,10 +40,27 @@ possible to use a non-standard ABI for specialized purposes.
...
-* `madvise(MADV_DONTNEED)`.
-* Shared memory, where a physical address range is mapped to multiple physical
-  pages in a single WebAssembly module as well as across modules.
+Provide access to safe OS-provided functionality including:
+* `map_file(addr, length, Blob, file-offset)`: semantically, this operation
+   copies the specified range from `Blob` into the range `[addr, addr+length)`
+   (where `addr+length <= memory_size`) but implementations are encouraged
+   to `mmap(addr, length, MAP_FIXED | MAP_PRIVATE, fd)`
+* `dont_need(addr, length)`: semantically, this operation zeroes the given range
+   but the implementation is encouraged to `madvise(addr, length, MADV_DONTNEED)`
```
jfbastien: I think we need to emphasize this way more: this allows applications to be good citizens and release *physical* memory. That's a huge win on memory-constrained devices, and is a pretty important feature for many platforms (otherwise the OS is stuck playing favorites with an OOM killer).

Modules.md
```diff
@@ -108,8 +108,9 @@ to allow *explicitly* sharing linear memory between multiple modules.
 ## Initial state of linear memory
 
 A module will contain a section declaring the linear memory size (initial and
-maximum size allowed by `sbrk`) and the initial contents of memory (analogous
-to `.data`, `.rodata`, `.bss` sections in native executables).
+maximum size allowed by [`resize_memory`](AstSemantics.md#resizing) and the
+initial contents of memory (analogous to `.data`, `.rodata`, `.bss` sections in
+native executables).
```
jfbastien: Do we want to dictate these, or let wasm handle them but restrict where they can go? As discussed in #285 we can get a no-signal non-page-table but highly-efficient zero-page and `.rodata` implementation.

Nondeterminism.md
```diff
@@ -31,6 +31,7 @@ currently admits nondeterminism:
    nondeterministic.
  * Out of bounds heap accesses *may* want
    [some flexibility](AstSemantics.md#out-of-bounds)
+ * The [`page_size` global constant](AstSemantics.md#resizing)
```
jfbastien: Clarify that this is a size that's guaranteed to be allocatable, and that passing less than this to relevant functions cause them to trap.

FAQ.md
lukewagner: If changes over time, it was nondeterministic the whole time (determinism being time invariant).  Also, I think that these two anecdotes are quite relevant since we're in roughly a similar situation as a virtual platform.

FAQ.md
lukewagner: done

FutureFeatures.md
lukewagner: good point

FAQ.md
sunfishcode: I agree. Part of the point of avoiding nondeterminism in this feature is to prevent the creation of our own versions of those anecdotes.

Modules.md
lukewagner: In a sense, `.rodata` *is* letting wasm handle them but restricting where they can go.  Having the data declared in a global section also allows the wasm compiler to bake in more constant address info compared to a fully dynamic API that returned dynamic pointers to the data (both with and without dynamic loading).  That all being said, there's a lot more design space to consider on this subject, but it seems like a topic for a separate PR/issue.

Nondeterminism.md
lukewagner: done

FAQ.md
jfbastien: Take the example of a hash map: if iteration order is unspecified then people may rely on it, then when it changes it breaks code. Those are the above anecdotes. If instead the hash maps throws in a different seed every time then it's *always* non-deterministic, and people can't rely on iteration order, ever. This IMO is *good* non-determinism in many cases. Memory allocation order can go that way too: always randomize.

That's why I think the anecdotes are irrelevant: they're one exploration of what could have happened. Yes we should learn from them, but the learning isn't "kill all non-determinism with fire". If anything, things like ASLR are things that we want!

Having this discussion focuses us on these specific example, instead of seeing the bigger picture. We can't generalize these anecdotes to say "and that's why we don't do non-determinism, kids!".

sunfishcode: "always randomize" is plausible when one has sufficient flexibility over the range of possible values. However with address spaces, some implementations aren't going to want to randomize, because it could fragment their own address space, because it could leak information about the layout of their own address space, or because they don't have enough free address space to do meaningful randomization.

ASLR is an interesting question inside a system where host virtual addresses are not exposed, return addresses aren't in the address space, function bodies aren't in the address space, JIT code (when such features are added) isn't in the address space, and indirect calls can't jump anywhere but the start of a function. There may be uses for it, but I've been assuming ASLR inside the WebAssembly address space would be implemented by WebAssembly producers and libraries rather than being built into the platform.

lukewagner: Just to emphasize what @sunfishcode is saying: if the whole point of noncontiguity is to allow the engine to just call the OSes `mmap`, then it won't be randomizing.  Also, this isn't listed as an absolute reason we cannot, just as a con.

kripken:
I still believe it would be better to remove memory resizing from the MVP, since we know that we can't polyfill it well (@pizlonator for jsc, and last I heard v8 still had issues, but I'm not sure if that is still true - @titzer?).

I feel that the MVP is far more compelling when it just uses features we know polyfill and polyfill performantly. Since we plan to allow memory resizing after the MVP, this does not cause any long-term downsides - and people can still use asm.js if they need memory resizing, until browsers have native wasm support; there are a few other features similarly limited in the MVP anyhow, for which we say that.

lukewagner:
First, I think it makes sense to require some module-level flag to opt-in to allowing heap-resizing.  This allows various optimizations even with native wasm support.  For the polyfill, this would control whether the changeHeap (resize) function was emitted.  This would give developers the control of when the functionality was worth the hit.

As I understood it, the massive [slowdown we saw in TurboAsm](https://code.google.com/p/v8/issues/detail?id=3907) was *not* just because the singleton/closure analysis failed, but because the fallback was all the way back to ICs instead of the dynamic (hoisted) loads and length checks that were being emitted by CrankShaft.  Has this since been fixed @titzer?

@pizlonator When you say in [your comment](https://github.com/WebAssembly/design/pull/285#issuecomment-125776304) "We support it, you just get punished for doing it.  But your code *will* run.", I assume it's not the massive punishment mentioned above, but a fallback do dynamic loads of the heap base/length.  Is that right?

FAQ.md
jfbastien: That wasn't the point of my latest reply. I'm showing why the two links are interesting anecdotes that are misleading without full context. ASLR and `mmap` aren't relevant here besides being illustrative. I'm not suggesting ASLR (or advocating against it), I'm saying the anecdotes sound very convincing to someone without context (I find this misleading), it's not worth explaining more context, I therefore suggest removing the anecdotes.

jfbastien:
I'm not a fan of module-level flags, especially when we don't have dynamic linking and therefore have all the information up-front (list of module imports).

FAQ.md
lukewagner: What exactly is the missing context here?  It sounds like you're trying to say: if we had nondeterminism *and* had a *de facto* practice in at least some browsers of handing out very random memory, then people wouldn't be able to get away with depending on the nondeterminism and in that way the anecdotes are different, because they were regular enough that people were able to depend on a particular behavior?  If so, I feel like that isn't "context", but, rather "an attempt to partially mitigate the effects of nondeterminism".

lukewagner:
I don't think we need an actual flag.  Rather, based on what is currently in the design doc:
* the main module gets to declare the min/max linear memory size
* `resize_memory` calls (in any linked module) fail that would go above/below max/min
* so disabling resize means having min = max

FAQ.md
jfbastien: "history has clear examples" is loaded phrasing. Yes, people messed up in the past and got bit, but nondeterminism is much more complex than just these two anecdotes. It's desirable in many cases, e.g. beyond ASLR see UCI research on code randomization or the Matasano JIT mitigation paper.

Having loaded sentences against nondeterminism isn't something I appreciate because I'll then get unsubstantiated pushback (or rather, pushback backed by anecdotes) for further nondeterminism where it's reasoned and grounded in research.

The takeaway shouldn't be "nondeterminism is bad", it should be "Careful with That Axe, Eugene".

jfbastien:
"declaring" a value means there's magic in the platform that just "does the right thing". By doing this we're not giving the tools to developers to build things we don't anticipate, which goes against the Extensible Web's approach.

Agreed that sometimes implementation constraints mean that's what needs to be done, but I thought that we'd settled on exposing lower-level capabilities in this case?

FAQ.md
lukewagner: The entire sentence is "history has clear examples of memory allocator nondeterminism leading to real-world bustage" which means that this statement isn't refering to "all nondeterminism" but nondeterminism of a specific kind.  Furthermore, the placement of these anecdotes scope them to the this one kind of nondeterminism.  Perhaps, then, what you're really asking for is that the scope should be further reduced to "history has clear examples of memory allocator almost-determinism leading to real-world bustage" since the "axe" here is behaving deterministically enough to get broken code depending on it.

lukewagner:
@jfbastien The min/max-size discussion really depends on what optimizations and impl techniques it allows.  But that discussion should be in a separate issue since this PR is not adding it.  Anyhow, still waiting to hear whether heap resizing in asm.js is a major slowdown in v8/jsc.

FAQ.md
lukewagner: Updated.

lukewagner:
Rebased.  Any other comments before merging?

jfbastien:
As [we discussed](https://github.com/WebAssembly/design/pull/288#discussion_r36268627) I still think that the links aren't useful to this discussions: they're biased anecdotes, and explaining a fuller context wouldn't serve much of a purpose. I'd be happy with the patch minus the links.

kripken:
My comment from before still stands about removing memory resizing from the MVP. Last I can see you were waiting for two people to respond, has there been further discussion there?

lukewagner:
I was still waiting for a response to my [response](https://github.com/WebAssembly/design/pull/288#discussion_r36441378) and [update to more narrowly scope the anecdotes](https://github.com/WebAssembly/design/commit/fee4183555ccef8adda43d343d11de42fdbe77f4).

lukewagner:
Resizing was already in the spec repo so I think the question of whether to remove it should be decoupled from all the issues being clarified here.  An outcome of that discussion could be to move things from AstSemantics to FutureFeatures, although I hope that won't be necessary.

kripken:
Fair enough, filed #294.

lukewagner:
Phrasing `page_size` as a global constant (which is quite unlike real user globals) was confusing, so rephrased as a nullary constant operation (so AST node) which makes more sense anyway.

lukewagner:
It's not worth holding up for some links in the FAQ, so removed and merged based on several lgtm.

jfbastien:
Thanks, official lgtm stamp of approval.

PR #296, opened 2015-08-13
Elaborate on the semantics of `page_size`.

sunfishcode:
This fills out the specification of the `page_size` operator a little.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -232,6 +232,10 @@ support for page sizes requiring a type of `int64` may be added. The page size
 is an unsigned integer which is a power of 2 and at least 1024. `page_size`
 traps if the actual page size cannot be represented under these constraints.
 
+(Note that the `page_size` value need not reflect the actual internal page size
+of the implementation; it just needs to be a value suitable for use with
+`resize_memory`)
+
 ## Local variables
 
 Each function has a fixed, pre-declared number of local variables which occupy a single
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -227,9 +227,8 @@ MVP, there are [future features](FutureFeatures.md#finer-grained-control-over-me
 proposed to allow setting protection and creating mappings within the
 contiguous linear memory.
 
-In the MVP, the result type of `page_size` is `int32`; in the future, support
-support for page sizes requiring a type of `int64` may be added. The page size
-is an unsigned integer which is a power of 2.
+The result type of `page_size` is `int32` for wasm32 and `int64` for wasm64.
+The result value of `page_size` is an unsigned integer which is a power of 2.
 
 (Note that the `page_size` value need not reflect the actual internal page size
 of the implementation; it just needs to be a value suitable for use with
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -229,8 +229,8 @@ contiguous linear memory.
 
 In the MVP, the result type of `page_size` is `int32`; in the future, support
 support for page sizes requiring a type of `int64` may be added. The page size
-is an unsigned integer which is a power of 2 and at least 1024. `page_size`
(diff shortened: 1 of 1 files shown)


AstSemantics.md
```diff
@@ -227,6 +227,11 @@ MVP, there are [future features](FutureFeatures.md#finer-grained-control-over-me
 proposed to allow setting protection and creating mappings within the
 contiguous linear memory.
 
+In the MVP, the result type of `page_size` is int32`; in the future, support
```
jfbastien: One missing `backtick` here.

AstSemantics.md
```diff
@@ -227,6 +227,11 @@ MVP, there are [future features](FutureFeatures.md#finer-grained-control-over-me
...
+support for page sizes requiring a type of `int64` may be added. The page size
+is an unsigned integer which is a power of 2 and at least 1024. `page_size`
```
jfbastien: Mention that this is bytes?
Also, something along the lines of "it's not necessarily the actual page size, it's a suggested assumption from the runtime".

AstSemantics.md
sunfishcode: Fixed.

titzer:
What can page_size be used for in the MVP?

sunfishcode:
@titzer [The `resize_memory` operation](https://github.com/WebAssembly/design/blob/master/AstSemantics.md#resizing).

AstSemantics.md
sunfishcode: Fixed.

titzer:
What if we just limited memory sizes to powers of 2?

jfbastien:
@titzer going from 2GiB to 4GiB seems like a big jump :)

Though changing `page_size` to be `page_size_log2` would be interesting.

titzer:
page_size = 1gb, problem solved :-)

sunfishcode:
@titzer After thinking about it more, I agree. I removed the minimum page size.

jfbastien:
lgtm

titzer:
Sorry if I wasn't clear. What if we just limited memory sizes to powers of
2, which might obviate the need for page_size?

sunfishcode:
On 32-bit devices, powers of 2 are very big steps relative to the underlying address space. A device may be able to easily allocate more than 128 MiB contiguous address space to a wasm app, but not all of 256 MiB, for example.

jfbastien:
@titzer the OS we run on imposes a minimum page size. I guess we can over-allocate, and tell the user where memory now ends?

lukewagner:
With asm.js, we started with a power of 2 allocation and got pushback that even the 128, 256, 512, 1024 quantas were too coarse, particularly on 32-bit, so we loosened the restriction to "multiple of 16mb" (for the benefit of ARM operand2 immediate encoding).  This power-of-2 limitation would be exacerbated in the many-small-wasm-modules-in-one-web-app use case we've discussed before.

jfbastien:
Could we move this PR forward? I was looking for it in the repo, and just realized it was still uncommitted.

Small nit along the lines of #354, should this say "same type as the address' index"?

jfbastien:
@sunfishcode reminded me that @titzer may want to chime in some more before we move this forward. Let's wait on him :)

titzer:
I think we should just drop the language about int64 page sizes and
trapping, since I don't think there are any machines on the horizon that
don't fit the current requirements.

sunfishcode:
WRT trapping, I agree. That's part of an older idea that is now removed.

WRT int64 page sizes, we can now reword this in terms of wasm32 and wasm64, which I think makes this clearer. What do you think of the updated text?

titzer:
lgtm although the larger discussion around wasm32/wasm64 is probably still
not quite settled.

jfbastien:
lgtms abound! Merging.

Issue #127, opened 2015-06-09
Define function signatures


steveluc:
It would be helpful in reasoning about ABIs if the Call section of the AST document could give some examples of function signature specifications.  

How would this proposal handle variable-length parameter lists?
How do compilers place arrays on the stack?  Looks like all of the local variable types are scalar.

sunfishcode:
#278 is a pull request which defines function signatures.

sunfishcode:
Function signatures are defined in #307, which is now merged, along with a note about C varargs. There are also some example functions [in the spec test directory](https://github.com/WebAssembly/spec/tree/master/ml-proto/test).

PR #307, opened 2015-08-20
Tidy up function return types.

sunfishcode:
This replaces the `void` type with the concept of a return type sequence which may be empty, and it defines *signature*.

This brings in the parts of #278 not connected to function pointers.

It addresses #127 but not #89.


jfbastien:
lgtm

lukewagner:
lgtm too

Issue #281, opened 2015-07-23
Non-storable bool type


jfbastien:
AstSemantics.md currently uses integers to hold comparison results. I'd much rather have a non-storable boolean type (call it `i1`) which is returned by comparisons. It's convertible to/from other integer types, but cannot be loaded or stored.

It can be used for `select`, and once we have SIMD you can construct non-storable `i1` vectors.

This is nice because it keeps the type information in the IR, making things easier to track and in the case of SIMD easier to optimize because different ISAs represent SIMD comparisons differently.

sunfishcode:
If we add a scalar bool type, I suggest `bool` instead of `i1` or `int1' because I don't believe we'd plan to give it basic integer operations like `add`, `sub`, etc.

I like this because it's human-friendly, it enables some fun (though not game-changing) optimizations, and it eliminates a pesky asymmetry where we just use `int32` to hold booleans, however I don't feel I have any arguments which conclusively outweigh the downsides of adding another type. In particular, it's feasible to have a SIMD bool type without having a scalar bool type.

jfbastien:
`bool` sgtm.

I would allow bitwise operations on `bool`.

Agreed the optimizations aren't game-changing, but they're useful enough and make the code generator ever-so-slightly easier that it's worth it. I do like the symmetry for SIMD bool types :-)

titzer:
I don't see a real value-add here. It means more local types, which means
more signatures, requires explicit conversions between ints and bools, and
requires more operators on booleans. Overall, this just means more
complexity with marginal static checking possibilities. Actually, I see
bool as more of a hassle for wasm producers and consumers than anything
else. I'm not sure what optimizations this enables. E.g. TurboFan already
does plenty of optimizations, e.g. reasoning about boolean negation,
without the extra help of a boolean type.

sunfishcode:
I agree that symmetry with SIMD is also in favor of a scalar bool type. WebAssembly has a pretty good chance of achieving symmetry between all operators which will have both scalar and SIMD forms, but a lack of a bool type would compromise this aesthetic property.

[...]

PR #308, opened 2015-08-21
Add a `bool` type.

sunfishcode:
Following up on #281, this is a PR for a `bool` type. `bool` is a local type, for local variables, parameters, and return values, but it is not directly storable in linear memory.

The advantages of a `bool` type over just `int32` are admittedly subjective, but I think it's worth considering:
 - Some minor optimization opportunities:
    - `bool` makes it easier for implementations to decide how to represent arguments and return values
    - on x86, the SETcc family of instructions write to a byte-sized register. For cases when a boolean value needs to be materialized in a register, producing a full 32-bit register value requires an extra instruction.
 - Symmetry with SIMD comparisons and selects. For example, SIMD less-than is expected to return a SIMD bool type, so it's nice to have the corresponding scalar SIMD less-than also return a bool type.
 - Symmetry between `int32` and `int64` for comparisons and selects.
 - Human-friendliness.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -256,12 +256,14 @@ are statements.
   * `if`: if statement
   * `do_while`: do while statement, basically a loop with a conditional branch
     (back to the top of the loop)
-  * `forever`: infinite loop statement (like `while (1)`), basically an
+  * `forever`: infinite loop statement (like `while (true)`), basically an
     unconditional branch (back to the top of the loop)
   * `continue`: continue to start of nested loop
   * `break`: break to end from nested loop or block
   * `return`: return zero or more values from this function
   * `switch`: switch statement with fallthrough
+
+The condition operand of `if` and `do_while` has type `bool`.
 
 Loops (`do_while` and `forever`) may only be entered via fallthrough at the top.
 In particular, loops may not be entered directly via a `break`, `continue`, or
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -550,6 +550,10 @@ Truncation from floating point to integer where IEEE 754-2008 would specify an
 invalid operation exception (e.g. when the floating point value is NaN or
 outside the range which rounds to an integer in range) traps.
 
+Note that there are no conversion operators involving `bool`. Conversion from
+an integer type to `bool` can be performed by a comparison operator. Conversion
+from `bool to an integer type can be performed using the `select` operation.
+
 ## Feature test
 
 To support [feature testing](FeatureTest.md), an AST node would be provided:
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -505,7 +505,8 @@ is a `bool` value determining which of the other two operands to return.
 
   * `bool.and`: binary logical and
   * `bool.ior`: binary inclusive logical or
-  * `bool.xor`: binary exclusive logical or
+  * `bool.eq`: compare equal
+  * `bool.ne`: compare unequal
   * `bool.not`: unary logical negation
 
 ## Datatype conversions, truncations, reinterpretations, promotions, and demotions
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -34,9 +34,9 @@ exceeds the available callstack space at any time, a trap occurs.
 
 ## Types
 
-### Local Types
+### Basic Types
 
-The following types are called the *local types*:
+The following types are called the *basic types*:
 
   * `i32`: 32-bit integer
   * `i64`: 64-bit integer
@@ -47,11 +47,17 @@ Note that the local types `i32` and `i64` are not inherently signed or
 unsigned. The interpretation of these types is determined by individual
 operations.
 
-Parameters and local variables use local types.
-
 Also note that there is no need for a `void` type; function signatures use
 [sequences of types](#calls) to describe their return values, so a `void`
 return type is represented as an empty sequence.
+
+### Local Types
+
+*Local types* are a superset of the basic types, adding the following:
+
+  * `bool`: boolean
+
+Parameters and local variables use local types.
 
 ## Linear Memory
 
@@ -397,14 +403,15 @@ results into the result type.
   * `i32.clz`: sign-agnostic count leading zero bits (defined for all values, including zero)
   * `i32.ctz`: sign-agnostic count trailing zero bits (defined for all values, including zero)
   * `i32.popcnt`: sign-agnostic count number of one bits
+  * `int32.select`: sign-agnostic ternary `?:` operator (non-short-circuiting)
 
 Shifts interpret their shift count operand as an unsigned value. When the shift
 count is at least the bitwidth of the shift, `shl` and `shr_u` produce `0`, and
 `shr_s` produces `0` if the value being shifted is non-negative, and `-1`
 otherwise.
 
-All comparison operations yield 32-bit integer results with `1` representing
-`true` and `0` representing `false`.
+All comparison operations yield `bool` results. The first operand to `select`
+is a `bool` value determining which of the other two operands to return.
 
 ## 64-bit integer operations
 
@@ -415,8 +422,8 @@ The same operations are available on 64-bit integers as the those available for
 
 Floating point arithmetic follows the IEEE 754-2008 standard, except that:
  - The sign bit and significand bit pattern of any NaN value returned from a
-   floating point arithmetic operation other than `neg`, `abs`, and `copysign`
-   are not specified. In particular, the "NaN propagation"
+   floating point arithmetic operation other than `neg`, `abs`, `copysign`,
+   and `select` are not specified. In particular, the "NaN propagation"
    section of IEEE 754-2008 is not required. NaNs do propagate through
    arithmetic operations according to IEEE 754-2008 rules, the difference here
    is that they do so without necessarily preserving the specific bit patterns
@@ -482,11 +489,22 @@ implementations of the remaining required operations.
   * `f64.sqrt`: square root
   * `f64.min`: minimum (binary operator); if either operand is NaN, returns NaN
   * `f64.max`: maximum (binary operator); if either operand is NaN, returns NaN
+  * `float64.select`: sign-agnostic ternary `?:` operator (non-short-circuiting)
 
 `min` and `max` operations treat `-0.0` as being effectively less than `0.0`.
 
 In floating point comparisons, the operands are *unordered* if either operand
 is NaN, and *ordered* otherwise.
+
+All comparison operations yield `bool` results. The first operand to `select`
+is a `bool` value determining which of the other two operands to return.
+
+## Boolean operations
+
+  * `bool.and`: binary logical and
+  * `bool.ior`: binary inclusive logical or
+  * `bool.xor`: binary exclusive logical or
+  * `bool.not`: unary logical negation
 
 ## Datatype conversions, truncations, reinterpretations, promotions, and demotions
 


rossberg-chromium:
This change would complicate the type structure quite a bit, which the current text completely glosses over. In particular, would all conditionals now take a bool? How do you convert between an int32 and a bool, e.g. when you need to get/set/load/store it? Are these conversions explicit? If yes, it seems you would need them often. If not, you'd end up in the rabbit hole of implicit conversions, which seems highly undesirable.

sunfishcode:
> Would all conditionals now take a bool?

By 'conditional', do you mean 'comparison'? Then no. If you mean 'if-else', 'do-while', then yes. I can make that more explicit in the text.

> How do you convert between an int32 and a bool?

int32->bool: Equality comparison with zero.
bool->int32: Select between 0 and 1.

> Are these conversions explicit?

Yes.

titzer:
I don't see the added value of bools, especially if we start introducing
operations like conjunction, disjunction, xor, etc on bools that are just
expressible as integer operations. There won't be any code quality
improvements (e.g. TF just uses int32 for bools internally and reasons
about boolean formulas just fine), and we're just increasing the surface
area for no expressiveness benefit.

Maybe we can revisit this when we discuss SIMD vectors of bools, but I'd
prefer to avoid adding complexity drip by drip.

rossberg-chromium:
>
> How do you convert between an int32 and a bool?
>
> int32->bool: Equality comparison with zero.
> bool->int32: Select between 0 and 1.
>
> Are these conversions explicit?
>
> Yes.
>

So, every access to a boolean stored in a variable requires either a
compare or a select/conditional then?

sunfishcode:
@titzer If we wait until SIMD, it will be difficult to change how something as common as comparison operators work. Also, implementations can easily translate `bool` into `int32` if they wish to ignore it.

For what it's worth, this is the last major AST "change to consider" on my present TODO list.

@rossberg-chromium It depends on what the value is being used for, but if it's needed in a boolean context, then yes.

titzer:
On Fri, Aug 21, 2015 at 7:02 PM, Dan Gohman <notifications@github.com>
wrote:

> @titzer <https://github.com/titzer> If we wait until SIMD, it will be
> difficult to change how something as common as comparison operators work.
> Also, implementations can easily translate bool into int32 if they wish
> to ignore it.
>
Why would we want to change the existing operators for SIMD? We'll be
adding new local types for SIMD, one of which is presumably a vector of
bools. Packing and unpacking those vectors of bools can work with int32.

sunfishcode:
@titzer To keep the scalars operators symmetric with their SIMD counterparts, which isn't necessarily a direct optimization advantage, but it is a nice property to have from a big-picture perspective.

jfbastien:
I like that `bool` and its SIMD equivalent doesn't need a storage size definition, yet preserves the type information in the cfg *and* in/out of call sites.

lukewagner:
Another mild reason: I think `bool` would help a dumb compiler generate better code.  E.g., if you had `if (a < b) { ... }`, and use the strategy that "bool expressions set condition flags, not registers", then it's easy to compile to "cmp a b; jge" without a pattern match.  Otherwise (without pattern-matching), the `a < b` would naively need to compare and set a register and `if` would need to test its operand's register.  None of these pattern-matches we've discussed in this other issues breaks the bank, but it's a different form of the drip-by-drip argument.  I personally lean on the side of having the abstract insn set reflect the hardware insn set when possible (which seems to be the case for bool).

titzer:
One big advantage of preorder encoding is that "if (a < b)" will appear in
the stream as "[if][lt][a][b]", so by looking ahead one byte, the dumb
compiler can already pattern match many kinds of ifs. So I don't think bool
helps there.

kg:
Could we just introduce the concept of ```bool``` in the type system, but have it be equivalent to an ```int32``` with the caveat that the valid values are 0 and 1? This gives runtimes the option to optimize more aggressively but doesn't require a bunch of explicit conversions (unless you're actually intending to convert, say, ```3``` to a bool).

I think having comparison operators produce raw ```int32``` is gross, and it's more gross to be able to use raw ```int32``` as the variable for an ```if``` or a loop.

Would there be security risks if we did it this way? I.e. if the compiler promises it's generating code that only puts 0s or 1s into ```bool```-type values, is there a risk from the VM believing it (at least in release mode)? If the VM is simple and treats it as an int32, I think the codegen would be the same in both cases... but the (compiler) promise that it's 0 or 1 lets you use bitfields or condition flags or a single byte in the VM if you want to.

I've dealt with systems in the past that handled bools like this, and it was a little awkward but it turned out okay. The only corner case I remember was that if you tried to compare a ```bool``` directly with a literal you had to be careful (or lean on the compiler for help) - i.e. ```x == true``` might not evaluate the way you expect if someone crammed a ```7``` into the bool. But that sort of pattern is pretty easy to detect and convert into the correct comparison (```x != false```), because you know the types on the left and right-hand sides.

sunfishcode:
@kg I appreciate seeking hybrid approaches :-). However, I think making bool an int32 with nondeterminism if it's not 0 or 1 would have fewer advantages than a proper bool type -- it wouldn't achieve symmetry between scalar and SIMD, nor between int32 and int64 -- and it would also introduce the considerable disadvantage of significant nondeterminism in a core part of the spec.

rossberg-chromium:
What is the disadvantage of just making bool a regular type? You wouldn't
need to specify a storage size either, because it is unobservable.
Implementations could do whatever they see fit, as long as there are only
two observable values.

The only way a bool representation would potentially be observable is
through memory stores. You either do not provide those (requiring explicit
conversion to int), or you introduce bool _memory types_ of specified size
(like SIMD), and make the semantics such that it only observably writes 0
or 1.

sunfishcode:
@rossberg-chromium My proposal here seeks to do exactly that. If it's not clear, I'd like to clarify it.

The patch introduces a *basic type* category, with *local type* and *memory type* being distinct supersets. Local types are for local variables and parameters, and memory types are for loads and stores. Does that make sense, or is there a better way to organize this?

rossberg-chromium:
> @rossberg-chromium <https://github.com/rossberg-chromium> My proposal
> here seeks to do exactly that. If it's not clear, I'd like to clarify it.
>
> The patch introduces a *basic type* category, with *local type* and *memory
> type* being distinct supersets. Local types are for local variables and
> parameters, and memory types are for loads and stores. Does that make
> sense, or is there a better way to organize this?
>

Ah, okay, sorry, I somehow managed to misread the details of the type
hierarchy. On rereading, it is actually clear enough. (Though tbh, I'm not
convinced it is worth introducing the auxiliary notion of basic_type.)

So, why not have memory type(s) for bool?

rossberg-chromium:
Also not convinced of the utility of adding a family of select operators,
which are just conditionals. Either way, they are probably a separable
proposal.

sunfishcode:
I'm not attached to the "basic type" concept. If people find it cleaner to have fewer type categories and just describe what types can go where in the operator descriptions, I'm ok with that.

Should bool be non-storable? This avoids the mild awkwardness of defining a storage representation for it, and the possibility of needing atomic operations for it. And it avoids the need to implicitly canonicalize bool values -- a bool load followed by a store of the bool value would not be able to be a bit-preserving operation.

Should we have a non-short-circuit select? It aligns with how I tend to think of conditional-move instructions in hardware (select) as compared to branches (the short-circuit conditional operator), but I'd be ok removing this if people don't like it.

titzer:
If you can't load or store bools to memory, then dealing with any bools
stored in the memory requires conversions, which are totally useless. It
actually makes the code generated by dumb compilers worse and doesn't make
any difference on good compilers.

So you end up with if(int32ne(int32load_int8_(addr), #0)) and
int32store[int8](addr, (bool_toint32(x))).

sunfishcode:
I don't actually expect it to make code generated by dumb compilers worse, because it's what the underlying machine code is going to do even if we don't have a bool type. Load the byte, do a comparison, and branch. And for the store case: canonicalize the value to 0/1 (since without a bool type, a dumb compiler often always know what the value is), and store a byte.

Generating good code for the canonicalize requires a pattern match, but it's just one operation with some constant operands. Also, the canonicalize can easily be optimized away in the same cases that it could otherwise be easily avoided in slightly smarter compilers.

sunfishcode:
Actually, I'm wrong about the store case. If we don't have a bool type, wasm producers are the ones that can omit the canonicalization if they don't need it. If we do have a non-storable bool type, producers have to insert a canonicalization somewhere. Although there are clever ways to arrange this to allow it to be optimized away, this is not ideal. Let me think about this some more.

sunfishcode:
Ok, the answer is: Producers would have the flexibility to choose where they do the conversions. The problem of avoiding a canonicalization for a store only happens when the producer has passed a bool value across something like a function boundary, and then needs to store it. A C++ compiler for example would probably choose to pass bool values around as an optimization when it knows it doesn't need to do a store.

If a boolean value is consumed as part of an expression tree, there's no overhead to having a bool type in WebAssembly. This is a common case:
```
if (x < y)
  foo();
```

If a boolean value is stored to a variable or memory, having a bool type in WebAssembly would require converting it to integer and back:
```
bool b = x < y; // convert to integer via (x < y) ? 1 : 0
...
if (b) // convert to bool via b != 0
  foo();
```
Dumb compilers would be able to emit the same code that they would without those conversions.

jfbastien:
IMO there's also niceness across function boundaries, where the backend can have a neat calling convention about bools. It's impl-defined, but means compilers don't have to do full IPO to keep values in flag registers, or in predicate registers if the ISA has those (I miss having 8 predicates for the taking).

e.g.:
```
bool p = x < y;
bool r = foo(p);
```
Here both p and r can hang around in the flags if the implementation decides to be smart about the calling convention.

rossberg-chromium:
I've come around to liking this proposal. It makes the treatment of
Booleans a bit more abstract, which is probably a good thing. I don't see
any particular downside, other than the addition of one more type as such.
An implementation that is not interested in the distinction can just treat
it as a synonym for int32, but the extra bit of information will
potentially be handy in a couple of places.

That is under the assumption that there are also going to be appropriate
memory types for bool, so that there is a proper story for loads and
stores. I don't think it makes sense without that.

One operator that seems to be missing in the PR is `bool.eq`, btw. You
probably also want `bool.neq` for symmetry, in which case you can strike
`bool.xor`, because it is the very same thing.

sunfishcode:
Bool.eq and bool.ne sound good to me. I updated the PR to do that.

The proposal here is to specifically omit memory types for bool. In the spirit of keeping Booleans more abstract, this proposal seeks to avoid embedding a particular serialization of bool in the platform. As noted above, languages like C++ that have a bool type that can be stored to memory will be expected to convert "escaping" Boolean values to integer eagerly. Also, this is consistent with the anticipated SIMD Boolean types, which won't be directly storable either.

AndrewScheidecker:
Is there consensus to merge this?

titzer:
I think there are unresolved questions here. E.g. are we going to add loads
and stores for bools?

AndrewScheidecker:
I think that what's here is useful without memory operations.

I also think that the memory encoding is best left to the user through the integer memory ops. There are multiple useful ways to store bools in memory, and it doesn't seem very useful to only support one of them.

taisel:
Bool can exist across CPU status flags, and often set inside an arch-dependent way, or indirectly as the result of a comparsion operation, as well as push'd and pop'd in an arch dependent way to and from an arch dependent status register that may or may not exist or be allowed to be touched. I always see it as an optimization to allow bool, because arch-dependent optimizations can hoist bools to the status flags, rather than constantly spilled as integers that have to be re-checked via comparison instructions. Plus what a bunch mentioned about being able to store bools in a more compressed fashion than ints that may or may not be faster on some archs.

sunfishcode:
Closing; see additional discussion in #281.

PR #315, opened 2015-08-27
Reference IEEE 754-2008 specifically.

sunfishcode:
IEEE 754-2008 contains several useful improvements over IEEE 754-1985.
And should there be a future IEEE 754 document, we can update the
references at the time that we make any necessary semantic changes.


jfbastien:
Good idea to reference a specific version. lgtm, merging since this is straightforward.

Issue #292, opened 2015-08-10
Textual format: floating-point representation


jfbastien:
I'm opening up a can of bikeshed: how should the **textual** WebAssembly format represent floating-point numbers?

**This has nothing to do with the binary format.**

Keep in mind:
* Currently we support `f32` and `f64`.
* We'll eventually support vectors of these types.
* Must support NaN properly.
  - Do we support a leading sign?
  - We don't differentiate between different NaN types at the moment. Do we allow *n-char-sequence* after NaN as `printf` does?
  - How do we capitalize NaN (if we represent it as text at all)? `nan` `NaN` `NAN`.
* Must support `±infinity`.
  - How do we spell it (if we represent it as text at all)? `infinity` `INFINITY` `inf` `INF` `∞`.
* Must support `±0`.
* Must support denormals, but that's likely a non-issue.
* Not all numbers are accurately representable as `f32` or `f64`: decimal (e.g. `3.141592`) and exponential (e.g.  `6.022140e23`) often have such non-representable numbers that need to be rounded somehow.
* We can support multiple inputs for programmer's convenience, but then we're making the text format parser more complex (which doesn't really matter).

A few ideas:
* little-endian hexadecimal representation of the IEEE-754 number (e.g. `0x7FF8000000000000` is positive NaN with `0` bits). This is effectively a `reinterpret_cast` from integer to floating-point.
* C99 hexadecimal floating-point (e.g. `0x1p1` is 2.0).
* Precisely-representable decimal and exponential notations only.
* Any decimal and exponential notation, with a spec-mandated rounding.

Or a combination of the above.

kg:
Some thoughts:

* Syntax shouldn't have any unintentional similarity with integer literals. For example, ```0x7FF...``` is undesirable because you could treat it as an integer literal. Bit/byte-pattern representations like that should have distinct syntax like (strawman) ```[aa bb cc dd ee ff 00 11]``` that isn't used for any purpose other than raw byte literals.

* Keywords should be folded into a higher level scheme of keywords. Our textual format will already have other keywords, like types (```int32```, ```float32```, etc). Things like nan and infinity should be keywords of the same sort, but contextually valid only in places where a float literal is acceptable. I'm not convinced case should matter.

* I think we should just allow leading ```+``` or ```-``` signs on all signed literals (both int and float) and handle them appropriately. Much better than nuanced rules where the sign is only valid in some cases. This would include ```0``` and whatever infinity looks like.

[...]

PR #318, opened 2015-08-28
Text format: not unique, but precise

jfbastien:
Addresses the conclusion reached in #292.

### DIFF
--- a/TextFormat.md
+++ b/TextFormat.md
@@ -30,6 +30,6 @@ can be broken out into smaller pieces in the text format.
 
 The text format is precise in that values that cannot be accurately represented in the
 binary format are considered invalid text. Floating-point numbers are therefore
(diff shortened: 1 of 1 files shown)


jfbastien:
Small nit from talking with @sunfishcode : hex float doesn't address signaling versus quiet NaNs. I'm thinking about bringing that up with the C committee, it shouldn't be hard to fix.

jfbastien:
Updated to reflect that IEEE-754-2008 section "5.12.3 External hexadecimal-significand character sequences representing finite numbers 5.12 3.0" also standardizes this, and refers to ISO/IEC 9899:1999(E) Programming languages — C (C99).

jfbastien:
Assuming this is good to go based on the issue discussion.

PR #321, opened 2015-09-01
Be more explicit about module vs. instance

lukewagner:
Although the ml-proto is explicit about the distinction between (static) [module](https://github.com/WebAssembly/spec/blob/master/ml-proto/src/ast.ml#L130) and (dynamic) [instance](https://github.com/WebAssembly/spec/blob/master/ml-proto/src/eval.ml#L20), some clarification could help in the design repo (this has come up a few times).  In particular, this PR mentions that, with dynamic linking, a single instance could contain code from multiple modules.  There's more work to describe the dynamic-linking interaction in DynamicLinking.md; this PR is just a starter.

On the naming:
* `process` has a nice connotation wrt "multiple modules, one address space", but process also implies a more complete separation and so I think would be a better term to refer to the [multi-process support future feature](https://github.com/WebAssembly/design/blob/master/FutureFeatures.md#multiprocess-support) (wherein a wasm process could actually be expected to get a new OS process)
* `module_instance` is the current name in ml-proto, but this name suggests a 1:1 relationship between modules and instances, which won't be the case w/ dynamic linking.

Open to other names, though.

### DIFF
--- a/Modules.md
+++ b/Modules.md
@@ -1,7 +1,12 @@
 # Modules
 
 The distributable, loadable, and executable unit of code in WebAssembly
-is called a **module**. A module contains:
+is called a **module**. At runtime, a module can be loaded by a runtime
+to produce an **instance** which encapsulates all the state directly
+manipulated by running WebAssembly code. A WebAssembly instance's initial state
+is determined by the module it was loaded from.
+
+A module contains:
 * a set of [imports and exports](Modules.md#imports-and-exports);
 * a section defining the [initial state of linear memory](Modules.md#initial-state-of-linear-memory);
 * a section containing [code](Modules.md#code-section);
@@ -10,6 +15,13 @@ is called a **module**. A module contains:
 * possibly other sections in the future.
 Sections declare their type and byte-length. Sections with unknown types are
 silently ignored.
+
+An instance contains:
+* a [linear memory](AstSemantics.md#linear-memory);
+* [global variable](AstSemantics.md#global-variables) state;
+* (when [threading](PostMVP.md#threads) is added) thread-local state;
+* (when [dynamic linking](DynamicLinking.md) is added) the code of multiple modules
+  that have been linked into the same instance;
 
 While WebAssembly modules are designed to interoperate with ES6 modules
 in a Web environment (more details [below](Modules.md#integration-with-es6-modules)),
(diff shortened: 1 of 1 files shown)


rossberg-chromium:
LGTM

Modules.md
```diff
@@ -11,6 +16,13 @@ is called a **module**. A module contains:
 Sections declare their type and byte-length. Sections with unknown types are
 silently ignored.
 
+An instance contains:
+* a [linear memory](AstSemantics.md#linear-memory);
+* [global variable](AstSemantics.md#global-variables) state;
+* (when [threading](PostMVP.md#threads) is added) thread-local state;
```
jfbastien: I think it would be useful to clarify this a bit. "Thread-local state" sounds a lot like TLS, and I'm not sure that's what you mean. It's implementation detail, but worker context is the type of thread state that the embedder provides, not WebAssembly IIUC.

I guess this will get hairy... Maybe leave it as "thread-specific state" to avoid the TLS ambiguity, but leave it unspecified otherwise?

Modules.md
```diff
@@ -11,6 +16,13 @@ is called a **module**. A module contains:
...
+* (when [dynamic linking](DynamicLinking.md) is added) the code of multiple modules
+  that have been linked into the same instance;
```
jfbastien: What do you mean by "code"? The instance doesn't contain the main module's code, so mentioning the code from loaded modules is asymmetric.

Maybe: import/export tables, and globals?

sunfishcode: I think the instance *should* be considered to contain the main module's code, and that this is missing in the current module/instance split.

Modules.md
lukewagner: I did mean TLS as in, what you get if you use `thread_local`; this definitely seems like something wasm should provide and not be embedding-specific.

Modules.md
lukewagner: @sunfishcode Yes, that makes sense, and it reflects what's actually in `module_instance`.

jfbastien: Could you also mention imports/exports list?

Modules.md
jfbastien: That would be worth discussion with @lars-t-hansen and @juj, c.f.:
https://github.com/lars-t-hansen/ecmascript_sharedmem/issues/14

lukewagner: That conversation isn't quite relevant: asm.js *already has TLS* in the form of plain old asm.js global variables (which are, by semantics of JS, thread-local).  In fact, the implementation of asm.js global access in Odin literally uses a pinned (TLS) register.  Jukka didn't realize this initially but I have since pointed it out to him so I think he's using them now.  Thus, it makes sense for Lars to say it doesn't make sense to put TLS *in SAB*; JS already has TLS in spades :)

However, in wasm, we're not defining our globals in terms of JS, so we don't get TLS "for free" like we do in asm.js; we have to define it (otherwise wasm will have a distinct regression compared to asm.js).  FWIW, I think TLS can be quite symmetric with shared globals such that global declarations can be any element of the product `{shared, thread-local}` x `{aliased, unaliased}`; I was talking about this with @dschuff the other day in the context of dynamic linking.

Modules.md
lukewagner: Yes, makes sense.

Modules.md
jfbastien: That sounds good to me. I indeed filed the above issue with an eye towards wasm, without full understanding of how asm.js implemented TLS. It looks like we agree on what wasm would do (including w.r.t. dynamic linking + TLS), so that's good.

I do think it's worth clarifying this point though: call TLS out explicitly. Is there anything else worth mentioning? We will have some form of shadow stack that can be inspected programatically (it will become per-thread eventually). There's also some implicit context that's provided by the embedder.

lukewagner:
Added some more examples; I don't want to exhaustively list everything possible, so I added a "and other semantically-visible state added by other future features" at the end of the list.

jfbastien:
Thanks!

PR #323, opened 2015-09-01
Move "polyfill" support from MVP to FutureFeatures

lukewagner:
One big missing part of the MVP design that we've talked about a few times in passing and only has one super-vague bullet in the design repo is "polyfilling".  The basic idea is that, even if we have [feature testing](https://github.com/WebAssembly/design/blob/master/FeatureTest.md) (as an AST node, presumably), code like
```
(if (feature-test "foo") (foo ...))
```
will fail to validate if `foo` isn't supported (how would the engine decode the `foo` node if it has no idea what a `foo` is?).  The hand-wavy answer is: well, if we're allowing modules to declare their own local opcode spaces for [backwards compatibility reasons](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#backwards-compatibility), we could also allow modules at the same time to declare enough structure about conditionally-supported AST nodes (list of immediates and operands) that they could be generically decoded (into either a fault or a module-chosen call to a function that has a matching signature).

On closer inspection, this hand-wavy strategy seems problematic.  For one thing, it only makes sense for eager expressions; things like statements and types would require different mechanisms.  In general, every polyfillable kind of AST node would require explicit support to be polyfillable.  This could add up to a lot of work which we'd be adding to the MVP with no use cases and no real tests of the design.  Sounds like a fail on usual "MVP?" litmus test.

At the same time, it seems pretty likely that normal uses of WebAssembly (at least in the MVP) will involve user-space decoding (of [generic and/or specific compression](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md) into the raw layer 0 format).  If user code is decoding wasm op-by-op *anyway*, that seems like a fine place to implement polyfilling: the user code would have complete control, freeing it to do arbitrarily sophisticated polyfiling.  After some experience with user-space layer 1, we could later standardize.

Based on these this reasoning, this PR removes the mention of polyfill support in BinaryEncoding.md and creates a new FutureFeatures.md section.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -140,10 +140,6 @@ conflict-avoidance practices surrounding string names:
     * So, as a general rule, no magic numbers in the spec (other than the literal [magic number](https://en.wikipedia.org/wiki/Magic_number_%28programming%29)).
   * Instead, a module defines its *own* local index spaces of opcodes by providing tables *of names*. 
     * So what the spec *would* define is a set of names and their associated semantics.
-  * If the implementation encounters a name it doesn't implement, by default an error is thrown while loading.
-    * However, a name *may* include a corresponding polyfill function (identified by index 
-      into the function array) to be called if the name isn't natively implemented. (There are a lot
-      more details to figure out here.)
   * To avoid (over time) large index-space declaration sections that are largely the same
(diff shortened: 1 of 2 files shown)


sunfishcode:
lgtm

kg:
lgtm

PR #324, opened 2015-09-01
Say "instance" instead of "module" in a few places.

sunfishcode:
With #321 we now have a distinction between modules and instances. Updated a few more places to use "instances".


lukewagner:
lgtm, thanks!

sunfishcode:
Merging with lgtm, and since this is a fairly obvious follow-up to the module/instance naming change.

Issue #300, opened 2015-08-17
Atomicity: lock-free guarantees


jfbastien:
TL;DR: make all accesses of 1, 2 and 4 bytes "always" lock-free, wider access "maybe".

C++11 has a concept of "atomicity" where an operations is either entirely observable, or not at all. This guarantee is on a per object basis, during the object's lifetime—for our purpose this is equivalent to per memory location. This is irrespective of the object's size, but does require proper natural alignment (UB if not naturally aligned).

Not all hardware have load/store of wide sizes (indeed, atomics can have any size!), the language therefore offers an `is_lock_free` method on the `std::atomic` class. This is a per-type property, but it's effectively based on size. This method is determined at runtime (i.e. it is not `constexpr`), but it's effectively known at load time because it can't change during program execution.

Using an `atomic` of non-atomic size relies on the runtime using a global lock (or a shard of global locks) to seamlessly prevent partial reads or writes. Normal loads/stores are unaffected, only atomics of unusual sizes go through this lock.

C has `ATOMIC_*_LOCK_FREE` macros for `BOOL`, `CHAR`, `CHAR16_T`, `CHAR32_T`, `WCHAR_T`, `SHORT`, `INT`, `LONG`, `LLONG`, `POINTER`, which can be used at compile time. These macros are a tribool: never lock-free, sometimes lock free, always lock free.

There is a proposal to add [`constexpr atomic::is_always_lock_free`](http://wg21.link/N4509) to the languages.

Note: things are a bit more complicated because C11 decided to standardize `is_lock_free` as taking a pointer to the object as well. Implementations ignore that pointer, I suggest we do the same.

In all 3 cases we (WebAssembly implementors) need to decide what we want to guarantee. The runtime checks are easy (look at current machine, return that), the shard of global locks is easy (it's an implementation detail), but the compile-time guarantee of "yes this will always be lock free on all platforms WebAssembly will ever support" isn't that easy to make, and it's binding **forever**.

Having such a guarantee is pretty important: it's silly to write a lock-free algorithm if the datastructure is atomic through a lock (i.e. isn't actually lock free!). A good example for 64-bit lock-free being useful is a doubly-linked list where 2 pointers use up 64 bits.

@sunfishcode points out:
 * ppc32, ppc64, x86-32, x86-64, sparcv9, systemz, bpf, mips32, mips64, arm32, and arm64 can all do int32 lock-free.
[...]

PR #327, opened 2015-09-02
Lock-free guarantees

jfbastien:
Addresses #300.

### DIFF
--- a/Portability.md
+++ b/Portability.md
@@ -39,6 +39,9 @@ characteristics:
   processes executing on the same machine.
 * An execution environment which offers forward progress guarantees to all
   threads of execution (even when executing in a non-parallel manner).
+* Availability of lock-free atomic memory operations, when naturally aligned, for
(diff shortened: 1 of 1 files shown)


Portability.md
```diff
@@ -39,6 +39,9 @@ characteristics:
   processes executing on the same machine.
 * An execution environment which offers forward progress guarantees to all
   threads of execution (even when executing in a non-parallel manner).
+* Availability of lock-free atomic memory operations, when naturally aligned, for
+  8- 16- and 32-bit accesses. At a minimum this must include an atomic
+  compare-and-exchange operation (including load-linked/store-conditional).
```
sunfishcode: "(including load-linked/store-conditional)" could be misconstrued as requiring ll/sc. How about "(load-linked/store-conditional is permitted)" or so?

sunfishcode:
lgtm, with nit above.

PR #332, opened 2015-09-05
Consistent naming conventions for signedness, addendum

rossberg-chromium:
This would fit nicely with the convention and is more informative.


sunfishcode:
lgtm.

I can see arguments both ways here; thinking of arithmetic right shift as a signed integer operation entails thinking of it as performing a division operation that rounds differently from `div_s` for negative values, but it works.

titzer:
lgtm

jfbastien:
I updated LLVM accordingly:
http://reviews.llvm.org/rL247028

PR #339, opened 2015-09-11
Add feature test AST node

lukewagner:
We've discussed having a way to feature test from within wasm code (including in FeatureTest.md).  AST node makes it easy to const-fold.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -531,3 +531,12 @@ round-to-nearest ties-to-even rounding.
 Truncation from floating point to integer where IEEE 754-2008 would specify an
 invalid operation exception (e.g. when the floating point value is NaN or
 outside the range which rounds to an integer in range) traps.
+
+## Feature test
+
+To support [feature testing](FeatureTest.md), an AST node would be provided:
+
+  * `has_feature`: return whether the given feature is supported, identified by string
+
+In the MVP, `has_feature` would always return false. As features were added post-MVP,
+`has_feature` would start returning true.
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -539,4 +539,6 @@ To support [feature testing](FeatureTest.md), an AST node would be provided:
   * `has_feature`: return whether the given feature is supported, identified by string
 
 In the MVP, `has_feature` would always return false. As features were added post-MVP,
-`has_feature` would start returning true.
+`has_feature` would start returning true. `has_feature` is a pure function, always
+returning the same value for the same string over the lifetime of a single
+instance and other related (as defined by the host environment) instances.
--- a/FeatureTest.md
+++ b/FeatureTest.md
@@ -4,8 +4,8 @@ Any feature not present in [the MVP](MVP.md) will have a corresponding feature
 test. This allows polyfills to elegantly emulate a feature, and allows
 developers to fallback to another implementation if they so desire.
 
-Feature tests will be available from WebAssembly itself, as well as from
-JavaScript.
+Feature tests will be available from
+[WebAssembly itself](AstSemantics.md#feature-test), as well as from JavaScript.
 
-Details
-[are still be hashed out](https://github.com/WebAssembly/design/issues/90).
+See also [better feature testing support](FutureFeatures.md#better-feature-testing-support)
+in future features.


AstSemantics.md
```diff
@@ -531,3 +531,12 @@ round-to-nearest ties-to-even rounding.
 Truncation from floating point to integer where IEEE 754-2008 would specify an
 invalid operation exception (e.g. when the floating point value is NaN or
 outside the range which rounds to an integer in range) traps.
+
+## Feature test
+
+To support [feature testing](FeatureTest.md), an AST node would be provided:
+
+  * `has_feature`: return whether the given feature is supported, identified by string
```
jfbastien: "and is guaranteed to always return the same result during an instance's lifetime, and between different instances whose lifetime overlap."

jfbastien:
Should `has_feature` return a `bool`? Or an `int32` / `int64`/`f32`/some string type? E.g. can it tell me that SIMD width goes up to 256? WebGL version is 3.0?

AstSemantics.md
sunfishcode: I get being always the same within a given instance. What's the motivation for requiring different instances whose lifetime overlap to be the same? Clearly instances on different computers at the same time won't require their implementations to communicate, so where would we draw the line?

AndrewScheidecker:
What features do you envision being exposed this way? From @jfbastien's examples, SIMD capabilities seem like a better use case than WebGL: an API call is fine for checking GL capabilities. However, I would argue we should just generate SIMD code by default, and make sure the operations can be lowered to efficient scalar code for architectures that don't support SIMD.

One thing you couldn't handle with a has_feature operation is using operations that not all runtimes can decode. If it was instead ```(if_feature "feature name" a b)```, then the binary encoding of that operation could include an offset to skip over ```a``` without needing to know how to decode it.

sunfishcode:
AndrewSheidecker: SIMD code lowered to scalars is quite often slower than plain scalar code, because SIMD introduces several constraints that scalar doesn't have -- no branching (selects or even vector predication are less expressive), peel loops, increased register pressure, a need to structure data structures and loops to achieve SOA (vs AOS) even when that isn't otherwise beneficial, and so on. While [SIMD is not in the MVP](https://github.com/WebAssembly/design/blob/master/PostMVP.md) and doesn't need to be designed right away, we do expect it will be valuable to give applications some guidance on whether to use it and how to use it. I think string operands and boolean results will be sufficient for this.

@JSStats: queries that want to return more information can use dedicated AST nodes. Just as we have `page_size` instead of `sysconf` because `sysconf` is just a multiplex on top of separate queries anyway, we can add new operators instead of overloading `feature_test` if we have more needs.

Also, the proposal to have `string` be the operand is just at the conceptual AST level; the binary encoding would probably use a string table and indices in the obvious way.

AstSemantics.md
lukewagner: Agreed that once we have something bigger than an instance, things get fuzzy.  However, it does seem valuable to know that if module A probes a feature and chooses to load module B, B will get the same result.  I think we could bound this by saying "within the same user agent session" which is admittedly fuzzy, but probably good enough for now.

lukewagner:
Agreed with @sunfishcode that we want to keep this a bool value (literally `bool` if we add that type or `int32` if not) and scoped to just "feature testing", not "the generic hook for probing the browser for values" which could end up being a pandora's box of non-standard browser extension.

lukewagner:
> One thing you couldn't handle with a has_feature operation is using operations that not all runtimes 
> can decode.

@AndrewScheidecker That's a good question; see [this future feature](https://github.com/WebAssembly/design/blob/master/FutureFeatures.md#better-feature-testing-support) for a short discussion on this.

JSStats:
@sunfishcode An example, using page_size. Lets say some hot code was aligning indexes to the page size, and doing this in a lot of code, and doing so by masking off the low bits. The mask would be derived from the page_size, but it would help reduce the binary size if the code did not need to repeat this derivation on each reference. This could be done in asm.js by importing a constant global for the page mask initialized with the derived value. Adding dedicated AST nodes does not seem to support this well. Perhaps global constants would handle this better, and could also be used for features too, and yield a more general design.

sunfishcode:
@JSStats In that case, it would be an application with highly specialized needs, and it would be marginally suboptimal, and I'd consider it accordingly. Such an application would likely have more pressing concerns, such as the fact that the `page_size` operation doesn't return the actual page size of the hardware (as an aside, I'm wonder if we should rename it to avoid this confusion). We'd really want to consider what the application's underlying needs are, rather than hoping that speculative attempts at addressing the symptoms of its needs will be appropriate.

lukewagner:
Constant global imports and non-feature-testing implementation-defined values are a separate topic.  Let's stay focused here.

lukewagner:
Updated to mention purity throughout instance and left it up to the host environment to which other instances that purity extends.  @jfbastien look good?

AstSemantics.md
jfbastien: Agreed that we need to bound where lifetime observability extends! Clearly different machines don't need to return the same value :-)

What @lukewagner said sounds good.

AndrewScheidecker:
> no branching (selects or even vector predication are less expressive), peel loops, increased register pressure, a need to structure data structures and loops to achieve SOA (vs AOS)

Good points, though I can't see an application changing what data structures it uses based on a feature check.

Still, it seems like the biggest variable between architectures is not whether they support SIMD, but the specifics: vector width, alignment, what operations are supported, etc.

> Perhaps global constants would handle this better, and could also be used for features too, and yield a more general design.

The need for feature tests over global constants (like page_size or the base address of a relocatable data segment) is that a feature test has a value for all features, not just features the runtime knows about. If it's something every runtime implements, a global constant should be preferred.

So I can imagine code like this:
```(if_feature "SIMD" (global @simd_i32_vector_width) 1)```

Another thing feature testing needs to handle is syscall imports that are only defined if a feature is supported by the runtime.

jfbastien:
I agree with @sunfishcode on SIMD, and was about to write the same thing :-)

@AndrewScheidecker makes a good point on AST undefined nodes. As @lukewagner pointed out we've discussed it before, and this PR doesn't address the issue. I think we'll want to resolve this issue separately.


@lukewagner thanks for the update, this PR lgtm.

kg:
LGTM

jfbastien:
Merging based on lgtms.

@AndrewScheidecker could you open an issue on what you brought up, and refer to this PR?

JSStats:
@sunfishcode The use of the page_size was an example. A counter claim that the contrived use in this example is 'highly specialized' is really not fair, in fact this seems to have become a pattern of behaviour used by a few participants and I think some people need to consider their conduct. I would remind everyone of the code of conduct for this forum which includes 'Treat each other with respect, professionalism, fairness ...'

There seem to be good technical reasons that the use cases in the page_size example be considered, not because of any technical matter about page sizes in particular - the uses are similar to the need for a pre-process in C, and the technical challenges are to deal with this efficiently. If this were considered then the solution would be more general than this boolean features proposal and subsume it. Perhaps it's just my personal preference to have a design with more general support and to avoid overlapping and unnecessary specialized features, but technical matters have been ignored such as the efficiency of references to values derived from features or runtime values such as the page_size.

jfbastien:
@JSStats in general we try to evolve this repo with incremental steps, with some rare big shifts. I think this is an incremental step but agree with you that we may eventually want to evolve feature testing to provide more than just booleans in some way. I think we can evolve this, and base it on actual implementation experience.

It's taken us a while to find a good balance in how we work, and I think we've all learned to let things go before they're perfect, settle for incremental improvement, and accept revisiting things as they need polishing.

PR #342, opened 2015-09-11
Mention opcode name syntax for alignment

lukewagner:
This patch explains how alignment shows up in the opcode name, so design matches ml-proto.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -149,6 +149,12 @@ Each linear memory access operation also has an immediate positive integer power
 of 2 alignment attribute. An alignment value which is the same as the memory
 attribute size is considered to be a *natural* alignment.
 
+The linear memory operation names listed above specify natural alignment.
+To specify unnatural alignment, the opcode name can be suffixed with `/n` for
+`n` any integer power of 2. For example, `float64.load/2` specifies a `float32`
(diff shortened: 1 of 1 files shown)


AstSemantics.md
```diff
@@ -149,6 +149,12 @@ Each linear memory access operation also has an immediate positive integer power
 of 2 alignment attribute. An alignment value which is the same as the memory
 attribute size is considered to be a *natural* alignment.
 
+The linear memory operation names listed above specify natural alignment.
+To specify unnatural alignment, the opcode name can be suffixed with `/n` for
+`n` any integer power of 2. For example, `float64.load/2` specifies a `float32`
+load with 2-byte alignment; `int32.load16_s/1` specifies a signed 2-byte load
+that is unaligned (1-byte aligned).
```
jfbastien: I think this is a better fit for text format, since it's implicit in the bytecode format?

lukewagner: Do you mean that you think this para should go in TextFormat.md?  I kinda see that point, but opcode names are not just a detail of the text format; they also show up in the SExpr language as the names of AST nodes and will show up in the module local opcode table (which is part of the binary format).  AstSemantics is sortof the central hub, so I think it makes sense for it to define the names of opcodes.

jfbastien: Yeah, but I see your point too. lgtm then :-)

PR #343, opened 2015-09-11
Clarify alignment + addressing

jfbastien:


AstSemantics.md
```diff
@@ -149,6 +149,10 @@ Each linear memory access operation also has an immediate positive integer power
 of 2 alignment attribute. An alignment value which is the same as the memory
 attribute size is considered to be a *natural* alignment.
 
+When a memory access uses addressing as defined above and also specifies a
+non-natural alignment, then the alignment applies to the effective address
+(not merely the address operand).
```
sunfishcode: The alignment attribute *always* applies to the effective address, and never merely the address operand. This is described in the immediately following sentance, "If the effective address [...]".

jfbastien: Agreed, I'm trying to make that absolutely clear! You sounds like you though it already was?

sunfishcode: I did. I'm ok clarifying things, however the new paragraph here confused me a little. "When X then Y" in casual language may suggest "When not X then not Y", because if it doesn't then perhaps you should have just said "Y" or "Y, even when X".

jfbastien: Oh you and your formal-logic-English!

I clarified the clarification :-)

jfbastien:
Merging based on @sunfishcode's strong "OK".

PR #344, opened 2015-09-11
Move globals out of the MVP

lukewagner:
Based on a lot of discussion in #154, I'm proposing we remove globals from the MVP.  This PR also discusses the different things we want post-MVP: immutable global pointers for dynamic linking, mutable thread-local variables for threads, and a new future feature for real mutable global variables that captures the use cases.

### DIFF
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -376,3 +376,18 @@ is to allow a WebAssembly decoder to decode "through" an AST node that it knows
 nothing about. There are a number of ways to achieve this and more concrete
 experience with the realities of polyfilling is necessary to suggest the right
 design.
+
+## Mutable global variables
+
+In the MVP, there are no global variables; C/C++ global variables are stored in
+linear memory and thus accessed through normal
+[linear memory operations](AstSemantics.md#linear-memory-operations).
+[Dynamic linking](DynamicLinking.md) will add some form of immutable global
+variable analogous to "symbols" in native binaries. In some cases, though,
+it may be useful to have a fully mutable global variable which lives outside
+linear memory. This would allow more aggressive compiler optimizations (due to
+better alias information). If globals are additionally allowed array types,
+significant portions of memory could be moved out of linear memory which could
+reduce fragmentation issues. Langauges like FORTRAN which limit aliasing would be
+one use case. C/C++ compilers could also determine that some global variables never
+have their address taken.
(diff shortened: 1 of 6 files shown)


kripken:
lgtm.

FutureFeatures.md
```diff
@@ -364,3 +364,18 @@ is to allow a WebAssembly decoder to decode "through" an AST node that it knows
...
+In the MVP, there are no global variables; C/C++ global variables are stored in
+linear memory and thus accessed through normal
+[linear memory operations](AstSemantics.md#linear-memory-operations).
+[Dynamic linking](DynamicLinking.md) will add some form of immutable global
+variable analogous to "symbols" in native binaries. In some cases, though,
+it may be useful to have a fully mutable global variable which lives outside
+linear memory. This would allow more aggressive compiler optimizations (due to
+better alias information). If globals are additionally allowed array types,
+significant portions of memory could be moved out of linear memory which could
+reduce fragmentation issues. Langauges like FORTRAN which limit aliasing would be
```
jfbastien: Langauge fail.

jfbastien: Nowadays it's Fortran. No more yelling.

lukewagner: Done

jfbastien:
lgtm, but it would be good to leave this open until early next week, so @titzer and others who are supposed to be on vacation can take a look :-)

kg:
If we do this we should consider adding a way to explicitly set a symbol's numeric value. That will let us assign the heap locations of global state names in our test cases (and make it possible to round-trip them in a decompile/compile loop).

lukewagner:
At first I was confused by your use of "symbol" (interpreting "symbol" in the context of the hunk added to DynamicLinking.md by this PR) since immutable global pointers are post-MVP.  But I think you're talking about "debugging symbols", as in the separate, optional module section which is purely added to assist projection of binary to text.  In that case, I totally agree; I just added a section to TextFormat.md describing this interaction.

jfbastien:
We (@ncbray, @flagxor, @lukewagner, @kripken, @sunfishcode) had a chat about this offline, and agree that this patch can indeed land as-is. We had concerns that for MVP we may want a more restricted form of globals to address specific performance concerns, and Mozilla folks expressed their willingness to revisit new-MVP-globals if/when data shows that we need to revisit.

Issue #54, opened 2015-05-12
Draft an email for the LLVM community


jfbastien:
This documentation is great, but it's a bit TMI for the LLVM community and isn't the best way to say "hi, we want to work with you". I'll draft an email that we can send to the LLVM mailing list when we go public. The goal will be to extract specifically what they'll be interested in, to make sure we don't turn people off from the start.

The email will also serve as the main discussion point by LLVM folks, and I'm hoping to get good feedback from them early. We have good ideas about how we can avoid PNaCl and Emscripten's mistakes, we can write a good rationale, but I'm sure LLVM folks will find things to improve.

We want to involve them early, offer to change the design if needed, propose to upstream, and sign up to maintain the wasm backend.

PR #348, opened 2015-09-14
Add call_import to call section

lukewagner:
This patch tweaks the call section to match [#54](https://github.com/WebAssembly/spec/pull/54) in the spec repo.  As [explained](https://github.com/WebAssembly/spec/pull/54#issue-105841895), instead of attempting to inject imports into the function table (shared with normal function definitions), imports are given their own table called via `call_import` (instead of `call_direct`).  From what I can tell, this simplifies basically everything (spec, impl, code generator).

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -306,10 +306,14 @@ Direct calls to a function specify the callee by index into a function table.
 
   * `call_direct`: call function directly
 
-Calls must match the function signature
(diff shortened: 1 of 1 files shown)


jfbastien:
lgtm, but I'd leave open so other can chime in.

lukewagner:
Merging based on lgtm and no other objections.

PR #351, opened 2015-09-15
Memory type nit

jfbastien:
"Different superset" different from what? Just "superset" seems cleaner.


sunfishcode:
lgtm; this was left behind by earlier changes.

Issue #356, opened 2015-09-16
Tweak: switch to short type names in op names


lukewagner:
One significant remaining divergence between ml-proto and design is that ml-proto uses short type names (`i32` instead of `int32`).  We've agreed that we want opcode names to exactly match AstSemantics.md, so one side needs to change.  I actually like the short names better (shorter and no added ambiguity), so rather than changing ml-proto to match, how about we change design to use the short type names?

sunfishcode:
sgtm

jfbastien:
Agreed.

lukewagner:
Making a new issue for this was probably overkill.  I'll close this issue and we can discuss in #357.

PR #357, opened 2015-09-16
Use short type names (i32 instead of int32)

lukewagner:
As proposed in #356.


PostMVP.md
```diff
@@ -40,7 +40,7 @@ demonstrated in [PNaCl's SIMD][] and [SIMD.js][].
 
 SIMD adds new local types (e.g., `float32x4`) so it has to be part of the core
 semantics. SIMD operations (e.g., `float32x4.add`) could be either builtin
```
sunfishcode: It'd be nice to rename the SIMD types too -- `f32x4` etc.

sunfishcode:
lgtm, provided the SIMD types are likewise updated :-).

jfbastien:
lgtm

PR #361, opened 2015-09-19
Official text format, or lack thereof

jfbastien:
The official documents haven't really clarified what we're doing with the text format, and that leads to confusion:
http://stackoverflow.com/questions/32666337/will-web-assembly-wasm-have-its-own-syntax/

This edit mentions all the current text formats.

### DIFF
--- a/TextFormat.md
+++ b/TextFormat.md
@@ -34,7 +34,35 @@ represented as hexadecimal floating-point as specified by the C99 standard, whic
 IEEE-754-2008 section 5.12.3 also specifies. The textual format may be improved to also
 support more human-readable representations, but never at the cost of accurate representation.
 
-## Debug symbol integration
+# Official Text Format
+
+WebAssembly currently doesn't have a final, official, text format. As detailed above the
+main purpose of the text format will be for human consumption, humans feedback on readability
+will therefore factor into standardizing a text format.
+
+There are, however, prototype syntaxes which are used to bring up WebAssembly: it's easier
+to develop using a text format than it is with a binary format, even if the ultimate
+WebAssembly format will be binary. Most of these prototypes use [s-expressions][] because they
+can easily represent expression trees and [ASTs](ASTSemantics.md) (as opposed to CFGs).
+
+  [s-expressions]: https://en.wikipedia.org/wiki/S-expression
+
+* [Prototype specification][] consumes an s-expression syntax.
+* [WAVM backend][] consumes compatible s-expressions.
+* [sexpr-wasm prototype][] consumes compatible s-expressions, and works closely with the [V8 prototype][].
+* [LLVM backend][] (the `CHECK:` parts of these tests) emits compatible s-expressions.
+* [ilwasm][] emits compatible s-expressions.
+* [wassembler][] consumes a different syntax, and works closely with the [V8 prototype][].
+
+  [prototype specification]: https://github.com/WebAssembly/spec/tree/master/ml-proto/test
+  [LLVM backend]: https://github.com/llvm-mirror/llvm/tree/master/test/CodeGen/WebAssembly
+  [WAVM backend]: https://github.com/AndrewScheidecker/WAVM/tree/master/Test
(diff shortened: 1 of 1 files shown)


TextFormat.md
```diff
@@ -34,7 +34,35 @@ represented as hexadecimal floating-point as specified by the C99 standard, whic
 IEEE-754-2008 section 5.12.3 also specifies. The textual format may be improved to also
 support more human-readable representations, but never at the cost of accurate representation.
 
-## Debug symbol integration
+# Official Text Format
+
+WebAssembly currently doesn't have a final, official, text format. As detailed above the
+main purpose of the text format will be for human consumption, humans feedback on readability
```
kripken: either `human` or `humans'`, I think. but `feedback from humans` might be clearer

jfbastien: Done.

TextFormat.md
```diff
@@ -34,7 +34,35 @@ represented as hexadecimal floating-point as specified by the C99 standard, whic
...
+can easily represent expression trees and [ASTs](ASTSemantics.md) (as opposed to CFGs).
+
+  [s-expressions]: https://en.wikipedia.org/wiki/S-expression
+
+* [Prototype specification][] consumes an s-expression syntax.
+* [WAVM backend][] consumes compatible s-expressions.
+* [sexpr-wasm prototype][] consumes compatible s-expressions, and works closely with the [V8 prototype][].
+* [LLVM backend][] (the `CHECK:` parts of these tests) emits compatible s-expressions.
+* [ilwasm][] emits compatible s-expressions.
+* [wassembler][] consumes a different syntax, and works closely with the [V8 prototype][].
```
sunfishcode: I'm unclear about the purpose of this list. As the previous paragraph says, none of these represent the official text format.

TextFormat.md
```diff
@@ -34,7 +34,35 @@ represented as hexadecimal floating-point as specified by the C99 standard, whic
...
+# Official Text Format
+
+WebAssembly currently doesn't have a final, official, text format. As detailed above the
+main purpose of the text format will be for human consumption, feedback from humans on
+readability will therefore factor into standardizing a text format.
+
+There are, however, prototype syntaxes which are used to bring up WebAssembly: it's easier
+to develop using a text format than it is with a binary format, even if the ultimate
+WebAssembly format will be binary. Most of these prototypes use [s-expressions][] because they
+can easily represent expression trees and [ASTs](ASTSemantics.md) (as opposed to CFGs).
```
sunfishcode: I would say that we are prototyping using s-expressions because s-expressions transcend syntax.

TextFormat.md
jfbastien: People want examples, something to build off of. It's also great at showing divergences!

TextFormat.md
jfbastien: Done.

TextFormat.md
sunfishcode: But isn't the whole point of the previous paragraphs that these aren't necessarily what the official text format is going to look like?

jfbastien: Exactly. It's better to mention the divergences, than let people hunt for them and come to the wrong conclusions!

sunfishcode: If that's the purpose, then I suggest adding "Here is a list of things that are *NOT* the official text format, or anything like it:" immediately above the list, to clarify the purpose of the list.

jfbastien: Done :-)

sunfishcode: It still reads like it's trying to say "here's what the official format will look like; we just haven't finalized it yet. Can you preserve the "or anything like it" sentiment of my suggestion? :-)

jfbastien: Update.

sunfishcode:
I still don't really agree that we want a list of text formats that *aren't* related the official text format in a section about "the official text format", but since there appears to be a strong desire to include it, and since the text now at least isn't misleading if people read it carefully, I won't oppose it further.

flagxor:
lgtm, clearer than it was anyhow.

PR #364, opened 2015-09-20
Corrected memory-types for i32.load-instructions

pjuftring:
I guess someone copied the "sign/zero-extend i8 to i32" for i32.load16 but forgot to change i8 to i16 so I corrected that.
Furthermore I changed the "uint8" to "i8", maybe you want to have a "unsigned" here, but "uint8" is definitely not a defined memory-type.


jfbastien:
Thanks!

PR #366, opened 2015-09-23
Update AstSemantics.md

titzer:
It looks like loads and stores of i64 with sizes < 64bits got dropped on the floor. This PR adds them back and revises the text to no longer refer to memory types.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -118,8 +118,8 @@ Stores write only the bytes in the linear memory corresponding to their size.
   * `f32.store`: (no conversion) store 4 bytes
   * `f64.store`: (no conversion) store 8 bytes
 
-In addition to storing to memory, store instructions produce a value which is their
-input operand without conversion.
+In addition to storing to memory, store instructions produce a value which is their 
+`value` input operand without conversion.
 
 ### Addressing
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -84,10 +84,7 @@ will be possible to share linear memory between threads of execution when
 
 ### Linear Memory Accesses
 
-Linear memory access is accomplished with explicit `load` and `store` operations that
-accept as input an `index` operand that is interpreted as a unsigned byte offset into the
-linear memory.
-
+Linear memory access is accomplished with explicit `load` and `store` operations.
 Integer loads can specify a size which is smaller than the result type as well as a 
 signedness. Such integer loads perform an implicit sign- or zero-extension as specified below.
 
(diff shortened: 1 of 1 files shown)


AstSemantics.md
```diff
@@ -82,38 +82,47 @@ only communicate through messaging, e.g. in browsers using `postMessage`. It
...
-Wrapping of integers simply discards any upper bits; i.e. wrapping does not
-perform saturation, trap on overflow, etc.
-
-In addition to storing a value to linear memory, store instructions also
-reproduce their value operand, with no conversion applied.
+### Linear Memory Accesses
+
+Linear memory access is accomplished with explicit `load` and `store` operations that
+accept as input an `index` operand that is interpreted as a unsigned byte offset into the
+linear memory.
```
sunfishcode: They accept as input an `index` and an `offset` which are described in the [Addressing](https://github.com/WebAssembly/design/blob/master/AstSemantics.md#addressing) section immediately following this section.

AstSemantics.md
```diff
@@ -82,38 +82,47 @@ only communicate through messaging, e.g. in browsers using `postMessage`. It
...
+  * `i64.load8_u`: load 1 byte and zero-extend i8 to i64
+  * `i64.load16_s`: load 2 bytes and sign-extend i16 to i64
+  * `i64.load16_u`: load 2 bytes and zero-extend i16 to i64
+  * `i64.load32_s`: load 4 bytes and sign-extend i32 to i64
+  * `i64.load32_u`: load 4 bytes and zero-extend i32 to i64
+  * `i64.load`: load 8 bytes as i64
+  * `f32.load`: load 4 bytes as f32
+  * `f64.load`: load 8 bytes as f64
+
+In addition to an `index` operand, stores have an additional input operand which is the
```
sunfishcode: Ditto.

sunfishcode:
lgtm, other than a comment above.

titzer:
I just removed the references to indexing, since they are redundant with
the text in the next section.

lukewagner:
This PR undoes the explicit change in #326/#341/[spec/#55](https://github.com/WebAssembly/spec/pull/55) that @rossberg-chromium proposed and everyone (including [you](https://github.com/WebAssembly/design/issues/326#issuecomment-139465449)) lgtm'd which was to avoid including every possible coercion-on-load except those that were necessary to avoid introducing unwanted intermediate types (int8/int16).  I don't have a strong opinion either way, but the change was intentional and we'll need to revise spec/ (again) with this design change.

titzer:
It was an oversight on my part to not notice that the coercions for i64
were removed in rossberg's proposal.

I think it's important to keep them since otherwise load/stores working
with i64 and smaller sizes would require another explicit coercion.

lukewagner:
Ok; this is just the usual pattern-match vs. EITBI argument.  I generally lean to the latter, so I guess I'm actually happy to see this PR; I mostly just wanted to point out that the change wasn't accidental (at least for some).  I can update `spec/` to match this.

lukewagner:
Btw: if you search AstSemantics.md there is only one more dangling reference to memory type (in Addressing) and then we could remove "Memory type".

PR #373, opened 2015-09-25
Tidy up references to memory type after #366

lukewagner:
As suggested at the end of #366 and matching the analogous change in [spec/#86](https://github.com/WebAssembly/spec/pull/86).


jfbastien:
Overall lgtm, though *memory size* versus `memory_size` is slightly confusing IMO. Could we call the English concept *storage size* instead?

lukewagner:
Ah, good point, agreed.

PR #377, opened 2015-09-29
Rename "Literals" to "Immediates" and spell out operators.

sunfishcode:
"Immediate" is the more common term to use in assembly languages.

Also, remove a broken link to the binary encoding page. The binary
encoding is not yet determined, but AstSemantics.md doesn't need to
reference it right now anyway.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -346,10 +346,15 @@ function that returns multiple values will likely have to be a statement that
 specifies multiple local variables to which to assign the corresponding return
 values.
 
-## Literals
+## Immediates
 
-Each local type allows literal values directly in the AST. See the
-[binary encoding section](BinaryEncoding.md#constant-pool).
+Each local type allows immediate values directly in the AST. All possible
+values of all types are supported.
+
+  * `i32.imm`: i32 immediate
+  * `i64.imm`: i64 immediate
+  * `f32.imm`: f32 immediate
+  * `f64.imm`: f64 immediate
 
 ## Expressions with control flow
 


rossberg-chromium:
I don't think literals are the same as immediates. An immediate in assembly language is part of a given instruction. But the literal expressions in Wasm are not part of the surrounding opcode, they are (at least logically) their own, stand-alone instructions, "loading" constants. Moreover, they may or may not be implemented with CPU instructions having the constant as an immediate (for example, a float constant may need to be loaded from memory on some architectures, I believe; some constants can also be loaded from special registers, or be produced via specific opcodes).

So I disagree with this change; IMHO it is conflating abstraction levels.

titzer:
On Tue, Sep 29, 2015 at 12:55 PM, rossberg-chromium <
notifications@github.com> wrote:

> I don't think literals are the same as immediates. An immediate in
> assembly language is part of a given instruction. But the literal
> expressions in Wasm are not part of the surrounding opcode, they are (at
> least logically) their own, stand-alone instructions, "loading" constants.
> Moreover, they may or may not be implemented with CPU instructions having
> the constant as an immediate (for example, a float constant may need to be
> loaded from memory on some architectures, I believe; some constants can
> also be loaded from special registers, or be produced via specific opcodes).
>
> So I disagree with this change; IMHO it is conflating abstraction levels.
>

+ 1

lukewagner:
FWIW, I do think we have immediates, in the form of the integer operands to AST nodes (e.g., the index of `get_local` or the alignment of `load`).

sunfishcode:
That works for me. Immediates are things that are "part of the opcode", and we can call our "load immediate" instructions "literals" to avoid ambiguity. I submitted https://github.com/WebAssembly/spec/issues/88 to implement this in the spec.

PR #380, opened 2015-09-29
Spell out the "literal" operators.

sunfishcode:
Also, remove a broken link to the binary encoding page. The binary
encoding is not yet determined, but AstSemantics.md doesn't need to
reference it right now anyway.

See also https://github.com/WebAssembly/spec/pull/88

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -350,7 +350,7 @@ values.
 
 These opcodes have an immediate operand of their associated type which is
 copied into their result value. All possible values of all types are
-supported.
+supported (including NaN values of all possible bit patterns).
 
   * `i32.literal`: produce the value of an i32 immediate
   * `i64.literal`: produce the value of an i64 immediate
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -348,8 +348,14 @@ values.
 
 ## Literals
 
-Each local type allows literal values directly in the AST. See the
-[binary encoding section](BinaryEncoding.md#constant-pool).
+These opcodes have an immediate operand of their associated type which is
+copied into their result value. All possible values of all types are
+supported.
+
+  * `i32.literal`: produce the value of an i32 immediate
+  * `i64.literal`: produce the value of an i64 immediate
+  * `f32.literal`: produce the value of an f32 immediate
+  * `f64.literal`: produce the value of an f64 immediate
 
 ## Expressions with control flow
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -346,16 +346,16 @@ function that returns multiple values will likely have to be a statement that
 specifies multiple local variables to which to assign the corresponding return
 values.
 
-## Literals
+## Constants
 
 These opcodes have an immediate operand of their associated type which is
 produced as their result value. All possible values of all types are
 supported (including NaN values of all possible bit patterns).
 
-  * `i32.literal`: produce the value of an i32 immediate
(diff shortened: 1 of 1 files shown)


lukewagner:
Looks good to me, although I wonder if we could get by with the slightly more compact `lit`.

sunfishcode:
I'm ok with `lit` myself.

rossberg-chromium:
A bit more terminology nitpicking from me :) : "literal" usually refers to the textual representation of something. Again this is not really the right level to talk about these opcodes. I would rather call them constants.

sunfishcode:
"Constant" means unchanging, which can include local variables or other things. "Literals" (or even "immediates") properly express the property that the value is being provided immediately with the instruction.

rossberg-chromium:
That's "constant" as an adjective. Constant as a noun denotes a constant
value, I'd say. ;)

sunfishcode:
I frequently use "constant" as a noun to refer to non-immediate constants. If "literal" sounds too much like a text thing, that's why it's my second choice after "immediate" ;).

AstSemantics.md
```diff
@@ -348,8 +348,14 @@ values.
...
-Each local type allows literal values directly in the AST. See the
-[binary encoding section](BinaryEncoding.md#constant-pool).
+These opcodes have an immediate operand of their associated type which is
+copied into their result value. All possible values of all types are
+supported.
+
+  * `i32.literal`: produce the value of an i32 immediate
+  * `i64.literal`: produce the value of an i64 immediate
+  * `f32.literal`: produce the value of an f32 immediate
+  * `f64.literal`: produce the value of an f64 immediate
```
jfbastien: Should we explicitly point out that this preserve NaN sign, signal/quiet, and payload?

sunfishcode: Sounds reasonable. I added a mention of NaN values of all possible bit patterns.

WebAssembly presently has no concept of signaling vs. quiet NaN, so I think it's sufficient to just say all possible bit patterns.

jfbastien:
lgtm with either term.

I'll suggest `cons` and walk away slowly.

rossberg-chromium:
On 29 September 2015 at 21:20, JF Bastien <notifications@github.com> wrote:

> I'll suggest cons and walk away slowly.
>

That would of course have very confusing connotations with the major
S-expression languages. But with that extra `t` I'd be happy. :)

FWIW, 'constant' also is the terminology that both the JVM (ldc and
{i,l,f,d}const instructions) and the CLR (ldc instruction) use in this
context.

jfbastien:
> > I'll suggest cons and walk away slowly.
>
> That would of course have very confusing connotations with the major
> S-expression languages. But with that extra `t` I'd be happy. :)

If that wasn't clear, I was trolling FP programmers, and it looks like I hit my mark dead on :-)

titzer:
"const" also sounds fine to me. v8-native-prototype uses internal names
"I8Const", "I32Const", etc.

On Wed, Sep 30, 2015 at 5:26 PM, rossberg-chromium <notifications@github.com

sunfishcode:
Ok. I added a patch to this PR to use `const`.

jfbastien:
lgtm, though it's a bit confusion to talk about immediates in the English text, and `const` as the opcode and header :)

sunfishcode:
The value is given in an immediate field of the `const` instruction :).

rossberg-chromium:
lgtm. Thanks!

PR #387, opened 2015-10-05
Add a paragraph describing the prohibition on implicit TCO.

sunfishcode:
This corresponds with https://github.com/WebAssembly/spec/pull/103

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -42,7 +42,9 @@ WebAssembly will add some form of stack-introspection functionality, in which
 case such optimizations would be directly observable.
 
 Support for explicit tail calls is planned in
-[the future](FutureFeatures.md#general-purpose-proper-tail-calls).
+[the future](FutureFeatures.md#general-purpose-proper-tail-calls),
+which would add an explicit tail-call operation with well-defined effects
+on stack introspection.
 
 ## Types
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -36,13 +36,13 @@ Implementations are not permitted to do implicit tail-call optimizations.
 Semantically, implementations must have an internal maximum call stack size,
 and every call must take up some resources toward exhausting that size. Of
 course, dynamic resources may be exhausted much earlier. This rule exists
-because otherwise programs could be written to depend on tail call optimizations
-in some implementations and become unportable to other implementations. Also, in
-the future, it is expected that WebAssembly will add some form of
-stack-introspection functionality, in which case such optimizations would be
-observable.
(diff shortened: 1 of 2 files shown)


sunfishcode:
Wordsmith integrated. FutureFeatures.md section added. I rewrote "likely to be added" with "planned for the future" which is stronger without quite suggesting more commitment from all parties involved than I'm comfortable making in this PR.

AstSemantics.md
```diff
@@ -32,6 +32,18 @@ Callstack space is limited by unspecified and dynamically varying constraints
...
+Semantically, implementations must have an internal maximum call stack size,
+and every call must take up some resources toward exhausting that size. Of
+course, dynamic resources may be exhausted much earlier. This rule exists
+to avoid programs depending on tail call optimizations which may not be
+portable across implementations. Also, in the future, it is expected that
+WebAssembly will add some form of stack-introspection functionality, in which
+case such optimizations would be directly observable.
+
+Support for explicit tail calls is planned in
+[the future](FutureFeatures.md#general-purpose-proper-tail-calls).
```
lukewagner: To address the above objection, perhaps append "which would add an explicit tail-call operation with well-defined effects on stack introspection."?

sunfishcode:
Sounds good to me. Updated.

jfbastien:
I discussed this with @titzer and @flagxor: we'd rather specify this positively instead of negatively. It would be better to say that the transform should eventually hit a limit, so an optimizer can do TCO but has to either prove that the corresponding loop terminates or keep an implementation-defined limit to model resource exhaustion (as detailed in Nondeterminism.md).

sunfishcode:
@jfbastien If I remove the first sentence of the paragraph, does the rest satisfy your concern?

I added the first sentence as a "plain language" version of what follows, since the theoretical mechanism being inventing here is somewhat abstract and subtle, and I didn't want users or implementors to miss the main point. But I'm open to suggestions on how to clarify it.

jfbastien:
@sunfishcode yes, dropping the first sentence is good.

I'd also clarify this one:
> This rule exists to avoid programs depending on tail call optimizations which may not be portable across implementations.

Calls are already not portable because of their implementation-defined limit! I'd instead say something like:
> This rule exists to avoid transforming a program which necessarily runs out of resources into one which cannot possibly exhaust any resource.

sunfishcode:
@jfbastien Ok, I removed the first sentence and rewrote the other sentence. Does this work for you?

jfbastien:
lgtm

sunfishcode:
Merging with lgtm above and rough consensus in WebAssembly/spec#103.

Issue #384, opened 2015-10-02
Move `dont_need` into MVP?


lukewagner:
Weighing costs/benefits, I was thinking there's little reason not to bump [`dont_need`](https://github.com/WebAssembly/design/blob/master/FutureFeatures.md#finer-grained-control-over-memory) into the MVP: it's trivial to implement (if you don't want to do the implied sycall, you can just memzero), it'd be immediately used by the toolchain's malloc, and it would be immediately useful on mobile for reducing RSS and thus probability of oom-kill.

JSStats:
+1

kg:
sgtm

JSStats:
One challenge for current asm.js implementations will be to ensure that the array buffer elements start aligned to a page boundary. I don't think any other alignment is useful and that app developers should be able to depend on this alignment. This was practical on Odin last time I looked, but I have not looking in this for v8 and jsc implementations?

Could this alignment requirement be added to the wasm spec.

lukewagner:
Yes, this is vaguely implied in the [current text](https://github.com/WebAssembly/design/blob/master/FutureFeatures.md#finer-grained-control-over-memory) by "The `addr` and `length` parameters above would be required to be multiples of `page_size`." and would be made more explicit.

titzer:
I can see this being harmless, but it seems like it's a member of a class
of memory-control operations that should be designed to work together, like
memcpy, memove, memset, etc. Maybe it makes sense to decide for all of them
at once and do a chunked approach as opposed to a one-by-one approach?

lukewagner:
I think `dont_need` is in a fundamentally different category from those ops: it's not about performance, it's about exposing the raw ability to release physical pages associated with virtual memory (thereby decreasing RSS, etc).

Now I was thinking that maybe memcpy allows exposing the underlying OS functionality to remap pages (e.g., `mremap`), but if it has plain C semantics, I don't think we'll be able to use `mremap` or anything on Windows (but maybe there's an angle I'm missing).  Furthermore, with SIMD.wasm, I would think we'd be able to get pretty good performance for memcpy/memmove/memset from pure wasm (sure native can do AVX 512-bit copies, but I wonder if we're not just memory-bound by that point and so SIMD.wasm's 128-bit is sufficient).  Lastly, `dont_need` can be used to optimize `memset(0)` since semantically it leaves the pages 0.  So while I think we *definitely* want `dont_need`, I'm not sure we need builtin memset/memcpy/memmove.

[...]

PR #388, opened 2015-10-06
Rename `dont_need` to `discard` and mention `decommit` as an option

lukewagner:
As discussed in #384.

### DIFF
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -22,17 +22,19 @@ Provide access to safe OS-provided functionality including:
    copies the specified range from `Blob` into the range `[addr, addr+length)`
    (where `addr+length <= memory_size`) but implementations are encouraged
    to `mmap(addr, length, MAP_FIXED | MAP_PRIVATE, fd)`
-* `dont_need(addr, length)`: semantically, this operation zeroes the given range
-   but the implementation is encouraged to `madvise(addr, length, MADV_DONTNEED)`
-   (this allows applications to be good citizens and release unused physical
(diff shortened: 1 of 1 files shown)


FutureFeatures.md
```diff
@@ -22,17 +22,19 @@ Provide access to safe OS-provided functionality including:
...
+   POSIX)
 * `shmem_create(length)`: create a memory object that can be simultaneously
   shared between multiple linear memories
 * `map_shmem(addr, length, shmem, shmem-offset)`: like `map_file` except
   `MAP_SHARED`, which isn't otherwise valid on read-only Blobs
 * `mprotect(addr, length, prot-flags)`: change protection on the range
   `[addr, addr+length)` (where `addr+length <= memory_size`)
+* `decommit(addr, length)`: equivalent to `mprotect(addr, length, PROT_NONE)`
+  followed by `discard(addr, length)` and potentially more efficient than
+  performing these operations in sequence.
```
jfbastien: Getting `PROT_NONE` has a bigger meaning that decommitting IMO. I'm not sure what name would be better... "void" has other meanings.

lukewagner: I'm not sure what you mean.  Are you commenting on the naming of `decommit`?

jfbastien: Yeah, but without a better name to suggest I'll lgtm and hope someone else comes up with a nicer name before you commit :-)

lukewagner: Hah, thanks.  As we see here with `dont_need`=>`discard`, we can always rename.

Issue #385, opened 2015-10-04
Specify resize_memory to not be able to shink below the minimum memory.


JSStats:
It does not appear to be specified that `resize_memory` can not shrink the linear memory below the modules defined minimum. Odin makes good use of a known minimum and so could other wasm compilers. It can help avoid bounds checks for access to locations below the minimum, and some of these are expected to be hot such as locations of global variables.

lukewagner:
This sounds fine to me.  Technically what we have now is `initial`; I'd be inclined to have a separate `min` (with `min` defaulting to `initial`, just like `max`).

AndrewScheidecker:
Does this need an explicit minimum? It seems like a runtime that wants to implement this optimization can always reserve some minimum address-space (probably derived from the module initial memory size), and then rely on page faults to catch out-of-bounds accesses below that size.

lukewagner:
@AndrewScheidecker Depending on impl, that could require extra bookkeeping metadata and code paths for constant-offset accesses that otherwise couldn't possibly fail.

JSStats:
@AndrewScheidecker Yes, but if the implementation needs to keep this area of VM reserved anyway then resizing below the minimum does not free up any VM, so there seems little practical difference to using `dont_need`. Some implementations will choose inline bounds checks and it will help these to have a guaranteed minimum. If an application really wants to be able to shrink the memory below the initial size then a separate `minimum` and `initial` size seems to handle this case.

jfbastien:
At some point we had discussed making `resize_memory` return the value that the memory was actually resized to. This allows implementations to overallocate when it makes sense for them, and would also allow them to do what you describe (enforce a sensible minimum).

Could we instead do this?

lukewagner:
@jfbastien There was not agreement on whether to allow that nondeterminism in `resize_memory`.

jfbastien:
@lukewagner indeed, but this seems like a great reason to revisit. :-)

JSStats:
@jfbastien Yes that would seem to be fine to. So the code could ask `resize_memory` to shrink the memory and `resize_memory` could return the new size with zero difference rather than failing.

lukewagner:
@jfbastien As we've agreed in so many other cases, nondeterminism has to be *justified* by some significant improvement over what you can do deterministically and there seems to be no strong new reason here.

flagxor:
It seems like having an implicit contract with the code generator and the wasm implementation over magical sizes that result in better performance is undesireable.

[...]

PR #389, opened 2015-10-06
Only allow memory growth in MVP

lukewagner:
As discussed at the end of #385.


lukewagner:
Oh, I meant to add: this PR renames `resize_memory` to `grow_memory` so that, if we do add shrinking, it could be a separate operation, perhaps that is gated by some module-level declaration that avoids "surprise" shrinking (main module never shrinks but dylib does).  Even if we don't do that gating at the design level, having separate ops for growth/shrinkage could be useful for engines that want to statically observe the absence of shrinking.

jfbastien:
lgtm, splitting grow/shrink seems nice.

titzer:
lgtm too

PR #390, opened 2015-10-06
Add Rationale.md

jfbastien:
Move some of AstSemantics.md to that document. I left things pretty much as-is for now, and will defer more contended discussions to subsequent (and more targetted) edits.


Rationale.md
```diff
@@ -0,0 +1,215 @@
...
+* [Polyfill prototype][] shows simple and efficient translation to asm.js.
+
+  [JSZap]: https://research.microsoft.com/en-us/projects/jszap/
+  [Slim Binaries]: https://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.108.1711
+  [Polyfill prototype]: https://github.com/WebAssembly/polyfill-prototype-1
+
+
+## Basic Types Only
+
+WebAssembly only represents [a few types](AstSemantics.md).
```
lukewagner: Could you link explicitly to AstSemantics.md#types?

Rationale.md
```diff
@@ -0,0 +1,215 @@
...
+  language compiler to express its own types in terms of the basic machine
+  types. This allows WebAssembly to present itself as a virtual ISA, and lets
+  compilers target it as they would any other ISA.
+* These types are efficiently executed by all modern architectures.
+* Smaller types (such as `i8` and `i16`) are no more efficient and are only
+  semantically meaningful for memory accesses.
+* Other types (such as `f16`, `f80`, `i128`) aren't widely supported by existing
+  hardware and can be supported by runtime libraries if developers wish to use
+  them. They can be added to WebAssembly later without compromising MVP.
+* More complex object types aren't sematcically useful for MVP. They may become
```
lukewagner: *semantically

AstSemantics.md
```diff
@@ -156,53 +149,12 @@ when considering alignment.
...
- * Aligned accesses with less than natural alignment may be somewhat slower
-   (think: implementation makes multiple accesses, either in software or
-    in hardware).
- * Misaligned access of any kind may be *massively* slower
-   (think: implementation takes a signal and fixes things up).
-
-Thus, it is recommend that WebAssembly producers align frequently-used data
-to permit the use of natural alignment access, and use loads and stores with
-the greatest alignment values practical, while always avoiding misaligned
-accesses.
```
sunfishcode: The discussion of the affects of alignment wasn't intended as rationale; it's guidance to developers to know what to expect from alignment fields, and to implementors to know what they're expected to do with alignment fields.

AstSemantics.md
```diff
@@ -220,12 +172,9 @@ MVP, there are [future features](FutureFeatures.md#finer-grained-control-over-me
...
 contiguous linear memory.
 
-The result type of `page_size` is `int32` for wasm32 and `int64` for wasm64.
-The result value of `page_size` is an unsigned integer which is a power of 2.
+The result type of `page_size` is `i32` for wasm32 and `i64` for wasm64.  The
+result value of `page_size` is an unsigned integer which is a power of 2.
 
-(Note that the `page_size` value need not reflect the actual internal page size
-of the implementation; it just needs to be a value suitable for use with
-`resize_memory`)
```
sunfishcode: This clarification is also guidance, much like footnotes are used in the C++ standard. The markdown formatting we're using here doesn't support footnotes, but this content feels to me like it belongs in this document regardless.

AstSemantics.md
```diff
@@ -286,10 +236,7 @@ Each function has a *signature*, which consists of:
   * Return types, which are a sequence of local types
   * Argument types, which are a sequence of local types
 
-Note that WebAssembly itself does not support variable-length argument lists
-(aka varargs). C and C++ compilers are expected to implement this functionality
-by storing arguments in a buffer in linear memory and passing a pointer to the
-buffer.
```
sunfishcode: This is also footnote-like material which feels like it belongs here, though perhaps it could be fully parenthesized to indicate the informative nature.

Rationale.md
```diff
@@ -0,0 +1,215 @@
...
+
+You, the reader, may also be interested in understanding why
+[Abstract Syntax Tree (AST)](AstSemantics.md) semantics are the way they are,
+and digging through github's issues and pull requests becomes difficult.
+
+This rationale document tries to list how decisions were made, and where
+tradeoffs were made for the sake of language ergonomics, portability,
+performance, security, and Getting Things Done.
+
+The rationale for [limited local nondeterminism](Nondeterminism.md) is details
```
sunfishcode: typo: is details

lukewagner:
Moving the rationale out of AstSemantics.md into its own doc is great.  There's a number of FAQ entries that could perhaps be all moved to Rationale.md, replacing them with a Question "Why is _ specified as _?" who's Answer is "For the rationale behind various technical decisions in AstSemantics.md, see Rationale.md".

Rationale.md
jfbastien: Done.

Rationale.md
jfbastien: Done.

AstSemantics.md
jfbastien: Done.

Rationale.md
```diff
@@ -0,0 +1,215 @@
...
+
+WebAssembly was designed incrementally, and desisions were made with a solid
+rationale which may turn out to be incorrect in the future. Sometimes, decisions
+were made without the luxury of data, and the designers would like to revisit
+certain decisions once real end-to-end usecases are met and Science can be
+performed.
+
+You, the reader, may also be interested in understanding why
+[Abstract Syntax Tree (AST)](AstSemantics.md) semantics are the way they are,
+and digging through github's issues and pull requests becomes difficult.
```
sunfishcode: Are the above two paragraphs necessary? It's a rationale document; it's own purpose should be self-evident. And, I don't fully agree with the characterization of the process given here.

AstSemantics.md
jfbastien: Done.

Rationale.md
```diff
@@ -0,0 +1,215 @@
...
+immediate use. The following primitives provide AST nodes that express control
+flow and thus allow more opportunities to build bigger expression trees and
+further reduce `set_local`/`get_local` usage (which constitute 30-40% of total
+bytes in the
+[polyfill prototype](https://github.com/WebAssembly/polyfill-prototype-1)).
+Additionally, these primitives are useful building blocks for
+WebAssembly-generators (including the JavaScript polyfill prototype).
+
+
+## Limited Local Nondeterminism
```
sunfishcode: In the intro, it says that the rationale for limited local nondeterminism is in its own document, but then there's a section here too.

AstSemantics.md
jfbastien: Done.

Rationale.md
jfbastien: Done.

Rationale.md
```diff
@@ -0,0 +1,198 @@
...
+  hardware and can be supported by runtime libraries if developers wish to use
+  them. They can be added to WebAssembly later without compromising MVP.
+* More complex object types aren't semantically useful for MVP. They may become
+  useful to support other languages, especially when considering
+  [garbage collection](GC.md).
+
+
+## Load/Store Addressing
+
+Load/store instructions include and immediate offset used for
```
kripken: looks like a typo on "and"

Rationale.md
jfbastien: I think it's useful :-)

Which part don't you agree with? I'm open to changing it, but I do think the text makes stuff easier to understand.

Rationale.md
```diff
@@ -0,0 +1,198 @@
...
+
+The current specification is fairly strict when it comes to
+[limited local nondeterminism](Nondeterminism.md) of operations: it tries to
+specify all possible corner cases.
+
+As WebAssembly gets implemented and tested with multiple languages on multiple
+architectures there may be a need to revisit some of the decisions:
+
+* When all relevant hardware implement features the same way then there's no
+  need to get creative.
```
sunfishcode: What areas of the spec do you consider to be "creative"?

Rationale.md
jfbastien: Ah yes, I just updated that first paragraph, and added a link from Nondeterminism.md to here. The goal is to expand that section of Rationale.md later, since it may be a bit more contended.

Rationale.md
```diff
@@ -0,0 +1,198 @@
...
+* When different languages have different expectations then it's unfortunate if
+  WebAssembly measurably penalizes one's performance by enforcing determinism
+  which that language doesn't care about, but which another language may want.
```
sunfishcode: How about something like "The MVP's feature set is focused primarily on C/C++, but other source languages have different semantics in some cases. As we pursue support for more languages, we'll likely need to add more features, and generalize existing ones." ?

Rationale.md
jfbastien: Done.

Rationale.md
jfbastien: I added a clarification to this.

Rationale.md
jfbastien: I'm not sure that's quite true: if it were the case we'd avoid specifying e.g. shift and divide the way we do! I'm really hoping that we start of with something that seems good, and before MVP we adjust what we have in preparation for other languages.

Rationale.md
sunfishcode: If you think people reading this don't know what a rationale is, how about just "This document explains some of the decisions made in the WebAssembly design."

There has been a lot of Science going on around here, and this introduction feels like we're saying we've regularly sourced major parts of the spec from thin air.

And finally, saying "desisions were made with a solid rationale" feels like something we wouldn't say this unless there's doubt as to whether it is true.

Rationale.md
```diff
@@ -0,0 +1,201 @@
...
+specify all possible corner cases.
+
+As WebAssembly gets implemented and tested with multiple languages on multiple
+architectures there may be a need to revisit some of the decisions:
+
+* When all relevant hardware implement features the same way then there's no
+  need to get creative and add nondeterminism to WebAssembly when realistically
+  there's only one mapping from WebAssenbly expression to ISA-specific
+  operations. One such example is floating-point, where at a high-level most
+  basic instructions follow IEEE-754 semantics.
```
sunfishcode: WebAssembly already does take extensive advantage of IEEE-754. What are you implying that it's missing here?

Rationale.md
```diff
@@ -0,0 +1,201 @@
...
+[polyfill prototype](https://github.com/WebAssembly/polyfill-prototype-1)).
+Additionally, these primitives are useful building blocks for
+WebAssembly-generators (including the JavaScript polyfill prototype).
+
+
+## Limited Local Nondeterminism
+
+The current specification is fairly strict when it comes to
+[limited local nondeterminism](Nondeterminism.md) of operations: it tries to
+specify all possible corner cases.
```
sunfishcode: "tries to" suggests that we don't have a lot of confidence in how well we've done here.

The rationale for the handling of nondeterminism in WebAssembly is:
 - Nondeterminism is non-portable, and it complicates the task of reasoning about the possible states an instance could be in, so we avoid it by default.
 - Sometimes, eliminating nondeterminism is costly in performance, resource usage, or implementation complexity, and we accept nondeterminism when there are compelling reasons to do so.
 - When we do permit nondeterminism, we keep it limited and local in order to minimize the impact on portability and predictability.

Rationale.md
jfbastien: I'm saying all relevant HW does IEEE-754, so there's no point to spec otherwise (even if forth would like us to).

sunfishcode: We already take advantage of IEEE 754 (with only a few specific details otherwise, but that's a different topic), so this doesn't seem like something there "may be a need to revisit".

AstSemantics.md
```diff
@@ -286,10 +255,8 @@ Each function has a *signature*, which consists of:
...
+WebAssembly doesn't support variable-length argument lists (aka varargs), they
+are instead supported through explicit accesses to linear memory.
```
sunfishcode: Please parenthesize or otherwise demark the note about using linear memory to implement varargs in a way that makes it clear that this is just informative content, and not something that WebAssembly's own semantics do themselves.

AstSemantics.md
```diff
@@ -316,14 +283,11 @@ in the function table.
 
 Function-pointer values are comparable for equality and the `addressof` operator
 is monomorphic. Function-pointer values can be explicitly coerced to and from
-integers (which, in particular, is necessary when loading/storing to memory
-since memory only provides integer types). For security and safety reasons,
-the integer value of a coerced function-pointer value is an abstract index and
-does not reveal the actual machine code address of the target function.
+integers.
```
titzer: This section will conflict with https://github.com/WebAssembly/design/pull/392. Maybe just leave as is?

Rationale.md
```diff
@@ -0,0 +1,201 @@
...
+
+
+## Control Flow
+
+See [#299](https://github.com/WebAssembly/design/pull/299).
+
+
+## Locals
+
+Address-taken variables in the call stack are expected to be in linear memory,
```
titzer: I know this is a copy-over, but this might be a good opportunity to revise this section to make clear that address-taken is a concept only in certain languages, e.g. C/C++.

Rationale.md
jfbastien: I think that's what I'm saying?

AstSemantics.md
jfbastien: I'll merge when #392 lands (I think it'll make it first).

Rationale.md
jfbastien: Done.

Rationale.md
jfbastien: I moved some of the rationale to this document, and expanded on what you suggested.

I do think that we "try to"! Threads are a pretty big failure ;-)

AstSemantics.md
jfbastien: Done.

Rationale.md
sunfishcode: "Failed" suggests we tried to do something but didn't succeed. That's not what happened with threads. We looked at threads, and alternatives, and deliberately chose to make the tradeoff of accepting nondeterminism for performance in that case.

jfbastien: I see what you mean. How about "it specifies all possible corner cases when the WebAssembly authors deem it sensible"?

I guess it's better than ["I tried but I gave up"](https://www.youtube.com/watch?v=yVpw1SwJRBI).

Rationale.md
sunfishcode: I'm reading it and that's not what I'm getting?

> there may be a need to revisit some of the decisions:

> When all relevant hardware implement features the same way then there's no need to get creative and add nondeterminism to WebAssembly when realistically there's only one mapping from WebAssenbly expression to ISA-specific operations.

To me, this reads like you're saying that WebAssembly currently has places where we got "creative" and "there may be a need to revisit" this "decision".

"One such example is floating-point, where at a high-level most basic instructions follow IEEE-754 semantics."

And this suggests that we made up our own floating point semantics and that in the future "there may be a need to revisit" this "decision" and adopt IEEE 754.

Rationale.md
```diff
@@ -0,0 +1,216 @@
...
+bytes in the
+[polyfill prototype](https://github.com/WebAssembly/polyfill-prototype-1)).
+Additionally, these primitives are useful building blocks for
+WebAssembly-generators (including the JavaScript polyfill prototype).
+
+
+## Limited Local Nondeterminism
+
+The current specification is fairly strict when it comes to
+[limited local nondeterminism](Nondeterminism.md) of operations: it tries to
```
sunfishcode: The spec is also more than just "fairly strict". As of right now, there is no unlimited or non-local nondeterminism in the spec. And, we keep an explicit list of all the nondeterminism that we do have.

Rationale.md
```diff
@@ -0,0 +1,216 @@
...
+specify all possible corner cases.
+
+Ideally, WebAssembly would be fully deterministic because a fully deterministic
+platform is easier to reason about, implement and test portably. There are a few
+obvious exceptions where nondeterminism is essential to the API, like random
+number generators, date/time functions or input events. Nondeterminism is only
```
sunfishcode: API concerns are off-topic here.

Rationale.md
```diff
@@ -0,0 +1,216 @@
...
+specified as a compromise when there is no other practical way to achieve
+[portable](Portability.md) native performance, or when implementation leeway is
+desirable to allow usage of new hardware features or to security-harden certain
+usecases.
```
sunfishcode: Does WebAssembly currently have any nondeterminism motivated by security-hardening?

Rationale.md
sunfishcode: How about something closer to what I suggested [above](https://github.com/WebAssembly/design/pull/390#discussion_r41326616)?

Rationale.md
```diff
@@ -0,0 +1,216 @@
...
+## Basic Types Only
+
+WebAssembly only represents [a few types](AstSemantics.md#Types).
+* More complex types can be formed from these basic types. It's up to the source
+  language compiler to express its own types in terms of the basic machine
+  types. This allows WebAssembly to present itself as a virtual ISA, and lets
+  compilers target it as they would any other ISA.
+* These types are efficiently executed by all modern architectures.
+* Smaller types (such as `i8` and `i16`) are no more efficient and are only
+  semantically meaningful for memory accesses.
```
sunfishcode: It might be worth acknowledging that some 8-bit operations are actually faster than their 32-bit counterparts on some x86 chips (Agner). But (a) it's not a *lot* faster, and (b) in C/C++ all arithmetic operands are promoted to int anyway so the only time we see narrower arithmetic happen from C/C++ is when the optimizer proved it was safe, and in many of those cases, it's possible for a wasm backend to prove that it's safe to narrow for the same reason. And (c) these types don't add much over the existing types in terms of functionality, and leaving them out simplifies implementations.

Rationale.md
```diff
@@ -0,0 +1,216 @@
...
+* Other types (such as `f16`, `f80`, `i128`) aren't widely supported by existing
+  hardware and can be supported by runtime libraries if developers wish to use
+  them. They can be added to WebAssembly later without compromising MVP.
```
sunfishcode: I'd take `f80` out of this list, or if you really want to mention it, address it separately, because there would be significant hurdles to adding it to WebAssembly later, if it even could be considered.

x86-style `f80` is not portable to non-x86 machines, and even on x86 machines it requires specialized codegen support because of the stack-based register file, so it's a significant implementation burden. And as a soft-float type, it's less desirable than other soft-float types, in part because it has "unsupported" bit patterns sometimes known by names as pseudo-NaNs.

Rationale.md
```diff
@@ -0,0 +1,216 @@
...
+* More complex object types aren't semantically useful for MVP. They may become
+  useful to support other languages, especially when considering
+  [garbage collection](GC.md).
```
sunfishcode: "aren't semantically useful" doesn't explain the rationale. This would be a good place to explain that WebAssembly is seeking to provide the primitive building blocks upon which higher-level constructs can be built.

Rationale.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Control Flow
+
+See [#299](https://github.com/WebAssembly/design/pull/299).
+
+
+## Locals
+
+C/C++ makes it possible to take the address of a function's local values and
+pass this pointer to callees or other threads. Such address-taken variables in
```
sunfishcode: wording suggestion: "to callees or to other threads".

Rationale.md
```diff
@@ -0,0 +1,216 @@
...
+the call stack are expected to be in the WebAssembly linear memory, since they
+cannot be represented as locals. This prevents WebAssembly from performing
```
sunfishcode: They're more than just "expected" to be in linear memory; that's the only way they can work. Perhaps something like "Since WebAssembly's local variables are outside the address space, C/C++ compilers implement address-taken variables by creating a separate stack data structure within the linear memory. This stack is sometimes called the "aliased" stack, since it is used for variables which may be pointed to by pointers." ?

Rationale.md
```diff
@@ -0,0 +1,216 @@
...
+clever optimizations on the stack and liveness of such variables, but this loss
+isn't expected to be consequential.
```
sunfishcode: It's also worth noting that common C compiler optimizations such as what LLVM calls GVN effectively split address-taken variables into parts, shrinking the range where they actually need to have their address taken, and creating new ranges where they can be allocated as local variables in wasm.

Rationale.md
```diff
@@ -0,0 +1,216 @@
...
+
+Conversely, non-address taken values which are usually on the stack are instead
+represented as locals inside functions. This effectively means that WebAssembly
+has an infinite set of registers, and can choose to spill values as it sees fit
+in a manner unobservable to the hosted code. This implies that there's a shadow
+stack, which is also used to spill return values. This allows strong security
```
sunfishcode: The term "shadow stack" has many meanings and could be misinterpreted. Additional wording suggestion: "Since compilers tend to treat wasm's local variables as virtual registers, WebAssembly can be thought of has having an infinite virtual register file".

Rationale.md
```diff
@@ -0,0 +1,216 @@
...
+properties to be enforced, but does mean that two stacks are maintained (one by
+the VM, the other by the compiler which targets WebAssembly) which can lead to
+some inefficiencies.
+
+Local values can also be pre-colored, meaning that multiple incoming SSA values
```
sunfishcode: Coloring here is something which doesn't have an obvious meaning except for someone coming from an SSA background. Perhaps this could start by saying "local variables are not in SSA form; their value can be changed by a `set_local`. From an SSA perspective, this means that multiple independent values can share a local variable in wasm, which is effectively a kind of pre-coloring that clever wasm producers can use to pre-color variables and give hints to wasm implementation register allocation algorithms."

Rationale.md
sunfishcode: Ah, this was pre-existing text. Sorry for the noise.

Rationale.md
jfbastien: Removed for now, will move it to its own PR.

Rationale.md
jfbastien: Ah I see what you mean. That's not what I wanted to get across, I hope this new update is better.

Rationale.md
jfbastien: I refactored this heavily, and I think it includes everything you mentioned.

Rationale.md
jfbastien: Dropped "fairly".

Rationale.md
jfbastien: I think it's useful still: it's not what we're discussing, but "nondeterminism" means *many* things to different people, as we've been very good at proving! This paragraph explicitly puts API discussion out of scope.

Rationale.md
jfbastien: The discussions around function pointers have brought up this topic repeatedly. I don't want to dive into it for this PR, but that's one of the strong reasons to allow nondeterminism.

Rationale.md
jfbastien: I added a bit to this. I'd like to address this more in-depth in a separate PR, get this one in first.

Rationale.md
jfbastien: Done.

Rationale.md
jfbastien: Done.

Rationale.md
jfbastien: Done.

Rationale.md
jfbastien: Done.

Rationale.md
jfbastien: Done.

Rationale.md
jfbastien: I dropped shadow, but I already mentioned the inf. reg. thing so my rewording went a bit differently.

Rationale.md
jfbastien: Done.

jfbastien:
Alright this should be ready to land as I've moved all the contended stuff out and will file separate PRs for them. Good to go? @sunfishcode / @lukewagner

Rationale.md
```diff
@@ -0,0 +1,230 @@
...
+Dynamic linking further complicates this: WebAssembly cannot simply standardize
+on fixed-size function tables since dynamically linked code can add new
+functions, as well as remove them.
+
+
+## Expressions with Control Flow
+
+Expression trees offer significant size reduction by avoiding the need for
+`set_local`/`get_local` pairs in the common case of an expression with only one,
+immediate use. The following primitives provide AST nodes that express control
```
lukewagner: Need to reword b/c there are no "following primitives".

Rationale.md
```diff
@@ -0,0 +1,230 @@
...
+
+* Reason about.
+* Implement.
+* Test portably.
+
+Nondeterminism is only specified as a compromise when there is no other
+practical way to:
+
+* Achieve [portable](Portability.md) native performance.
+* Lowering resource usage.
```
lukewagner: s/Lowering/Lower/

lukewagner:
lgtm with two tiny nits above

Rationale.md
jfbastien: Done.

Rationale.md
jfbastien: Done.

Rationale.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Multiple Return Values
+
+TODO
+
+
+## Indirect Calls
+
+The exact semantics of indirect function calls, function pointers, and what
+happens when calling the wrong function, are still being discussed.
```
titzer: This is OK to land (since it was just yanked from elsewhere), but it will need to be revised because it's outdated.

titzer:
lgtm; we can wordsmith individual sections in follow up PRs.

sunfishcode:
lgtm for now; we can continue to iterate in follow-up PRs.

PR #394, opened 2015-10-08
Contributing: affiliation

jfbastien:
We were asked to follow this affiliation rule a while ago, had been applying it, but I'd forgotten to write it down!


lukewagner:
Ah, right.

PR #397, opened 2015-10-09
Update FAQ

jfbastien:
I was going to move some of it to rationale.md (as discussed in #390), but I figure I should format things similarly, and update some of what's now outdated first.


lukewagner:
lgtm

PR #400, opened 2015-10-12
Added else child to the if statement.

Teemperor:
Without an else branch we would need a temp local for writing normal if/else statements. If someone just wants an normal if statement he/she can just use a empty block as an else.

Other suggestion if changing `if` is not up to debate: Add an `ifelse` and keep `if` the way it is.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -237,7 +237,7 @@ WebAssembly offers basic structured control flow. All control flow structures
 are statements.
 
   * `block`: a fixed-length sequence of statements
-  * `if`: if statement with `then` and `else` child
+  * `if`: if statement with `then` and `else` children
   * `do_while`: do while statement, basically a loop with a conditional branch
     (back to the top of the loop)
   * `forever`: infinite loop statement (like `while (1)`), basically an
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -237,7 +237,7 @@ WebAssembly offers basic structured control flow. All control flow structures
 are statements.
 
   * `block`: a fixed-length sequence of statements
-  * `if`: if statement
+  * `if`: if statement with `then` and `else` child
   * `do_while`: do while statement, basically a loop with a conditional branch
     (back to the top of the loop)
   * `forever`: infinite loop statement (like `while (1)`), basically an


kg:
This is already the case for ml-proto, so lgtm.

rossberg-chromium:
Actually, ml-proto allows else-less if (and treats it as `else nop`).

AstSemantics.md
```diff
@@ -237,7 +237,7 @@ WebAssembly offers basic structured control flow. All control flow structures
 are statements.
 
   * `block`: a fixed-length sequence of statements
-  * `if`: if statement
+  * `if`: if statement with `then` and `else` child
```
jfbastien: Children?

Teemperor:
Thanks JF, fixed the spelling!

jfbastien:
Is this mandating that both be present? Or is "else" optional?
Or should we have "ifelse" and "ifthenelse"?
@sunfishcode?

Teemperor:
IMHO: The problem with an optional `else` branch is that it's weird to encode in the binary format (in a sexpr it is no problem obviously). An `ifthen` and `ifthenelse` could be shorter in binary (even though I have no idea about the planned compression which has a big influence on that).

rossberg-chromium:
I'm not sure I understand the issue. This seems to be a concern about the binary format, not the AST semantics itself.

Teemperor:
The current AST encoding is to my knowledge something like `opcode {opcode data} {child-node}` and ifthen/ifthenelse would save us many `nop` child nodes in the else child.

So instead of `if-opcode condition then-node **nop**` for a simple if, one could use `ifthen-opcode condition then-node` and save the byte for the `nop` opcode.

I know the discussion is not about the binary format, but the preliminary format would suffer from that issue (and I assume it will not change that much).

kg:
```(if (...) (nop))``` and ```(if (nop) (...))``` aren't especially difficult to encode, will stream compress okay, and shouldn't amount to that much of an executable's size. It's not really worth worrying about it at this stage, I think. We can introduce specialized if opcodes later.

Or we could introduce them now as syntax sugar, I guess, and reserve the right to turn them into actual opcodes if we need to.

lukewagner:
lgtm to land as is and we can refine layer as we get serious about the binary encoding.

sunfishcode:
With #427 merged, there are now separate `if` and `if_else` operators.

Issue #399, opened 2015-10-12
Order of evaluation for child nodes is not specified.


Teemperor:
It seems the spec is missing how child nodes are evaluation. Problems that could arise through that are:

1 . Child nodes with side effects would make the program behave differently on different platforms like in:

    (set_local $x (i32.const 8))
    (int32.div (get_local $x) (set_local $x (i32.const 4))

div could either return 1 or 0 in this case.

2 . Unnecessary locals to clarify the order of evaluation. Example from above with clear order but unnecessary locals:

    (set_local $x (i32.const 8))
    (set_local $right (set_local $x (i32.const 4))
    (int32.div (get_local $x) (get_local $right)

AndrewScheidecker:
I thought I saw it define evaluation to occur in lexical order somewhere, but I can't find the reference now. Maybe it got lost in the shuffle?

pjuftring:
[TestingTodo](https://github.com/WebAssembly/spec/blob/master/ml-proto/TestingTodo.md) says, that operands have to be evaluated from left to right.

jcbeyler:
Tests for this already exist by the way:
https://github.com/WebAssembly/spec/pull/119

Not saying this should not be stated :); just saying we have tests for it now

lukewagner:
The intention is indeed for evaluation order to be [deterministic](https://github.com/WebAssembly/design/blob/master/Nondeterminism.md), in the order defined by the [spec](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/eval.ml#L211).  I agree it would be good to note this explicitly in AstSemantics.md, though.

lukewagner:
Merged #401.

PR #401, opened 2015-10-12
Added order of evaluation

Teemperor:
Added order of evaluation notice for arithmetic operations. All other operations are to my knowledge either unary or have an obvious ordering (e.g. `if`, `do_while`).

Also see issue #399 .

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -179,6 +179,9 @@ Thus, it is recommend that WebAssembly producers align frequently-used data to
 permit the use of natural alignment access, and use loads and stores with the
 greatest alignment values practical, while always avoiding misaligned accesses.
 
+### Order of evaluation
+
+All operands of arithmetic operations are evaluated left-to-right.
 
 ### Out of Bounds
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -181,7 +181,11 @@ greatest alignment values practical, while always avoiding misaligned accesses.
 
 ### Order of evaluation
 
-Operand evaluation order is deterministic. All nodes other than control flow constructs need to evaluate their child nodes in the order they appear in the AST.
+Operand evaluation order is deterministic. 
+All nodes other than control flow constructs need to evaluate their child nodes in the order they appear in the serialized AST.
+
+For example, this s-expression presentation of a i32.add node `(i32.add (i32.const 2) (get_local $x))`
+would first evaluate the `(i32.const 2)` node and afterwards the `(get_local $x)` node.
 
 ### Out of Bounds
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -181,7 +181,7 @@ greatest alignment values practical, while always avoiding misaligned accesses.
 
 ### Order of evaluation
 
-All operands of arithmetic operations are evaluated left-to-right.
+All operands of arithmetic operations and the arguments of function calls are evaluated left-to-right.
 
 ### Out of Bounds
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -181,7 +181,7 @@ greatest alignment values practical, while always avoiding misaligned accesses.
 
 ### Order of evaluation
 
-All operands of arithmetic operations and the arguments of function calls are evaluated left-to-right.
+All statements other than control flow constructs need to evaluate their child nodes in the order they appear in the AST.
 
 ### Out of Bounds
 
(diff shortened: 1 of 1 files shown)


rossberg-chromium:
The other relevant case are call arguments.

Teemperor:
Oh true, fixed that. Thanks!

sunfishcode:
And store operands :-). And possibly more things in the future too. How about something like "AST nodes other than control flow constructs"?

And I know I wrote "left to right" in TestingTodo.md, but thinking about this more, "left" and "right" are relative to a textual presentation of the AST. How about "in order from first to last"?

Teemperor:
Both binary/text presentation of the AST are an ordered sequence of symbols, so I wrote: "All statements other than control flow constructs need to evaluate their child nodes in the order they appear in the AST.".

sunfishcode:
Looks good; I like that wording. Except that we haven't officially decided whether statements and expressions are the same thing or not yet :-}. How about 'nodes'?

lukewagner:
Perhaps you could start with the sentence "Operand evaluation order is deterministic." (so that it's clear that even control flow nodes have a deterministic evaluation order, it's just not quite as simple as the other exprs).

AstSemantics.md
```diff
@@ -179,6 +179,9 @@ Thus, it is recommend that WebAssembly producers align frequently-used data to
 permit the use of natural alignment access, and use loads and stores with the
 greatest alignment values practical, while always avoiding misaligned accesses.
 
+### Order of evaluation
+
+Operand evaluation order is deterministic. All nodes other than control flow constructs need to evaluate their child nodes in the order they appear in the AST.
```
titzer: This wording isn't very clear; e.g. there are a number of different "orderings" for tree walks.

Suggest wording the first sentence as "When evaluating an operation, operands are always evaluated in 'left-to-right' order" and then given an example, such as int32_add.

Teemperor:
Yeah, the wording wasn't clear. I should have written"Operand evaluation order is deterministic. All nodes other than control flow constructs need to evaluate their child nodes in the order they appear in the **serialized** AST.". Both text and binary will probably imply the node order by the order in which they appear in the file, so this should prevent confusion.

I will add an example.

AstSemantics.md
```diff
@@ -179,6 +179,13 @@ Thus, it is recommend that WebAssembly producers align frequently-used data to
 permit the use of natural alignment access, and use loads and stores with the
 greatest alignment values practical, while always avoiding misaligned accesses.
 
+### Order of evaluation
+
+Operand evaluation order is deterministic. 
+All nodes other than control flow constructs need to evaluate their child nodes in the order they appear in the serialized AST.
+
+For example, this s-expression presentation of a i32.add node `(i32.add (i32.const 2) (get_local $x))`
+would first evaluate the `(i32.const 2)` node and afterwards the `(get_local $x)` node.
```
jfbastien: Neither of these are observable, it would be better to have function calls as both operands.

jfbastien:
It would be good to also add that a function is executed top-to-bottom, or something like that to express how control flows. Expressions aren't operands to control-flow operations, so it doesn't just derive from the current definition.

jfbastien:
This will be contended when we add atomics: those can be evaluated in a non-linear order. You can try to forbid it at the compiler level, but fundamentally a program with a race can be observed at the hardware level so it's silly to forbid the compiler from reordering operations when the hardware can already do it. Notably, it prevents interesting loop optimizations, as well as vectorization.

I know threads aren't spec'd yet, but I'd like this section to already document that loads and store can be reordered. Hardware will do it anyways, and we'd be leaving significant performance on the table otherwise.

I'm happy to move this to a follow-up PR if it's contended.

AstSemantics.md
```diff
@@ -179,6 +179,13 @@ Thus, it is recommend that WebAssembly producers align frequently-used data to
 permit the use of natural alignment access, and use loads and stores with the
 greatest alignment values practical, while always avoiding misaligned accesses.
 
+### Order of evaluation
+
+Operand evaluation order is deterministic.
```
jfbastien: Two newlines between paragraphs.

sunfishcode:
@jfbastien Within a single thread, the thread's own view of its own operations is deterministic. Loads and stores (atomic or otherwise) can be reordered, but only within the purview of the as-if rule. Interactions between different threads is a separate question.

jfbastien:
Sounds like @sunfishcode and I agree, but I'm happy to move this to a separate PR nonetheless once this one lands to make sure the wording is solid. I did have a few other comments that needed to be addressed.

Teemperor:
I'm using now "The evaluation order of child nodes is deterministic.".

Also the example is using now `set_local` so that the evaluation order is easier to observe.

PR #402, opened 2015-10-12
Added identifying information to the LICENSE.

Teemperor:
I'm not a lawyer, but I'm pretty sure that this LICENSE is not valid without a year and a name.


jfbastien:
This is in the section "APPENDIX: How to apply the Apache License to your work.". The line you're editing is what we should put inside of source files. I don't think this needs to change.

Teemperor:
True, nevermind

PR #403, opened 2015-10-13
Add `br_if` and `br_switch`.

sunfishcode:
This adds `br_if` and `br_switch` to AstSemantics.md, essentially as suggested in https://github.com/WebAssembly/design/pull/299 . The semantic description here is very brief; for full details see the corresponding spec PR, https://github.com/WebAssembly/spec/pull/131 .

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -246,6 +246,11 @@ are statements.
   * `break`: break to end from nested loop or block
   * `return`: return zero or more values from this function
   * `switch`: switch statement with fallthrough
+  * `br_if`: conditional branch to end from nested loop or block
+  * `br_switch`: a list of pairs of nested loop or block references and
+     immediate indices, a default loop or block reference, and an index operand.
+     Branch to the end of the indexed loop or block if in bounds, otherwise the
+     default loop or block
 
 Loops (`do_while` and `forever`) may only be entered via fallthrough at the top.
 In particular, loops may not be entered directly via a `break`, `continue`, or


rossberg-chromium:
Do we actually have consensus on `br_switch`? As far as I recall the discussion in #299, the only thing people could agree to (and not particularly enthusiastically) was adding `break_if`/`break_continue`. But `br_switch` is a significantly more complicated operator. It's fine changing the switch operator in certain ways (e.g., making it fallthrough by default, and/or making it table switch over a range), but I doubt we should have several switch operators.

In general, I'm still not sure where all this is going, and whether I agree with the direction.

lukewagner:
@rossberg-chromium The argument that convinced me in #299 is that the natural way to transform a CFG into structured control flow is to generate tons of if-break's and switch-break's and so it makes sense to have these available as primitives (less ops to decode and no need to pattern-match to fold branches means faster decoding and better performance for dumb compilers).  `br_if` and `br_switch` seem symmetric in this regard, as they are two pieces of the same algo; it'd be weird to only have one.  In terms of engine complexity, I think it will be negligible (both `switch` and `br_switch` would decode into the same engine-internal IR op).

Also, with `br_switch`, you get fallthrough "for free", and so `switch` could keep its simple, high-level definition.

sunfishcode:
After several offline discussions, I'm now working on a new hybrid proposal that attempts to merge the best of both approaches. I'll open a new PR when the new plan is ready.

Issue #223, opened 2015-06-24
Eliminate Expression/Statement distinction in AST?


titzer:

qwertie:
While I haven't worked this out in detail, I know from designing [Enhanced C#](https://github.com/qwertie/Loyc/wiki/Enhanced-C%23) (unfinished) and [LES][] that languages in the Algol tradition (C/C++/Java etc) can be viewed as purely expression-based languages like LISP or Haskell.

As I was saying elsewhere, I think all executable code (if not more of the system) should be expression-based.

- If `void` (`unit`, `()`) is understood as a proper data type, then `if` (without else) and `do_while` can be seen as expressions that return `void`, and if-else can be seen as equivalent to the conditional operator `?:`
- Side effects can occur inside an expression, and the comma operator provides sequences. Everyone is already familiar with this effect in C, e.g. `(x = y * y, (x += y) + 5)`. Discarding a value happens naturally whenever one operation (whose result you want to discard) is followed by any other operation. However if the operation whose result you want to discard is the last operation in a block, you have to add a dummy 'void' (`foo(), bar(), void`)
- `forever` can be seen as an expression whose arguments are the statements inside the loop, and `break` could include a 'return value', which will be the value of an enclosing `forever` loop.
- `return` is not needed; return values mid-function can be simulated by enclosing the function in a `forever` that uses `break` to specify the return value.

Merging statements and expressions into a single concept called 'expressions' seems to produce a strict superset of the current plan. I think it would be a simpler and more capable model than what is currently proposed, and it won't substantially alter code generation for the Algol branch of languages (or, if it turns out that it does, perhaps the language could be tweaked until it doesn't) The translation to asm.js could be more complicated, admittedly, but remember that this is a short-term problem--we won't use a polyfill forever.

jfbastien:
Yes. It doesn't seem to provide much at the moment.

qwertie:
I'd like to be able to write code like this:

    x += switch(...) {
    case 1: break foo();
    case 2: break bar();
    case 3: break baz();
    }

or this:

    foo(12345, if (...) {
            foo()
        } else {
            y = 0;
            forever {
                y += ...
                if (...) break(y);
            }
        });

Things like this are possible in LISP and Haskell (except that Haskell doesn't allow mutable variables) and I'm sure I've seen it in at least one other (otherwise traditional) imperative language.

MikeHolman:
[...]

PR #407, opened 2015-10-13
Clarified that implicit return is allowed

Teemperor:
Current spec doesn't specify what happens in non-void functions without any return.

ml-proto is already doing it as described in the paragraph: https://github.com/WebAssembly/spec/blob/master/ml-proto/test/f32.wast#L2

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -325,6 +325,10 @@ function that returns multiple values will likely have to be a statement that
 specifies multiple local variables to which to assign the corresponding return
 values.
 
+If a function with a non-void return type finishes evaluating and no `return`
+was issued during the evaluation, the value of the function node is
+returned.
+
 ## Constants
 
 These opcodes have an immediate operand of their associated type which is


rossberg-chromium:
This is just a consequence of being an expression language, so probably hinges on https://github.com/WebAssembly/design/issues/223 being resolved.

kg:
If I don't end with a ```return```, but my last expression has a type that doesn't match the return type of my function, do we detect that and reject it as an error?
Is it reasonable to do that analysis statically when loading/compiling a wasm module at runtime, since it's 'does the function ever invoke return along this path of execution'? You can certainly do it but I'm not sure how fast it would be.

It might be simpler to just obligate all non-```void``` functions to have a 'last chance' ```return``` as their last expression, and prohibit jumping past it with ```break```. Most code generators will produce it anyway and that constraint should be much easier to enforce.

rossberg-chromium:
>
> If I don't end with a return, but my last expression has a type that
> doesn't match the return type of my function, do we detect that and reject
> it as an error?
>
Of course. Already the case.


> Is it reasonable to do that analysis statically when loading/compiling a
> wasm module at runtime, since it's 'does the function ever invoke return
> along this path of execution'? You can certainly do it but I'm not sure how
> fast it would be.
>
The function return type is explicit, so this check is completely trivial,
regardless of what returns a function uses. (In fact, it is simpler than
checking returns.)

titzer:
I agree with @rossberg-chromium; if we resolve #223 then this is a non-issue.

PR #409, opened 2015-10-14
Fixed formatting in floating point list

Teemperor:
Two new lines create a paragraph in the list which creates a funny looking spacing around the `f32.max` and `f64.add` item. This should fix it.


jfbastien:
I'd rather avoid HTML in .md files. Could you instead add newlines and text that says:

> 64-bit floating point operations are as follows:

Or something like that?

Teemperor:
Ok, done!

PR #412, opened 2015-10-15
Provide more details in FeatureTest.md

lukewagner:
As requested in [spec/#120](https://github.com/WebAssembly/spec/pull/120#issuecomment-148409196), this PR adds some more details on how polyfilling and feature testing could work in practice in the MVP.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -530,3 +530,5 @@ In the MVP, `has_feature` would always return false. As features were added post
 `has_feature` would start returning true. `has_feature` is a pure function, always
 returning the same value for the same string over the lifetime of a single
 instance and other related (as defined by the host environment) instances.
+See also [feature testing](FeatureTest.md) and
+[better feature testing](FutureFeatures.md#better-feature-testing-support).
--- a/FeatureTest.md
+++ b/FeatureTest.md
@@ -3,7 +3,7 @@
 The [MVP](MVP.md) allows an application to query which post-MVP features are
 supported via [`has_feature`](AstSemantics.md#feature-test). This accounts for
 the pragmatic reality that features are shipped in different orders at different
-times by different engines on the Web.
+times by different engines.
 
 Since some WebAssembly features add operators and all WebAssembly code in a
 module is validated ahead-of-time, the usual JS feature detection pattern:
@@ -21,12 +21,16 @@ Instead, applications may use one of the following strategies:
 1. Compile several versions of a module, each assuming different feature support
    and use `has_feature` tests to determine which version to load.
 
-2. During [layer 1 decoding](BinaryEncoding.md), which will happen in user code
-   in the MVP *anyway*, use `has_feature` to determine which features are
-   supported and then translate unsupported feature use into either a polyfill
+2. During the ["specific" layer decoding](BinaryEncoding.md), which will happen
+   in user code in the MVP *anyway*, use `has_feature` to determine which features
+   are supported and then translate unsupported feature use into either a polyfill
    or a trap.
 
-To illustrate, consider 3 representative examples:
+Both of these options could be automatically provided by the toolchain and
+controlled by compiler flags. Since `has_feature` is a constant expression,
+it can be constant-folded by WebAssembly engines.
+
+To illustrate, consider 4 examples:
 
 * [`i32.min_s`](FutureFeatures.md#additional-integer-operations) - Strategy 2
   could be used to translate `(i32.min_s lhs rhs)` into an equivalent expression
@@ -34,7 +38,8 @@ To illustrate, consider 3 representative examples:
 * [Threads](PostMVP.md#threads) - If an application uses `#ifdef` extensively
   to produce thread-enabled/disabled builds, Strategy 1 would be appropriate.
   However, if the application was able to abstract use of threading to a few
-  primitives, Strategy 2 could be used to patch in the right primitive.
+  primitives, Strategy 2 could be used to patch in the right primitive 
+  implementation.
 * [`mprotect`](FutureFeatures.md#finer-grained-control-over-memory) - If engines
   aren't able to use OS signal handling to implement `mprotect` efficiently,
   `mprotect` may become a permanently optional feature. For uses of `mprotect`
@@ -43,7 +48,37 @@ To illustrate, consider 3 representative examples:
   correctness but an alternative strategy existed that did not rely on
   `mprotect`, `mprotect` could be replaced with an `abort()` call, relying on
   the application to test `(has_feature "mprotect")` to avoid calling the
-  `abort()`.
+  `abort()`. The `has_feature` query could be exposed to C++ code via
+  the existing `__builtin_cpu_supports`.
+* [SIMD](PostMVP.md#fixed-width-simd) - When SIMD operations have a good-enough
+  polyfill, e.g., `f32x4.fma` via `f32x4.mul`/`add`, Strategy 2 could be used 
+  (similar to the `i32.min_s` example above). However, when a SIMD feature has no
+  efficient polyfill (e.g., `f64x2`, which introduces both operators *and*
+  types), alternative algorithms need to be provided and selected at load time.
+  With toolchain integration, function attributes could be used to annotate
+  functions as feature-dependent optimized versions of other functions:
+  ```
+  #include <xmmintrin.h>
+  void foo(...) {
+    __m128 x, y;           // -> f32x4 locals
+    ...
+    x = _mm_add_ps(x, y);  // -> f32x4.add
+    ...
+  }
+  void foo_f64x2(...) __attribute__((optimizes("foo","f64x2"))) {
+    __m256 x, y;           // -> f64x2 locals
+    ...
+    x = _m_add_pd(x, y);   // -> f64x2.add
+    ...
+  }
+  ...
+  foo(...);                 // calls either foo or foo_f64x2
+  ```
+  In this example, the toolchain could emit both `foo` and `foo_f64x2` and
+  the load-time polyfill would replace `foo` with `foo_f64x2` if
+  `(has_feature "f64x2")`. Many other strategies are possible to allow finer or
+  coarser granularity substitution. Since this is all in userspace, the
+  strategy can evolve over time.
 
 See also the [better feature testing support](FutureFeatures.md#better-feature-testing-support)
 future feature.
--- a/FeatureTest.md
+++ b/FeatureTest.md
@@ -6,7 +6,8 @@ the pragmatic reality that features are shipped in different orders at different
 times by different engines.
 
 Since some WebAssembly features add operators and all WebAssembly code in a
-module is validated ahead-of-time, the usual JS feature detection pattern:
+module is validated ahead-of-time, the usual JavaScript feature detection
+pattern:
 ```
 if (foo)
     foo();
--- a/FeatureTest.md
+++ b/FeatureTest.md
@@ -56,10 +56,12 @@ To illustrate, consider 4 examples:
   efficient polyfill (e.g., `f64x2`, which introduces both operators *and*
   types), alternative algorithms need to be provided and selected at load time.
 
-As a hypothetical (not implemented) example of the SIMD example, the toolchain
-could provide a new function attribute that indicated that the function
-was a feature-dependent optimized version of another function (similar to
-`ifunc`s, but without the callback):
+As a hypothetical (not implemented) example polyfilling the SIMD `f64x2`
+feature, the C++ compiler could provide a new function attribute that indicated
+that one function was an optimized, but feature-dependent, version of another
+function (similar to the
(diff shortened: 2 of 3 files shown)


FeatureTest.md
```diff
@@ -1,11 +1,49 @@
 # Feature Test
 
-Any feature not present in [the MVP](MVP.md) will have a corresponding feature
-test. This allows polyfills to elegantly emulate a feature, and allows
-developers to fallback to another implementation if they so desire.
+The [MVP](MVP.md) allows an application to query which post-MVP features are
+supported via [`has_feature`](AstSemantics.md#feature-test). This accounts for
+the pragmatic reality that features are shipped in different orders at different
+times by different engines on the Web.
```
jfbastien: I'd drop "on the Web" since this applies equally to non-Web (node, or any other usage of wasm).

FeatureTest.md
```diff
@@ -1,11 +1,49 @@
...
-in future features.
+Instead, applications may use one of the following strategies:
+
+1. Compile several versions of a module, each assuming different feature support
+   and use `has_feature` tests to determine which version to load.
+
+2. During [layer 1 decoding](BinaryEncoding.md), which will happen in user code
+   in the MVP *anyway*, use `has_feature` to determine which features are
+   supported and then translate unsupported feature use into either a polyfill
+   or a trap.
```
jfbastien: Mention that this can be part of the default toolchain, so developers don't have to roll their own?

FeatureTest.md
```diff
@@ -1,11 +1,49 @@
...
+  primitives, Strategy 2 could be used to patch in the right primitive.
+* [`mprotect`](FutureFeatures.md#finer-grained-control-over-memory) - If engines
+  aren't able to use OS signal handling to implement `mprotect` efficiently,
+  `mprotect` may become a permanently optional feature. For uses of `mprotect`
+  that are not necessary for correctness (but rather just catching bugs),
+  `mprotect` could be replaced with `nop`. If `mprotect` was necessary for
+  correctness but an alternative strategy existed that did not rely on
+  `mprotect`, `mprotect` could be replaced with an `abort()` call, relying on
+  the application to test `(has_feature "mprotect")` to avoid calling the
+  `abort()`.
```
jfbastien: I think it would be useful to have a SIMD example as well, since it adds new types as well.

FeatureTest.md
```diff
@@ -1,11 +1,49 @@
...
+```
+won't work in WebAssembly (if `foo` isn't supported, `foo()` will fail to
+validate).
 
-See also [better feature testing support](FutureFeatures.md#better-feature-testing-support)
-in future features.
+Instead, applications may use one of the following strategies:
+
+1. Compile several versions of a module, each assuming different feature support
+   and use `has_feature` tests to determine which version to load.
```
jfbastien: Mention dynamic linking?

FeatureTest.md
jfbastien: It would be nice to explain what things could look like from a developer perspective, in their own C/C++ code. I think LLVM/GCC's `__builtin_cpu_supports` is a good fit for feature detection.

FeatureTest.md
lukewagner: Even before dynamic linking:
```
var script = document.createElement('script');
if (has_feature('a'))
   script.url = "foo_with_a.wasm";
else
  script.url = "foo_without_a.wasm";
```
Dynamic linking is basically orthogonal.

AstSemantics.md
```diff
@@ -530,3 +530,5 @@ In the MVP, `has_feature` would always return false. As features were added post
 `has_feature` would start returning true. `has_feature` is a pure function, always
 returning the same value for the same string over the lifetime of a single
 instance and other related (as defined by the host environment) instances.
```
jfbastien: `has_feature` can't just be a function, if we want to be able to have SIMD types in a function signature. It has to be a structural construct that contains s-expression children.

FeatureTest.md
mtrofin: We should make it clear that has_feature is expected to be invariant (for a given input) at runtime (I'm assuming that's the case, isn't it?)

FeatureTest.md
mtrofin: Since we're contrasting with feature detection in javascript, I think we should highlight that both options 1 and 2 are pre-runtime mechanisms. It may help a reader if we compared the proposed mechanism with conditional compilation directives in C/C++ (\#ifdef...\#endif)

FeatureTest.md
mtrofin: I agree with JF, we should think through what the C/C++ developer experience will be. I think using dynamic checks, though, may be sending the wrong message (or at least confusing) - since the result of the check isn't going to change dynamically at runtime. Preprocessor directives seem like a better fit from that point of view, albeit they may be too fine-grained.

I can understand if "the ship has sailed", though, given the precedent he's mentioning.

FeatureTest.md
mtrofin: Could option 1 be seen as option 2 at the coarsest granularity, the whole Module?

FeatureTest.md
mtrofin: I like the examples, I think we should actually start with them and perhaps categorize them, from the perspective of the developer. It seems like there are at least 2 broad categories: developer-authored and toolchain-authored. The i32.min_s example, if I understand it correctly, falls into the latter: the toolchain is aware of new capabilities, but produces backwards compatible bytecode. I think we should highlight this distinction, if it's intentional, for clarity.

FeatureTest.md
mtrofin: It may help to elaborate on why it matters that layer 1 decoding will happen in user code in the MVP; and what will happen after the MVP.

Currently, I wasn't able to find a reference to a "layer 1 decoding" notion in BinaryEncoding.mp. Should there be one?

lukewagner: I guess that's just what we've been informally calling it (with layer 0 = "raw", layer 1 = "specific", layer 2 = "generic compression"); I'll update to "specific layer".  With the corrected name, the reader should be able to read the link and see that the reason it would happen in user code is because the specific layer isn't standardized.

FeatureTest.md
lukewagner: AstSemantics.md does specifically mention the invariance of `has_feature`, but I can restate here.  There is a sortof philosphical difference, though, between serving 1 file and 2.  I've seen multiple devs (e.g., Unity) say they'd prefer to serve only 1 file.

AstSemantics.md
lukewagner: "function" in the mathematical sense, not literally a wasm function.  I don't know what you mean about s-expression children, but `has_feature` is nullary, so it wouldn't have children.

FeatureTest.md
jfbastien: It does mention it on line 530.

lukewagner:
Ok, posted update to address most comments.

FeatureTest.md
```diff
@@ -1,11 +1,88 @@
...
-developers to fallback to another implementation if they so desire.
+The [MVP](MVP.md) allows an application to query which post-MVP features are
+supported via [`has_feature`](AstSemantics.md#feature-test). This accounts for
+the pragmatic reality that features are shipped in different orders at different
+times by different engines.
 
-Feature tests will be available from
-[WebAssembly itself](AstSemantics.md#feature-test), as well as from JavaScript.
+Since some WebAssembly features add operators and all WebAssembly code in a
+module is validated ahead-of-time, the usual JS feature detection pattern:
```
jfbastien: "JavaScript" :)

AstSemantics.md
jfbastien: What I mean is that we can't `#ifdef` out a function if it has a SIMD type in its signature. If SIMD isn't supported then none of the `.wasm` file can use it in a signature. If `has_feature` were instead a structural container for sub-s-expressions (including top-level ones) then this would be possible.

lukewagner:
@jfbastien @mtrofin Some of these ideas sound like they could be followups you could propose.  The goal here is just to add some more detail and clarification, not to be the last word on polyfilling.  Perhaps you could evaluate this PR just on what it has and then file new PRs to further improve?

jfbastien:
lgtm

JSStats:
1. Knowledge of which features were probed is valuable input for a cache key. If there are cache-able outputs then enforcing some data flow would be necessary to make this usable. For example a distinct layer-1 stage, for which the probed features are noted, allow the caching of the layer-1 output or downstream output.

2. This is more effective if the probed 'features' are fine-grain. Perhaps this could be added to the rationale, so that we don't have feature detection functions added that return a bit set of orthogonal features (for which the cache would not know which were actually used, and which the output depends on). Perhaps also encourage developers to not probe for a feature unless the output really depends on it.

lukewagner:
@JSStats In FF, asm.js code cache entries simply include the buildid in the key (since any new FF binary may change struct layouts baked into the machien code).  Assuming `has_feature` only changes result with new buildids (which should be the case), then no extra addition to the key is needed.

JSStats:
@lukewagner `has_feature` is not defined to only change between runtime builds so the key is needed. The key will also need other inputs such as the negotiated memory size when a fixed size has been requested, so it is natural to also include the probed features.

lukewagner:
@JSStats I'm talking about the engine-internal code cache.

JSStats:
@lukewagner I am also 'talking about the engine-internal code cache'. The runtime engine needs to be able to reason about the provenance of the wasm code it is compiling in order to manage this cache. The `has_feature` results are inputs into layers in the pipeline, for example a layer-1 that emits difference wasm depending on the features, and `has_feature` is only defined to be static per-instance, not per-runtime-build.

The upper layers need to be deterministic even if user defined so that the runtime will know that the result of decompression does not change if the compressed source does not change, and that the product of the layer-1 code writing does not change if the inputs (including the `has_feature` results) do not change, and it can reuse cached compiled code without having to re-run the upper layers of the pipeline.

The runtime can note the calls made to `has_feature` when the upper layers run (on a cache miss), plus other inputs, and add these to the cache key. Only those features probed need be noted, so the cache key can be as general as possible to keep the hit rate as high as possible. By noting the probe features when the upper layers run, the set of features that the code depends on can vary based on the features.

For example, consider two orthogonal features A and B. The code might only need feature B if feature A is available, so it probes A first and only probes B if true. If A is not available then the key need not depend on B.

lukewagner:
@JSStats The internal cache is keyed on the content, which would be the output of layer 1.  `has_feature` is not going to change for a given buildid.  Thus, (content, buildid (, cpuid)) is sufficient.

JSStats:
@lukewagner Your response misses the key points:

1. I repeat that `has_feature` is not defined to be static for a given build ID. I can think of many uses for per-instance constants that the runtime may want to vary. If people want a `has_build_feature` that is static across instances then add that, but it would be a very narrow feature that is subsumed by more general instant-constants and really is not even worth wasting time on.

2. We do not want to be re-computing the content input into the final stage, so suggesting that a wasm compiler cache needs the raw input (the output from the upstream layers) condemns wasm to be continually re-applying the upper stages on any use of the wasm source.

We want the runtime to be able to note that the input source (the compressed and pre-transformed code) has not changed, and that the inputs into the various stages in the pipeline have not changed, and then (without re-computing the input into the final stage) be able to reason that the compiled code will not changed and to reuse it.

In some ways getting this right is more important than even the binary encoding. Developers can deal with differences in the binary code between implementations using transforms at upper layers, but an inefficient pipeline cache blocks client side decompression and transforms in user defined code.

lukewagner:
@JSStats (1) I realize that, but the engine can choose to have it be.  I do not see us adding arbitrary constants, but rather only static, boolean detection of features (it's in the name; this isn't `sysconf`).  (2) Not if outputs of the upper layers are cached (SW/Cache API or IDB).  As discussed elsewhere, the long-term goal is that these upper layers get standardized which would allow the whole pipeline to be a pure function of the over-the-wire bits + fixed context.  Before that, though, it's just arbitrary JS execution.

JSStats:
@lukewagner 

(1.1) If features are only defined to be static per-instance then an engine can choose to change them per-instance, and would still want to implement an effective cache. wasm should not be designed to only work efficiently with an engine with per-build static features.

(1.2) There is a need for per-instance static constants input into the upper layers. Very significantly the memory size to allow the upper layers to emit code optimized for a static size. Even if you can answer the performance challenges of eliminating bounds checks without a fixed memory size (and no one here has), as mentioned many times in discussions there are applications that can naturally take advantage of a fixed memory size - and they need a negotiated size to work on a range of instances with different amounts of available vm.

(2) In order for the outputs of upper layers for be cached, the inputs need to be part of the cache key. The runtime should be able to reason across layers in the data flow and not have to do a cache lookup on each stage and compare the output to the key of the next stage - it's often referred to as 'provenance'.

It's news to me that the 'long-term goal is that these upper layers get standardized'.  Part of the appeal of supporting user defined upper layers was the flexibility to meet a range of uses. I get the impression that web developers are frustrated by being locked into a fixed stack, and are desperate for lower level access (that can also be used efficiently). You'll be bike-shedding for the next decade over a lost cause spec'ing a stack. Please consider supporting a user defined stack and this might mean some support for managing the data flow (which I think would be a much better investment).

PR #413, opened 2015-10-15
Defined behavior of clz/ctz

Teemperor:
Using 32 because ml-proto and ARM are behaving like that for a 32 bit input.


jfbastien:
lgtm

PR #414, opened 2015-10-15
Nondeterminism: has_feature

jfbastien:


sunfishcode:
lgtm

kripken:
Perhaps overly pedantic, but this is nondeterministic in the same sense that calling an outside function is nondeterministic, and that isn't mentioned on the list, because it is obvious. Not opposed to this, but if we are going for a full list here, do we need external methods too? Or is there a subtle difference I am missing?

sunfishcode:
`has_feature` is presently spec'd as an opcode in the platform itself, which to me is what qualifies it for inclusion here, as opposed to API calls to the outside world.

jfbastien:
I moved the API discussion to Rationale.md in a previous PR:

> There are a few obvious cases where nondeterminism is essential to the API, such as random number generators, date/time functions or input events. The WebAssembly specification is strict when it comes to other sources of limited local nondeterminism of operations: it specifies all possible corner cases, and specifies a single outcome when this can be done reasonably.

kripken:
Ah, I see, thanks.

Issue #322, opened 2015-09-01
Define the behavior of "switch"


AndrewScheidecker:
The design is vague about how switches work. The spec interpreter does this:
- A switch is an operation with N>=2 operands
- The first operand is an integer of any type.
- The last operand is an expression that yields the same type as the switch that is used by default if no case is matched.
- The other operands are cases.
- Each case has a literal integer and an expression (the syntax allows 0 or many expressions that yield nop or block expressions).
- Each case has an optional fallthru flag that determines whether it will fall through to the following case (or default case). The syntax sugar for a case without an expression implies fallthru.
- If a case is fallthru, then its expression is void type ([] in the spec interpreter).
- If a case isn't fallthru, then its expression is the same type as the switch.

More relevant context about the spec interpreter:
- All operations are expressions: they may yield a value, and so have an associated type
- Break/return/loop operations are considered to yield a value of the bottom type, representing that they never yield a value to their enclosing expression, and may occur as a subexpression where any type is expected.
- Labels are expressions, so the break operation needs an operand of the appropriate type for the label to yield.

This issue is not about whether there should be a C-like concept of statements, which seems to be an open issue. If the spec ends up making switch a statement, this is irrelevant. But if not:

As it's defined in the spec interpreter, there isn't a direct translation to and from C-style switches: the default case must be the equivalent of the final case, and so cannot fall through to other cases. I believe the rationale for this design is to ensure a switch always yields a value of the type expected by its context.

Another way to achieve the same goal would be:
- Give switch an implicit label.
- Add a way to declare a default case, and get rid of the final default expression.
- Make all but the last case the equivalent of "fallthru". That means the last case expression would need to be same type as the switch, but the other case expressions would be any typed.

Instead of each non-fallthru case being an exit from the switch, the exits from this switch would be explicit breaks and the final case. This makes switch equivalent in flexibility to C-style switches, but allows it to still be used as an expression.

For example:
```
...
(param %key i64)
(local %a f64)
(local %b i32)
(setlocal %a (switch.i64 #switch %key
    (case 0)
    (case 1 (break #switch (constant.f64 100.0)))
    (default (setlocal %b 1))
    (case 2 (break #switch (constant.f64 200.0)))
    (case 3 (constant.f64 300.0))
    ))
```
[...]

PR #427, opened 2015-10-23
Flexible control-flow operators.

sunfishcode:
This introduces `br`, `br_if`, and so on, while also preserving the
high-level `if` and `if_else` operators. It also converts `switch` into
`tableswitch` and generalizes it to support labels in enclosing scopes.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -247,39 +247,43 @@ others, etc.
 
 ## Control flow structures
 
-WebAssembly offers basic structured control flow. All control flow structures
-are statements.
+WebAssembly offers basic structured control flow with the following constructs.
+All control flow structures, except `case`, are statements.
 
-  * `block`: a fixed-length sequence of statements
-  * `if`: if statement
-  * `do_while`: do while statement, basically a loop with a conditional branch
-    (back to the top of the loop)
-  * `forever`: infinite loop statement (like `while (1)`), basically an
-    unconditional branch (back to the top of the loop)
-  * `continue`: continue to start of nested loop
-  * `break`: break to end from nested loop or block
-  * `return`: return zero or more values from this function
-  * `switch`: switch statement with fallthrough
+ * `block`: a fixed-length sequence of statements with a label at the end
+ * `loop`: a fixed-length sequence of statements with a label at the end
+           and a loop header label at the top
+ * `if`: if statement with then body
+ * `if_else`: if statement with then and else bodies
+ * `br`: branch to a given label in an enclosing construct (see below)
+ * `br_if`: conditionally branch to a given label in an enclosing construct
+ * `tableswitch`: a jump table transferring which may jump either to enclosed
+                  `case` blocks or to labels in enclosing constructs (see below
+                  for a more detailed description)
+ * `case`: must be an immediate child of `tableswitch`; has a label declared
+           in the `tableswitch`'s table and a body (as above, see below)
+ * `return`: return zero or more values from this function
 
-Loops (`do_while` and `forever`) may only be entered via fallthrough at the top.
-In particular, loops may not be entered directly via a `break`, `continue`, or
-`switch` destination. Break and continue statements can only target blocks or
-loops in which they are nested. These rules guarantee that all control flow
-graphs are well-structured.
+References to labels must occur within an *enclosing construct* that defined
+the label. This means that references to an AST node's label can only happen
+within descendents of the node in the tree. For example, references to a
+`block`'s label can only happen from within the `block`'s body. In practice,
+one can arrange `block`s to put labels wherever one wants to jump to, except
+for one restriction: one can't jump into the middle of a loop from outside
+it. This restriction ensures the well-structured property discussed below.
 
-Structured control flow provides simple and size-efficient binary encoding and
-compilation. Any control flow—even irreducible—can be transformed into structured
-control flow with the
-[Relooper](https://github.com/kripken/emscripten/raw/master/docs/paper.pdf)
-[algorithm](http://dl.acm.org/citation.cfm?id=2048224&CFID=670868333&CFTOKEN=46181900),
-with guaranteed low code size overhead, and typically minimal throughput
-overhead (except for pathological cases of irreducible control
-flow). Alternative approaches can generate reducible control flow via node
-splitting, which can reduce throughput overhead, at the cost of increasing
-code size (potentially very significantly in pathological cases).
-Also,
-[more expressive control flow constructs](FutureFeatures.md#more-expressive-control-flow)
-may be added in the future.
+`tableswitch` instructions have a zero-based array of labels, a label index,
+a "default" label, an index operand, and a list of `case` nodes. A `tableswitch`
+selects which label to branch to by looking up the index value in the label
+array, and transferring control to that label. If the index is out of bounds,
+it transfers control to the "default" label.
+
+`case` nodes can only appear as immediate children of `tableswitch` statements.
+They have a label, which must be declared in the immediately enclosing
+`tableswitch`'s array, and a body which can contain arbitrary code. Control
+falls through the end of a `case` block into the following `case` block, or
+the end of the `tableswitch` in the case of the last `case`.
+
 
 ## Calls
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -267,7 +267,7 @@ All control flow structures, except `case`, are statements.
 References to labels must occur within an *enclosing construct* that defined
 the label. This means that references to an AST node's label can only happen
 within descendents of the node in the tree. For example, references to a
-`block`'s label can only happen from within the `block`'s body. In practice,
+`block`'s label can only occur within the `block`'s body. In practice,
 one can arrange `block`s to put labels wherever one wants to jump to, except
 for one restriction: one can't jump into the middle of a loop from outside
 it. This restriction ensures the well-structured property discussed below.
--- a/Rationale.md
+++ b/Rationale.md
@@ -114,7 +114,19 @@ developer.
 
 ## Control Flow
 
-See [#299](https://github.com/WebAssembly/design/pull/299).
+Structured control flow provides simple and size-efficient binary encoding and
+compilation. Any control flow—even irreducible—can be transformed into structured
+control flow with the
+[Relooper](https://github.com/kripken/emscripten/raw/master/docs/paper.pdf)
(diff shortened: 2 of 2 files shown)


sunfishcode:
Updated PR to be independent of statements=expressions.

rossberg-chromium:
I'd like to understand the motivation for tableswitch better. Isn't that simply a more convoluted way of writing a C-style switch? As far as I can tell, it is completely equivalent, the only differences being stylistic. Namely, (1) all cases have to be written with an indirection through an extra label instead of using `case` directly as the label, (2) it forces to list all cases between 0 and the max label explicitly instead of having omitted ones go to default implicitly, and (3) it can jump to outside labels directly instead of using the existing branch construct for that.

(1) and (2) are effectively just changes in syntax, not semantics. Of course, this representation makes sense on the encoding level, but on the AST level, it just makes the construct (significantly) more verbose and more complicated to describe and spec. (3) on the other hand is just syntactic sugar, analogous to `br_if`.

So in terms of AST semantics and spec factorisation, tableswitch is strictly inferior to a C-style switch, while being functionally equivalent. I think we should treat the table representation as an encoding concern, not a semantics concern.

sunfishcode:
This `tableswitch` has one area of flexibility beyond what a C-style switch has: it can directly branch to labels outside its own body.

It also has a notable restriction in that it requires the index space to be zero-based and dense. This makes it more directly correspond to an actual jump table, as reflected by the rename to `tableswitch`.

rossberg-chromium:
Yes, I understand, but as I tried to point out above, that seems more like an encoding consideration, and not like something that is necessarily relevant on the semantic level of the AST.

sunfishcode:
I don't see how the semantics of where a `tableswitch` can branch to is an encoding consideration. Similarly, the index space determines how `tableswitch` interprets its operand, so it's also part of the semantics.

rossberg-chromium:
As mentioned above, the ability to jump to an external label is merely syntactic sugar over a break (and limited one, because there is no way you can make it transfer a value). As such, it is primarily an encoding consideration as well.

Can you explain what you are referring to as index space, and how it affects the operand?

qwertie:
We discussed `tableswitch` earlier in #322. It is advantageous over the old switch because

1. `tableswitch` is able to represent the vast majority of C switch statements, unlike the old `switch` (IIUC, [duff's device](https://en.wikipedia.org/wiki/Duff%27s_device) is still not allowed)
2. Using a dense rather than sparse jumptable allows a Wasm JIT or interpreter to very simple, which maximizes startup time. The optimization problem of how to implement the switch statement (as a collection of jump tables, breaks and if-elses) is shifted from JIT-time to compile-time.

sunfishcode:
@rossberg-chromium One can express any reasonable switch-like construct as syntax sugar on top of any other plus extra nodes. As such, I don't see why this is a reason to prefer any one over any other.

Also, Statements == Expressions is an orthogonal concern. This `tableswitch` can still transfer values, just as `br` can, if we choose to do that.

The index space of `tableswitch` is the integer sequence starting at 0 and continuing for some finite length. A `tableswitch` operator defines a mapping from values in its sequence to labels, and it uses this mapping to compute a label, given its operand.

rossberg-chromium:
@qwertie, yes, I was comparing with a C-style switch, which I'm happy moving towards. Density is an independent consideration.

@sunfishcode, re index space, that's what I guessed, but then I still don't understand your earlier remark on that.

It may help to point out that we are actually discussing syntax here rather than semantics. I am fine with the goal of increasing the flexibility of the switch operator. But I am puzzled how the explicit syntactic indirection through extra labels helps at the AST level.

Let's be concrete. Consider a switch as described in this proposal:
```
$label: (switch (operand)
  ($case0 $case1 $case2 $label $case4 $case5) $default
  (case $case2 (a))
  (case $case1 (b))
  (case $case5 (c))
  (case $case4)
  (case $default (d))
  (case $case0 (e))
)
```
Here is how I would design the concrete S-expr syntax:
```
$label: (switch (operand)
  (case 3 br $label)
  (case 2 (a))
  (case 1 (b))
  (case 5 (c))
  (case 4)
  (default (d))
  (case 0 (e))
)
```
That is expressing _exactly the same_ structure but with much less repetition. You can still represent it as a table of labels in the binary encoding, but as a textual or AST-level representation it is superior in just about every way I can imagine, for both producers and "consumers", including the spec. 

Note that the density requirement is independent of this choice of syntax. I don't see a particularly strong reason to require listing the `(case 4)` explicitly in the latter form, but there is nothing preventing us from requiring it if we want to emphasise the table nature.

Note also that I even made the break to the outer label special syntax that's part of `case`, and different from using a separate `br` operator (which would be parenthesised). It obviously is redundant to introduce that, but there you go.

So, again, this is really about syntax, not semantics. AFAICT, the latter way of structuring the operator syntax is more convenient on all levels. I'd like to understand what advantage you see in the former notation. Or would you be fine with the latter as well?

sunfishcode:
I've now added a patch to this PR converting it to what I think you're asking for. Is this what you had in mind?

In my opinion, this patch makes the prose harder to read for humans, makes it harder for me to convince myself that the semantics aren't incomplete or inconsistent, and it puts greater distance between the design-document AST representation and what I might imagine we'll want in the binary encoding.

rossberg-chromium:
@sunfishcode,cool, thanks! And for your pleasure, I already implemented it: https://github.com/WebAssembly/spec/pull/153 :)

sunfishcode:
I decided I didn't like it, so I removed it now.

rossberg-chromium:
I see. :(  Well, please have a look at the spec PR. AFAICS, with this structure, it is as simple as it will get. Tests also remain more readable. I hear what you are saying re prose and distance to binary encoding, but I think those are far outweighed by familiarity and structural simplicity, respectively.

AstSemantics.md
```diff
@@ -247,39 +247,43 @@ others, etc.
...
+                  `case` blocks or to labels in enclosing constructs (see below
+                  for a more detailed description)
+ * `case`: must be an immediate child of `tableswitch`; has a label declared
+           in the `tableswitch`'s table and a body (as above, see below)
+ * `return`: return zero or more values from this function
+
+References to labels must occur within an *enclosing construct* that defined
+the label. This means that references to an AST node's label can only happen
+within descendents of the node in the tree. For example, references to a
+`block`'s label can only happen from within the `block`'s body. In practice,
```
titzer: "happen from within" => "occur within"

Rationale.md
```diff
@@ -114,7 +114,19 @@ developer.
 
 ## Control Flow
 
-See [#299](https://github.com/WebAssembly/design/pull/299).
+Structured control flow provides simple and size-efficient binary encoding and
+compilation. Any control flow—even irreducible—can be transformed into structured
```
titzer: dashes "--", not hyphens.

titzer:
This mostly looks good modulo some wordsmithing.

titzer:
@rossberg-chromium You've mentioned a C-style switch as a basis for the tableswitch design. Here we are being a little bit more explicit with labels, which pushes a bit of work up to producers, at the gain of being explicit for backends and engines. Is this a reasonable tradeoff in your opinion? Or are you proposing that the syntax of the AST is C-like and the binary encoding is table-switch like?

kg:
@sunfishcode Have you done much research into how this would impact pre-compression and post-compression file size (in the asmjs packer, for example?) The significant changes to ASTs this implies could mean a big file size regression (or a big file size win, in which case that alone is a compelling reason to make your changes!)

sunfishcode:
@kg Looking at some asm.js code of some miscellaneous applications, `if` statements are over 6x more frequent than loops, and over 300x more frequent than switches. This iteration of the proposal still has `if` and `if_else`. And, I expect no difference in the size of loops in the common case of optimized code. Consequently, I'm assuming this doesn't introduce a big file size regression or win.

Rationale.md
sunfishcode: This is pre-existing text that I'm just moving here, but those are Unicode em-dash characters, not hyphens, so I believe this is correct usage.

sunfishcode:
Wordsmithing changes applied.

kg:
The question is about the binary encoding. Are you saying ```if``` and ```if_else``` would be opcodes in the binary representation? I thought they were syntax sugar?

JSStats:
@kg I think this has been answer elsewhere - syntax sugar is a spec matter and there are still opcodes for the sugar.

rossberg-chromium:
On 26 October 2015 at 18:55, titzer <notifications@github.com> wrote:

> @rossberg-chromium <https://github.com/rossberg-chromium> You've
> mentioned a C-style switch as a basis for the tableswitch design. Here we
> are being a little bit more explicit with labels, which pushes a bit of
> work up to producers, at the gain of being explicit for backends and
> engines. Is this a reasonable tradeoff in your opinion? Or are you
> proposing that the syntax of the AST is C-like and the binary encoding is
> table-switch like?
>

Yes, the latter. There is no real benefit in introducing explicit labels as
jump targets on the AST level -- quite the opposite, in fact. A
C/JavaScript-style switch notation is much more manageable when formulating
semantics, or writing/reading examples.

rossberg-chromium:
https://github.com/WebAssembly/spec/pull/153 now matches everything described by this PR (I think), modulo the notation for tableswitch.

I did some more pondering over the latter, esp looking into how the proposed notation would affect the spec. Honestly, it would be quite ugly. :( In addition to tracking the extra indirection, it would require distinguishing two different kinds of (raw) labels in the AST (since tableswitch addresses its inner targets differently from external targets) as well as different scoping rules for those labels' symbolic representation.

In other words, my assessment still stands that this notational factorisation for tableswitch is suboptimal. Unless we have strong technical reasons to do it that way, I highly recommend more streamlined notation. @sunfishcode, do you think you can still warm up to that?

sunfishcode:
My attempt above at implementing what I thought you were asking for made the prose about twice as long, and I've now discovered, as I feared, that it did have a bug hidden in the extra complexity: it didn't address whether case and default could fall through to case_br and default_br.

Right now, I am very confused as to the intended relationship between AstSemantics.md, ml-proto are s-expression syntax, what we might expect to call the binary format AST, and what we might expect in the actual binary format. This confusion is making it difficult for me to figure out what you're asking for here, and what it means for other parts of the system.

I propose we merge this PR as a starting point, and then you can submit a new PR after it proposing the specific changes what you would like to see. I think that will help me understand better how you want this to look, rather than having me try to write up what I think you want. Does that make sense?

rossberg-chromium:
@sunfishcode, okay, sounds good to me.

sunfishcode:
Ok. Merging.

PR #428, opened 2015-10-23
Select operators.

sunfishcode:
This adds select (analogous to conditional move) operators on all the local types.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -530,6 +530,10 @@ Truncation from floating point to integer where IEEE 754-2008 would specify an
 invalid operator exception (e.g. when the floating point value is NaN or
 outside the range which rounds to an integer in range) traps.
 
+## Type-parameterized operators.
+
+  * `select`: select one of two operands according to a boolean condition
+
 ## Feature test
 
 To support [feature testing](FeatureTest.md), an AST node would be provided:
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -533,9 +533,9 @@ outside the range which rounds to an integer in range) traps.
 ## Type-parameterized operators.
 
   * `select`: a ternary operator with a boolean (i32) condition and two
-    additional operands with the parameterized type. `select` returns the
-    the first type-parameterized operand if the condition operand is non-zero,
-    or the second type-parameterized operand otherwise.
+    additional operands, which must have the same type as each other. `select`
+    returns the the first of these two operands if the condition operand is
+    non-zero, or the second otherwise.
 
 ## Feature test
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -532,7 +532,10 @@ outside the range which rounds to an integer in range) traps.
 
 ## Type-parameterized operators.
 
-  * `select`: select one of two operands according to a boolean condition
+  * `select`: a ternary operator with a boolean (i32) condition and two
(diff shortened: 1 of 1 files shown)


sunfishcode:
It was pointed out that `select` can be expressed as a type-parametric operator. Updated.

rossberg-chromium:
Isn't this simply (a strict version of) `if`?

titzer:
@rossberg-chromium yes, indeed. It can be implemented as a conditional move instruction.

sunfishcode:
Updated to describe the operand order and semantics of `select` in more detail.

titzer:
lgtm

rossberg-chromium:
@titzer, sure, but i assume you'd use conditional move instructions for suitable `if`s as well? So what does a separate operator add?

sunfishcode:
@rossberg-chromium When a consumer sees an `if` operator, it can immediately start setting up the bookkeeping needed to create basic blocks or an SSA scope or whatever it needs to prepare to decode the rest of the body. `select` can be translated directly to a conditional move, which is a very common hardware instruction.

qwertie:
+1 (btw, unto whom is destowed the power of the lgtm?)

rossberg-chromium:
@sunfishcode, what you're saying sounds like it's wanted only for the sake of micro-optimising some decoders. Is that really the primary motivation?

qwertie:
I see this as shifting optimization work (the decision between branch and conditional moves) from browser-time to compile-time. As long as compilers tend to make the same decision for different ISAs, it makes sense.

sunfishcode:
It's also convenient for producers which already have a strict-evaluation select operation [eg.](http://llvm.org/docs/LangRef.html#i-select) [eg.](http://pp.ipd.kit.edu/firm/Nodes_latest#Mux). It makes it easier to work across multiple levels of the system (producer, wasm, consumer, machine code) if we can minimize the number of intermediate semantic translations/optimizations needed in common cases.

titzer:
The use of a select operator is a clear signal from a producer to a consumer that a conditional move is a best implementation strategy for the given condition. It doesn't compel the consumer to do so, and a simple and correct implementation is to just emit an normal high-level if after evaluating all operands. 

Implementation burden is not high, maps to a single instruction on multiple machines, expresses clear optimization hint, so, I think it makes sense to add this.

AstSemantics.md
```diff
@@ -534,6 +534,13 @@ Truncation from floating point to integer where IEEE 754-2008 would specify an
 invalid operation exception (e.g. when the floating point value is NaN or
 outside the range which rounds to an integer in range) traps.
 
+## Type-parameterized operators.
+
+  * `select`: a ternary operator with a boolean (i32) condition and two
+    additional operands with the parameterized type. `select` returns the
+    the first type-parameterized operand if the condition operand is non-zero,
+    or the second type-parameterized operand otherwise.
```
lukewagner: I think it'd be simpler to not say "parameterized" at all and just have a sentence at the end saying the parameters must have the same value type.

lukewagner:
Generally lgtm

sunfishcode:
Updated to avoid saying "type-parameterized operands", and to just say that the two operands to select between must have the same type as each other.

sunfishcode:
Merging, with lgtm's and +1's above.

PR #431, opened 2015-10-27
Change linear memory max to be a hint

lukewagner:
(Following up on #370.)  Got to discuss this issue a bit more with @titzer and @dschuff.  We found two root use cases:
* giving the VM a hint so that it can preallocate space to avoid realloc'ing in `grow_memory`
* telling the VM there will never be growth (via initial=max) so it can optimize based on this knowledge

For the former use case, `max` actually isn't a great hint because it must be set quite conservatively (greater than the peak usage of the app for all possible executions).  So a hint that could be non-fatally exceeded would be more accurate and thus useful to the VM.  Also, a hint doesn't punish the dev/user for failing to anticipate a corner case where memory peaks higher than what the dev observed during testing when all the dev wanted was a little perf boost.  Lastly, if the engine supports resizing, then it can be more aggressive about pre-allocating the max size since, if virtual address space gets tight in the process, the engine can release the extra preallcoated space and fall back on realloc.

For the latter use case, what's needed is just an optional flag `no_resize` (which constrains both the main module and all future dynamically-linked modules).  After some impl discussion, there doesn't appear to be any major known needs in the current anticipated engine impls so it seems best to leave it out of the MVP until there is a demonstrated need.  In a totally separate discussion about how JS can alias wasm linear memory, we found a pretty significant use case for `no_resize` though: with threads, wasm linear memory could move at any time so we can't simply hand out a SharedArrayBuffer to JS (and the rest of the browser).  Rather, we would need to restrict aliasing to only when `no_resize` is set and provide a threadsafe, but less efficient, copy in/out API fallback.

### DIFF
--- a/Modules.md
+++ b/Modules.md
@@ -8,7 +8,7 @@ is determined by the module it was loaded from.
 
 A module contains:
 * a set of [imports and exports](Modules.md#imports-and-exports);
-* a section defining the [initial state of linear memory](Modules.md#initial-state-of-linear-memory);
+* a section defining [linear memory](Modules.md#linear-memory-section);
 * a section containing [code](Modules.md#code-section);
 * after the MVP, sections containing [debugging/symbol information](Tooling.md) or
   a reference to separate files containing them; and
@@ -122,12 +122,22 @@ shared `malloc` and coordinated global address ranges). Instead, the
 [dynamic linking future feature](DynamicLinking.md) is intended
 to allow *explicitly* injecting multiple modules into the same instance.
 
-## Initial state of linear memory
+## Linear memory section
 
-A module will contain a section declaring the linear memory size (initial and
-maximum size allowed by [`grow_memory`](AstSemantics.md#resizing) and the
-initial contents of memory, analogous to `.data`, `.rodata`, `.bss` sections in
-native executables (see [binary encoding](BinaryEncoding.md#global-structure)).
+A module may contain an optional section declaring the use of linear memory
+by the module. If the section is absent, the linear memory operators
+`load`, `store`, `memory_size`, and `grow_memory` may not be used in the module.
+
+The linear memory section declares the initial [memory size](AstSemantics.md#linear-memory)
+(which may be subsequently increased by [`grow_memory`](AstSemantics.md#resizing)).
+
+The initial contents of linear memory are zero by default. However, the memory
+section contains a possibly-empty array of *segments* (analogous to `.data`)
+which can specify the initial contents of fixed ranges of memory.
+
+The linear memory section may also contain an optional hint declaring the maximum
+heap usage. This hint is not semantically visible but can help a WebAssembly engine
+to optimize `grow_memory`.
 
 ## Code section
 
--- a/Modules.md
+++ b/Modules.md
@@ -133,11 +133,12 @@ The linear memory section declares the initial [memory size](AstSemantics.md#lin
 
 The initial contents of linear memory are zero by default. However, the memory
 section contains a possibly-empty array of *segments* (analogous to `.data`)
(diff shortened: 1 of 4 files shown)


Modules.md
```diff
@@ -122,12 +122,22 @@ shared `malloc` and coordinated global address ranges). Instead, the
...
+`load`, `store`, `memory_size`, and `grow_memory` may not be used in the module.
+
+The linear memory section declares the initial [memory size](AstSemantics.md#linear-memory)
+(which may be subsequently increased by [`grow_memory`](AstSemantics.md#resizing)).
+
+The initial contents of linear memory are zero by default. However, the memory
+section contains a possibly-empty array of *segments* (analogous to `.data`)
+which can specify the initial contents of fixed ranges of memory.
+
+The linear memory section may also contain an optional hint declaring the maximum
```
sunfishcode: This might say "the expected maximum" to emphasize the non-normative nature of the hint.

Modules.md
```diff
@@ -122,12 +122,22 @@ shared `malloc` and coordinated global address ranges). Instead, the
...
+A module may contain an optional section declaring the use of linear memory
+by the module. If the section is absent, the linear memory operators
+`load`, `store`, `memory_size`, and `grow_memory` may not be used in the module.
+
+The linear memory section declares the initial [memory size](AstSemantics.md#linear-memory)
+(which may be subsequently increased by [`grow_memory`](AstSemantics.md#resizing)).
+
+The initial contents of linear memory are zero by default. However, the memory
+section contains a possibly-empty array of *segments* (analogous to `.data`)
+which can specify the initial contents of fixed ranges of memory.
```
sunfishcode: It'd be good to mention that segments are specified with an offset too.

sunfishcode:
lgtm, other than some minor wording suggestions above.

lukewagner:
Good points; updated as suggested.

Modules.md
```diff
@@ -122,12 +122,23 @@ shared `malloc` and coordinated global address ranges). Instead, the
...
-## Initial state of linear memory
+## Linear memory section
 
-A module will contain a section declaring the linear memory size (initial and
-maximum size allowed by [`grow_memory`](AstSemantics.md#resizing) and the
-initial contents of memory, analogous to `.data`, `.rodata`, `.bss` sections in
-native executables (see [binary encoding](BinaryEncoding.md#global-structure)).
+A module may contain an optional section declaring the use of linear memory
+by the module. If the section is absent, the linear memory operators
+`load`, `store`, `memory_size`, and `grow_memory` may not be used in the module.
```
jfbastien: I'm not sure I understand what purpose this serves.

lukewagner: This is actually summarizing a [previous spec PR](https://github.com/WebAssembly/spec/pull/107) which forgot to update design repo to match.

jfbastien: Could you mention this in Rationale.md?

lukewagner: Do you mean adding a link to the spec PR or copying the summary from the spec PR to Rationale.md?

lukewagner: Other places in Rationale.md link to GH, so I'll do the former.

jfbastien: Sorry, was hanging out with @sunfishcode :-)
I just meant to explain why you can't validate memory operations when you don't have memory declared.

Rationale.md
```diff
@@ -112,6 +112,11 @@ to communicate their own implementation details in a useful manner to the
 developer.
 
 
+## Linear memory disabled if no linear memory section
+
+See [#107](https://github.com/WebAssembly/spec/pull/107).
```
jfbastien: Looks good, the PR explains things pretty well :-)

PR #432, opened 2015-10-28
Move 'Order of evaluation' section

lukewagner:
Just tidying up: "Order of evaluation" section was accidentally put as a subsection of "Linear Memory".


jfbastien:
Order of evaluation was in the wrong order? Very nice. :-)

PR #433, opened 2015-10-28
Move low-memory faulting to FutureFeatures.md

lukewagner:
There was general consensus in #204 and again in recent discussions to have low-memory faulting be a post-MVP future feature since it fits in well with the other [fine-grained memory control](https://github.com/WebAssembly/design/blob/master/FutureFeatures.md#finer-grained-control-over-memory) features.  This PR moves mention of low-memory faulting from AstSemantics.md (which is the MVP) to FutureFeatures.md and expands on the option mentioned in #204 of treating low-memory separate from `mprotect` (since it allows more efficient impl when hw memory protection isn't available).


sunfishcode:
lgtm

titzer:
lgtm

PR #437, opened 2015-11-02
Control flow clarifications

kripken:
Two clarifications for things I noticed while implementing control flow in wasm-emscripten.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -242,7 +242,9 @@ All control flow structures, except `case`, are statements.
 
  * `block`: a fixed-length sequence of statements with a label at the end
  * `loop`: a fixed-length sequence of statements with a label at the end
-           and a loop header label at the top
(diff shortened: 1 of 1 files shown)


AstSemantics.md
```diff
@@ -242,7 +242,8 @@ All control flow structures, except `case`, are statements.
 
  * `block`: a fixed-length sequence of statements with a label at the end
  * `loop`: a fixed-length sequence of statements with a label at the end
-           and a loop header label at the top
+           and a loop header label at the top (note: this does not loop
+           by itself, a break to the top is required)
```
sunfishcode: `break` has been renamed to `br` and is not the only way to form a backedge. How about: "note: this does not loop by itself, so one would often combine this with a `br_if` at the end to form a branch back to the top"?

kripken: updated.

sunfishcode:
lgtm, thanks!

PR #441, opened 2015-11-03
Full IEEE 754-2008 compliant NaN bit pattern propagation.

sunfishcode:
The current NaN propagation rules leave NaN payloads nondeterministic after most operators, which makes them inefficient to use for NaN-boxing in language interpreters running on top of WebAssembly.

This PR addresses this use case by changing WebAssembly to use full IEEE 754-2008 conforming NaN bit propagation to WebAssembly. This eliminates almost all of the non-determinism, though a little bit remains (binary operators that receive multiple NaN operands have to pick one to form their return value -- IEEE 754-2008 itself has this nondeterminism too). It also goes beyond IEEE 754-2008 in that it specifies the bit-pattern of a NaN generated from an invalid condition.

The algorithms described here for computing NaN result values admittadly look somewhat elaborate, but this is what x86, ARM (in its default mode, though not in "Default NaN" mode), Power, and MIPS (in IEEE 754-2008 mode, though not in legacy mode) all do. A possible exception is conversions on MIPS, where the documentation I have is vague -- it suggests that the bits are preserved in a conversion, but it doesn't seem to say exactly how -- however in the worst case a MIPS implementation could check for NaN after a conversion and handle it explicitly in a cold path if necessary.

This patch leaves the `min` and `max` operators in place, even though they'll be harder to implement on x86 under the new NaN rules, since they'll remain easy on ARM and Power, and are desirable instructions to have in an ISA in general.

### DIFF
--- a/Nondeterminism.md
+++ b/Nondeterminism.md
@@ -30,9 +30,10 @@ currently admits nondeterminism:
    shared memory, nondeterminism will be visible through the global sequence of
    API calls. With shared memory, the result of load operators is
    nondeterministic.
- * NaN bit patterns in floating point
-   [operators](AstSemantics.md#floating-point-operators) and
-   [conversions](AstSemantics.md#datatype-conversions-truncations-reinterpretations-promotions-and-demotions)
+ * Except when otherwise specified, when an arithmetic operator with multiple
+   floating point operand types and a floating point result type recieves
+   multiple NaN input values with differing bit patterns, it is nondeterminsitic
+   which bit pattern is used as the basis for the result.
  * [Fixed-width SIMD may want some flexibility](PostMVP.md#fixed-width-simd)
    - In SIMD.js, floating point values may or may not have subnormals flushed to
      zero.
--- a/Nondeterminism.md
+++ b/Nondeterminism.md
@@ -31,7 +31,7 @@ currently admits nondeterminism:
    API calls. With shared memory, the result of load operators is
    nondeterministic.
  * Except when otherwise specified, when an arithmetic operator with multiple
-   floating point operand types and a floating point result type recieves
+   floating point operand types and a floating point result type receives
    multiple NaN input values with differing bit patterns, it is nondeterminsitic
    which bit pattern is used as the basis for the result.
  * [Fixed-width SIMD may want some flexibility](PostMVP.md#fixed-width-simd)
--- a/Nondeterminism.md
+++ b/Nondeterminism.md
@@ -33,7 +33,8 @@ currently admits nondeterminism:
  * Except when otherwise specified, when an arithmetic operator with multiple
    floating point operand types and a floating point result type receives
    multiple NaN input values with differing bit patterns, it is nondeterminsitic
-   which bit pattern is used as the basis for the result.
+   which bit pattern is used as the basis for the result (as it is in
+   IEEE 754-2008).
  * [Fixed-width SIMD may want some flexibility](PostMVP.md#fixed-width-simd)
    - In SIMD.js, floating point values may or may not have subnormals flushed to
      zero.
--- a/Polyfill.md
+++ b/Polyfill.md
@@ -70,6 +70,10 @@ Some divergences that we've identified as potentially desirable:
   standard behavior:
   - Return zero when conversion from floating point to integer fails;
   - Optionally canonicalize NaN values.
+* **[NaN bit-pattern propagation](AstSemantics.md#floating-point-operators)**:
+  Regardless of WebAssembly behavior, an asm.js polyfill will follow its
+  standard behavior:
+  - Optionally canonicalize NaN values.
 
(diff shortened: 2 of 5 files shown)


JSStats:
For the use case noted, wouldn't it only be necessary to preserve the NaN across bit copy operations, for example arguments and loads and stores? If the interpreter sees a NaN tagged value then I presume it will not be applying float operators to it, rather interpreting the bits as another type of data, a pointer or other immediate.

What are the implications for an asm.js polyfill?

sunfishcode:
The current semantics already do preserve NaN bit patterns across loads and stores. This PR allows a JIT using NaN boxing to further assume that if the input to a sequence of arithmetic instructions are all canonical values, that the output will also be a canonical value, so it can avoid having to re-canonicalize whenever it needs canonical values.

The polyfill would not be able to guarantee this behavior. I added a note to Polyfill.md about this.

JSStats:
Thank you for the clarification - similar words might be useful to add to the rationale etc.

Odin canonicalizes NaNs so will not preserve a non-canonical NaN across loads and stores?

If the use case here is to define canonical NaNs to be preserved, then Odin seems compatible, assuming the Odin and the wasm apps canonical representation is the same?

sunfishcode:
ECMAScript doesn't make any guarantee about which NaN pattern an implementation might canonicalize to, so I don't want to say anything stronger about the polyfill here.

Nondeterminism.md
```diff
@@ -30,9 +30,10 @@ currently admits nondeterminism:
    shared memory, nondeterminism will be visible through the global sequence of
    API calls. With shared memory, the result of load operators is
    nondeterministic.
- * NaN bit patterns in floating point
-   [operators](AstSemantics.md#floating-point-operators) and
-   [conversions](AstSemantics.md#datatype-conversions-truncations-reinterpretations-promotions-and-demotions)
+ * Except when otherwise specified, when an arithmetic operator with multiple
+   floating point operand types and a floating point result type recieves
```
jfbastien: Typo: recieves

Nondeterminism.md
```diff
@@ -30,9 +30,10 @@ currently admits nondeterminism:
...
+   multiple NaN input values with differing bit patterns, it is nondeterminsitic
+   which bit pattern is used as the basis for the result.
```
jfbastien: "as per IEEE-754-2008"

Polyfill.md
```diff
@@ -70,6 +70,10 @@ Some divergences that we've identified as potentially desirable:
   standard behavior:
   - Return zero when conversion from floating point to integer fails;
   - Optionally canonicalize NaN values.
+* **[NaN bit-pattern propagation](AstSemantics.md#floating-point-operators)**:
+  Regardless of WebAssembly behavior, an asm.js polyfill will follow its
```
jfbastien: It's not really because the polyfill is asm.js that NaN behavior isn't specified. Just say "JavaScript polyfill" here instead?

jfbastien:
Could you add something to Rationale.md explaining why NaN bit pattern preservation matters.

IIRC MIPS uses a different NaN encoding for s/q than other ISAs, and recent chips changed (maybe with a mode switch) to follow what everyone else did. I'm not sure this is exactly in line with what you wrote in the PR's description, could you clarify?

Polyfill.md
sunfishcode: This is just copied from similar text in other parts of the paragraph.

sunfishcode:
The MIPS mode switch you're thinking of is the IEEE 754-2008 mode, which is what I referred to in the PR's description. Specifically, it's the FCSR NAN2008 flag.

Polyfill.md
jfbastien: Hmm, that's unfortunate: I think the polyfill should detail what's different because JS semantics are that way, versus what's different because asm.js' subset of JS is that way. e.g. the alignment stuff could "just work" with JS but perf would suffer because asm.js.

jfbastien:
Ah yes, the MIPS thing is right. Could you also mention these aspects in rationale.md (what you have in the PR's description)?

Otherwise lgtm, though it would be nice to also tidy up the polyfill stuff to express JS- versus asm.js-imposed restrictions.

Polyfill.md
sunfishcode: JS has the same restrictions as asm.js with respect to alignment.

sunfishcode:
Added a rationale paragraph mentioning MIPS and the NAN2008 configuration flag. We can address JS vs asm.js in a separate PR, since it'd be an unrelated change.

jfbastien:
lgtm

sunfishcode:
Merging, with LGTM above. Of course, this is all currently supported by a survey of available hardware documentation for popular architectures. If we discover actual hardware which doesn't implement these rules, we should open a new issue and consider changing the spec.

PR #449, opened 2015-11-05
Application development scenario

mtrofin:
Added an application development scenario.
Organized the scenarios as a broader section, added an introduction.


JSStats:
I don't think this is useful, sorry. It is a bit hard for me to articulate why, but it might be the lack of a solid technical problem, the lack of a solid effort to solve these problems, the raising of new problems and proposed solutions to them embedded in such examples rather than raising them in issues, and the inability to assess them on technical grounds. For example, the 'unchecked' feature is not something I have even seen discussed, and it might have misunderstood the purpose that other people had in mind for the non-determinism.

mtrofin:
The Unchecked stuff shouldn't be here, I separated that away, it's a scribble I was working on. If it matters it shows in the history, I can take care of that.

JSStats:
A `git rebase -i` might help clean up the noise.

What I would fine useful are benchmarks for a codec use case and other cases. Even then I think it would be a bit too much to just dump them on the designers, and that some effort to analyse performance problem and to optimize for wasm would be expected.

mtrofin:
Thanks for the git tip - I'll wait for more feedback and then clean it up accordingly and resubmit.

Interesting comment about the designers' experience with perf. Could you expand on what the problem would be, I think it would make a great scenario to add to the scenario pack!

JSStats:
It seems a courtesy to reviewers to clean this up now - now that you know how.

These scenarios don't seem useful to me as they are not sufficient to make any wasm design decisions, and I don't think they really explain the rationale, and they seem well into bike shedding territory. I don't see designers writing these stories in annotated language design documents. I suggest leaving such comments out of the spec and rationale, unless they relate to a specific design decision.

I think everyone would agree that wasm should support high performance codecs, and they are working away at it. Adding benchmarks and analysing performance and filing performance issues seem like constructive activities.

I've said my piece on this, and I'm not going to buy into this again. Someone will problem just ltgm it.

Rationale.md
```diff
@@ -275,6 +275,60 @@ architectures there may be a need to revisit some of the decisions:
...
+Some scenarios may closely relate to a design area, such as feature testing,
+while others may crosscut design areas and even get out of the scope of the
+runtime standardization effort. This is intentional, in order to ensure we have
+a clear picture of the broader context in which WebAssembly will be used.
+
+## Application Development Scenarios
+
+### Plugin-free Application Deployment
+A startup wants to offer a new 4K video streaming service to a great breadth of
+consumers, across any device - from mobile, to browser, to TV appliance. Their
```
jfbastien: Em dash :D

jfbastien:
I'm not really a huge fan of user narratives. We have UseCases.md which is pretty short & sweet, I feel like this is kind of TL;DR at this point.

I agree that the current site doesn't do a great job of explaining *why* someone would want wasm if it's not already obvious to them, but my gut feel is that to convince these skeptics we'd need a narrative such as what you wrote that speaks directly to them. That seems better left to specialized parts of the internet (e.g. a gaming website writes an article about wasm, or a codec enthusiast forum writes a wasm breakdown, an Amazon reviewer writes a 5 star review for wasm, ...).

mtrofin:
I think there's value in spelling out the "why": makes coming in and contributing easier, helps maintain focus, helps keep the big picture in the mind of participants, and I personally found it a good exercise in better understanding the problem space, because it adds nuance.

That's my $0.02, and I'm happy to close the PR if the arguments above aren't compelling enough.

In that case, do we see value in capturing the known problem space for more specific, in-flux areas, like feature testing? Maybe I'm taking your comment about user narratives out of context, but I have trouble seeing how else to effectively capture a problem than via a narrative: somebody wants to do this thing and hits a problem - it's hard/expensive/can't be done/etc.

lukewagner:
Agreed with what @jfbastien and @JSStats said above; I'm not sure this is the right place for this info.

mtrofin:
Happy with moving it elsewhere. Should we have a separate .md file where we capture our understanding of the problem domain (i.e. User-focused scenarios)?

PR #451, opened 2015-11-05
remove unneeded comma

chicoxyzzy:


sunfishcode:
This is a [serial comma](https://en.wikipedia.org/wiki/Serial_comma), which is accepted usage in numerous popular style guides.

PR #455, opened 2015-11-09
Refine linear memory aliasing

lukewagner:
This PR adds some more details to how linear memory is exported to JS from WebAssembly in Web.md to reflect more recent discussions concerning the hazards of resizing, `mprotect` and threads.

Heap *exports* also begs the question of heap *imports* which this PR takes the liberty of adding to FutureFeatures.md.

### DIFF
--- a/Modules.md
+++ b/Modules.md
@@ -140,6 +140,14 @@ The linear memory section may also contain an optional hint declaring the expect
 maximum heap usage. This hint is not semantically visible but can help a
 WebAssembly engine to optimize `grow_memory`.
 
+The linear memory section may optionally declare that the instance's
+linear memory is *externally aliasable*. How linear memory is aliased is up
+to the host environment (as with all module exports). The 
+[Web](Web.md#aliasing-linear-memory-from-JS) would reflect exported linear
+memory to JS as an `ArrayBuffer`. The MVP does not currently provide for
+*importing* linear memory though this may be added 
+[in the future](FutureFeatures.md#importing-linear-memory).
+
 ## Code section
 
 The WebAssembly spec defines the code section of a module in terms of an
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -399,3 +399,13 @@ significant portions of memory could be moved out of linear memory which could
 reduce fragmentation issues. Languages like Fortran which limit aliasing would be
 one use case. C/C++ compilers could also determine that some global variables never
 have their address taken.
+
+## Importing linear memory
+
+In the MVP, functions and [linear memory](Modules.md#linear-memory-section) can
+be exported, but only functions can be imported. This feature would additionally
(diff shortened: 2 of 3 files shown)


lukewagner:
Oops, forgot about this.  Any objections?

titzer:
lgtm

Modules.md
```diff
@@ -140,6 +140,14 @@ The linear memory section may also contain an optional hint declaring the expect
 maximum heap usage. This hint is not semantically visible but can help a
 WebAssembly engine to optimize `grow_memory`.
 
+The linear memory section may optionally declare that the instance's
+linear memory is *externally aliasable*. How linear memory is aliased is up
+to the host environment (as with all module exports). The 
+[Web](Web.md#aliasing-linear-memory-from-JS) would reflect exported linear
+memory to JS as an `ArrayBuffer`. The MVP does not currently provide for
+*importing* linear memory though this may be added 
+[in the future](FutureFeatures.md#importing-linear-memory).
```
jfbastien: I'm pretty wary of how this will interact with multi-threading and SAB. In C++ it's a bit of a fiasco to use shared memory:
* `volatile` should be used to denote external modification, so the compiler remains conservative.
* `std::atomic` should be used to enforce ordering and no-tear (yes, in addition to `volatile`).
* Non-lock-free atomics are guaranteed to work in an address-free way *within the same process* but definitely do *not* work when using shared memory across processes.

This makes me wary of specifying things as you've outlined here without properly understanding what this implies for threads in WebAssembly. Does it entirely prevent optimizations? Valid reordering? 64-bit atomics on some platforms?

I'm OK specifying this as you've done fi you think it'll move stuff forward, but I insist on opening an issue with the above text, to revisit *before* MVP so we're not stuck when we want to add threads.

WDYT?

lukewagner: I'm having trouble understanding your specific concern.  At the top you mention wasm+SAB interactions but then the rest of your comment seems to be general memory model concerns that are not specific to SAB.  There are some real concerns about resize+mprotect/SAB interactions outlined in Web.md, but those are quite different than what you're talking about here.  Moreover, we have agreed that, because threads (not just SAB interactions) are such a cross-cutting concern, that we should have a good experimental prototype up and working before v.1 is finished; that's #104.  So I'm not sure what specifically goes in a new issue.

jfbastien: The text reads as "oh I can just say externally alias and then everything works" but I see it as an undefined hole in the memory model. A hole we can plug, but will require some pretty serious design work that I'm not sure we want to tackle for MVP considering C++ hasn't gotten it right either.

That being said, maybe we can go with it for now, I can open an issue, and we can revisit when we have a clearer idea backed by code?

lukewagner: sgtm, happy to discuss more in that issue.

jfbastien: Thanks, I opened #485.

Issue #445, opened 2015-11-03
Break / Branch?


kripken:
This repo currently has `br: branch to a given label in an enclosing construct`. I assume the intention is that

````
(block $block
  (branch $block) // goes to $more
)
(block $more)
)
````

Is my understanding correct?

If so, isn't this more of a "break" than a "branch"? We not are not actually branching to `$block`, we are going to `$more` in fact. But the name "branch" implies to me "branch **to**". It seems nicer to do either

1. `(branch $more)`, which says "branch to `$more`" (where we actually branch to), or
2. `(break $block)`, which says "break on `$block`" (just like a C/JS/etc. break on a labeled block),

over the current state, which says `(branch $block)` but actually does not branch to `$block`.

Obviously 1 is a forward goto, which is bad. How about 2?

**Comment by kripken** [comment]
(bonus: we don't need to rename the `br` ;)

**Comment by sunfishcode** [comment]
Another approach is to rearrange the syntax. Instead of putting the label at the top of a block, we could put it at the bottom:

```
(block
  (blah)
  (blah)
  (blah)
  $label
)
```

which makes it quite clear where control goes on a branch with that label.

**Comment by rossberg-chromium** [comment]
I agree that "break" is the more adequate term.

@sunfishcode, labels are symbolic names for AST nodes, not "positions" in sequential code. The syntax you propose would not be good fit for that.

**Comment by sunfishcode** [comment]
@rossberg-chromium It's a new language, so we can make labels be whatever we want. This issue appears subjective at this point.

**Comment by lukewagner** [comment]
@rossberg-chromium You could consider the syntax for the `block` nodes to be `(block ... $label)`, so it's still just syntax for an AST node.

Furthermore, if we do the analogous thing for loop:
```
(loop $back
  (blah)
  (blah)
  $forward)
```
then it seems 10x more readable and intuitive than `(loop $back $forward (blah) (blah))`.  Also, I do think "branches" is the right metaphor here as we're designing an assembly language (it's in the name!).

rossberg-chromium:
@sunfishcode, well, if it involves changing the semantics of labels then it would be a change to the language as it has been spec'ed and implemented so far (e.g. in v8-proto). So in that sense, it's not just subjective.

[...]

PR #457, opened 2015-11-10
Split the text format

kripken:
Followup for #445:

 * Introduce a split between a higher and lower text format, one more assembly-like, the other more web-like, as proposed in the middle of [this comment](https://github.com/WebAssembly/design/issues/445#issuecomment-154987087). 
 * Update the FAQ on View Source following recent discussions.

Obviously there was a lot of debate in #445, this isn't meant to imply there is already a consensus. This pull is more to give us something more concrete to discuss.

### DIFF
--- a/TextFormat.md
+++ b/TextFormat.md
@@ -16,14 +16,6 @@ The text format will be standardized, but only for tooling purposes:
 * Browsers will not parse the textual format on regular web content in order to
   implement WebAssembly semantics.
 
-Given that the code representation is actually an
-[Abstract Syntax Tree](AstSemantics.md), the syntax would contain nested
-statements and expressions (instead of the linear list of instructions most
-assembly languages have).
-
-There is no requirement to use JavaScript syntax; this format is not intended to
-be evaluated or translated directly into JavaScript.
-
 The text format isn't uniquely representable. Multiple textual files can assemble
 to the same binary file, for example whitespace isn't relevant and memory initialization
 can be broken out into smaller pieces in the text format.
@@ -34,7 +26,39 @@ represented as hexadecimal floating-point as specified by the C99 standard, whic
 IEEE-754-2008 section 5.12.3 also specifies. The textual format may be improved to also
 support more human-readable representations, but never at the cost of accurate representation.
 
-# Official Text Format
+# Lower and Higher text formats
+
+Note that there is some tension in the above purposes and goals: On the one
+hand, WebAssembly is designed for the **Web**, and should allow View Source to
+let people inspect code on the sites they visit in as clear and as natural
+a way as possible. On the other hand, WebAssembly is an **assembly** language,
+low-level and designed as a compiler-target; it is not a high-level language
+intended for ease of human readability.
+
+To bridge that gap, we plan to have *two* text formats:
+
+ * "Lower" or "Assembly" text format: A close parallel to the binary format,
+   very assembly-like. This is what compiler writers targeting WebAssembly
+   would look at.
+ * "Higher" or "Web" text format: Designed to maximize readability,
+   this format can include various forms of syntax sugar. This is
+   what would be used in browsers' View Source.
+
+The two text formats will be directly translatable beween each other, and
+hence concepts in each must must have direct parallels in the other. However,
+they may differ in how they represent those concepts. For example, in the
+Higher/Web text format we might have loops, breaks and continues, while in
+the Lower/Assembly text format breaks and continues might be replaced with
+branches and labels; given that the code representation is actually an
+[Abstract Syntax Tree](AstSemantics.md), translating between breaks and
+branches is straightforward.
+
+There is no requirement to use JavaScript syntax; these formats are not intended to
+be evaluated or translated directly into JavaScript. However, where possible,
+it is beneficial to remain as familiar as possible in the Higher/Web text
+format for web developers.
+
+# Status
 
 WebAssembly currently doesn't have a final, official, text format. As detailed above the
 main purpose of the text format will be for human consumption, feedback from humans on
--- a/FAQ.md
+++ b/FAQ.md
@@ -142,17 +142,21 @@ work with them on ABI matters.
 
 ## Will WebAssembly support View Source on the Web?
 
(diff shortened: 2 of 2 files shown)


TextFormat.md
```diff
@@ -16,14 +16,6 @@ The text format will be standardized, but only for tooling purposes:
 * Browsers will not parse the textual format on regular web content in order to
   implement WebAssembly semantics.
 
-Given that the code representation is actually an
```
kripken: (these two paragraphs are moved lower down, they aren't gone)

kg:
I'm not okay with putting this in the public-facing design documents with the degree of finality in this PR. I don't think we've proven that this is a good idea.

TextFormat.md
```diff
@@ -16,14 +16,6 @@ The text format will be standardized, but only for tooling purposes:
 * Browsers will not parse the textual format on regular web content in order to
   implement WebAssembly semantics.
```
kripken: (these two paragraphs are moved lower down, they aren't gone)

kripken:
As discussed on irc, I can move it to a side .md file if that's preferred. But also I think we could just leave this open for as long as the discussion takes - obviously we should only merge this, with its changes to the core design docs, after we are all happy with it.

JSStats:
I think it would also help to add. 'Supporting the structure of the high level view-source text format is a matter that should be considered in the design of the AST and binary format and that this might have some grounds to be assessed on technical terms for the computational model and the pipeline, and supporting code transforms in the pipeline might also have technical implications, but the low level structure could support a wide range of higher level text formats and these choices might be made on terms that do not impact the technical operation of wasm so have been split from the technical matters.'

kripken:
@JSStats: I think there are a few good points in there, but (1) I'd prefer to not add anything here right now, before seeing more responses - I'd like to only remove or fix things, as we already have a lot to discuss here. And (2) I think the points you make could be split out separately, as one long sentence it is a little hard for me to follow.

AndrewScheidecker:
Why not just make the "web format" C++? It seems more useful than defining a new format that tries to make WebAssembly human-friendly.

The way I picture this working is that the browser view source still gives you WebAssembly, and then we make sure there's a good WebAssembly->C++ decompiler available.

The advantage to this approach is that it gives you access to all the great C++ tools out there without burdening WebAssembly tools with supporting two different text formats.

JSStats:
@AndrewScheidecker C++ might not naturally express the wasm AST making it rather convoluted to define the transform back from the view-source to the binary code. For example, multiple return values, the planned GC object support, the linear memory access effective address calculation, etc. Also C++ might have undefined behaviour in areas that wasm demands it to be well defined. I expect it could be all handled in C++ with suitable abstractions, but then it's not really raw C++ but a language defined by the abstractions added to C++.

kripken:
@AndrewScheidecker: As @JSStats says, the WebAssembly AST is not a direct parallel to C++, it is both lower-level than C++, and intended to be a compiler target for other languages as well. However, I agree that where possible, we might as well make the "web" text format as familiar as possible to developers, including C++ developers.

AndrewScheidecker:
> multiple return values

This translates to functions returning a struct.

> the linear memory access effective address calculation

This seems no more complicated than e.g. decompiling an x86 memory access.

> the planned GC object support

For the foreseeable future, the people producing WebAssembly will be compiling it from C++, so it seems like this feature would need to be exposed via C++ somehow for it to be useful.

> the WebAssembly AST is not a direct parallel to C++, it is both lower-level than C++, and intended to be a compiler target for other languages as well

My point is that a different syntax for the WebAssembly AST isn't very helpful. The nicest syntax won't be very readable if it doesn't reverse engineer structs, local arrays, global variables, string literals, pointer types, gotos, etc.

I suggest C++ (or C) because it's plausible that the decompiler output could be recompiled into a program with the same behavior. It would be hard to get the same binary out of it, but I don't think that's a problem.

kripken:
I think we mostly agree. Yes, as much as is feasible, if we can infer structs, control flow, etc., we should do that in the "higher" text format. I just wouldn't say "C++" because

1. It will be a tiny, tiny subset of C++.
2. We will have things like `i32` for compatibility with the spec, which I doubt we'd want to replace with `int32_t` or `uint32_t` (it's also not directly comparable to those, because it lacks a sign).
3. It's useful to be familiar to not just C++ developers, but other relevant  groups, like current web developers.
4. We need something as forward-compatible as possible, so GC support, while not relevant in the short term, will matter. It's very possible that GC support will not be exposed to C++, unless it's via some API like Boehm, but we would want a more natural way to express GC objects in the text format (optimally something like JavaScript, C#, Java, D, etc.).

JSStats:
@AndrewScheidecker It seems that it would be a requirement for a view-source format that it can be parsed and encoded back into the wasm binary format. I think it would be difficult or fragile to try and achieve this by re-compiling view-source C++ with standard tools, so a dedicated wasm-c++ to wasm-binary compiler would be needed.

I think this discussion demonstrates the problem with trying to define a text format - 'familiar' is subjective and it can't be resolved on technical merit. I don't think this should block or distract the core technical work. It could be deferred by allowing plugin encoder/decoder pairs, and these could be developed separately from the wasm specification by people accepting their development and maintenance cost versus their utility and the value they assign to being 'familiar'.

Personally s-expressions are familiar for me, and I know they are light weight, but I have a conflict of interest and can not buy into the choice of format, and I don't many people would make this choice. Can someone rally support for a de facto standard, develop the encoder/decoder etc, and do so separately from the wasm core technical work.

kripken:
There doesn't seem sufficient interest in discussing this - perhaps it is too early. I plan to do some practical experiments with text formats soon, and maybe we'll get back to this splitting proposal then. Meanwhile, closing this.

jfbastien:
(catching up to things post-vacation)

I'm also not a huge fan of having two formats when one would do. I get the rationale, but I'd really like to avoid two formats! I think we'll get a better picture as we go into beta and developers start giving us feedback. In the meantime my weak preference is to go with assembly-like syntax that's precise, readable, but not particularly writable.

Issue #422, opened 2015-10-22
Include byteswap and rotates in the MVP


mbodart:
Of the integer operations listed in FutureFeatures.md, it seems bswap and the rotates belong in the MVP.  I imagine they have as much utility and direct hardware support as popcnt, clz and ctz.

They seem trivial to support, and increase the richness of the MVP, so I don't see why we would defer them.

I created PR https://github.com/WebAssembly/spec/pull/149 for support in the spec interpreter.

Of course, AstSemantics and other tools would need updating as well.

What say all?

JSStats:
+1 to bringing low hanging fruit into the MVP.

jcbeyler:
+1 too :)

jfbastien:
Hmm, reading @sunfishcode's comments on #121 I'm a bit confused... It sounds like he added bswap based on a comment, but the description says otherwise because it's hard to polyfill. I don't quite recall.

sunfishcode:
The observation is: we don't need to worry about the polyfill being slow for operators like these, because it'll be no worse than what apps that need this functionality would do instead.

These operators were removed from the MVP due to a general desire to keep the MVP focused. bswap in particular is tricky because 32-bit and 64-bit bwap begs a 16-bit bswap too. It'd probably have to have the extra behavior of zeroing the high 16 bits, and it'd be asymmetric between i32 and i64. And, since some architectures fold bswaps into loads and stores (e.g. x86's movbe instruction, and other architectures have similar things), would it be better to have an optional endian attribute on regular loads and stores, instead of a standalone integer operator?

Rotates on the other hand are simple and have prominent use cases. :-)

titzer:
Since this operation is pretty common and there are architectures with this
explicit instruction, this seems OK to me, but I'm not sure if it's
pressing enough to add to MVP. We'll probably implement this "just for fun"
but not sure all engines need to in MVP.

mbodart:
As with any large project I can understand the desire to avoid feature creep in the MVP,
towards keeping schedules manageable and allowing sufficient time for good design.

But bswap and rotate seem straight forward to define and implement, certainly no more difficult,
nor less pressing I would imagine, than popcnt/clz/ctz.  Low hanging fruit, as @JSStats put it.
So I don't see what holding them back buys anyone.

The 16-bit argument seems generically applicable to all integer operations.
Adding 32/64-bit flavors now does not inhibit adding 16-bit flavors later.

[...]

PR #462, opened 2015-11-12
Move rotate instructions from FutureFeatures to AstSemantics.

mbodart:
This is the follow up to design issue #422, adding rotate instructions to the MVP (but not bswap at this time).

You'll notice I added a description of wrapping behavior on the rotate count.
I don't think this is strictly needed as rotates, unlike shifts, have well defined
and consistent mathematical behavior with large counts.  And arguably even
negative counts would "just work" if treated as a rotate in the opposite direction.

For simplicity I think it's reasonable to define the count as unsigned.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -369,6 +369,8 @@ results into the result type.
   * `i32.shl`: sign-agnostic shift left
   * `i32.shr_u`: zero-replicating (logical) shift right
   * `i32.shr_s`: sign-replicating (arithmetic) shift right
+  * `i32.rotl`: sign-agnostic rotate left
+  * `i32.rotr`: sign-agnostic rotate right
   * `i32.eq`: sign-agnostic compare equal
   * `i32.ne`: sign-agnostic compare unequal
   * `i32.lt_s`: signed less than
@@ -387,6 +389,11 @@ Shifts counts are wrapped to be less than the log-base-2 of the number of bits
 in the value to be shifted, as an unsigned quantity. For example, in a 32-bit
 shift, only the least 5 significant bits of the count affect the result. In a
 64-bit shift, only the least 6 significant bits of the count affect the result.
+
+Rotate counts are treated as unsigned.  A count value greater than or equal
+to the number of bits in the value to be rotated yields the same result as
+if the count was wrapped to its least significant N bits, where N is 5 for
+an i32 value or 6 for an i64 value.
 
 All comparison operators yield 32-bit integer results with `1` representing
 `true` and `0` representing `false`.
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -241,8 +241,6 @@ use cases:
 * The following operators can be built from other operators already present,
   however in doing so they read at least one non-constant input multiple times,
   breaking single-use expression tree formation.
-  * `i32.rotr`: sign-agnostic bitwise rotate right
-  * `i32.rotl`: sign-agnostic bitwise rotate left
   * `i32.min_s`: signed minimum
   * `i32.max_s`: signed maximum
   * `i32.min_u`: unsigned minimum


sunfishcode:
I think it's reasonable to define the count as unsigned too. This looks good to me.

jfbastien:
On the issue @titzer didn't oppose adding rotate, but suggested that we may still want to keep it post-MVP. Maybe to quantify the "cost" of rotate it would be useful to do a PR to https://github.com/WebAssembly/v8-native-prototype and add the operation there? If it's truly trivial to add to V8 then I think we'd have a pretty good rationale to add it to MVP from at least one engine's point of view.

WDYT @titzer?

titzer:
What's the process for moving FutureFeatures into the MVP? At some point we
have to cut off the feature set for MVP and FutureFeatures was one
mechanism we had to do that. I'm not particularly concerned about rotates,
but are they MVP-urgent, balanced against their cost?

kg:
One of our repeatedly stated design goals is that we want it to be very fast and easy to iterate on the spec by adding new features. Given this, it seems like rotates would need to be extremely high-priority for us to pull them in the MVP - they're a small, independent feature that could easily be specced and shipped at any point after launch.

sunfishcode:
The small, independent nature of this feature is what makes it low cost to include in the MVP. I wouldn't describe them as urgent, but useful and not complicated.

mbodart:
Given the nature of this project it is difficult to quantify the cost/benefit of most instructions.
Pattern recognition by an engine is often feasible, which would argue for keeping the ISA small.
But why not take advantage of C/C++ compilers' ability to do the recognition?  This would lead to smaller wasm, and be one less feature that compilers have to worry about versioning later.

It's hard to imagine the implementation cost for rotates being any more expensive than popcnt/clz/ctz, or the recently added select operator.  How is rotate different from these instructions?

Rotates by themselves are not critical for the MVP.  But in general the richer we can make the MVP, the more likely it is to be adopted.

kg:
> Pattern recognition by an engine is often feasible, which would argue for keeping the ISA small.
But why not take advantage of C/C++ compilers' ability to do the recognition? This would lead to smaller wasm, and be one less feature that compilers have to worry about versioning later.

Features aren't versioned, they're feature-detected. And part of the feature detection design assumes a way to polyfill some subset of unsupported instructions. Any 'rotate' instruction in the opcode table at the top of a module could also include a simple wasm polyfill that implements the rotate.

This isn't a matter of 'rotate is hard' or 'rotate is easier than popcnt', it's more the fundamental question: When are we going to stop pulling features into the MVP? How do we want to draw the line? Arguably, anything nonessential can be bumped until shortly after the MVP.

> in general the richer we can make the MVP, the more likely it is to be adopted.

A MVP need only meet the minimum threshold to be viable. Being adopted everywhere is something you optimize for once you've got a working product out the door.

jfbastien:
My opinion: this is small, simple, and useful. Having a feature test for it *forever* seems to outweigh the saving of putting it in FutureFeatures. I suggest validating this assumption by implementing it in v8-proto before moving to MVP.

This is a very ad-hoc process, but I don't think we need to overthink it either :-)

lukewagner:
No strong opinion on rotate, but just wanted to point out that features missing from the MVP won't have to be feature-tested *forever* just like people don't feature test Web APIs forever: at some point a feature reaches ubiquity and can just be assumed.

mbodart:
Per the suggestion from @jfbastien the rotates were prototyped in https://github.com/weilianglin/v8-native-prototype/tree/rotate

The changes were straightforward.  The line count is misleading as insertion of opcode numbers resulted in shifting many others by 1.

I believe this demonstrates the low effort/cost of adding this support,
and respectfully request approval of this PR.

The spec PR 149 was also updated to remove bswap, and rename the rotate instructions to rotl/rotr.

sunfishcode:
Ping; this PR adds the rotate instructions, which were received favorably in offline discussions.

titzer:
lgtm

@mbodart once had a patch to add these to V8. If they can forward port it to V8 today, I'll land that too :-)

jfbastien:
Merging, since we seem to have consensus on these.

weilianglin:
@titzer sorry for missing it. I will submit a patch to V8 today.

PR #463, opened 2015-11-12
Mention consistency on the web as a factor in the text format

kripken:
Based on recent discussions, I think now might be a good time to start exploring and prototyping text formats, to get some experience before more organized discussions on that topic. I plan to do so myself soon. Thinking about how to get started, I think this addition to TextFormat.md is useful.

### DIFF
--- a/TextFormat.md
+++ b/TextFormat.md
@@ -22,7 +22,14 @@ statements and expressions (instead of the linear list of instructions most
 assembly languages have).
 
 There is no requirement to use JavaScript syntax; this format is not intended to
-be evaluated or translated directly into JavaScript.
+be evaluated or translated directly into JavaScript. There may also be
+substantive reasons to use notation that is different than JavaScript (for
+example, WebAssembly has a 32-bit integer type, and it should be represented
+in the text format, since that is the natural thing to do for WebAssembly,
+regardless of JavaScript not having such a type). On the other hand,
+when there are no substantive reasons and the options are basically
+bikeshedding, then it does make sense for the text format to match existing
+conventions on the Web (for example, curly braces, as in JavaScript and CSS).
 
 The text format isn't uniquely representable. Multiple textual files can assemble
 to the same binary file, for example whitespace isn't relevant and memory initialization


kg:
I would suggest that a good place to start would just be refinements to the s-expression format. If we can find a pleasing set of adjustments to s-expressions, we can eventually just iterate towards that and then we still have One Format.

I'm not sure whether that would ever be nice enough for people to be happy with it, though.

kripken:
I agree, s-expressions are a good starting point. The change in this pull is more as a longer-view goal to try to reach, as opposed to where to start. In fact, I was going to start with the s-expression support in binaryen and work from there. But I did hope to be able to get to a place where it feels consistent with the web - too early to tell, of course.

kripken:
(Btw, one specific thing I like about s-expressions is that there is no ambiguity of operator precedence, and we do have potential conflicts here, e.g., operator precedence is not the same between JS and C, so that's a hard choice, but s-expressions might avoid it. But, hard to know before experimenting with concrete proposals, I think.)

jfbastien:
I'm not sure that the text format really needs to be close to the web's expectations as a design criteria. I think being non-alien does matter, but so does avoiding misleading syntax (e.g. the syntax is precise, `==` isn't the same as JS' `==` which may be surprising, etc). We should avoid being *too much* like JS lest @BrendanEich be blamed *again* for silly things. I do like curlies, but are we going to go with ASI? ;-)

I think it would be better to express the basic goals of the syntax, and then we can bikeshed preferences like curlies versus whitespace versus parens.

WDYT?

kripken:
Sure, I agree that `==` should never be the same as JS - I tried to cover that with "when reasonable to do so," and obviously I don't think any of us would think it is reasonable to match that behavior. Do you think the text is too ambiguous? Happy to re-word if that's the concern, and not a disagreement on the substance.

I'm basically just trying to say in this section that overall platform consistency is a nice thing to have, of course as one factor among many, and just when reasonable to do so. In other words, I agree that it makes sense to start by expressing the basic goals of the syntax, as you said, and I am saying that overall platform consistency should be one of those goals. It seems like an obvious goal, and I think we already imply it in the FAQ, but re-reading the FAQ and the Text Format document, I thought it might be nice to say it directly. We keep getting questions about view-source, and saying something like in this pull might help address people's concerns, I think.

jfbastien:
Gotcha, that does seem to make sense. I'd be less worried if we had a few guiding principles, and said something like "when things don't actually matter, the shed shall be painted Web-colored" (which I think is what you're going for).

kripken:
Yeah, you're saying it better than I did. I rewrote to something more along the lines of that. How does it look now?

jfbastien:
lgtm

kripken:
Merging with lgtm and no objections.

PR #464, opened 2015-11-18
Remove the statement/expression distinction.

titzer:
This catches the design repo up to the spec and implementation repositories where all AST nodes are expressions.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -231,7 +231,6 @@ WebAssembly offers basic structured control flow with the following constructs.
 Since all AST nodes are expressions in WebAssembly, control constructs may yield
 a value and may appear as children of other expressions.
 
- * `nop`: empty operator that does not yield a value
  * `block`: a fixed-length sequence of expressions with a label at the end
  * `loop`: a block with an additional label at the beginning which may be used to form loops
  * `if`: if expression with a *then* expression
@@ -256,7 +255,7 @@ loop from outside it. This restriction ensures all control flow graphs are well-
 
 ### Yielding values from control constructs
 
-The `nop`, `if`, `br`, `br_if`, `case`, and `return` constructs do not yield values.
+The `if`, `br`, `br_if`, `case`, and `return` constructs do not yield values.
 Other control constructs may yield values if their subexpressions yield values:
 
 * `block`: yields either the value of the last expression in the block or the result of an inner `br` that targeted the label of the block
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -1,14 +1,13 @@
 # Abstract Syntax Tree Semantics
 
-WebAssembly code is represented as an Abstract Syntax Tree (AST) that has a
-basic division between statements and expressions. Each function body consists
-of exactly one statement. All expressions and operators are typed, with no
-implicit conversions or overloading rules.
+WebAssembly code is represented as an Abstract Syntax Tree (AST) where each node
+represents an expression. Each function body consists of exactly one expression. 
+All expressions and operators are typed, with no implicit conversions or overloading rules.
 
 Verification of WebAssembly code requires only a single pass with constant-time
 type checking and well-formedness checking.
 
-WebAssembly offers a set of operators that are language-independent but closely
+WebAssembly offers a set of language-independent operators that closely
 match operators in many programming languages and are efficiently implementable
 on all modern computers.
 
@@ -66,15 +65,13 @@ WebAssembly has the following *value types*:
   * `f32`: 32-bit floating point
   * `f64`: 64-bit floating point
 
+Each parameter and local variable has exactly one value type. Function signatures
+consist of a sequence of zero or more parameter types and a sequence zero or more return
+types. (In the MVP, a function can have at most one return type).
+
 Note that the value types `i32` and `i64` are not inherently signed or
 unsigned. The interpretation of these types is determined by individual
 operators.
-
-Parameters and local variables have value types.
-
-Also note that there is no need for a `void` type; function signatures use
-[sequences of types](AstSemantics.md#calls) to describe their return values, so
-a `void` return type is represented as an empty sequence.
 
 ## Linear Memory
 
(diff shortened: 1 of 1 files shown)


AstSemantics.md
```diff
@@ -66,16 +65,14 @@ WebAssembly has the following *value types*:
   * `f32`: 32-bit floating point
   * `f64`: 64-bit floating point
 
+Each parameter and local variable has exactly one value type. Function signatures
```
lukewagner: Perhaps link "value type" to AstSemantics.md#types.

AstSemantics.md
```diff
@@ -66,16 +65,14 @@ WebAssembly has the following *value types*:
...
+consist of a sequence of zero or more parameter types and a sequence zero or more return
+types. (In the MVP, a function can have at most one return type).
```
lukewagner: To be clear that "parameter types" and "return types" are not separate kinds, perhaps say "parameter value types" and "return value types".  Also missing "of" in "sequence zero".

AstSemantics.md
```diff
@@ -231,23 +228,22 @@ others, etc.
...
- * `block`: a fixed-length sequence of statements with a label at the end
- * `loop`: a fixed-length sequence of statements with a label at the end
-           and a loop header label at the top (note: this does not loop by
-           itself, so one would often combine this with a br_if at the end
-           to form a branch back to the top)
- * `if`: if statement with then body
- * `if_else`: if statement with then and else bodies
- * `br`: branch to a given label in an enclosing construct (see below)
+Since all AST nodes are expressions in WebAssembly, control constructs may yield
+a value and appear as children of other expressions.
```
lukewagner: Perhaps "*may* appear"

titzer:
PTAL

AstSemantics.md
```diff
@@ -231,23 +228,22 @@ others, etc.
...
+
+ * `block`: a fixed-length sequence of expressions with a label at the end
+ * `loop`: a block with an additional label at the beginning which may be used to form loops
```
AndrewScheidecker: I also think it's worth leaving the note in there about loops needing an explicit branch to the top.

AstSemantics.md
```diff
@@ -256,14 +252,25 @@ one can arrange `block`s to put labels wherever one wants to jump to, except
 for one restriction: one can't jump into the middle of a loop from outside
 it. This restriction ensures the well-structured property discussed below.
 
-A `tableswitch` has a zero-based array of targets, a "default" target, an index
-operand, and a list of `case` nodes. It jumps to the target indexed in the
-array, or the default if the index is out of bounds. `tableswitch` targets may
-be either labels or `case` nodes.
+### Yielding values from control constructs
+
+The `if`, `br`, `br_if`, `case`, and `return` constructs do not yield values.
```
lukewagner: IIUC, `br` and `return` can be thought of as yielding *any* value type.  That's rather different than `if` and `br_if` which have a void return type (or, using the above definition, "empty sequence of return value types") since you *can* write `(set_local $i (br $foo))` but not `(set_local $i (if ...))`.

Also, `case` is a rather weird case (pun!): I don't really consider it a first-class AST node, but rather an extension of the `tableswitch` (even if `case` does end up getting an opcode in the binary format, you could think of it as an in-order encoding of the parent `tableswitch`).  Perhaps we could fold `case` into `tableswitch` in the list above and remove it from this list here?

titzer:
Yeah, I was careful here to not get into discussing typechecking and what types expressions have. The reason for that is that one can either do top-down type checking (like the sepc) or bottom-up type checking (like decoders) and to describe the former requires type variables and to describe the latter requires a "bottom" or "dead end" type and unification.

lukewagner:
Ah, ok.  I guess we don't really talk about validation other than the "everything fully typed" handwaving at the top, so what you wrote makes sense.  Still `case` is an oddball, but I guess that's preexisting and could be fixed independently.  So generally lgtm.

AstSemantics.md
rossberg-chromium: How about saying that `if` and `br_if` yield an "empty" value (and define this as short for empty sequence of values), while the others abort their evaluation context, so do not yield anything at all (not even an empty value).

rossberg-chromium:
@titzer, agreed, but to nitpick on your comment: top-down does not require type variables, you just pass down the expected type, which (by design) is always known from the context in this language.

rossberg-chromium:
Btw, it should probably be mentioned that `br` (and `br_if`) take a sequence of parameters (or in the MVP, at most one) to be passed to the target.

Also, perhaps a good opportunity to include `nop`, which also yields the empty value.

titzer:
PTAL. I avoided the language about "zero or more subexpressions" for `br` and `br_if` since I'm not sure there's consensus on having multi-br yet.

rossberg-chromium:
LGTM

lukewagner:
I know we have `nop` in `Kernel`, but if we have separate concrete `Ast` nodes for `if`/`if_then` (and, by extension, separate opcodes in the binary format), what's the motivation for a `nop` in `Ast` (and, by extension, the binary format)?  Alternatively, I see now that `Block` is required to be non-empty in `check.ml`... would it make sense to have empty blocks where an empty block has the `None` type?

titzer:
Yeah, we can probably remove it, since I can see how this can be a binary format encoding issue. In v8-native, `br` and `br_if` require a value argument, so this usually `nop`. Remove?

rossberg-chromium:
I'm fine either way.

titzer:
I've removed `nop`.

lukewagner:
Since nullary `br`/`br_if` should be super-common, it seems like they'd benefit from having separate opcodes for nullary vs. unary (vs. perhaps n-ary in the future).

lukewagner:
Thanks, lgtm.

binji:
This discussion (https://github.com/WebAssembly/spec/pull/173) a week ago seemed to conclude with keeping nop. Why remove it now?

titzer:
Ah, yeah. Forgot about that discussion. I can add it back.

kg:
```nop``` should not be removed. If we remove ```nop``` people will just start using meaningless zero-side-effect combinations of opcodes to encode it (and then once deduplication happens those combinations of opcodes will get folded into a macro... that one could call ```nop```). ```nop``` is a requirement for code transforms and probably for some less-sophisticated compilers.

sunfishcode:
Can we have a binary format design discussion, rather than a "should we have nop" discussion? :-)

kg:
Since we have many new community group participants since the last in-depth binary format discussions, that sounds like a good idea.

My (now very old) rough document is worth reading, as much of it still applies. Maybe I should revise it and split it out and integrate it with the current binary format document.
https://github.com/kg/design/blob/master/LayeredBinaryEncoding.md

jcbeyler:
@kg: that link does not work for me for some reason

alkchr:
@jcbeyler Apparently, this is correct link:  https://github.com/kg/design/blob/58196d8fddc38db2483e32c03138f26a87383b25/LayeredBinaryEncoding.md

jcbeyler:
Thanks :)

PR #467, opened 2015-11-19
Add nop

titzer:


binji:
lgtm

sunfishcode:
This is the kind of thing where it'd be really great if we had a more explicit process around the binary format design, because the utility of a nop is highly dependent on the binary encoding strategy.

lukewagner:
So then what was the use case for nop?

jfbastien:
Rationale.md would be a good place for an explanation of not-super-obvious things such as this?

kg:
```nop```'s utility should be obvious for people who are familiar with the history of similar instructions in native instruction sets, or people who have written compilers. I'm kinda shocked that we keep discussing whether to keep it or not, given that by its definition we don't need to do any work to implement it other than have the ast node lying around. It doesn't do anything.

I'll try to put together a PR that adds a rationale to Rationale.md.

lukewagner:
> nop's utility should be obvious for people who are familiar with the history of similar instructions in 
> native instruction sets, or people who have written compilers.

Let's keep this polite.

PR #468, opened 2015-11-19
Rename `loop` to `do`.

sunfishcode:
@dherman suggested this name, and it seems to make sense to me.

This makes the fact that it doesn't automatically loop less surprising.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -233,7 +233,7 @@ a value and may appear as children of other expressions.
 
  * `nop`: an empty operator that does not yield a value 
  * `block`: a fixed-length sequence of expressions with a label at the end
- * `loop`: a block with an additional label at the beginning which may be used to form loops
+ * `do`: a block with an additional label at the beginning which may be used to form loops
  * `if`: if expression with a *then* expression
  * `if_else`: if expression with *then* and *else* expressions
  * `br`: branch to a given label in an enclosing construct
@@ -246,7 +246,7 @@ a value and may appear as children of other expressions.
 ### Branches and nesting
 
 The `br` and `br_if` constructs express low-level branching.
-Branches that exit a `block`, `loop`, or `tableswitch` may take a subexpression
+Branches that exit a `block`, `do`, or `tableswitch` may take a subexpression
 that yields a value for the exited construct.
 Branches may only reference labels defined by an outer *enclosing construct*.
 This means that, for example, references to a `block`'s label can only occur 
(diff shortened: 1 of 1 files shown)


binji:
Hm, interesting. I kinda like it, though if you were scanning looking for the looping primitive, it might be easier to miss.

kripken:
I like this. Feels more appropriate (loops which might not loop are potentially confusing), and shorter too.

lukewagner:
+1

jfbastien:
do it! (heh)

rossberg-chromium:
Why? This name has no informative value. In particular, it does not suggest how the operator differs from `block`, and like @binji said, makes it non-obvious from code that this is _the_ looping construct.

Renaming `block` to `do` would make more sense to me.

titzer:
I'm not sure that `do` is really any more clear than `loop`; i.e. it does not really scream "this expression can be reevaluated multiple times"

kg:
Agreed on titzer's point, but technically ```block``` can be re-evaluated multiple times at this point if it's inside of a loop block anyway. Do we gain much by having a loop-block-that-doesn't-loop? They're functionally equivalent, and maybe the difficulty we're having naming it indicates that it shouldn't exist.

titzer:
On Fri, Nov 20, 2015 at 7:04 PM, Katelyn Gadd <notifications@github.com>
wrote:

> Agreed on titzer's point, but technically block can be re-evaluated
> multiple times at this point if it's inside of a loop block anyway. Do we
> gain much by having a loop-block-that-doesn't-loop? They're functionally
> equivalent, and maybe the difficulty we're having naming it indicates that
> it shouldn't exist.
>
It's an important distinction for the decoder, since it must prepare for a
loop at the beginning, before the body. That operation is expensive (since
it involves creating a conservative SSA-renaming of all variables at that
point), so it would be too costly if every block was a potential loop.

jcbeyler:
How about potential_loop (too long perhaps); loop_ish (not serious enough?)

What about bringing back then do_while?

At least a do_while has the idea that you enter automatically and will determine if you branch perhaps at the end.

sunfishcode:
From a C perspective, perhaps think of `do` as a `do_while` without the `while` part built into it, because that's what wasm's construct is :-).

From an OCaml perspective, perhaps think of `do` as the `do` part of a `while x do ... done` -- it's the thing which syntactically introduces the body of the loop, as opposed to the condition part of the loop.

sunfishcode:
Closing, since I don't have anything else to say here.

PR #469, opened 2015-11-20
Explicit the order of evaluation for the select operation.

jcbeyler:


AstSemantics.md
```diff
@@ -553,7 +553,10 @@ outside the range which rounds to an integer in range) traps.
   * `select`: a ternary operator with a boolean (i32) condition and two
     additional operands, which must have the same type as each other. `select`
     returns the the first of these two operands if the condition operand is
-    non-zero, or the second otherwise.
+    non-zero, or the second otherwise. The order of evaluation is the left
```
kripken: I think the order is already specified, where it says `All nodes other than control flow constructs need to evaluate their child nodes in the order they appear in the serialized AST.` I feel that specifying it here redundantly might be confusing (people might think, why is this specified here, and why don't other ops say their order of evaluation).

Perhaps remove the first added sentence, and keep the last, replacing "therefore" with "note that"?

jcbeyler: Done in new version

kripken:
Thanks! But wait, re-reading that now, I'm confused - isn't the condition evaluated first, then the two values? (Maybe I have that wrong, and if so, my interpreter is wrong too ;)

AstSemantics.md
```diff
@@ -553,7 +553,9 @@ outside the range which rounds to an integer in range) traps.
   * `select`: a ternary operator with a boolean (i32) condition and two
     additional operands, which must have the same type as each other. `select`
     returns the the first of these two operands if the condition operand is
-    non-zero, or the second otherwise.
+    non-zero, or the second otherwise.  Note that, both of
```
kripken: no need for this comma

jcbeyler:
Good point, I don't know. It just seemed logical to me to do it this way :) But maybe someone has an opinion here?

AndrewScheidecker:
It seems like translation to ASM.JS requires fewer temporaries if the condition is evaluated after the other two subexpressions. e.g. `(select a.. b.. c..)` ->
```
var _a, _b, _c;
_a = a..;
_b = b..;
_c = c..;
_a ? _b : _c;
```
vs
```
var _a, _b;
_a = a..;
_b = b..;
c.. ? _a : _b;
```

It seems like it also makes it easier to generate machine code that feed condition flags directly from the condition into `cmov`.

However, if we change the order it would be nice to change the parameter order so it matches the general left-to-right rule.

jcbeyler:
So do we want the first sentence to say:

`select`: a ternary operator with two operands, which must have the same type as each other, and a a boolean (i32) condition.

rossberg-chromium:
The first section in this doc explicitly specifies left-to-right evaluation order. That sometimes requires additional temporaries, but that's not specific to `select`.

titzer:
I'm OK changing the evaluation order of select by defining the condition to
come last. It doesn't make a different to TurboFan if there are no side
effects; it will generate exactly the same code.

sunfishcode:
@AndrewScheidecker's idea about putting the condition last to better set up condition codes makes sense to me at first glance. I also agree that if we do this, we should change the operand order to that we don't have to make select an exception to the left-to-right rule.

jfbastien:
Agreed with what @sunfishcode [just said](https://github.com/WebAssembly/design/pull/469#issuecomment-159004714).

sunfishcode:
I created https://github.com/WebAssembly/design/pull/489 to propose reordering select's operands.

sunfishcode:
This is superseded by #489.

PR #470, opened 2015-11-20
Add a rationale for nop being retained in the instruction set

kg:


kripken:
I agree with this, but I do have a question about "Finding an appropriate opcode that does nothing but has the appropriate type for the node's location is nontrivial." - is the idea that nop will automatically have the proper type, and a default value, based on the context, so `(i32.add (nop) (nop))` would return `(i32 0)`, assuming 0 is the default value?

kg:
I'm not sure how we would define the behavior of ```nop``` inside of an expression like ```i32.add```. That might be illegal. It's mostly important since many of our control flow structures yield values, so replacing a switch-case or an arm of an ```if-else``` shouldn't break compilation. But then if the value is being used...

Maybe ```nop```'s type should be void. In that case, people could standardize on some particular does-nothing opcode combo and hope they don't regret it later. (The rest of the reasons for nop still stand.)

kripken:
Ok. I'm fine either way, was just curious.

kg:
Discussing this locally with some colleagues based on @kripken 's point, I'm not sure there's a valid way to define ```nop``` for wasm. I still think we need it, but I can't come up with a way to define it that works as long as control flow structures yield values. Throwing one out in favor of the other sucks (aside from the fact that I don't think CF should yield values, obviously other people want it).

Does anyone know of a robust way to solve this? Not having nop implies having to crawl all the way up the tree and rewrite nodes, which is awful. But having nop that produces some sort of default value is... kinda troubling. Maybe it's fine because we have a very simple machine model.

JSStats:
Just use a constant, zero.

kripken:
Yes, as @JSStats says, this could use a default value? That is, `nop` would automatically type for the context, and if the type is not null, it would use 0 (of that type).

The benefits for tooling are exactly as described in this pull. However, an argument could be made that such a `nop` could be an internal tooling feature, and not part of `wasm`. However, a counter-argument to that is that having `nop` in wasm means better interoperability and consistency between different wasm processing toolchains (otherwise, one might have a `nop` which is typed, and another not; and sending code between toolchains won't work, etc.).

Overall I think this could be useful, but I also see the downsides, so as I said before I'm fine either way.

AndrewScheidecker:
I'm having trouble imagining where you'd use a non-void `nop`. But one way to do that would be to add a `void.const` operation, and places that want to substitute a `nop` for a subtree would just use a `const` of the appropriate type.

lukewagner:
lgtm

jcbeyler:
"I agree with this, but I do have a question about "Finding an appropriate opcode that does nothing but has the appropriate type for the node's location is nontrivial." - is the idea that nop will automatically have the proper type, and a default value, based on the context, so (i32.add (nop) (nop)) would return (i32 0), assuming 0 is the default value?"

Isn't that exactly the same problem that we have for unreachable? Or are we assuming that because unreachable throws, we don't have that issue (because I had the issue with my system and had to do some fix-ups for unreachable but I might have just hacked it too badly in there :))

titzer:
On Fri, Nov 20, 2015 at 9:51 PM, Alon Zakai <notifications@github.com>
wrote:

> Yes, as @JSStats <https://github.com/JSStats> says, this could use a
> default value? That is, nop would automatically type for the context, and
> if the type is not null, it would use 0 (of that type).
>

I was actually thinking along similar lines that nop could be the "default
value" for any context. I ran into the problem that that basically dictates
a bi-directional typechecking strategy. That is, to build the compiler IR
for a nop expression (i.e. an Int32Constant 0, a Float32Constant 0.0, etc),
one has to know in what context it is used. Up to now, that hasn't been a
requirement.

JSStats:
It might not be such a burden for upper layer transforms to use the appropriate constant type for the context as they will likely be pattern matching the context anyway. There is no support shown for the claim that 'Finding an appropriate opcode that does nothing but has the appropriate type for the node's location is nontrivial.'. If there are contexts in which this would be difficult then perhaps their design could be modified to fit. Could some examples be offered to support this claim?

kg:
> Isn't that exactly the same problem that we have for unreachable? Or are we assuming that because unreachable throws, we don't have that issue

Unreachable is different, yeah. The fact that it traps means that everything it's nested within doesn't run to completion, so any value it might produce is unobservable. ```nop``` is different because it's something you would theoretically insert anywhere and it wouldn't halt execution.

> It might not be such a burden for upper layer transforms to use the appropriate constant type for the context as they will likely be pattern matching the context anyway.

What do you mean here by 'pattern matching'? There are many transforms one could run on code that don't pattern match, and some don't need detailed knowledge of enclosing context either.

> If there are contexts in which this would be difficult then perhaps their design could be modified to fit. Could some examples be offered to support this claim?

Opcodes have explicit return types but the operand types are defined in the spec. As a result, if you want to infer the correct constant opcode to use in a particular place to replace something you're hoisting out, you'll need to walk up the stack some and then do type inference as you walk back down to identify the type you need. It's feasible, but it's a bunch of extra logic that's somewhat unnecessary - the decoder can do this at decode time when it hits a ```nop``` because it's already doing the walk through the tree and building nodes. (In practice this may not matter, because the design of wasm may obligate every tool interacting with it to do exhaustive typing anyway.)

re: 'an appropriate opcode that does nothing', I have in mind the history of ```nop``` in architectures like x86.

In x86 there are numerous ways to encode a ```nop``` and all of them theoretically do nothing. However, as the architecture evolved, many of the encodings turned out to have undesirable side effects. For example:

A traditional ```xchg ax, ax``` or ```xchg eax, eax``` would potentially create a dependency (where code after the ```nop``` was dependent on the result of it) unless the processor's instruction decoder/etc were specialized to completely ignore all operations in that form. When you think about all the myriad ways to encode 'do nothing' it becomes nontrivial to identify and ignore every form of ```nop``` and you have to count on people to converge on 'standard' ones.

It gets worse: x86-64 defined the interaction between 32-bit code and 64-bit registers in such a way that certain encodings of ```nop``` would blow away the high bits contained by a register while others would not. This meant that some ```nop```s suddenly had observable behavior. In practice (as I understand it) what happened is that a couple particular encodings were 'blessed' as ```nop``` and are exempted from the new behavior.

I will follow up on my past comments though: The more I think about this, the more I don't know if there's a satisfactory way to solve this problem. I'm having a hard time thinking of transform use cases for ```nop``` that actually work in our current model.

kg:
I should mention that @AndrewScheidecker's suggestion for a ```void.const``` may be the right solution here. It serves as a reasonable way to replace void-typed operations and then for operations with a type, you replace them with some acceptable default value, and in contexts where a value is unacceptable you rewrite a subtree.

rossberg-chromium:
The `nop` operator is currently defined in both design and spec to produce "no value" -- just like a function that doesn't return anything. Accordingly, its type is the empty expression type (which you may call `void`, but there are no values of this pseudo type, so naming the operator to `void.const` would sound extremely weird to me).

But this thread actually seems to suggest generalising `nop` for other/all types. That has problems on several levels:
- For evaluation, it would require that every type is given some kind of default or zero value. That is easy for the few current types, but might not make any sense for types we may want to add in the future (e.g. structs).
- For typing, it means that `nop` would effectively be an overloaded constant. I really don't think we want overloading in Wasm!
- If we wanted to avoid overloading, however, then we'd have to introduce a family of `nop` operators prefixed by type. But at that point, you could just as well generate whatever type-specific constant you prefer.

AndrewScheidecker:
> Accordingly, its type is the empty expression type (which you may call void, but there are no values of this pseudo type, so naming the operator to void.const would sound extremely weird to me).

I think it would be worth calling the empty type sequence `void` if we added "type-indexed" void operations; `void.const`

lukewagner:
Chatting about the use case described in this PR (being able to easily backpatch wasm ops from a streaming/dumb compiler) earlier with @sunfishcode, it seems like you'd want more than just nullary `nop`, but also a unary `id` identity function and other arities too.

There is an extension to the [module opcode tables](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#global-structure) that I've been thinking about wherein the tables don't simply assign opcodes to operators, but they can also assign opcodes to operators + fixed immediates.   The idea is that we wouldn't have to pull a Java with `iload_0`...`iload_3`, but rather the table could arbitrarily map a single-byte opcode to `(i32.get_local x)` for any `x`.  `load`/`store` are other key examples.  This feature would allow modules to further specialize their opcode space to their contents.  Yes, the macro/specific layer could also do this, but at the cost of more memory traffic and decoding cycles which I expect to be significant for these demonstrably-super-common ops with immediates).  Getting back to the `nop` issue, though: this feature would *also* help the root use case presented here by allowing a single-byte opcode for `block 0` (block of length 0) to serve as `nop`, and a single-byte opcode for `block 1` to serve as the identity function.

kripken:
Here's a concrete toolchain use case for `nop`: the [`remove-imports` pass in binaryen](https://github.com/WebAssembly/binaryen/blob/master/src/passes/RemoveImports.cpp). We need to remove imports since imports from JS won't validate in the spec interpreter, but we do want to validate all the rest of the code. So the pass erases all `import`s, and replaces `call_import`s with either a `nop`, if the type is null, or a `const` of 0 of the right type.

As you can see in the code it's not hard to do this even if `nop` remains of type null. It's also not that hard to do if we remove `nop`, but then the pass would need to emit some other placeholder (suggestions?). In that case, different toolchains might emit different things, which is a slight annoyance for consistency in the entire ecosystem.

In any case, I see the reasons both ways, I'm personally fine however this gets resolved. Just posting this to show a little concrete use case.

jfbastien:
This is starting to sounds an awful lot like adding a `void` type, as well as `undef` for other types.

JSStats:
If supporting multi return values for expressions then wasm will want an operator to construct multiple values from a sequence of single value expressions. This could be used with 0 to n constants to fill any expression. Is this adequate, or do people also want something that fills the required number of values at compile-time.

Rationale.md
```diff
@@ -370,3 +370,12 @@ etc).
...
 consoles or server side have a similar sandboxing mechanism).
+
+
+## ```nop```
+
+Nop remains in the instruction set for mostly utilitarian reasons. Many compilers and tools that transform/generate code will need to replace instructions with a ```nop```, and the alternative would be for them to alter entire subtrees of nodes to eliminate the hole that this instruction would otherwise fill. Tree transforms of this type are more complex (thus, error prone) and more expensive (thus, inappropriate for JIT and debug builds).
+
+Without a ```nop``` instruction, code generators would use alternative *does-nothing* opcode patterns that take up space in a module and potentially aren't 0-cost at runtime. Finding an appropriate opcode that does nothing but has the appropriate type for the node's location is nontrivial. The existence of many different ways to encode ```nop``` - often mixed in the same module - would reduce the efficiency of compression algorithms.
+
+The history of ```nop``` instructions is long and there are numerous examples of their use in production software.
```
sunfishcode: This sentence is confusing because most of the long history of ```nop``` instructions is purposes that don't apply to wasm: alignment padding, hazard avoidance, vliw packing, branch delay slot stuffing, timing, and modifying code at runtime.

The one remaining obvious use, placeholders to preserve the relative offsets of bytes in surrounding code in a module, is even subject to some uncertainty if byte offsets come to be avoided in general, as is beginning to be explored in [recent observations about streaming use cases](https://github.com/WebAssembly/spec/issues/185).

binji:
I found another use for nop: when implementing tableswitch in sexpr-wasm, the fallthrough case requires an operation to branch to. The v8-native bytecode could be reorganized so this isn't necessary, but nop made it easy.

e.g.
```
(module
  (func
    (tableswitch (i32.const 0)
      (table (case 0) (case 1)) (case 2)
        (case (i32.const 0))
        (case) ;; fallthrough
        (case (i32.const 2)))))
```
is translated to
```
0000013: 08                                         ; OPCODE_TABLESWITCH
0000014: 0000                                       ; num targets
0000016: 0000                                       ; num cases
0000018: 0000                                       ; case index
000001a: 0100                                       ; case index
000001c: 0200                                       ; case index
0000016: 0300                                       ; FIXUP num cases
000001e: 09                                         ; OPCODE_I8_CONST
000001f: 00                                         ; u8 literal
0000020: 09                                         ; OPCODE_I8_CONST
0000021: 00                                         ; u8 literal
0000022: 00                                         ; WASM_OPCODE_NOP for fallthrough
0000023: 09                                         ; OPCODE_I8_CONST
0000024: 02                                         ; u8 literal
0000014: 0300                                       ; FIXUP num targets
0000011: 1200                                       ; FIXUP func body size
```

jbondc:
A ```nop``` could also serve useful as a form of software diversity to prevent against attacks (```nop``` instructions randomly inserted in shipped code) :
http://www.computer.org/csdl/mags/sp/2014/02/msp2014020028.html

Couldn't find a public version of the paper or if there's any patent pending ties but an interesting idea.

jbondc:
Reached out to ```Michael Franz``` about ```nop``` insertions, his reply:
> the idea is not patent pending. In fact, we have upstreamed some of the basic things into LLVM in order to hopefully make this standard practice one day.

jfbastien:
@jbondc please do not discuss patents in this forum.

jbondc:
Why not?
Donating ideas is as valuable as donating patents:
https://github.com/WebAssembly/design/blob/master/LICENSE#L73

JSStats:
@jbondc I guess because all patent claims would be disputed and no one here is going to make any admissions or help qualify the scope of any patents. Just discuss ideas, and use cases, and comply with any requirements of the agreement.

jfbastien:
> Why not?

@jbondc: because we want to focus on engineering, not legal issues. The W3C Community Group model provides a framework for engineers to avoid all of these issues, including patent donations being implicit in a member's participation. Yes it's valuable, but it doesn't need to be said. My further concern is when *non-CG member's patents* are being discussed.

jbondc:
ok :+1:

titzer:
Moving this over to #553

PR #481, opened 2015-11-25
Add a br_unless instruction.

sunfishcode:
This is the same as br_if, but with an inverted condition.

This is useful for low-level tools that wish to invert a branch condition in place, for example to rearrange basic blocks while being otherwise minimally invasive. Having both `br_if` and `br_unless` means that this is possible just by rewriting the opcode. If course, it is possible to invert a condition explicitly without this, although in many cases it requires inserting new AST nodes.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -262,11 +262,12 @@ a value and may appear as children of other expressions.
  * `br`: branch to a given label in an enclosing construct
  * `br_if`: conditionally branch to a given label in an enclosing construct
  * `br_table`: a jump table which jumps to a label in an enclosing construct
+ * `br_unless`: same as `br_if` but branches if the condition is false instead
  * `return`: return zero or more values from this function
 
 ### Branches and nesting
 
-The `br` and `br_if` constructs express low-level branching.
+The `br`, `br_if`, and `br_unless` constructs express low-level branching.
 Branches may only reference labels defined by an outer *enclosing construct*.
 This means that, for example, references to a `block`'s label can only occur 
 within the `block`'s body.
@@ -285,7 +286,7 @@ before any others.
 
 ### Yielding values from control constructs
 
-The `nop`, `if`, `br`, `br_if`, and `return` constructs do not yield values.
+The `nop`, `if`, `br`, `br_if`, `br_unless`, and `return` constructs do not yield values.
 Other control constructs may yield values if their subexpressions yield values:
 
 * `block`: yields either the value of the last expression in the block or the result of an inner `br` that targeted the label of the block


jfbastien:
I'm not sure I like this: compilers can have their own IR where this is easy but I don't think we necessarily want WebAssembly to have it. IMO the format isn't designed to support low-level tools: these have to understand the WebAssembly format already so having an internal representation and a writer isn't that onerous.

Similarly, LLVM has comparisons for eq/ne/... and it has true/false comparisons. I see this change as analogous to adding true/false comparisons: useful for low-level tools, but IMO wrong abstraction level for WebAssembly's opcodes.

WDYT?

kripken:
It seems that for consistency, if we have `br_if` and `br_unless`, we should also have `if` and `unless` (where `unless` is like `if`, but executes the code if the condition is false)? I'm not sure I like that.

Since we do have the negations of the relational ops (both `==` and `!=`, etc.), in which cases is another node actually needed? Just for the case of the `br_if` receiving a direct `i32`, I suppose, instead of a comparison result? If that's very common perhaps it could justify this.

sunfishcode:
We don't have the negations of all the floating-point comparisons (currently). And since we don't have a bool type, the condition is an i32, which means the value will sometimes come from a `get_local`, so the code for reversing the condition will have to know multiple tricks, and still have to insert new instructions in some cases.

It's also worth mentioning that this change closely echos constructs present in many ISAs, which commonly have both "branch on zero" and "branch on non-zero" instructions. They give compilers full flexibility to put blocks in whatever order they like (eg. to optimize for fall-through paths) without having to weigh the benefits against inserting new instructions.

I wouldn't necessarily be opposed to adding an `unless` instruction, which would be to `if` what `br_unless` if to `br_if`.

kripken:
Fair points. But I feel that the motivation here could equally support adding the missing floating-point comparisons, plus adding `i32.not`. That's just one new instruction instead of `br_unless` and `unless`, and `i32.not` could be useful for more things too.

But again, if `br_unless`/`unless` would be very common, then maybe avoiding the extra `not` would justify adding them.

In `asm2wasm`, btw, I could in some cases work around the lack of `i32.not`/`br_unless` by branching to a different label. But I agree that's annoying.

sunfishcode:
@kripken In my prototyping of this feature, `br_unless` is quite common. For example, in a CFG diamond, LLVM CodeGen chooses which block gets to be the fallthrough from the conditional branch by consulting profiling information if available, or static heuristics otherwise. The result is that conditions are quite often reversed. `br_unless` makes it very natural to perform these kinds of low-level block ordering optimizations since we don't have to worry about the tradeoffs of inserting `not`s while doing so.

kripken:
Interesting. Is it easy to see which proportion of those would actually require a not?

JSStats:
@kripken It may well help with the text source code to have separate opcodes here. If a canonicalization of the `br` and test occurs there would be an information loss.

sunfishcode:
I got an in-person lgtm from @jfbastien for this proposal.

jfbastien:
Virtual lgtm as well. @sunfishcode showed me the different code that gets emitted by LLVM with/without this feature and it's much cleaner to have both.

kripken:
I did some measurements for this, on bananabread converted with binaryen. I looked at a random 20 out of 3,225 `br_if`s. 90% could be flipped (< to >=, etc.), while 2 were direct checks of a getlocal, which could only be trivially flipped if we had `br_unless`.

JSStats:
@kripken Any change of opinion here?

kripken:
I agree this has reasonable motivations. My only real concern here is consistency with the rest of the spec. Maybe the simplest solution to keep that consistency is, as discussed above, to add both `br_unless` and `unless` (a reversed `if`), so that we maintain a parallel to `br_if` and `if`. How does that sound, @sunfishcode?

titzer:
IMO an "unless" instruction is not necessary, since one can simply express this with `if(!cond) nop else expr`

JSStats:
@titzer The `if` operation is defined to return no values whereas `if_else` returns a value. I recall a discussion about a `nop` in an `if_else` branch but it might not have been resolved and might not be valid if the other branch returns values so would need `(if (!cond) (nop) (block expr (nop)))`.

Neither `br_unless` or `unless` are necessary. The opening message in this PR mentions this and requests `br_unless` as a convenience to the producer. I see the `unless` pattern in asm2wasm generated code, and adding one opcode for this and using it decreases the v8 encoded binary size by 3 bytes per instance, and the brotli compressed size decreased too with a suitable number of the opcode.

Just wanted to test the consensus on this one again - would like it from a source-code point of view too.

JSStats:
@titzer Sorry, looking at the type issue I see it's not a problem and your proposal would work, at a savings of 2 bytes per instance. I thinks it's just requested as a convenience.

kripken:
@titzer: technically yes, however

1. Your exact argument can also work against `br_unless`. In other words, both `br_unless` and `unless` are not strictly necessary, but both are nice to have.
2. The spec seems odd with `if, br_if, br_unless`, since `if` and `br_if` are clearly parallel - both in name and in function - yet we would be missing a parallel for `br_unless`.
3. The construct you mention depends on `nop` remaining in the spec, which some people want to remove; depending on it here would add a constraint.

titzer:
@kripken 

1. Agree. At this point both br_unless and unless are encoder conveniences, since no one has yet built a simple JIT for which we seem to be designing.
2. Agree. I'm not sure what mode we are in here. Are we in encoding convenience mode or simplification mode? We've got outstanding proposals that go in both directions.
3. `nop` is also used for the value to expression "no value" for the value-producing `br_if` and `br`, and that's not just a binary format issue.

sunfishcode:
@titzer Following up on https://github.com/WebAssembly/design/issues/554#issuecomment-188920788 and offline discussions, would you mind taking another look at this PR?

rossberg-chromium:
With agreement on `not`, do we really need this?

titzer:
I won't fight strongly against `br_unless`; it won't be much work to support it.

lukewagner:
I like `br_unless` since it reflects the underlying machine and avoids adding the need to pattern match in a dumb compiler to fold the negation into the condition code check.  I expect this is fairly common so it might also give a small layer 0 size boost; it'd nice to measure this with the current binary encoding.

kripken:
Same rationale works for `unless`, plus it maintains symmetry and consistency. So let's add both, as suggested above?

titzer:
unless would only be useful in the 1-arm if case, since in the 2-arm if
case, you can simply swap the true and false cases. IMO that makes unless
less attractive.

kripken:
Agreed, that reduces the usefulness a little, but 1-arm cases are over 50% in my measurements. So this would be quite useful, in addition to maintaining consistency in the spec.

JSStats:
I believe my results also support the utility of `unless`. The results below are for the zlib benchmark, compiler with the llvm backend and then the br_if's translated to `if` and `if_else` for some familiar patterns, and finally a translation of `if` to `unless`. Note that `unless` has a higher frequency than `if_else`.

```
block                            819     2.62
br_if                            685     2.19
if                               551     1.77
br                               387     1.24
unless                           141     0.45
if_else                           88     0.28

```

qwertie:
Just a thought. I think I read somewhere that (x86) when the branch target buffer is missing a cache entry for a conditional branch, it assumes the branch will not be taken and moreover the btb is left unmodified if the prediction is correct. I've taken advantage of this when writing code like

~~~
if cond:
   f()
   return
else:
  g()
  return
~~~

What I do is, I put the most likely branch first. But if my compiler swaps "if" and "else" branches, this optimization is defeated. Therefore if we only pick one, I'd rather have a i32.eqz than an "unless" instruction, unless there is a matching "unless_else" with two branches.

pizlonator:
The compiler is extremely likely to flip the order of those basic blocks. The compilers that I'm familiar with will forget and then rebuild the order of blocks based on some tried-and-true heuristics. 

-Filip

JSStats:
Looking at the comments there does appear to be consensus to add br_unless, or no significant objections except requests to also add an `unless` operator to balance it but perhaps it is fair to split that out into a separate PR see https://github.com/WebAssembly/design/issues/565 and allow this to land.

pizlonator:
I'm not opposed to it. 

-Filip

kripken:
I strongly disagree, we should not split out `br_unless`.

`if` is a direct parallel to `br_if`, and all arguments in favor of adding an `unless` version to one also apply to the other. Literally. Therefore we should add `unless` to neither or to both, in order to keep the spec symmetrical, consistent, and unsurprising.

Furthermore, the parallel is not just conceptual, but also practical - binaryen for example has optimizations that can convert between `if`s and `br_if`s, so if we have an `unless` version of just one then we have an odd limitation there.

I am in favor of adding both, since I think a reasonable argument has been made in this pull.

sunfishcode:
I don't have a strong opinion about `unless`, so unless someone has strong opinions against it, I'll add it to this PR, once `if` and `if_else` are merged (https://github.com/WebAssembly/spec/pull/257, and we'll need a corresponding design PR too).

JSStats:
@sunfishcode If there is consensus to add both `br_unless` and `unless` then can we please do it now rather than waiting on future PR's - take the opportunity.

rossberg-chromium:
To be honest, I'm not fond of `br_unless`, and I'm even less fond of `unless`. There are many other special branch instructions that are directly supported by CPUs and which we could introduce along similar lines, but I don't think creating a zoo here is an improvement. IMO `br_unless` is taking us on that slippery slope. The motivation in the original comment also seems rather hypothetical.

Can we perhaps wait until after we gained some experience with the MVP before we broaden the operator set in this direction? I don't think there is pressure to do it now.

JSStats:
@rossberg-chromium They are just convenience operators, and not necessary, but there is evidence that they are both common patterns. What other 'special branch instructions that are directly supported by CPUs' do you suggest are in the same class of being so frequent?

It seems only a small step and not the start of adding endless convenience control flow operators. There might be some other patterns to consider such as `and` (sequential `if_else`) but it's not endless.

There is a chunk of code in the llvm backend to do this inversion as cleanly as possible, and I see a need for this when converting between `br_if` and `if` style, so it does not seem 'hypothetical'. They are common patterns.

There seems to otherwise be consensus. What results from the MVP 'experience' do you need to make a decision? Can we measure this for you now?

sunfishcode:
This idea did not achieve consensus.

PR #487, opened 2015-12-01
Clarify type inference for control structures.

Teemperor:
WIP.

This PR is supposed to clarify how the type inference system for control structures works.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -261,18 +261,11 @@ Other control constructs may yield values if their subexpressions yield values:
 
 * `block`: yields either the value of the last expression in the block or the result of an inner `br` or `br_if` that targets the label of the block
 * `loop`: yields either the value of the last expression in the loop or the result of an inner `br` or `br_if` that targets the end label of the loop
-* `if_else`: yields either the value of the true expression or the false expression. If both expressions yield values of different types, the `if_else` will not yield a value
+* `if_else`: yields either the value of the true expression or the false expression
 * `tableswitch`: yields either the value of the last case or the result of an inner `br` or `br_if` that targets the tableswitch
 
-Note that the yield value types of all inner `br` or `br_if` that target a control construct and the types of all yielding subexpressions of said control construct have to match. Subexpressions that consist of only a `return`, `br`, `br_if`, `return` or `unreachable` are exempt from this type check.
-
-This means for example that a block whose last expression is an `(i32.const 1)` can only be the target of a `br` and `br_if`
-with results of the type `i32`. A block whose last expression is an `(unreachable)` can be the target of a `br` and `br_if`
-with results from any type. 
-
-The expression `(if (i32.const 1) (i32.const 2) (f32.const 3.0))` will *not* yield a value while
-the expression `(if (i32.const 1) (unreachable) (f32.const 3.0))` will yield a value. The expression 
-`(if (i32.const 1) (block (unreachable)) (f32.const 3.0))` will not yield a value because the true expression is considered to be a `block` statement.
+Every yielding control structure with an expected type has to ensure that all possible yield values match
+its expected type.
 
 ### Tableswitch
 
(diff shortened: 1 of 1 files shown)


AndrewScheidecker:
I think it's simpler to explain it like ml-proto [implements it](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/check.ml#L114): as a top-down propagation of expected type rather than bottom-up type inference. That has a very simple rule for the `if_else` operands for example: the same type is expected of the value operands as of the `if_else` itself.

rossberg-chromium:
Agreeing with @AndrewScheidecker. In particular, I think the formulation that a conditional with inconsistent types "does not yield a value" is misleading. Instead, the expected type of the conditional controls the expected type of its branches. If that is the empty type, then and only then it happens to be fine for the branches to yield inconsistent results, because they are thrown away (locally) anyway. That's rather an aside, though, not something that seems worth putting much focus on.

jfbastien:
I think the text is worth clarifying. Should that be a different PR?

Teemperor:
@AndrewScheidecker that's indeed easier, thanks!

@jfbastien You mean a fresh PR that is not referencing the current commit?

JSStats:
I actually see some merit in the patch as-is for `if_else`. Consider the case that the condition is selecting between paths that operate on different types and the branches are naturally returning different types then the only valid interpretation in wasm is that the result type of the `if_else` is the empty type or a constant 0 etc. How else could it be expressed, perhaps `(if_else <test> (block <double-exp> 0) <int-exp>)`, where the `(block ... 0)` had to be added to keep the types consistent?

rossberg-chromium:
@JSStats, not sure I follow. When the expected type is empty, this is fine. When the expected type is not empty, it cannot possibly be fine.

JSStats:
@rossberg-chromium Thank you for the explanation, that sounds fine. The comment by @AndrewScheidecker appears to conflict with this - if the expected type was empty then if 'the same type is expected of the value operands' the this would not be possible, but the proposed patch seems to fix this?

kg:
One detail on my mind that doesn't seem to be addressed even if we merge this PR: What happens if control flow is used outside of a value context, but both of the arms yield values? The below addresses a subset of this issue, where only one of the arms yields a value:

> If both expressions yield values of different types, the if_else will not yield a value

But a somewhat common real-world case will be a top-level conditional (i.e. ```if-else```) that calls functions with return values. In C and many other languages a top-level invocation of a function with a return value is quite reasonable, and discarding the unused result can be zero-cost. Presently we have no way to express this "discard" operation in the AST and without that, I don't think control flow involving discarded results will type-check successfully.

Making ```if-else``` with mismatched types on each arm silently discard values also seems gross in this context - we're hacking around this problem instead of fixing it at the root.

rossberg-chromium:
@JSStats, sorry, I'm really confused what you're asking. Can you clarify?

@kg, a value is discarded whenever the expected type is empty. You are right that mismatching conditionals should not "silently discard" per se -- and they don't, which was my very complaint about the text in this PR.

JSStats:
@rossberg-chromium Just that your comment above seems worthy of adding to the spec too: 'the expected type of the conditional controls the expected type of its branches. If that is the empty type, then and only then it happens to be fine for the branches to yield inconsistent results, because they are thrown away (locally) anyway.'

Teemperor:
I'm using the idea of the expected type now. What's a good explanation what the expected type is (the type that the parent operator demands from a operand or the return type of a function if the control structure is the top-level construct in a function?).

Also, "all possible yield values" means that we require an certain amount of analysis of the CFG.
It is for example not really clear if we expect this example to be a type error or not:

    (i32.add 
         (i32.const 1) 
         (block $blocklabel
             (unreachable)
             (i32.const 2)
         )
    )

rossberg-chromium:
Yes, that's a reasonable explanation.

You should probably take a look at the actual spec. The rules are really straightforward, and don't leave anything open. Your example typechecks just fine, there is no attempt to detect or reject dead code.

Teemperor:
@rossberg-chromium Do you mean ml-proto with 'actual spec'?

rossberg-chromium:
Yes. :) That codifies the [type checking rules](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/check.ml#L114) (for the kernel language).

sunfishcode:
Closing; the text of (what is now) Semantics.md has been significantly rewritten since this PR was opened. If anyone finds the current text unclear, please file a new issue or PR.

Issue #480, opened 2015-11-25
Factor block_cascade out of tableswitch


AndrewScheidecker:
This idea isn't fully baked, but I thought I'd kick it out here and see what people think.

There are two components to `tableswitch`: a branch to a target taken from a table+index, and a set of cases that introduce special targets that may be referenced by the table. As I understand it, the purpose for the second component is to avoid highly nested cascades of AST nodes that would otherwise be necessary to represent the a C-style `switch` with many cases.

I think the second component could be split into a separate AST node: `block_cascade`. It would look kind of like this:
```
(block_cascade
    (tableswitch Index... (table #case0 #case1 #case2) #default)
    #case0
    A...
    #case1
    B...
    #case2
    C...
    #default
    D...
)
````
Which could be desugared to this:
```
(block #default
    (block #case2
        (block #case1
            (block #case0
                (tableswitch Index... (table #case0 #case1 #case2) #default)
            )
            A...
        )
        B...
    )
    C...
)
D...
```

One benefit to this is that it would be much easier to produce `block_cascade` by pattern matching than a `tableswitch` with cases: it can ignore the contents of the A, B, C, D syntactic variables above. However, some of that benefit could be captured by making the existing tableswitch cases desugar into the block cascade so it doesn't need to worry about whether those blocks-to-be-case-ified are used as generate branch targets.

rossberg-chromium:
Regular labels are only in scope _inside_ the expression they name. Case labels are very different beasts, they are only known to the respective tableswitch statement _enclosing_ them. Your suggestion would require reconciling these two notions somehow, and allow referencing expression labels from the outside. It is not clear what the rules for this should be without losing well-structuredness of control flow. Even if we found such rules, I'm pretty sure they'd be more complicated and costly to check than the current design, which guarantees well-structuredness by construction.

AndrewScheidecker:
That difference between a tableswitch's cases and a cascade of blocks is what makes it harder to transform the latter to the former.

The goal here is that the block_cascade has the same semantics as a cascade of nested blocks in terms of label visibility, so it maintains the structured control flow property. The difference is that it doesn't reflect that nesting in the syntax structure.

sunfishcode:
@rossberg-chromium The rule for this to preserve well-structuredness of control flow is: these labels can only be referenced from within the block_cascade above their definition. This can be proven by examining the effective rules that would be imposed by the equivalent stack of nested blocks.

[...]

PR #490, opened 2015-12-03
Factor a `multiblock` node out of `tableswitch`.

sunfishcode:
As discussed in https://github.com/WebAssembly/design/issues/480, this introduces `multiblock`, and simplifies `tableswitch` accordingly.

This also updates the wording around control flow operators in several areas.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -232,54 +232,66 @@ Since all AST nodes are expressions in WebAssembly, control constructs may yield
 a value and may appear as children of other expressions.
 
  * `nop`: an empty operator that does not yield a value 
- * `block`: a fixed-length sequence of expressions with a label at the end
+ * `block`: a sequence of expressions with a label at the end
  * `loop`: a block with an additional label at the beginning which may be used to form loops
+ * `multiblock`: a mixed sequence of expressions and `label` nodes, with a label at the end
+ * `label`: a node which defines a label and must be an immediate child of `multiblock`
  * `if`: if expression with a *then* expression
  * `if_else`: if expression with *then* and *else* expressions
- * `br`: branch to a given label in an enclosing construct
- * `br_if`: conditionally branch to a given label in an enclosing construct
- * `tableswitch`: a jump table which may jump either to an immediate `case`
-                  child or to a label in an enclosing construct
- * `case`: a case which must be an immediate child of `tableswitch`
+ * `br`: branch to a given in-scope label
+ * `br_if`: conditionally branch to a given in-scope label
+ * `tableswitch`: branch using a "jump table" of in-scope labels
  * `return`: return zero or more values from this function
 
 ### Branches and nesting
 
-The `br` and `br_if` constructs express low-level branching.
-Branches that exit a `block`, `loop`, or `tableswitch` may take a subexpression
-that yields a value for the exited construct.
-Branches may only reference labels defined by an outer *enclosing construct*.
-This means that, for example, references to a `block`'s label can only occur 
-within the `block`'s body.
+The `br`, `br_if`, and `tableswitch` constructs express low-level branching. The
+`block`, `loop`, and `multiblock` constructs provide labels that may be branched
+to.
 
-In practice, outer `block`s can be used to place labels for any given branching
-pattern, except for one restriction: one can't branch into the middle of a loop
-from outside it. This restriction ensures all control flow graphs are well-structured
+Branches may only reference labels that are *in scope*. The labels in `block`
+and `loop` nodes are in scope only within their subtrees. The labels in
+`multiblock` nodes are only in scope only within their subtrees, and only in
+subtrees earlier in the `multiblock` than their definition. This restriction
+ensures all control flow graphs are
+[reducible](http://dl.acm.org/citation.cfm?id=804919).
+
+In terms of control flow graphs, the reducible requirement imposes
+only one effective restriction: loops can only be entered from the
+top, and not from a branch into the middle. This restriction ensures
+all control flow graphs are well-structured
 in the exact sense as in high-level languages like Java, JavaScript, Rust and Go. To
 further see the parallel, note that a `br` to a `block`'s label is functionally
 equivalent to a labeled `break` in high-level languages in that a `br` simply
 breaks out of a `block`.
 
+Branches may take a subexpression as an optional first operand (before any
+other operands). When a `block`, `multiblock`, or `loop` is exited by a branch,
+or when a `label` is branched to by a branch, this operand of the branch is
+yielded as the result. When `br_if`'s condition is false, this operand is
+yielded as `br_if`'s result.
+
 ### Yielding values from control constructs
 
-The `nop`, `if`, `br`, `br_if`, `case`, and `return` constructs do not yield values.
+The `nop` and `if` constructs do not yield values.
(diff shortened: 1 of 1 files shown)


AstSemantics.md
```diff
@@ -232,48 +232,58 @@ Since all AST nodes are expressions in WebAssembly, control constructs may yield
...
+ * `multiblock`: a mixed sequence of expressions and `label` nodes, with a label at the end
+ * `label`: a node which defines a label and must be an immediate child of `multiblock`
  * `if`: if expression with a *then* expression
  * `if_else`: if expression with *then* and *else* expressions
- * `br`: branch to a given label in an enclosing construct
- * `br_if`: conditionally branch to a given label in an enclosing construct
- * `tableswitch`: a jump table which may jump either to an immediate `case`
-                  child or to a label in an enclosing construct
- * `case`: a case which must be an immediate child of `tableswitch`
+ * `br`: branch to a given label
```
kripken: removing "in an enclosing construct" makes this less clear imo

kripken: perhaps that comment wasn't clear enough as well ;) I mean, why not write "in a visible construct" using the new terminology?

AstSemantics.md
```diff
@@ -232,48 +232,58 @@ Since all AST nodes are expressions in WebAssembly, control constructs may yield
...
-Branches may only reference labels defined by an outer *enclosing construct*.
-This means that, for example, references to a `block`'s label can only occur 
-within the `block`'s body. In practice, outer `block`s can be used to place labels for any
-given branching pattern, except for one restriction: one can't branch into the middle of a
-loop from outside it. This restriction ensures all control flow graphs are well-structured.
+The `br`, `br_if`, and `tableswitch` constructs express low-level branching. The
+`block`, `loop`, and `multiblock` constructs provide labels that may be branched
+to.
+
+Branches may only reference *visible* labels. The labels in `block` and `loop`
```
kripken: "visible" seems confusing to me. how about "accessible", "reachable", or some terminology that mentions "structure" somehow? the latter seems best, but I don't have a concrete suggestion for how to use it yet

AstSemantics.md
```diff
@@ -232,48 +232,58 @@ Since all AST nodes are expressions in WebAssembly, control constructs may yield
...
+nodes are visible only from within their subtrees. The labels in `multiblock`
+nodes are only visible from within their subtrees, and only from above their
+definition.
+
+In practice, `block`s, `loop`s, and `multiblocks` can be used to place labels
```
kripken: this change weakens your point substantially, doesn't it? you don't need `multiblocks` to achieve that property. having this text suggests that you do.

AstSemantics.md
sunfishcode: I'm not a fan of "reachable" due to possible ambiguity with the graph-theory meaning of the word. "Visible" evokes the idea of scoping, and these are essentially scoping restrictions. I did add "visible" to the descriptions above though; does that help?

kripken: I don't see how "visible" evokes scoping, maybe there's a connection I don't see?

How about just saying "in scope"?

AstSemantics.md
```diff
@@ -232,48 +232,60 @@ Since all AST nodes are expressions in WebAssembly, control constructs may yield
...
-within the `block`'s body. In practice, outer `block`s can be used to place labels for any
-given branching pattern, except for one restriction: one can't branch into the middle of a
-loop from outside it. This restriction ensures all control flow graphs are well-structured.
+The `br`, `br_if`, and `tableswitch` constructs express low-level branching. The
+`block`, `loop`, and `multiblock` constructs provide labels that may be branched
+to.
+
+Branches may only reference *visible* labels. The labels in `block` and `loop`
+nodes are visible only from within their subtrees. The labels in `multiblock`
+nodes are only visible from within their subtrees, and only from above their
```
kripken: "above" has the same issues as "left"/"right". How about "before"?

kripken:
Would we actually still need `block`, if we had `multiblock`?

AstSemantics.md
sunfishcode: I guess I meant visible in the sense of "visibility" in the intro paragraph [here](https://en.wikipedia.org/wiki/Scope_%28computer_science%29).

Would "in scope" be better?

kripken: I think so.

AstSemantics.md
sunfishcode: Makes sense. Updated.

sunfishcode:
I imagine `block` and `multiblock` might be encoded differently, but otherwise no, `multiblock` could subsume `block`.

AstSemantics.md
sunfishcode: Ok, I updated it to say "in scope".

kripken:
Some more thoughts: overall I like this, but also feel like we should gather some data first. In particular, this construct can subsume block, if and if_else, so just adding it leads to some redundancy - which is perhaps justified by efficiency, but we need data to know.

lukewagner:
Since multiblock would not be the redundant op, could we consider the issue of whether we still need the other ops as a separate issue?

JSStats:
On 12/05/2015 10:06 AM, Alon Zakai wrote:
> Some more thoughts: overall I like this, but also feel like we should
> gather some data first. In particular, this construct can subsume block,
> if and if_else, so just adding it leads to some redundancy - which is
> perhaps justified by efficiency, but we need data to know.

I think it would be unfortunate to remove `if` and `if_else` and require them to be expressed only as `multiblock` and `br_if` and `br_unless` from the point of view of giving some consideration to the source code being readable and writeable. Also keep in mind that `br_if` always evaluates the result expression, and we still need `if` when wanting to conditionally evaluate the result.

kripken:
@lukewagner I personally would prefer to consider the issue as a whole, as opposed to adding all the possibilities and weeding out the unnecessary ones later. I say "all" because there are other control flow proposals as well, like `br_unless` and `unless`, perhaps other options for switch, etc. ,which are also relevant here.

And given that we should be able to get relevant data quite soon, focusing on that seems to make more sense to me.

But, I don't feel strongly about this.

@JSStats: I believe you can still conditionally evaluate a result using `multiblock`, you just `br` forward to jump over the computation, so it becomes conditional.

JSStats:
@kripken `multiblock` meets use cases expected to be common in some code, but not in other code. Hand written code might be much more structured, and make sparing use of `multiblock`, but it will be interesting to see where it can be used with clarity. Thus I support @lukewagner's call to move the discussion about other operators to new issues.

It's probably still computational complete, but how readable and writeable??

`x = 2 * a + {
                     br_if(, fn1(a) < 10, $next);
                     br(fn2(a, b), $done);
                     $next;
                     fn3(a, b);
                     $done;
                   }
                 + k;`
versus
`x = 2 * x + if_else((fn1(a) < 10), fn3(a, b), fn2(a, b)) + k;`

I would be surprised if this did not show up in encoding efficiency too.

sunfishcode:
A lot of the discussion here has focused on the `multiblock` construct. As a possible easier way forward, I've submitted #512 which is a subset of this PR that just removes `case` without introducing `multiblock`.

JSStats:
Perhaps introducing the `multiblock` before removing `case` would be an easier path.

sunfishcode:
There is not consensus to add `multiblock` at this time.

Issue #477, opened 2015-11-23
Default NaN bit pattern requirement not compatible with x86


sunfishcode:
wasm currently requires that when the platform has to produce a new NaN value, the sign bit is 0. @AndrewScheidecker pointed out that x86 sets the sign bit to 1. Other platforms, such as ARM, do use 0 though. It will likely be necessary to make wasm's sign bit be nondeterministic. I believe this is still compatible with NaN boxing techniques.

jcbeyler:
Are there tests done now in the spec implementation to ensure this by any chance?

AndrewScheidecker:
The tests in nan_propagation.wast cover this.

jcbeyler:
Interesting because I did nothing special in this regard and yet I pass that test file. I'll have to go check if it is valid then and do I take this into account :)

jcbeyler:
This can be closed now since it merged no?

PR #491, opened 2015-12-03
Make the sign bit of a new NaN value nondeterministic.

sunfishcode:
x86 produces NaN values with the sign bit set to 1, and other architectures
produce NaN values with the sign bit set to 0; wasm can model this by
just making the sign bit nondeterminisitc. This should still permit the
NaN-boxing technique that all this NaN bit-pattern language is intended to
support.

This addresses #477.

### DIFF
--- a/Rationale.md
+++ b/Rationale.md
@@ -292,7 +292,8 @@ However, since the publication of IEEE 754-2008, MIPS has added a configuration
 mode (NAN2008) which enables support for the new rules.
 
 In particular, the sign bit of generated NaNs is nondeterministic since x86
-NaNs with it set to 1 while other architectures generate NaNs with it set to 0.
+generates NaNs with it set to 1 while other architectures generate NaNs with it
+set to 0.
 
 
 ## Integer operations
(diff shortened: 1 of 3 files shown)


jfbastien:
lgtm, could you also add the x86 + NaN-boxing explanation to Rationale.md? It would be great to explain why NaNs work the way we specify them.

sunfishcode:
NaN boxing is already mentioned; I've now added a mention of x86's NaN sign bit behavior.

jfbastien:
ty, still lgtm

Rationale.md
```diff
@@ -291,6 +291,9 @@ notably MIPS, historically behaved differently than other architectures.
 However, since the publication of IEEE 754-2008, MIPS has added a configuration
 mode (NAN2008) which enables support for the new rules.
 
+In particular, the sign bit of generated NaNs is nondeterministic since x86
+NaNs with it set to 1 while other architectures generate NaNs with it set to 0.
```
titzer: "generates" is missing between "x86" and "NaNs".

titzer:
LGTM other than small nit

sunfishcode:
Merging with LGTMs above and nit fixed. Thanks!

PR #498, opened 2015-12-09
Clarify what a Wasm module is to be seen as

jcbeyler:
Add a clarification that a module is Wasm is to be seen as the equivalent
 of a whole C/C++ program for example. It is not just a single class for
 example.

### DIFF
--- a/CAndC++.md
+++ b/CAndC++.md
@@ -1,7 +1,8 @@
 # Guide for C/C++ developers
 
 WebAssembly is being designed to support C and C++ code well, right from
-the start in [the MVP](MVP.md). The following explains the outlook for
+the start in [the MVP](MVP.md). The intent is to have the full C/C++ code
+ported into a single [module](Modules.md). The following explains the outlook for
 C and C++ developers.
 
 ## Porting C and C++ code to WebAssembly
--- a/Modules.md
+++ b/Modules.md
@@ -32,6 +32,13 @@ in a Web environment (more details [below](Modules.md#integration-with-es6-modul
 WebAssembly modules are defined independently of JavaScript and do not require
 the host environment to include a JavaScript VM.
 
+From a C/C++ perspective, a WebAssembly module is not a single ```.o``` file
+or a single class implementation but rather the
+equivalent of a C/C++ program with all its ```.o``` files linked together. At
+runtime, a module can be loaded by a runtime system, which would load at runtime
+the module. If the module has a start definition, that definition gets called
+before any other function from the module.
+
 ## Imports and Exports
 
 A module defines a set of functions in its


jcbeyler:
while talking with JF, I noticed this was not clear to me entirely and therefore we should perhaps explicit it somewhere. Not sure I did it in the right places though :)

titzer:
lgtm

Modules.md
```diff
@@ -1,7 +1,10 @@
 # Modules
 
 The distributable, loadable, and executable unit of code in WebAssembly
-is called a **module**. At runtime, a module can be loaded by a runtime
+is called a **module**. In a C/C++ world, a WebAssembly module is not a
+single .o file or a single class implementation but rather the
```
jfbastien: Code-quote `.o`.

jfbastien:
lgtm besides nit, but I'd like an explicit lgtm from @lukewagner as well.

Modules.md
jcbeyler: done :)

jcbeyler:
Let me know if this works and seems like we want to wait for @lukewagner 's lgtm

Modules.md
```diff
@@ -1,7 +1,10 @@
 # Modules
 
 The distributable, loadable, and executable unit of code in WebAssembly
-is called a **module**. At runtime, a module can be loaded by a runtime
+is called a **module**. In a C/C++ world, a WebAssembly module is not a
+single ```.o``` file or a single class implementation but rather the
+equivalent of a C/C++ program with all its ```.o``` files linked together. At
```
lukewagner: This is true in the MVP, but won't be true once we have dynamic linking (a module would be equivalent to either an executable or dynamic-link-library).  The explanation of what an instance contains below mentions this.  At the end of this section, we talk about what a wasm module is from a Web perspective; perhaps you could move this sentence to the end to explain what a wasm module looks like from a C/C++ perspective (and add a bit about dynamic linking).

lukewagner:
(Sorry for the delay.)

jcbeyler:
Put a bit there but did not know how far you wanted me to go into details (and not sure what those details were (have we talked about this before?))

CAndC++.md
```diff
@@ -1,7 +1,8 @@
 # Guide for C/C++ developers
 
 WebAssembly is being designed to support C and C++ code well, right from
-the start in [the MVP](MVP.md). The following explains the outlook for
+the start in [the MVP](MVP.md). The intent is to have the full C/C++ code
```
sunfishcode: It's not clear to me what "the full C/C++ code" means.

Modules.md
```diff
@@ -32,6 +32,13 @@ in a Web environment (more details [below](Modules.md#integration-with-es6-modul
 WebAssembly modules are defined independently of JavaScript and do not require
 the host environment to include a JavaScript VM.
 
+From a C/C++ perspective, a WebAssembly module is not a single ```.o``` file
+or a single class implementation but rather the
+equivalent of a C/C++ program with all its ```.o``` files linked together. At
```
sunfishcode: Instead of "program", how about "executable or dynamic library"?

Modules.md
```diff
@@ -32,6 +32,13 @@ in a Web environment (more details [below](Modules.md#integration-with-es6-modul
...
+runtime, a module can be loaded by a runtime system, which would load at runtime
+the module. If the module has a start definition, that definition gets called
+before any other function from the module.
```
sunfishcode: It's not clear that we actually guarantee this, in the case of cyclic module dependencies. But that said, I wonder if saying "executable" above is enough clarification, and we don't need to mention the module start function separately here. It's covered in Modules.md, and is a sufficiently low-level detail that this high-level C/C++-oriented doc may not need.

sunfishcode:
@jcbeyler Friendly ping :). This PR adds some useful content and I think only needs a few minor clarifications.

sunfishcode:
Second friendly ping :).

sunfishcode:
Closing; anyone wishing to propose clarifications to how a wasm module is to be seen is welcome to submit a new issue or PR.

PR #500, opened 2015-12-09
Added a link to the FAQ to state why LLVM is chosen now

jcbeyler:
Thought this would be better to really push on the fact that LLVM is chosen because that is where our speciality/knowledge is but we would love to see more compilers embrace WebAssembly

### DIFF
--- a/HighLevelGoals.md
+++ b/HighLevelGoals.md
@@ -33,6 +33,6 @@
 4. Design to support [non-browser embeddings](NonWeb.md) as well.
 5. Make a great platform:
     * build a new LLVM backend for WebAssembly and an accompanying
-      clang port;
+      clang port ([why LLVM first?](FAQ.md#which-compilers-can-i-use-to-build-webassembly-programs));
     * promote other compilers and tools targeting WebAssembly; and
     * enable other useful [tooling](Tooling.md).


HighLevelGoals.md
```diff
@@ -34,5 +34,7 @@
 5. Make a great platform:
     * build a new LLVM backend for WebAssembly and an accompanying
       clang port;
+      ** see the [FAQ] (FAQ.md) why LLVM is being chosen and note it is not the only compiler
```
jfbastien: I'd rather downplay this, since it's our front-facing page. Instead I suggest just saying:

> build a new LLVM backend for WebAssembly and an accompanying clang port ([why LLVM first?](FAQ.md#which-compilers-can-i-use-to-build-webassembly-programs));

jcbeyler: Done :)

HighLevelGoals.md
```diff
@@ -33,6 +33,6 @@
 4. Design to support [non-browser embeddings](NonWeb.md) as well.
 5. Make a great platform:
     * build a new LLVM backend for WebAssembly and an accompanying
-      clang port;
+      clang port ([why LLVM first?](FAQ.md#which-compilers-can-i-use-to-build-webassembly-programs))
```
jfbastien: You dropped the semicolon :-)

jfbastien:
lgtm with semicolon;

HighLevelGoals.md
jcbeyler: put it back

jcbeyler:
put back the semi-colon, so it should be good now (hopefully)

PR #512, opened 2016-01-04
Remove cases from `tableswitch`.

sunfishcode:
Some of the feedback I got concerning https://github.com/WebAssembly/design/pull/490 was that it's making multiple changes, and that people might prefer if the changes were split up into multiple PRs.

This PR removes the `case` construct from `tableswitch`. This is desirable because it simplifies the `tableswitch` instruction and eliminates the need for a "target" concept which abstracts over labels and cases.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -238,9 +238,7 @@ a value and may appear as children of other expressions.
  * `if_else`: if expression with *then* and *else* expressions
  * `br`: branch to a given label in an enclosing construct
  * `br_if`: conditionally branch to a given label in an enclosing construct
- * `tableswitch`: a jump table which may jump either to an immediate `case`
-                  child or to a label in an enclosing construct
- * `case`: a case which must be an immediate child of `tableswitch`
+ * `tableswitch`: a jump table which jumps to a label in an enclosing construct
  * `return`: return zero or more values from this function
 
 ### Branches and nesting
@@ -262,24 +260,20 @@ breaks out of a `block`.
 
 ### Yielding values from control constructs
 
-The `nop`, `if`, `br`, `br_if`, `case`, and `return` constructs do not yield values.
+The `nop`, `if`, `br`, `br_if`, and `return` constructs do not yield values.
 Other control constructs may yield values if their subexpressions yield values:
 
 * `block`: yields either the value of the last expression in the block or the result of an inner `br` that targeted the label of the block
 * `loop`: yields either the value of the last expression in the loop or the result of an inner `br` that targeted the end label of the loop
 * `if_else`: yields either the value of the true expression or the false expression
-* `tableswitch`: yields either the value of the last case or the result of an inner `br` that targeted the tableswitch
+* `tableswitch`: yields the result of an inner `br` that targeted the tableswitch
 
 
 ### Tableswitch
 
-A `tableswitch` consists of a zero-based array of targets, a *default* target, an index
-operand, and a list of `case` nodes. Targets may be either labels or `case` nodes.
-A `tableswitch` jumps to the target indexed in the array or the default target if the index is out of bounds. 
-
-A `case` node consists of an expression and may be referenced multiple times
-by the parent `tableswitch`. Unless exited explicitly, control falls through a `case` 
-to the next `case` or the end of the `tableswitch`.
+A `tableswitch` consists of a zero-based array of labels, a *default* label,
+and an index operand. A `tableswitch` jumps to the label indexed in the array
+or the default label if the index is out of bounds.
 
 
 ## Calls


kg:
LGTM

kripken:
This looks nice to me.

How about the name `tablebranch`?

sunfishcode:
I'd be ok with `tablebranch` (or other things) but I'd like to keep this PR itself focused.

jfbastien:
It would be good to get input from @titzer / @rossberg-chromium on this, but I think they're still on vacation.

JSStats:
Cases within a table might still have value from the point of view of the source code being 'familiar' and 'readable', even if they are computationally redundant. There might be value in having multiple ways to specify the same computational flow to aid readability. A switch/case without breaks to blocks outside a table is very common, and I think programmers accept that it help understand the code knowing that there is a dispatch to multiple paths and that they all merge back to the same control flow point at the end of the table scope.

We still do not have a text format for us to be able to assess how readable code using a multi-block might be - how quickly can the reader identify the control flow. Perhaps with some source code conventions this patten can be easily read too, but this is not immediately obvious.

Some might propose that the view-source process recognize some common multi-block/tableswitch patterns and convert them to a familiar switch for presentation, but then we end up with a loss of information in the source code to the binary encoding (two source code forms that have the same binary encoding).

I think there might yet be value is keeping some baggage if it helps with view-source, and it's not clear that we are in a position to make these judgements just yet. I still think it unfortunate that discussion on this matter is forced to be spread out over issues and that we can not integrate it into the documents to improve productivity - my own understand has advanced working on a s-exp text format (I'll leave the web-familiar format to @kripken) yet there is no where to document and share this.

lukewagner:
lgtm

JSStats:
@lukewagner Could you please articular your reasons for supporting this PR? Thank you in advance for your patience.

lukewagner:
'case' was added as a hack to avoid the big stack-o-`block`s which I think is really only a problem if a decoder is recursive and thus blows its stack recursing into the big stack-o-`block`s.  This problem can instead be solved by saying that decoders should not use (non-PTC) recursion (big-stack-o-`block`s won't be the only source of over-recursion, so this sounds reasonable) and/or introducing `multiblock`.  Thus, it makes sense to remove `case` and then, independently, consider whether to add `multiblock`.  Since `multiblock`'s potential benefits are purely practical, it's also something we probably want to wait on to measure.  As for the view-source concerns, we've already discussed this theme a number of times in other issues but, as a brief version, wasm is a compiler target not a programming language; the text format should reflect the AST, but not change its design.

jfbastien:
Would it be worth documenting the final form of `tableswitch` on Rationale.md?

JSStats:
@lukewagner I strongly object to the position that 'wasm is a compiler target not a programming language; the text format should reflect the AST, but not change its design.' as you should well know. The use case of being a 'compiler target' is well accommodated, and would continue to be well accommodated if `case` were retained to support readability - I am not even making a strong case to retain it just suggest that we are not in a good position to make this decision.

I believe your position of excluding the 'readability' use case is not tenable for a web facing format so I have to ask if we are finish discussing the matter now and should it be taken back to the group for a decision? Given the strong refusal to accommodate this use case from yourself and I believe also JFB it would have to be a vote on the Chairs, so are you ready to stand by your position in a vote of the Chairs, are your ready to be the champion of excluding this use case? I appeal to you to please reconsider, and be a little more accommodating.

jfbastien:
AstSemantics.md documents the binary format of the AST, not the text format which is in TextFormat.md and not well documented for now. That's not very well explained, let me update AstSemantics.md to clarify.

We haven't settled on a textual format yet, and I agree that making it easy to use *without hindering the binary format* is a desirable goal.

lukewagner:
Readability of the text format is indeed a (really, the only) goal of the text format; the key sticking point is whether we go back and add stuff to the AST (and, thus, the binary encoding) just for the benefit of the text format.  Anyhow, at this point I've clearly voiced my support of the simplification to the AST/binary in this PR.  As always, happy to have @MikeHolman or @pizlonator chime in; browser consensus is the goal.

JSStats:
Withdraw my objection. Shall explore a strategy of using a flag in a source code meta data section to change the presentation if it helps readability and it seems well possible do so here.

titzer:
I think removing cases is making tableswitch lower level in a way that doesn't really offer any benefits but has definite drawbacks. The existing tableswitch construct can already express a "caseless" tableswitch, so if that's the style of code produced by LLVM, no problem.

Removing cases means that we can no longer encode the much more mainstream AST-like switch construct with nested cases, which is more compact, too. A "caseless" tableswitch requires the declaration of all the cases as enclosing blocks, which are just space overhead in the binary, and are actually quite weird to force all producers to work with.

sunfishcode:
The benefit is it makes `tableswitch` simpler. We had several discussions to reach the current "target" abstraction, and several people involved say it's still awkward. Also, the interaction between cases and statements=expressions is tricky, as demonstrated by the current design text being incorrect.

Encoding a mainstream AST-like switch is already something the current `tableswitch` falls short of; cases are expected to be dense, start at 0, and they fall through by default. I expect high-level AST-based compilers will use a helper library to lower mainstream AST-like switches to wasm regardless.

The helper library could take care of declaring labels, as it isn't complicated. I'm also planning to propose `multiblock` (or similar) (as in #490) which would also simplify label declaration.

Concern for `tableswitch`'s code size is low, with its low incidence.

lukewagner:
From my perspective, `tableswitch` is the most complicated control operator and a good half of that complexity is cases; with cases removed, `tableswitch` is just a slight extension of `br_if`.  If we implement the binary encoding I'm thinking about, I don't even think there will be a size difference but, if there was, I agree it would be negligible; `tableswitch` will normally account for <1% of total bytes.  If we're thinking about the "tiny/custom compiler to wasm" use case, note that we can support this use case more directly by writing a JS library that takes as input some Esprima-like AST and compiles that to wasm binary (that's the abstraction I'd want to use).  That AST can include all the high-level control flow operators.  This is the same perspective we've taken with the binary encoding: don't do in layer 0 (engine) what could be done pretty well in layer 1 (user code).

rossberg-chromium:
I would argue that the main complexity of the current tableswitch is that the jump table is heterogeneous. This PR would remove that, which I like. However, it feels like it is favouring and optimising for the odd case now. The other possible simplification would be in the opposite direction, i.e. have only cases, no break shortcuts -- a simple structured tableswitch. I think that is the majority use case by far.

lukewagner:
@rossberg-chromium Having cases be nested as child nodes that need to be recursively visited and then wired together in the context of `tableswitch` is more complicated than having a `tableswitch` node that simply has an immediate list of indices that are bound in the same way as `br`.

JSStats:
@lukewagner Could you please elaborate on the 'the binary encoding I'm thinking about'? People will need some context to appreciate the point you are making and will want to evaluate the conclusion and implications. I occasionally see people with a strong point of view based on some vision that has not been articulated and it makes it difficult to appreciate their perspective. For example I would like to evaluate it from the point or view of how practical it is augment it with meta information to make it a more useful source code.

sunfishcode:
@JSStats I'd like to keep this issue focused on the language feature rather than the binary encoding. As Luke mentioned, even if we pick an encoding where it has a size increase, tableswitch is relatively infrequent and it wouldn't have a significant impact on overall code size.

JSStats:
@sunfishcode Agreed, it would be better to elaborate on alternative binary encodings elsewhere, a new issue, but to the extent that this PR is a step in that direction then it would be nice to see the plan and to be able to independently evaluate it. My comments are not an objection to this PR - I'll try to work around it if there is otherwise consensus. For the zlib benchmark, `tableswitch` plus the case bodies accounts for over 20% of the v8 binary code, and the `tableswitch` structure alone around 8% - this seems a significant AST element.

sunfishcode:
@JSStats I'm assuming the case *bodies* would not change size significantly.

sunfishcode:
@JSStats Also, in my copy of the zlib benchmark, there are only 3 switches and 38 cases total. This is much less than 1% of the total nodes in the program.

sunfishcode:
High-level compilers that are translating from traditional high-level language switches will be better served by using a library to produce wasm than producing wasm directly, because a library can intelligently lower sparse cases, handle dense case ranges that don't start at 0, and make cases not fall through by default. Such libraries can also easily translate cases into labels too.

The proposed tableswitch here makes the language simpler than using explicit cases because it reuses the existing system that `br` and `br_if` use for labels, rather than requiring a separate system. And, it more closely models actual machine code that the construct will be translated into.

sunfishcode:
Friendly ping :-).

sunfishcode:
Merging with LGTMs above, and discussion offline. Removing cases makes tableswitch/br_table simpler and makes the control-flow mechanism more consistent with br and br_if.

PR #513, opened 2016-01-05
Clarify AST semantic's goals

jfbastien:


jfbastien:
As discussed in https://github.com/WebAssembly/design/pull/512#issuecomment-168870709, a clarification.

JSStats:
@jfbastien It's not immediately clear that this really helps the issue of concern to me, but seems neutral.  If you could add your comment in https://github.com/WebAssembly/design/pull/512 'We haven't settled on a textual format yet, and I agree that making it easy to use without hindering the binary format is a desirable goal.' then this would be a significant concession, thank you. So long as all the 'easy to use' parts are not just ignorable binary blobs! For example, adding `when` and `unless` improve the compression ratio here and subjectively improve the readability, and these affect the AST too - the point being that these are not all a zero-sum decisions. I do appreciate the great reluctance to complicate web runtime implementations, and this seems one big motivation for a minimalist design, and if so then might this also be mentioned as a balancing point.

AstSemantics.md
```diff
@@ -4,6 +4,11 @@ WebAssembly code is represented as an Abstract Syntax Tree (AST) where each node
 represents an expression. Each function body consists of exactly one expression. 
 All expressions and operators are typed, with no implicit conversions or overloading rules.
 
+This document explain the high-level design of the AST: its types, constructs, and
```
kripken: explain**s**

AstSemantics.md
```diff
@@ -4,6 +4,11 @@ WebAssembly code is represented as an Abstract Syntax Tree (AST) where each node
...
+semantics. For an implementation consult [the formal spec](https://github.com/WebAssembly/spec),
```
kripken: an inplementation => implementation details

jfbastien:
@JSStats:
> It's not immediately clear that this really helps the issue of concern to me, but seems neutral. If you could add your comment in #512 'We haven't settled on a textual format yet, and I agree that making it easy to use without hindering the binary format is a desirable goal.' then this would be a significant concession, thank you. So long as all the 'easy to use' parts are not just ignorable binary blobs!

We've been moving these types of rationales to Rationale.md, and moving AstSemantics.md / TextFormat.md / BinaryEncoding.md towards being statements of how things are instead. Think of Rationale.md as the "why?", and these 3 documents as the "what?".

This PR is a pure clarification of the "why?", explaining where to look for sub-parts of "why?" since it's split up into multiple documents. It's therefore not the right place to get the change you desire. You've tried getting that change in previously and met pushback, that pushback hasn't changed. I suggest you wait for the text format to take shape before trying to influence the tradeoffs, or make your own text format and show us the tradeoffs. We're trying to move things forward and theoretical discussions aren't our focus now, code is. Once we have something limping along we'll step back, look at what we've created, and talk theory a bit more but with the ability to do Science and quantify tradeoffs (compresses better, compiles fasted, VM simpler, ...).

JSStats:
@jfbastien Thank you, that sounds like a wise strategy.

AstSemantics.md
```diff
@@ -4,6 +4,11 @@ WebAssembly code is represented as an Abstract Syntax Tree (AST) where each node
 represents an expression. Each function body consists of exactly one expression. 
 All expressions and operators are typed, with no implicit conversions or overloading rules.
 
+This document explains the high-level design of the AST: its types, constructs, and
+semantics. For implementation details consult [the formal spec](https://github.com/WebAssembly/spec),
```
lukewagner: The spec doesn't give "implementation details"; it's goal is to actually *be* the spec, so I'd say "For full details".

AstSemantics.md
```diff
@@ -4,6 +4,11 @@ WebAssembly code is represented as an Abstract Syntax Tree (AST) where each node
...
+for low-level encoding details consult [binary encoding](BinaryEncoding.md),
```
lukewagner: How about just "for the binary / over-the-wire representation of the AST, consult [BinaryEncoding.md]".

AstSemantics.md
```diff
@@ -4,6 +4,11 @@ WebAssembly code is represented as an Abstract Syntax Tree (AST) where each node
...
+and for the human-readable representation consult [text format](TextFormat.md).
```
lukewagner: ... and, symmetrically, "for the human-readable text representation, consult [TextFormat.md]".

AstSemantics.md
jfbastien: You're arguing @kripken's words here, I'll pick what you suggest but @kripken let me know if you'd like something else as well.

AstSemantics.md
jfbastien: I went with "file-level", seems more specific.

kripken:
lgtm

lukewagner:
Thanks!  lgtm

AstSemantics.md
```diff
@@ -4,6 +4,11 @@ WebAssembly code is represented as an Abstract Syntax Tree (AST) where each node
 represents an expression. Each function body consists of exactly one expression. 
 All expressions and operators are typed, with no implicit conversions or overloading rules.
 
+This document explains the high-level design of the AST: its types, constructs, and
+semantics. For full details consult [the formal spec](https://github.com/WebAssembly/spec),
+for file-level encoding details consult [binary encoding](BinaryEncoding.md),
+and for the human-readable text representation consult [text format](TextFormat.md).
```
sunfishcode: If "binary encoding" and "text encoding" are being used as titles, they should be capitalized. Otherwise, it would feel more comfortable to say "consult the binary encoding document" and "consult the text encoding document".

sunfishcode:
lgtm with style nit.

PR #514, opened 2016-01-10
New logo for WA

armanx64:
New logo for WebAssembly. 
Fast and strong, simple and futuristic same time.
I made it in vector format, so you can recolor it as you want for any purpose. 
Licence is a MIT.


jfbastien:
@armanx64, thanks for the suggestion, but WebAssembly isn't ready to have a logo yet. It's a cool looking logo though! I suggest adding your suggestion to #112 for now, and we'll figure out what logo WebAssembly has when we get closer to MVP.

Please also make sure you're a member of the W3C community as explained in [Contributing.md](Contributing.md).

License-wise we probably want something else than MIT, likely the same as the HTML5 logos as detailed in #112.

armanx64:
Sorry about licence part. Which is needed, how you think?

jfbastien:
@armanx64 we haven't really discussed it, but it we want to go with an HTML5 logo feel then we should follow what they do and use Creative Commons Attribution 3.0. We may want something else too, it's too early to figure out.

armanx64:
Sorry again for disturbing.

jfbastien:
You're not disturbing at all! Your participation is very welcome, it's just
about keeping all the logos in that issue for now so it's easier to get
back to when we add a logo.

PR #532, opened 2016-01-30
Move dynamic linking from FutureFeatures.md to PostMVP.md

lukewagner:
This better reflects what I believe is the shared priority that dynamic linking is in the set of features to do "right after" we finish/ship the MVP.

### DIFF
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -11,10 +11,6 @@ available under [feature tests](FeatureTest.md).
 
 This is covered in the [tooling](Tooling.md) section.
 
-## Dynamic linking
-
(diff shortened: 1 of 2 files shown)


haxxpop:
I agree with this. Without dynamic linking, it's quite hard and inefficient to port some existing libraries and porting existing libraries is a big issue :)

lukewagner:
Merging since no objections and implicit agreement from everyone [elsewhere](https://github.com/WebAssembly/webassembly.github.io/pull/1).

jfbastien:
Oh sorry I missed this, lgtm :)

PR #536, opened 2016-02-05
Change start to receive any function index to be its start function

haxxpop:
Since this makes it more flexible and similar to the existing `(export <int>)` construct, I would like to change `start` to accept any index number to the function instead of allowing just zero `(start 0)`

### DIFF
--- a/Modules.md
+++ b/Modules.md
@@ -140,11 +140,11 @@ For example, a start node in a module will be:
 
 or
 
-```(start 0)```
+```(start <int>)```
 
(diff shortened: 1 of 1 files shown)


jfbastien:
`0` was meant as an example, the same way `$start_function` isn't the only possible start function. Could you instead change the example from `0` to say `42`, and in the text explain it's the function index (same as for `export`).

haxxpop:
I have changed it already. Is it ok now?

jfbastien:
lgtm, thanks.

PR #539, opened 2016-02-09
Make br_if return the value of its value operand.

sunfishcode:
This follows naturally from br_if's value operand being strict. We
evaluate the expression to produce a value even when the condition is
false, so it'd be nice to be able to use the value when the condition
is false too.

This change also makes sense from a low-level compiler perspective. The
value operand of a br_if represents a value "live out" of the br_if's
basic block. Since br_if is a simple conditional branch, the value
should be "live in" in both succesors.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -263,19 +263,22 @@ further see the parallel, note that a `br` to a `block`'s label is functionally
 equivalent to a labeled `break` in high-level languages in that a `br` simply
 breaks out of a `block`.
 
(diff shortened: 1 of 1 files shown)


sunfishcode:
https://github.com/WebAssembly/spec/pull/237 is a PR to the spec repo implementing this, which has the specific proposal for how to do the type checking.

sunfishcode:
Closing, as the spec PR is now closed.

PR #546, opened 2016-02-23
Binary: Condense format of local declarations

rossberg-chromium:
The current format for declaring locals has two disadvantages: it is a bit wasteful, and it is not compatible with the future addition of new types. This PR suggests a more compact and extensible vector of entries.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -130,11 +130,11 @@ must contain a function body. Imported and exported functions must have a name. 
 
 #### Local Entry
 
-Each local entry declares a number of local variables of one type.
+Each local entry declares a number of local variables of a given type.
 It is legal to have several entries with the same type.
 
 | Field | Type | Description |
-| ----- | ----- | ----- | ----- |
+| ----- | ----- | ----- |
 | count | `varuint32` | number of local variables of the following type |
 | type | `value_type` | type of the variables |
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -123,12 +123,20 @@ must contain a function body. Imported and exported functions must have a name. 
 | flags | `uint8` | always | flags indicating attributes of a function <br>bit `0` : a name is present<br>bit `1` : the function is an import<br>bit `2` : the function has local variables<br>bit `3` : the function is an export |
 | signature | `uint16` | always | index into the Signature section |
 | name | `uint32` | `flags[0] == 1` | name of the function as an offset within the module |
-| i32 count | `uint16` | `flags[2] == 1` | number of `i32` local variables |
-| i64 count | `uint16` | `flags[2] == 1` | number of `i64` local variables |
-| f32 count | `uint16` | `flags[2] == 1` | number of `f32` local variables |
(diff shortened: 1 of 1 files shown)


titzer:
LRTM (looks reasonable to me)

jfbastien:
This allows interleaving and reordering locals of different types, correct? e.g. 5 x i32, then 6 x i64, then 3 x i32? Or simply putting the i32 at lower indices than the f64. I just want to make sure this is intentional, it seems like it could be useful to change local indexing so they compress better.

rossberg-chromium:
Yes, absolutely. That's why I explicitly included the second sentence in the "Local Entry" section.

jfbastien:
Cool, lgtm.

mbebenita:
Ran a quick experiment to measure this change. Binaryen currently reorders locals and groups them by type (to match the previous design) so these results measure at most 4 local entries (but arbitrarily many locals).

| Name                             |    Samples |       Size |    % Total |   Avg. Size  |
|:---------------------------------|-----------:|-----------:|-----------:|-------------:|
| Previous Encoding                |       2669 |    20.852K |      0.887 |       8.0000 |
| New Encoding                     |       2669 |     9.085K |      0.389 |       3.4856 |
| Proposed Encoding                |       2669 |     6.254K |      0.268 |       2.3994 |

Samples = Number of functions.
Size = Size in bytes of the local encoding.
% Total = Percent of the entire
Avg. Size = Size / Samples

The `Proposed Encoding` encodes the local entry (`type` and `count`) as `varuint32(count << 3) | type`. It only uses 3 bits for the type. To encode a type greater than 6, then we'd encode type as 7 and write another byte, this encoding yields a ~ 30% decrease in size. If we use 4 bits for the type, then we're looking at a 23% decrease (6.963K).

In summary, the new encoding looks much better, and we can go a little further. If people are interested, I can run the experiment without reordering/grouping.

mbebenita:
I should add that I ran these numbers on BananaBread.

JSStats:
@mbebenita Interesting thanks. Would not worry about packing bits just yet, split that out into a separate issue. The big advances in this patch seem to be: 1. allowing arbitrary ordering. 2. support local variable type extensions. 3. leb128 encoding of the count. Hope people support this patch and implement it promptly.

Might want to consider an leb128 encoding for the type if people are planning to store unboxed objects in local variables and tuples etc, and to pass them around in function operands.

A compression predictor could accurately predict the type of local variables from the context and use separate codes for each - assuming a pre-order encoding ;) Still a lot to settle in this area so perhaps bump bit packing to a future issue.

lukewagner:
lgtm

JSStats:
ping: looks like there was consensus here and this could land.

titzer:
I'm starting to implement this, and I think it makes sense to move the declarations to be part of the body, in fact, a prefix of the body.

JSStats:
@titzer Thank you. Just had the same thought about moving the local variable declarations to be with the body, see https://github.com/WebAssembly/design/pull/568 If the body is OOL then it would not be as practical to back patch them into the section definition.

PR #552, opened 2016-02-25
Unknown sections

### LINKED ISSUE #208: Should WebAssembly files support metadata/comments?
Minified JavaScript files, such Angular or jQuery, typically still include file information inside an opening comment block. This might include the name and version number of the software, copyright and licensing details, and maybe a URL leading to the original source code form. For example:

``` javascript
/*
 ExampleJS v1.3.16
 (c) 2010-2014 Example Foundation. http://example.org
 License: MIT
*/
```

Will WebAssembly support some kind of comment or metadata format, for similar file identification purposes?

jfbastien:
Fixes #208.

A later PR will make sure all sections start with the same format (so all sections use bytes, instead of count, making it easy to know the byte count of unknown sections). We'll also use strings instead.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -187,13 +187,15 @@ stored to from dedicated instructions.
 | type | `uint8` | the type of the global, as a memory type |
 | exported | `uint8` | a boolean indicating whether the global variable is exported |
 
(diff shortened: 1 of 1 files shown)


titzer:
lgtm

PR #555, opened 2016-02-25
Add i32.eqz.

sunfishcode:
This adds a `bool.not` operator, consistent with the one in the [binary format](BinaryFormat.md).

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -397,6 +397,7 @@ results into the result type.
   * `i32.clz`: sign-agnostic count leading zero bits (All zero bits are considered leading if the value is zero)
   * `i32.ctz`: sign-agnostic count trailing zero bits (All zero bits are considered trailing if the value is zero)
   * `i32.popcnt`: sign-agnostic count number of one bits
+  * `i32.eqz`: compare equal to zero (return 1 if operand is zero, 0 otherwise)
 
 Shifts counts are wrapped to be less than the log-base-2 of the number of bits
 in the value to be shifted, as an unsigned quantity. For example, in a 32-bit


titzer:
lgtm

rossberg-chromium:
Bikeshed: I know that v8-proto uses the same name, but it never struck me as particularly consistent. We decided not to have a bool type, so why should the operator name pretend otherwise?

MikeHolman:
Agreed with @rossberg-chromium. I would prefer to call i32.logicalNot or something along those lines.

JSStats:
Please not `i32.logicalNot` because there is a common logical-not operation and this is not it!

titzer:
i32.eqz? Since it is literally `x == 0`?

lukewagner:
I like the symmetry of `i32.eqz` with `i32.eq` so sgtm.

sunfishcode:
Currently we say all i32 operators are also available in i64. `bool.*` kind of bypassed that, but I'm not attached to it. Any preferences for adding an i64 version of this operation too for consistency, or instead for making an exception to that rule to exclude an i64 version? I don't have an opinion at present.

Also, for the shade of purple of the bikeshed, any preference for "eq0" instead of "eqz"?

binji:
Well, we already have `i32.clz` and `i32.ctz`, so `i32.eqz` seems more consistent.

If we had `i64.eqz`, would its return type be `i32` or `i64`? `i32` would be consistent with other comparison ops, but `i64` would be consistent with other unary ops.

jfbastien:
My bikeshed is painted with stripes of `i32.eqz` and `i64.eqz`.

titzer:
I think we should only add i32.eqz (or really, just rename bool.not). After
all, i64's cannot be used in conditions directly.

sunfishcode:
Updated the PR to `i32.eqz` with a special exception to the rule for i64 operators.

rossberg-chromium:
I strongly vote for including `i64.eqz` as well. I think it is pretty valuable to have symmetric operator sets for both types. And I can totally see it being useful for i64 as well, because zero comparisons are a frequent operation on any integer type (even more so once we have a 64-bit Wasm, where it may e.g. be used for a nullptr check).

@binji, since it is a predicate operator, `i64.eqz` would return bool a.k.a. i32, of course.

binji:
@rossberg-chromium yes, that would make the most sense. Though it does mean that `i64.eqz` is the only unary operator that returns a different type than its parameter (aside from the conversion operators, of course). Not that that's a reason not to do it.

titzer:
I'm fine with i64.eqz for symmetry.

sunfishcode:
Updated to add both `i32.eqz` and `i64.eqz` now, which appears to be the generally preferred approach.

rossberg-chromium:
Sounds good, though it somehow doesn't show up in the diff.

sunfishcode:
That's because AstSemantics.md just says "The same operators are available on 64-bit integers as the those available for 32-bit integers." The change to add `i64.eqz` was to remove the special exception I had previously added to that sentence.

rossberg-chromium:
Ah, okay, got it. LGTM

JSStats:
@sunfishcode Could you check that the name of the opcode in the binary encoding is updated here too as it was missed in https://github.com/WebAssembly/design/pull/587

sunfishcode:
i32.eqz and i64.eqz appear to have sufficient consensus; merging.

PR #562, opened 2016-02-27
Clarify encoding of varuint32 values.

mbebenita:
When encoding section sizes, tools will usually emit a patchable varuint32 value like 0xFFFFFFFF, reserving up to 5 bytes in the stream. Once the size is known, this is patched with the real size of the section which many not necessarily to take up 5 bytes, thus tools will need to encode leading zeros bytes. LEB128 supports this, but may be surprising to people writing decoders, hence this clarification.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -39,7 +39,7 @@ A two-byte little endian unsigned integer.
 A four-byte little endian unsigned integer.
 
 ### varuint32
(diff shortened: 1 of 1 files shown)


titzer:
lgtm

JSStats:
Could you please also consider making it clear that redundant leading zeros have not semantic meaning in wasm and are free to be stripped and canonicalized in the pipeline. For example a content optimizing tool in the pipeline feeding a mobile device etc.

PR #572, opened 2016-03-01
Move local entries to be part of the function body.

titzer:
This moves the declarations of local variables out of the function declaration into the body itself.


rossberg-chromium:
lgtm

Issue #569, opened 2016-02-29
When embedded in the web, clarify how export/import names convert to JS strings


pizlonator:
When embedded in the web, the exports and imports of a module are represented as JS objects.  JS objects are keyed by JS strings, which can be any UCS16 (roughly) sequence.  WebAssembly uses sequences of bytes.  Therefore, we need to specify conversion rules from WebAssembly byte sequence names to JS strings.  There will be JS strings that don't convert gracefully to bytes and vice-versa, so we also need something sensible to happen when the conversion fails.

titzer:
SGTM, wanna take a stab at updating Web.md?

jfbastien:
Do we want simple failure on the web when conversion fails, or do we want to try to "fix" things? Failure is fine IMO.

pizlonator:
Yup!

-Filip

pizlonator:
I was thinking eager failure at module compile time.

-Filip

s3ththompson:
Fixed in 04c63fb2b26f25b2bfd88569baf8711f146e4d06

PR #573, opened 2016-03-01
clarify how export/import names convert to JS strings (#569)

pizlonator:


Web.md
```diff
@@ -20,6 +20,42 @@ WebAssembly's [modules](Modules.md) allow for natural [integration with
 the ES6 module system](Modules.md#integration-with-es6-modules) and allow
 synchronous calling to and from JavaScript.
 
+### Function Names
+
+A WebAssembly module imports and exports functions. WebAssembly names functions
+using arbitrary-length byte sequences. The null character is permitted inside
+WebAssembly function names. The most natural Web representation of a mapping of
```
jfbastien: "the null character and invalid Unicode code points are permitted ..."

Web.md
```diff
@@ -20,6 +20,42 @@ WebAssembly's [modules](Modules.md) allow for natural [integration with
...
+function names to functions is a JS object in which each function is a property.
+Property names in JS are UTF-16 encoded strings. A WebAssembly modulde may fail
```
jfbastien: typo "modulde"

jfbastien:
lgtm besides comments.

Web.md
pizlonator: I agree with what this is trying to say, but it's a bit confusing since WebAssembly's notion of names are not Unicode - they're just bytes.

How about: "any 8-bit values are permitted in a WebAssembly name, including the null byte and byte sequences that don't correspond to any Unicode code point regardless of encoding"

jfbastien: Agreed your phrasing is better.

JSStats:
Alternative suggestion: mangle any non ucs16 sequences so that all valid wasm names work on the web. Perhaps break to an escaped format if it fails the already documented conversion.

pizlonator:
To give this serious consideration, I think you’d have to propose a specific mangling, and then provide reasoning about whether that mangling round-trips.  That’s gets quite weird.  It’s not obvious how to safely escape things while providing 1-1 mapping between JS UCS16 sequences and wasm byte sequences.  The biggest risk if we go this route is that we get some aspect of this wrong with respect to either Unicode itself or some browsers implementation of it.  Also, it will take more work for us, the spec writers, to write down and agree to an escaping format that satisfies all of the constraints that we might be interested in.

On the other hand, the likelihood that you’d choose to import or export something that has a name whose byte sequence doesn’t transcode is extremely low.  It’s probably not something anyone would want to do.  Most clients will either use a UTF8 encoding of some Unicode string (in which case they’re fine) or they’ll use ASCII (also fine).  Therefore, making that unlikely-and-undesirable thing be an error saves the spec, and all implementors of the spec, from having to worry about escaping.

-Filip

jfbastien:
Any escaped sequence would presumably *also* be a valid export/import name, giving us a different problem where escaped name is the same as a confusingly-named export. I don't think it's really worth it :)

JSStats:
Good points, thank you.

AndrewScheidecker:
You could make the mangling bijective, but it would require mangling otherwise valid UCS-16 strings (e.g. by expanding each byte into two hexits).

But I don't think it's worth distinguishing a web and non-web environment for this. Why not just say import/export names must be valid UTF-8 strings?

pizlonator:
> On Mar 1, 2016, at 1:55 PM, Andrew Scheidecker <notifications@github.com> wrote:
> 
> You could make the mangling bijective, but it would require mangling otherwise valid UCS-16 strings (e.g. by expanding each byte into two hexits).
> 
> But I don't think it's worth distinguishing a web and non-web environment for this. Why not just say import/export names must be valid UTF-8 strings?
> 
> 

This is an interesting suggestion.  I think that it’s easier on clients and implementors if this is a web-only constraint.

Non-web implementers may have no other need for any kind of Unicode stuff.  I think that this would be the only mention of Unicode in WebAssembly.  A non-Web client wishing to implement accurate verification (i.e. if fails exactly when the spec it says it must fail) would have to do Unicode logic only in this one place, and only to support a verification rule that only helps the web.

On the other hand, the web embedding scenario will necessarily perform verifications that non-web clients have no need for.  We’ll probably have to also add rules for what kind of object the imports object can be, though these rules may be implicit (for example if we say that the imports object is queried at module instantiation time then we’re mandating that all things named by the imports section are accessible via [[Get]] or somesuch).  We’ll definitely have additional Web-specific export rules.  For example I just realized that we probably want to prohibit exporting a function named “__proto__” because that that would probably make the world burn.  Again, it would be weird if this was a rule that non-web users would have to follow.

Since we will anyway have to have Web-specific verification, and since the non-Web case is probably happy with names being sequences of bytes, I think that the approach I’ve proposed is least bad.

-Filip

jfbastien:
> Non-web implementers may have no other need for any kind of Unicode stuff.  I think that this would be the only mention of Unicode in WebAssembly.  A non-Web client wishing to implement accurate verification (i.e. if fails exactly when the spec it says it must fail) would have to do Unicode logic only in this one place, and only to support a verification rule that only helps the web.

Agreed, that's what we'd discussed to justify our approach. Add it to Rationale.md?

lukewagner:
Agreed on keeping the utf8 (and any other) requirement a Web-only thing.

(As for `__proto__`, it seems to be possible to `Object.defineProperty(o, '__proto__', ...)` (and shadow the `__proto__` accessor property defined on `Object.prototype`).  So I'm not aware of any special *need* to exclude this particular case, devilish though it may be)

pizlonator:
Oh that’s true!

Opinions on whether __proto__ should be allowed as a function name when embedding in the web?

(I vote “no”.  It would cause too much agony, and I can’t imagine a well-behaved client wanting to do it.)

-Filip

lukewagner:
It's hard to argue "yes" with any conviction, but unless we can find a case where it breaks or bothers some particular engine or tool (I think SM used to implement `__proto__` in a particularly magic way that might've had problems, but no longer), I'd vote "yes" purely on the basis of avoiding adding special cases.

pizlonator:
In JSC, the following evaluates to true:

	({__proto__: 42}).__proto__ == Object.prototype

It also evaluates to true in Firefox.  I didn’t try others.

That’s pretty weird!  It means that a function exported as __proto__ will not be accessible if the exports object is created in a manner that is logically equivalent to what a JS developer would reasonably expect (i.e. literal construction).

For this reason, I think that a special case that causes verification failure for __proto__ is consistent with having verification failure for byte sequences that cannot be converted to JS strings: the idea is that if you would not have been able to easily reference the function name from JS, then you probably messed up.

-Filip

lukewagner:
Ah interesting.  So object literals must not do define-property but rather something set-property-esque.  We currently use define-property to build our wasm export object (set-property could trigger setters!) so
```
js> var code = wasmTextToBinary('(module (func (result i32) (i32.const 42)) (export "__proto__" 0))'));
js> var o = wasmEval(code);
js> o.__proto__();
42
```
works, but I guess that's not your point.  It's weird, no doubt, but I'm wondering if people will thank us for rejecting this one particular string because there isn't an equivalent object literal.  I guess that's something asm.js couldn't polyfill ;)  Likely people will never notice, so if you're really worried about this causing problems, I'm fine rejecting it; we could always allow it later if it caused people trouble.

Web.md
```diff
@@ -20,6 +20,44 @@ WebAssembly's [modules](Modules.md) allow for natural [integration with
...
+function convertToJSString(array)
+{
+  // Perform the actual conversion.
+  var string = "";
+  for (var i = 0; i < array.length; ++i)
+    string += String.fromCharCode(array[i]);
+  var result = decodeURIComponent(escape(string));
+  
+  // Check for errors. This will throw if 'result' contains bad characters.
+  encodeURIComponent(result);
```
lukewagner: I'm trying to figure out what byte sequences will *not* throw during `decodeURIComponent(escape(string))` but *will* throw in `encodeURIComponent(result)`.  I tried both invalid utf8 byte sequences and also bad high/low surrogate pairs but `decodeURIComponent` catches them all.

Web.md
```diff
@@ -20,6 +20,44 @@ WebAssembly's [modules](Modules.md) allow for natural [integration with
...
+  
+  return result;
+}
+```
+
+This performs the UTF8 decoding (`decodeURIComponent(unescape(string))`) using
```
lukewagner: I think you mean `escape` instead of `unescape`?

Web.md
```diff
@@ -20,6 +20,44 @@ WebAssembly's [modules](Modules.md) allow for natural [integration with
...
+a common JS idiom, and uses the first part of the encoding idiom
```
lukewagner: Maybe markdown-hyperlink "idiom" to http://monsur.hossa.in/2012/07/20/utf-8-in-javascript.html?

Web.md
lukewagner: Ah hah, and poking around more, it would appear that it's simplify invalid utf8 to store a code point in the surrogate range \[0xd800, 0xdc00\] (I had been thinking that utf8 wouldn't care about surrogates, but I guess this is a concession to utf16 baked into utf8?), so your definition here is *precisely* what we want (if we want to reuse our existing Utf8ToUtf16 conversion routines which I very much do.  lgtm!

rossberg-chromium:
On 1 March 2016 at 23:47, pizlonator <notifications@github.com> wrote:

> In JSC, the following evaluates to true:
>
> ({__proto__: 42}).__proto__ == Object.prototype
>
> It also evaluates to true in Firefox. I didn’t try others.
>

You just fell into one of JavaScript's many corner case traps. This only
yields true because you used 42, which is not an object, nor is it
converted to one. Instead, it gets replaced by Object.prototype (see
https://tc39.github.io/ecma262/#sec-__proto__-property-names-in-object-initializers,
but don't ask why, this is JavaScript). In contrast,
```
({__proto__: {}}).__proto__ == Object.prototype
```
will yield false.

Also note that this is a special case in the semantics of object _literals_
only, which is not something that Wasm needs to be concerned with. Other
than that, and the presence of `Object.prototype.__proto__` (all of which
is Annex B crap anyway), there is no special treatment of the name
`__proto__` in ES6. In particular, Object.defineProperty, the core
operation for adding properties, has no special awareness of it.

Hence, I see no reason for any exceptions regarding `__proto__`. In fact, I
would advise against it.

Web.md
lukewagner: I should add: it has the right behavior, but I'd still like to know if the `encodeURIComponent` is necessary or if `decodeURIComponent` catches it all.

pizlonator:
Good catch!

I agree with your analysis.

-Filip

Web.md
pizlonator: decodeURIComponent should throw.  The reason why I didn’t just rely on that is that some of the public documentation documents encodeURIComponent as throwing but doesn’t say that decodeURIComponent should throw.  But having read the spec (both ES5.1 and ES6), I no longer think that the encodeURIComponent call is necessary.

I *think* that we can remove the encodeURIComponent call.  Objections?

-Filip

lukewagner:
lgtm

lukewagner:
Oops, looks like this never merged.  Any objections to merging now?

jfbastien:
lgtm as well, let's merge!

PR #580, opened 2016-03-02
Names section

lukewagner:
This PR adds two optional sections: one containing a sequence of names of functions and one containing a sequence of names of locals.  The goal is that, when present, these names are used instead of auto-generated names in any projected text and so this constitutes the wasm analogue of a "symbol section".  Locals are separate since they'll be a lot bigger and devs may want function names but not locals.  Later, we could consider alternative sections which instead contain only a URL but I think in the MVP it'll be much simpler to embed these in the module file.

A few choices which I thought about a bit but could be convinced otherwise on:

. Unlike exports/imports which we wanted to leave as unconstrained byte sequence, if the *entire purpose* of these names is to be displayed as text in the text format, then it seems justified to constrain them to be valid utf8-encoded, null-terminated strings.  Note that an engine doesn't have to validate or even look at this section if it doesn't want to, so it doesn't force utf8 decoding.  Also, even if we didn't constrain the byte sequence, I think we'd end up with this limitation in practice.  We could consider giving each string a preceding byte length for symmetry with export/import strings, but I think this would simply add validation work without benefit.

. These name sections don't declare the number of functions or number of locals, they instead implicitly refer to the already-declared number-of-functions/locals.  We could add lengths but, as above, I think this would just add work without benefit.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -165,9 +165,8 @@ must contain a function body. Imported and exported functions must have a name. 
 
 | Field | Type |  Present?  | Description |
 | ----- |  ----- |  ----- |  ----- |
-| flags | `uint8` | always | flags indicating attributes of a function <br>bit `0` : a name is present<br>bit `1` : the function is an import<br>bit `2` : the function has local variables<br>bit `3` : the function is an export |
+| flags | `uint8` | always | flags indicating attributes of a function <br>bit `1` : the function is an import<br>bit `2` : the function has local variables<br>bit `3` : the function is an export |
 | signature | `uint16` | always | index into the Signature section |
-| name | `uint32` | `flags[0] == 1` | name of the function as an offset within the module |
 | body size | `uint16` | `flags[0] == 0` | size of function body to follow, in bytes |
 | body | `bytes` | `flags[0] == 0` | function body |
 
@@ -233,6 +232,42 @@ This section must be preceded by a [Functions](#functions-section) section.
 | ----- |  ----- | ----- |
 | count | `varuint32` | count of entries to follow |
 | entries | `uint16*` | repeated indexes into the function table |
+
+### Function Names section
+
+ID: `function_names`
+
+| Field | Type | Description |
+| ----- |  ----- | ----- |
+| names | `utf8-C-string*` | sequence of null-terminated utf8-encoded strings |
+
+The number of names is determined by the number of declared functions. The
+sequence of names assigns a name to each function index.
+
+This section is optional and does not contribute to the semenatics of execution.
+A validation error in this section is not reported and is treated as the section
+being absent. The expectation is that, when a binary WebAssembly module is
+viewed in a browser or other development environment, the names in this section
+will be used as the names of functions in the [text format](TextFormat.md).
+
+### Local Names section
+
+ID: `local_names`
+
+| Field | Type | Description |
+| ----- |  ----- | ----- |
+| names | `utf8-C-string*` | sequence of null-terminated utf8-encoded strings |
+
+The number of names is determined by the sum of all functions' number of locals.
+The sequence of names is defined to be the concatentation of the sequence of
+local names (ordered by local index) for each function (ordered by function index).
+This sequence assigns a name to every local index in every function.
+
+This section is optional and does not contribute to the semenatics of execution.
+A validation error in this section is not reported and is treated as the section
+being absent. The expectation is that, when a binary WebAssembly module is
+viewed in a browser or other development environment, the names in this section
+will be used as the names of locals in the [text format](TextFormat.md).
 
 ### End section
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -244,8 +244,8 @@ ID: `function_names`
 The number of names is determined by the number of declared functions. The
 sequence of names assigns a name to each function index.
 
-This section is optional and does not contribute to the semenatics of execution.
-A validation error in this section is not reported and is treated as the section
+This section may occur 0 or 1 times and does not change observable semantics. A
+validation error in this section is not reported and is treated as the section
 being absent. The expectation is that, when a binary WebAssembly module is
 viewed in a browser or other development environment, the names in this section
 will be used as the names of functions in the [text format](TextFormat.md).
@@ -263,8 +263,8 @@ The sequence of names is defined to be the concatentation of the sequence of
 local names (ordered by local index) for each function (ordered by function index).
 This sequence assigns a name to every local index in every function.
 
-This section is optional and does not contribute to the semenatics of execution.
-A validation error in this section is not reported and is treated as the section
+This section may occur 0 or 1 times and does not change observable semantics. A
+validation error in this section is not reported and is treated as the section
 being absent. The expectation is that, when a binary WebAssembly module is
 viewed in a browser or other development environment, the names in this section
 will be used as the names of locals in the [text format](TextFormat.md).
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -235,39 +235,49 @@ This section must be preceded by a [Functions](#functions-section) section.
 
 ### Function Names section
 
-ID: `function_names`
-
-| Field | Type | Description |
-| ----- |  ----- | ----- |
-| names | `utf8-C-string*` | sequence of null-terminated utf8-encoded strings |
-
-The number of names is determined by the number of declared functions. The
-sequence of names assigns a name to each function index.
-
 This section may occur 0 or 1 times and does not change execution semantics. A
 validation error in this section is not reported and is treated as the section
 being absent. The expectation is that, when a binary WebAssembly module is
 viewed in a browser or other development environment, the names in this section
 will be used as the names of functions in the [text format](TextFormat.md).
 
-### Local Names section
-
-ID: `local_names`
+ID: `function_names`
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
-| names | `utf8-C-string*` | sequence of null-terminated utf8-encoded strings |
+| names | `name*` | sequence of names |
 
-The number of names is determined by the sum of all functions' number of locals.
-The sequence of names is defined to be the concatentation of the sequence of
-local names (ordered by local index) for each function (ordered by function index).
-This sequence assigns a name to every local index in every function.
+The number of names is determined by the number of declared functions. The
+sequence of names assigns a name to each function index. A name is:
+
+| Field | Type | Description |
+| ----- |  ----- | ----- |
+| size | varuint32 | string length, in bytes |
+| string | bytes | valid utf8-encoding of non-null code points |
+
+### Local Names section
 
 This section may occur 0 or 1 times and does not change execution semantics. A
 validation error in this section is not reported and is treated as the section
 being absent. The expectation is that, when a binary WebAssembly module is
 viewed in a browser or other development environment, the names in this section
 will be used as the names of locals in the [text format](TextFormat.md).
+
+ID: `local_names`
+
+| Field | Type | Description |
+| ----- |  ----- | ----- |
+| names | `name*` | sequence of names |
+
+The number of names is determined by the sum of all functions' number of locals.
+The sequence of names is defined to be the concatentation of the sequence of
+local names (ordered by local index) for each function (ordered by function index).
+This sequence assigns a name to every local index in every function. A name is:
+
+| Field | Type | Description |
+| ----- |  ----- | ----- |
+| size | varuint32 | string length, in bytes |
+| string | bytes | valid utf8-encoding of non-null code points |
 
 ### End section
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -263,7 +263,7 @@ functions.
 | local_names | `local_name*` | sequence of local names |
 
 The sequence of `local_name` assigns names to the corresponding local index. The
-count may be greater or less than the actual number of functions.
+count may be greater or less than the actual number of locals.
 
 #### Local name
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -246,8 +246,8 @@ functions and locals in the [text format](TextFormat.md).
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
-| count | `varuint32` | count of names to follow |
-| names | `function_names*` | sequence of names |
+| count | `varuint32` | count of entries to follow |
+| entries | `function_names*` | sequence of names |
 
 The sequence of `function_name` assigns names to the corresponding
 function index. The count may be greater or less than the actual number of
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -233,51 +233,44 @@ This section must be preceded by a [Functions](#functions-section) section.
 | count | `varuint32` | count of entries to follow |
 | entries | `uint16*` | repeated indexes into the function table |
 
-### Function Names section
+### Names section
+
+ID: `names`
 
 This section may occur 0 or 1 times and does not change execution semantics. A
-validation error in this section is not reported and is treated as the section
-being absent. The expectation is that, when a binary WebAssembly module is
-viewed in a browser or other development environment, the names in this section
-will be used as the names of functions in the [text format](TextFormat.md).
-
-ID: `function_names`
-
-| Field | Type | Description |
-| ----- |  ----- | ----- |
-| names | `name*` | sequence of names |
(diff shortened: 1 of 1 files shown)


lukewagner:
If separate "functions" into a separate "declarations" section (just a list of signatures) and a "bodies" section, and then order sections logically (small stuff (including "declarations"), bodies, data, names) then I think we won't need any out-of-line concept and all that entails.

JSStats:
@lukewagner As described I don't think that would be adequate, but with some changes it might. For example, if the runtime does not know the offset of the start of the function bodies section and does not know the size of each function body then it can not randomly seek to the body data for a function which might be very useful for debugging or viewing one function in many. Similarly for looking up the local names for a function. The same issue might also frustrate compilation dispatch to multiple threads and compilation resource planning.

If all these offsets could be declared sooner, for example in the functions section, then it might all work. For example add the offset of the function bodies section to the functions section plus the size of each function body, and similarly add the start of the local names section and the size for each function to the functions section.

btw: The string null termination and utf-8 requirement and the serial order of the names all look good.

lukewagner:
Yes, in this proposal, each function body would continue to describe the byte length of the body so they could be trivially skipped.  And, additionally, there was also discussion of keeping the forward offsets, as part of the function-body section, which could be viewed as simply a further optimization to make it easy to do parallel/lazy validation.  So 'yes', but that's all for another PR :)

JSStats:
@lukewagner Thank you. So if function bodies will be in a separate section then the locals names section can come after the bodies and there is no need for them to be OOL - sounds like a workable plan. But there still appears to be a need to include the byte length of the locals names for each function in the functions section and the start of the locals names section so that it is possible to find the locals names for a particular function without scanning from the start of the first function? Or are you suggesting that this too will be addressed in a follow up PR?

If maintaining sizes to lookup the locals names then it might be useful to lump all the debug info for a function in the same blob rather than each field in separate sections. i.e. include the function name and optional locals names in a single blob for each function and in a single debug section.

lukewagner:
> But there still appears to be a need to include the byte length of the locals names for each function in the 
> functions section and the start of the locals names section so that it is possible to find the locals names for
> a particular function without scanning from the start of the first function?

It's a good question and I'm not fully decided on it.  My expectation (and current SM implementation) is that an engine won't necessarily keep the whole .wasm binary around after it has finished compiling.  Rather, it would decode any metadata needed into various other internal data structures, including local names if they were present, and free the binary afterwards.  Given that assumption, I was expecting an engine to decode the whole names sections at once.  Even if an engine wanted to keep the .wasm binary and use that directly for local names, the engine could build an internal table of offsets once (up front or lazily).

titzer:
On Wed, Mar 2, 2016 at 7:19 AM, Luke Wagner <notifications@github.com>
wrote:

> But there still appears to be a need to include the byte length of the
> locals names for each function in the
> functions section and the start of the locals names section so that it is
> possible to find the locals names for
> a particular function without scanning from the start of the first
> function?
>
> It's a good question and I'm not fully decided on it. My expectation (and
> current SM implementation) is that an engine won't necessarily keep the
> whole .wasm binary around after it has finished compiling. Rather, it would
> decode any metadata needed into various other internal data structures,
> including local names if they were present, and free the binary afterwards.
> Given that assumption, I was expecting an engine to decode the whole names
> sections at once. Even if an engine wanted to keep the .wasm binary and use
> that directly for local names, the engine could build an internal table of
> offsets once (up front or lazily).
>
V8 doesn't currently keep any of the binary's bytes after compiling the
WASM module. It materializes names for exports but not for internal
functions, but this will change when we start reporting more accurate
stacktraces for WASM code.

BinaryEncoding.md
```diff
@@ -240,6 +239,42 @@ This section must be preceded by a [Functions](#functions-section) section.
...
+| Field | Type | Description |
+| ----- |  ----- | ----- |
+| names | `utf8-C-string*` | sequence of null-terminated utf8-encoded strings |
+
+The number of names is determined by the sum of all functions' number of locals.
+The sequence of names is defined to be the concatentation of the sequence of
+local names (ordered by local index) for each function (ordered by function index).
+This sequence assigns a name to every local index in every function.
+
+This section may occur 0 or 1 times and does not change observable semantics. A
```
titzer: s/observable/execution/, here and above

BinaryEncoding.md
```diff
@@ -240,6 +239,42 @@ This section must be preceded by a [Functions](#functions-section) section.
 | count | `varuint32` | count of entries to follow |
 | entries | `uint16*` | repeated indexes into the function table |
 
+### Function Names section
+
+ID: `function_names`
+
+| Field | Type | Description |
+| ----- |  ----- | ----- |
+| names | `utf8-C-string*` | sequence of null-terminated utf8-encoded strings |
```
titzer: Do we need to mandate utf8 null terminated? Could these also be length+bytes like other strings, for consistency?

lukewagner: To avoid the annoying (and sure to cause bugs) question of "what happens if null is in the middle of the string", I'd like to disallow that case regardless.  So if we wanted up-front length, then we'd also drop the null-terminator (otherwise it's just more work to make sure both expressions of length match) right?  That'd be fine with me.

lukewagner: (Done, PTAL)

rossberg-chromium: +1 for up-front length and no terminator. -1 for disallowing internal 0 -- an embedder may impose additional constraints like that, but I don't see why the Wasm binary format needs to.

For what's it worth, even JavaScript can deal with 0 characters in the middle of strings just fine:
```
let s = 'a\0b'
s.charCodeAt(1) === 0  // true
let o = {[s]: 7}
o[s] === 7  // true
```

lukewagner: I'm well aware :)

I guess my point was: the annoyance of internal null-terminators is not specific to the web (as you said it works just fine as a string); and the whole point of these strings is to inject into bigger strings.  I guess it just boils down to me being lazy: I'd rather deal with utf8 char* C-strings than {length, char*} pairs :)  So I guess I'm fine dropping this restriction too, if others want to allow.

lukewagner:
Ok, it does feel a bit arbitrary to specially disallow null, so I dropped that requirement.  I expect, in practice, the strings will get mangled if the code points are bizarre, but that's fine.  Look good to everyone?

JSStats:
Good to have these sections settled, so I appreciate the effort.

I guess the community can experiment with a separate section that is a directory of the offset to the start of the locals for each function, so this does not seem a big issue. If it is missing then a full scan may be necessary.

However it still seems a burden to have to name all the locals even if the developer only wants to name some. If there were a design principle that the burden should increase gracefully as more source code meta information is added (a principle I am using for webll) then this design would not meet this principle.

BinaryEncoding.md
```diff
@@ -234,6 +233,52 @@ This section must be preceded by a [Functions](#functions-section) section.
...
+| ----- |  ----- | ----- |
+| names | `name*` | sequence of names |
+
+The number of names is determined by the number of declared functions. The
+sequence of names assigns a name to each function index. A name is:
+
+| Field | Type | Description |
+| ----- |  ----- | ----- |
+| size | varuint32 | string length, in bytes |
+| string | bytes | valid utf8 encoding |
```
rossberg-chromium: Shouldn't we likewise drop any mentioning of/requirement about encoding? It seems inconsistent to require valid UTF8 when no engine is actually wanting to validate that. Wasm as such just doesn't care, it's random bytes.

BinaryEncoding.md
```diff
@@ -234,6 +233,52 @@ This section must be preceded by a [Functions](#functions-section) section.
...
+viewed in a browser or other development environment, the names in this section
+will be used as the names of functions in the [text format](TextFormat.md).
+
+ID: `function_names`
+
+| Field | Type | Description |
+| ----- |  ----- | ----- |
+| names | `name*` | sequence of names |
+
+The number of names is determined by the number of declared functions. The
```
rossberg-chromium: Note that this is context-sensitive in a very non-local manner again. I don't think it would hurt to include a count here (and below). And it would be nice to have a consistent encoding of all things vector.

BinaryEncoding.md
```diff
@@ -234,6 +233,52 @@ This section must be preceded by a [Functions](#functions-section) section.
...
+viewed in a browser or other development environment, the names in this section
+will be used as the names of locals in the [text format](TextFormat.md).
+
+ID: `local_names`
+
+| Field | Type | Description |
+| ----- |  ----- | ----- |
+| names | `name*` | sequence of names |
+
+The number of names is determined by the sum of all functions' number of locals.
```
rossberg-chromium: I wonder if this isn't too simplistic. A large app might have 10s of 1000s of locals. Perhaps some producers would like to include local names only for selected functions?

Also, how about combining both name tables into one Names section? That could have one entry for each function: its name, followed by an (optional) list of names for its locals.

BinaryEncoding.md
lukewagner: Regular encoding of all-things-vector is a good argument; gets me every time ;)

BinaryEncoding.md
lukewagner: The practical reason is that, if we don't mandate utf8, then it's possible tools will start generating ascii or utf16 or ... and I think that could tangibly hurt the wasm ecosystem.  It'd be html content-sniffing days all over again (except no MIME types or <meta> tags to help out).  Here, the purpose of "render to text" is baked into wasm; with import/export strings we're intentionally throwing up our hands and letting the host decide (and the contents might not even be text, but some structured data rep that, say, describes lookup parameters).

jfbastien: I agree with @rossberg-chromium that we should be consistent. We already have the gotcha @lukewagner mentions with section names, and it would be nice to allow VMs which are oblivious to Unicode :-)

rossberg-chromium: ASCII would be fine, even under the UTF8 requirement. :)  But I see your point. Can the spec somehow make UTF8 a "recommendation" or intention? It's weird to say it's a requirement when in fact none of the major consumers will check or otherwise care, nor are they expected to.

lukewagner: @jfbastien This was the point of the text above saying "A validation error in this section is not reported and is treated as the section being absent.": a WebAssembly implementation that doesn't care can simply skip over this section.  And if you *do* want to use the text, either it's definitely in utf8 (so you need the decoder) or it's not and we have the fractured-ecosystem-with-no-way-to-tell-what-the-encoding-is problem.

BinaryEncoding.md
lukewagner: Yes, previously I broke it up b/c locals are too big.  BUT, if we do lengths anyway and then fuse the two lists together so that each function has a name and list of locals (that can be less than the actual number of locals) that solves it all nicely.  I think I'll do that.

BinaryEncoding.md
lukewagner: @rossberg-chromium Hehe, yeah, I guess I should've said latin1.

What do you think about the current text above which says "you don't have to validate, but if you do at all, it has to be real utf8".  I think that models what we'd do in FF: if present (and maybe only if a devtools flag is set) decode as utf8.  If that fails, you don't get symbols and maybe we print a warning to the console saying invalid names section.

lukewagner:
Thanks all, updated to address comments, PTAL.

BinaryEncoding.md
```diff
@@ -234,6 +233,45 @@ This section must be preceded by a [Functions](#functions-section) section.
...
+validation error in this section does not cause validation for the whole module
+to fail and is instead treated as if the section was absent. The expectation is
+that, when a binary WebAssembly module is viewed in a browser or other
+development environment, the names in this section will be used as the names of
+functions and locals in the [text format](TextFormat.md).
+
+| Field | Type | Description |
+| ----- |  ----- | ----- |
+| count | `varuint32` | count of names to follow |
+| names | `function_names*` | sequence of names |
```
rossberg-chromium: Nit: perhaps rename `names` to `entries` now?

BinaryEncoding.md
```diff
@@ -234,6 +233,45 @@ This section must be preceded by a [Functions](#functions-section) section.
...
+
+The sequence of `function_name` assigns names to the corresponding
+function index. The count may be greater or less than the actual number of
```
rossberg-chromium: Is there a specific reason you want to allow greater than?

BinaryEncoding.md
```diff
@@ -234,6 +233,45 @@ This section must be preceded by a [Functions](#functions-section) section.
...
+
+| Field | Type | Description |
+| ----- |  ----- | ----- |
+| fun_name_len | `varuint32` | string length, in bytes |
+| fun_name_str | `bytes` | valid utf8 encoding |
+| local_count | `varuint32` | count of local names to follow |
+| local_names | `local_name*` | sequence of local names |
+
+The sequence of `local_name` assigns names to the corresponding local index. The
+count may be greater or less than the actual number of functions.
```
rossberg-chromium: s/number of functions/number of locals/. Also, same question as above.

BinaryEncoding.md
lukewagner: Nothing specific, just thinking: if we're going to drop the equality requirement, might as well drop it all the way.

BinaryEncoding.md
rossberg-chromium: I guess we all agree on the intention, it just comes down to legal word smithing. So I'm fine with leaving as is for now.

jfbastien:
Sorry to be dense: I still don't understand the UTF8 thing. The table says:

> valid utf8 encoding

And you description says:

> Note that an engine doesn't have to validate or even look at this section if it doesn't want to, so it doesn't force utf8 decoding.

It seems like these are contradicting each other?

lukewagner:
The first para of the Names section says that if the names section is invalid, the module isn't invalid (so you can simply skip over the whole section without looking at a byte).  The "valid utf8 encoding" requirement contributes to whether the names section is valid which means that.

JSStats:
Thank you. Combining it into one section will work well with a directory as a single directory can be used to find the function name and the locals names.

What would you think about defining these locals names as a prefix and that where there are duplicates the consumer is expected to add a unique suffix?

Doing so would allow unnamed locals to be zero length and the suffix would define the entire name, except if there were only one and could that also be defined as needing a suffix? It also resolves the problem of duplications being potentially invalid.

lukewagner:
I expect there will be a fair amount of text duplication and potential for savings, but I would also imagine that network-level gzip/brotli would do really well about squeezing this down.

JSStats:
Yes, I think compression could really help if the locals names were defined to be prefixes. If unnamed locals are zero length then blocks of them can compress very well. Also if the producer can take advantage of the prefix encoding and allow the runtime to add the unique suffix then there might also be blocks of names with the same prefix that compress well.

What would you also think about moving all the counts and string lengths before all the string data? This would then give an effective directory at the start of the section that could be used to determine the offset in the section of a name. It would likely also help compression - a layer-1 benefit that also gives a directory useful at layer 0.

If there is a relative efficient encoding for unnamed locals (zero length prefix) then duplicating the local_count  might not be so important and the actual locals count from the functions section might be usable. It might still be useful to have a single bit flag per function to indicate if there are any locals names, just to reduce the number of uncompressed lengths where some functions have no locals names. So the section could be:

| ----- |  ----- | ----- |
| flags | `uint8*` | debug info flags for each function in the order defined in the functions section. Bit 0 - function is named. Bit 1 - locals are named with the count defined in the functions section. |
| name_lengths | `varuint32*` | lengths of function name strings and locals names strings as defined by the flags for each function and in the order defined in the functions section.
| name_str | `bytes*` | valid utf8 encoded names |

Perhaps the flags could be packed too, as it's not clear if this encode would be extensible anyway.

JSStats:
I see some merit in splitting the data into separate sections, a function names section and a locals names sections, from the point of view of extensibility. If more debug info is added it would go in a separate section and a runtime that did not understand it would just ignore it - this property might not hold if all pack per-function in a single section. Sorry for the round-about.

So could each section have a `flags` `name_lengths` and `name_str` block with the lengths up front to provide the directory. It might even be ok to drop the `flags` and just depend on compression to pack blocks of unnamed names in which case it would be very similar to your original proposal but with the lengths up front and the names defined to be prefixes and zero length names defined as unnamed.

lukewagner:
How about we merge the current state, which is pretty simple and gets the `name` out of function entries (that's my motivation for pushing on this) and then consider further improvements to the names section in a separate PR before the MVP.

@titzer @rossberg-chromium lgty?

titzer:
lgtm

PR #585, opened 2016-03-04
Br table

titzer:


kripken:
Was the first commit included here intentionally? It's from another pull which is still being debated, and doesn't look necessary for this change.

titzer:
Accidentally branched off a non-main branch.

PR #586, opened 2016-03-04
Replace tableswitch with br_table.

titzer:


sunfishcode:
I'd like to propose splitting the default label out of the table and making it a separate immediate. The default label is not "part of the table" in wasm, from an AST perspective (as it is currently proposed), or from an implementation perspective -- the default is what the instruction does when the index is out of bounds in the actual table.

lukewagner:
+1 (I was in the middle of writing the same comment)

titzer:
I guess it's just about what the table_count field stores. If it stores the
count of non-default targets, then the default target can immediately
follow the target table. Same bytes, except the table_count is 1 less.

rossberg-chromium:
@titzer, good point. I guess I'd prefer the count to not include the default, since that avoids introducing an illegal useless value for it.

lukewagner:
@titzer Agreed, and also prefer not including default in the count.

titzer:
PTAL

lukewagner:
lgtm

sunfishcode:
lgtm

PR #589, opened 2016-03-04
Remove the nonstandard load_global and store_global opcodes.

sunfishcode:


titzer:
lgtm.

Obviously we still use these internally but we can renumber them to be invalid opcodes and only support them in our internal API.

sunfishcode:
Yep; that's what we did as well.

PR #590, opened 2016-03-07
Store section data inline in the section

lukewagner:
This PR stores export/import strings and data section bytes inline instead of having offsets into other parts of the module.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -218,6 +218,7 @@ a `data_segment` is:
 | ----- |  ----- | ----- |
 | offset | `uint32` | the offset in linear memory at which to store the data |
 | size | `uint32` | the size of the data segment (in bytes) |
+| data | `bytes` | a sequence of `size` bytes |
 
 ### Indirect Function Table section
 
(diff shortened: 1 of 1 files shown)


BinaryEncoding.md
```diff
@@ -213,12 +212,12 @@ A module may only contain one data segments section.
...
-  - ```uint32```: The base address of the data segment in memory.
-  - ```uint32```: The offset of the data segment's data in the file.
-  - ```uint32```: The size of the data segment (in bytes)
-  - ```uint8```: ```1``` if the segment's data should be automatically loaded into memory at module load time.
+a `data_segment` is:
+
+| Field | Type | Description |
+| ----- |  ----- | ----- |
+| offset | `uint32` | the offset in linear memory at which to store the data |
+| size | `uint32` | the size of the data segment (in bytes) |
```
rossberg-chromium: Shouldn't this include a row for the actual data (presumably of type `bytes`)?

lukewagner: Correct, will update in a sec

lukewagner:
@rossberg-chromium @titzer PTAL

rossberg-chromium:
lgtm

MikeHolman:
lgtm

titzer:
lgtm

PR #591, opened 2016-03-07
Split Functions section into Function Signatures and Function Bodies

lukewagner:
This PR splits the Functions section into a Function Signatures section that goes near the beginning of the module and a Function Bodies section that would go at the end (but before Data Segments and Names).  This allows all function signatures to be known when compiling function bodies.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -145,28 +145,31 @@ A module may contain at most one import table section.
 | function_len | `varuint32` | function string length |
 | function_str | `bytes` | function string of `function_len` bytes |
 
-### Functions section
+### Function Signatures section
 
-ID: `functions`
+ID: `function_signatures`
 
-The Functions section declares the functions in the module and must be preceded by a [Signatures](#signatures-section) section. A module may contain at most one functions section.
+The Functions Signatures section declares the signatures of all functions in the
+module and must be preceded by the [Signatures](#signatures-section) section. A
+module may contain at most one functions section.
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
-| count | `varuint32` | count of function entries to follow |
-| entries | `function_entry*` | repeated function entries as described below |
+| count | `varuint32` | count of signature indices to follow |
+| signatures | `varuint32*` | sequence of indices into the Signature section |
 
-#### Function Entry
+### Function Bodies section
 
-Each function entry describes a function that can be optionally named, imported and/or exported. Non-imported functions
-must contain a function body. Imported and exported functions must have a name. Imported functions do not contain a body.
+ID: `function_bodies`
 
-| Field | Type |  Present?  | Description |
+The Function Bodies section assigns a body to every function in the module and
+must be preceded by the [Function Signatures](#function-signatures-section) section.
+The count of function signatures and function bodies must be the same.
+
+| Field | Type |  Description |
 | ----- |  ----- |  ----- |  ----- |
-| flags | `uint8` | always | flags indicating attributes of a function <br>bit `1` : the function is an import<br>bit `2` : the function has local variables<br>bit `3` : the function is an export |
-| signature | `uint16` | always | index into the Signature section |
-| body size | `uint16` | `flags[0] == 0` | size of function body to follow, in bytes |
-| body | `bytes` | `flags[0] == 0` | function body |
+| count | `varuint32` | count of function bodies to follow |
+| bodies | `function_body*` | sequence of [Function Bodies](#function-bodies) |
 
 ### Export Table section
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -187,7 +187,6 @@ This section must be preceded by a [Functions](#functions-section) section.
 #### Export entry
 | Field | Type | Description |
 | ----- |  ----- | ----- |
-| sig_index | `uint16` | signature index of the export |
 | func_index | `uint16` | index into the function table |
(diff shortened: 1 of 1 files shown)


BinaryEncoding.md
```diff
@@ -147,28 +147,31 @@ A module may contain at most one import table section.
...
-### Functions section
+### Function Signatures section
 
-ID: `functions`
+ID: `function_signatures`
 
-The Functions section declares the functions in the module and must be preceded by a [Signatures](#signatures-section) section. A module may contain at most one functions section.
+The Functions Signatures section declares the signatures of all functions in the
+module and must be preceded by the [Signatures](#signatures-section) section. A
+module may contain at most one functions section.
```
Cellule: `functions signatures section`?

BinaryEncoding.md
```diff
@@ -147,28 +147,31 @@ A module may contain at most one import table section.
...
 
 
-### Functions section
+### Function Signatures section
 
-ID: `functions`
+ID: `function_signatures`
 
-The Functions section declares the functions in the module and must be preceded by a [Signatures](#signatures-section) section. A module may contain at most one functions section.
+The Functions Signatures section declares the signatures of all functions in the
```
Cellule: Maybe worth mentioning that the order of the signatures must match the order of bodies

BinaryEncoding.md
lukewagner: good catch!

BinaryEncoding.md
lukewagner: Sounds good; I'll put the note in the Function Bodies section, where it already talks about the length correspondence.

BinaryEncoding.md
```diff
@@ -147,28 +147,32 @@ A module may contain at most one import table section.
...
+The Function Bodies section assigns a body to every function in the module and
+must be preceded by the [Function Signatures](#function-signatures-section) section.
+The count of function signatures and function bodies must be the same and the `i`th
+signature corresponds to the `i`th function body.
+
+| Field | Type |  Description |
 | ----- |  ----- |  ----- |  ----- |
-| flags | `uint8` | always | flags indicating attributes of a function <br>bit `1` : the function is an import<br>bit `2` : the function has local variables<br>bit `3` : the function is an export |
-| signature | `uint16` | always | index into the Signature section |
-| body size | `uint16` | `flags[0] == 0` | size of function body to follow, in bytes |
```
binji: function body now needs a terminator or length, right?

lukewagner: The existing body byte size is sufficient to know when the body is done.  @rossberg-chromium was suggesting to add a body terminator.  This PR reflects what's in SM which I had thought is what's in V8 as well.  Perhaps we could add the body terminator in a separate PR?

titzer: On Mon, Mar 7, 2016 at 2:18 PM, Luke Wagner <notifications@github.com>
wrote:

> In BinaryEncoding.md
> <https://github.com/WebAssembly/design/pull/591#discussion_r55282783>:
>
> >  | ----- |  ----- |  ----- |  ----- |
> > -| flags | `uint8` | always | flags indicating attributes of a function <br>bit `1` : the function is an import<br>bit `2` : the function has local variables<br>bit `3` : the function is an export |
> > -| signature | `uint16` | always | index into the Signature section |
> > -| body size | `uint16` | `flags[0] == 0` | size of function body to follow, in bytes |
>
> The existing body byte size is sufficient to know when the body is done.
> @rossberg-chromium <https://github.com/rossberg-chromium> was suggesting
> to add a body terminator. This PR reflects what's in SM which I had thought
> is what's in V8 as well. Perhaps we could add the body terminator in a
> separate PR?
>
The byte terminator seems somewhat redundant to me, and even in-person
discussions with @rossberg-chromium haven't convinced me otherwise yet :-)

lukewagner: Ok, let's hold off for now :)

JSStats:
This change seems to address quite a different use case than moving the bodies OOL would have. Namely it gives only the signatures early. It might also be useful to note that the signatures sections defines the number of functions. The body sizes and their position in the file are still not known early. If the bodies had just been move into a later section (keeping their sizes as-is) then this would have looked quite different and might still have been named the `functions` section rather than `function_signatures`.  As-is it leaves the work of moving the bodies OOL. I see some merit in placing the signatures all in one section as they might compress better together.

As-is the function body sizes seem undefined, as @binji questioned. One alternative option is to make the body an effective block with a length rather than adding a byte size - it does seem a little inconsistent to me to have to test the byte length when decoding the top level expression sequence of a function but having an expression count in other places. This would make the body size redundant, and if a `function_body_sizes` section were added to provide an directory then it might be optional.

lukewagner:
Oops, I had intended to move the body size down into the function body, sorry, I'll update in a sec.  I'd like to discuss adding forward offsets in a separate PR as that could just be a (possibly optional?) table inside and at the beginning of the function bodies section.

lukewagner:
@binji Ok, it's back!

JSStats:
@lukewagner Ok, thank you. Moving all the body sizes to the start of the `function_bodies` section would also help address the function body 'directory' use case and probably even help compression, but such options can be explored in a separate PR and a sections offset directory will also be needed.

Cellule:
I think the precedence of the `Indirect Function Table section` needs updating as well

lukewagner:
Good catch; I updated a few other dangling references to the "Function section" as well.

MikeHolman:
lgtm

PR #592, opened 2016-03-07
Convert (most) remaining fixed-width immediates to LEB128

lukewagner:
This PR builds on previous ones so it's probably best to just look at the last commit.  After the PR, the only fixed-width integers are the magic/version words (kept for simplicity and since they are "special") and the `exported` `uint8` in the Memory section, since it should be going away in a future PR to export (and import) memory from the exports/imports section.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -145,28 +145,31 @@ A module may contain at most one import table section.
 | function_len | `varuint32` | function string length |
 | function_str | `bytes` | function string of `function_len` bytes |
 
-### Functions section
+### Function Signatures section
 
-ID: `functions`
+ID: `function_signatures`
 
-The Functions section declares the functions in the module and must be preceded by a [Signatures](#signatures-section) section. A module may contain at most one functions section.
+The Functions Signatures section declares the signatures of all functions in the
+module and must be preceded by the [Signatures](#signatures-section) section. A
+module may contain at most one functions section.
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
-| count | `varuint32` | count of function entries to follow |
-| entries | `function_entry*` | repeated function entries as described below |
+| count | `varuint32` | count of signature indices to follow |
+| signatures | `varuint32*` | sequence of indices into the Signature section |
 
-#### Function Entry
+### Function Bodies section
 
-Each function entry describes a function that can be optionally named, imported and/or exported. Non-imported functions
-must contain a function body. Imported and exported functions must have a name. Imported functions do not contain a body.
+ID: `function_bodies`
 
-| Field | Type |  Present?  | Description |
+The Function Bodies section assigns a body to every function in the module and
+must be preceded by the [Function Signatures](#function-signatures-section) section.
+The count of function signatures and function bodies must be the same.
+
+| Field | Type |  Description |
 | ----- |  ----- |  ----- |  ----- |
-| flags | `uint8` | always | flags indicating attributes of a function <br>bit `1` : the function is an import<br>bit `2` : the function has local variables<br>bit `3` : the function is an export |
-| signature | `uint16` | always | index into the Signature section |
-| body size | `uint16` | `flags[0] == 0` | size of function body to follow, in bytes |
-| body | `bytes` | `flags[0] == 0` | function body |
+| count | `varuint32` | count of function bodies to follow |
+| bodies | `function_body*` | sequence of [Function Bodies](#function-bodies) |
 
 ### Export Table section
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -187,7 +187,6 @@ This section must be preceded by a [Functions](#functions-section) section.
 #### Export entry
 | Field | Type | Description |
 | ----- |  ----- | ----- |
-| sig_index | `uint16` | signature index of the export |
 | func_index | `uint16` | index into the function table |
 | function_len | `varuint32` | function string length |
 | function_str | `bytes` | function string of `function_len` bytes |
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -303,6 +303,7 @@ nodes (if any).
 
 | Name | Opcode |Description |
 | ----- | ----- | ----- |
+| body size | `varuint32` | size of function body to follow, in bytes |
 | local count | `varuint32` | number of local entries |
 | locals | `local_entry*` | local variables |
 | ast    | `byte*` | pre-order encoded AST |
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -178,7 +178,7 @@ ID: `export_table`
 
 The export table section declares all exports from the module.
 A module may contain at most one export table section.
-This section must be preceded by a [Functions](#functions-section) section.
+This section must be preceded by the [Function Signatures](#function-signatures-section) section.
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
@@ -197,7 +197,7 @@ This section must be preceded by a [Functions](#functions-section) section.
 ID: `start_function`
 
 A module may contain at most one start fuction section.
-This section must be preceded by a [Functions](#functions-section) section.
+This section must be preceded by a [Function Signatures](#function-signatures-section) section.
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
@@ -228,8 +228,7 @@ a `data_segment` is:
 ID: `function_table`
 
 The indirect function table section declares the size and entries of the indirect function table, which consist
-of indexes into the [Functions](#functions-section) section.
-This section must be preceded by a [Functions](#functions-section) section.
+of indexes into the [Function Signatures](#function-signatures-section) section (which must come before).
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -26,14 +26,8 @@ implementations.
 
 # Data types
 
-### int8
-A single-byte signed integer.
-
 ### uint8
 A single-byte unsigned integer.
-
-### uint16
-A two-byte little endian unsigned integer.
 
 ### uint32
 A four-byte little endian unsigned integer.
@@ -120,7 +114,7 @@ A module may contain at most one signatures section.
 #### Signature entry
 | Field | Type | Description |
 | ----- |  ----- | ----- |
-| param_count | `uint8` | the number of parameters to the function |
+| param_count | `varuint32` | the number of parameters to the function |
 | return_type | `value_type?` | the return type of the function, with `0` indicating no return type |
 | param_types | `value_type*` | the parameter types of the function |
 
@@ -139,7 +133,7 @@ A module may contain at most one import table section.
 #### Import entry
 | Field | Type | Description |
 | ----- |  ----- | ----- |
-| sig_index | `uint16` | signature index of the import |
+| sig_index | `varuint32` | signature index of the import |
 | module_len | `varuint32` | module string length |
 | module_str | `bytes` | module string of `module_len` bytes |
 | function_len | `varuint32` | function string length |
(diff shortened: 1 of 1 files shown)


BinaryEncoding.md
```diff
@@ -213,12 +209,12 @@ A module may only contain one data segments section.
...
-  - ```uint32```: The base address of the data segment in memory.
-  - ```uint32```: The offset of the data segment's data in the file.
-  - ```uint32```: The size of the data segment (in bytes)
-  - ```uint8```: ```1``` if the segment's data should be automatically loaded into memory at module load time.
+a `data_segment` is:
+
+| Field | Type | Description |
+| ----- |  ----- | ----- |
+| offset | `varuint32` | the offset in linear memory at which to store the data |
+| size | `varuint32` | the size of the data segment (in bytes) |
```
binji: where is the data segment data?

lukewagner: oops, let me add that that back :)

BinaryEncoding.md
```diff
@@ -390,10 +387,10 @@ out of range, `br_table` branches to the default target.
 
 The `memory_immediate` type is encoded as follows:
 
-| Name | Type | Present? | Description |
-| ---- | ---- | ---- | ---- |
-| flags | `uint8` | always | a bitfield where<br>bit `4` indicates an offset follows<br>bit `7` indicates natural alignment<br>other bits are reserved for future use |
-| offset | `varuint32` | `flags[4] == 1` | the value of the offset |
+| Name | Type | Description |
+| ---- | ---- | ---- |
+| offset | `varuint32` | the value of the offset |
```
binji: As I suggested in IRC, maybe swap the order of offset and align? If we assume that the memory access is often naturally aligned, then it should be better for compression (using nullary macros or otherwise) paired with the memory access opcode that precedes it.

binji: Also, do we want to add atomicity?

BinaryEncoding.md
```diff
@@ -390,10 +387,10 @@ out of range, `br_table` branches to the default target.
...
+| alignment | `varuint32` | the value of the alignment |
```
titzer: I'm still not convinced of the utility of alignments other than natural and unaligned. Can we leave this one as-is for now?

JSStats: Please also see the suggestion in https://github.com/WebAssembly/design/pull/571 which optimizes for the common case of natural alignment and for the  power-of-two values and the expected maximum alignment of the page size. One key property is that the flag is zero for natural alignment and does not change - whereas if the actual alignment is encoded then it will change based on the natural alignment. Encoding the powers of two keeps the values smaller, and if small values are move frequent then this might help compression. A custom compressor could easily map the proposed leb128 values to short codes, but if depending on a generic compressor then it should help to map the frequent case to a consistent code and map the values to high frequency values. Also keep in mind that Brotli only looks back at most two bytes so packing can still help.

lukewagner:
Updated `memory_immediate` section.  PTAL @titzer

BinaryEncoding.md
```diff
@@ -335,9 +332,9 @@ The `br_table` operator has an immediate operand which is encoded as follows:
 
 | Field | Type | Description |
 | ---- | ---- | ---- |
-| target_count | `uint16` | number of targets in the target_table |
-| target_table | `uint16*` | target entries that indicate an outer block or loop to which to break |
-| default_target | `uint16` | an outer block or loop to which to break in the default case |
+| target_count | `varuint32` | number of targets in the target_table |
+| target_table | `varuint32*` | target entries that indicate an outer block or loop to which to break |
```
binji: Maybe these should stay as u16s? It's nice to have the `br_table` data be fixed size for fast and simple parsing. Also, when running an interpreter, we'll want to index into this list dynamically, so having each element be fixed size is convenient.

titzer: Yeah, unfortunately an interpreter would have to decode the entire target_table to figure out where a given dynamic index has to go.

titzer: So in this case I support keeping a uint16_t.

rossberg-chromium: Hah, bummer. How about this: use leb128, but with the extra constraint that all values in the table must use an encoding with the same width (using leading zero as necessary)? That way, an interpreter would only have to look at the first index to know the size of the table.

sunfishcode: Wouldn't an interpreter interested in performance want to translate the table to a set of bytecode offsets anyway?

Also, uint16 gives us 65536 levels. There exist programs with switches with more arms than that. If this is to be a fixed size, I'd advocate for at least uint32_t.

lukewagner: Yeah, definitely seems like we'll hit `uint16_t` limit as time goes to infinity.  I think there's a famous example of this from Microsoft lore back in the day.  The binary format is already not directly interpretable with any efficiency and wants a translation to bytecode which could normalize.  Within a compiler, decoding a sequence of LEB128 is no big deal.

BinaryEncoding.md
lukewagner: Just to be clear, with the most recent update to this PR, alignment takes a fixed number of bits (bounded by `log2(memory width)`) and so, when we want to add any ordering flags, they could go after.  LEB128 lets us have an infinite bitfield :)

BinaryEncoding.md
titzer: I support making a uint32_t then. The direct interpretation case is actually pretty easy in postorder, so it'd be a shame to require a rewrite just for tableswitch.

lukewagner: I was going to make the argument that `loop`/`block`/`if`/`else`/`end` make direct interpretation infeasible, but the memoized jump table you've brought up would probably be effective.  If we want to officially change our stance on direct interpretation from "infeasible, requires compilation to bytecode" to "not super-fast, but feasible, requires jump table" *and allow this to affect our design*, it'd be good to have a working proof of concept before the MVP (that has no separate bytecode, just side tables).  Is this what you built with your debug interpreter?  If so, cool!

I would make an argument for size advantages of LEB128 over `uint32` (given that the common case will be 1 byte in LEB128), but this is `br_table` which accounts for <.1% of total bytes, so I won't :)

So `uint32` seems like a fine compromise, I'll update now and we can reconsider later if necessary.

lukewagner:
I also made `default_target` a `uint32`, just for symmetry.

lukewagner:
PTAL

BinaryEncoding.md
titzer: On Tue, Mar 8, 2016 at 8:04 PM, Luke Wagner <notifications@github.com>
wrote:

> In BinaryEncoding.md
> <https://github.com/WebAssembly/design/pull/592#discussion_r55409170>:
>
> > @@ -335,9 +332,9 @@ The `br_table` operator has an immediate operand which is encoded as follows:
> >
> >  | Field | Type | Description |
> >  | ---- | ---- | ---- |
> > -| target_count | `uint16` | number of targets in the target_table |
> > -| target_table | `uint16*` | target entries that indicate an outer block or loop to which to break |
> > -| default_target | `uint16` | an outer block or loop to which to break in the default case |
> > +| target_count | `varuint32` | number of targets in the target_table |
> > +| target_table | `varuint32*` | target entries that indicate an outer block or loop to which to break |
>
> I was going to make the argument that loop/block/if/else/end make direct
> interpretation infeasible, but the memoized jump table you've brought up
> would probably be effective. If we want to officially change our stance on
> direct interpretation from "infeasible, requires compilation to bytecode"
> to "not super-fast, but feasible, requires jump table" *and allow this to
> affect our design*, it'd be good to have a working proof of concept
> before the MVP (that has no separate bytecode, just side tables). Is this
> what you built with your debug interpreter? If so, cool!
>
Yes, I had that working for preorder, and it's even easier for postorder.
You just need a sidetable for if, else, br, br_if, and br_table opcodes.


> I would make an argument for size advantages of LEB128 over uint32 (given
> that the common case will be 1 byte in LEB128), but this is br_table
> which accounts for <.1% of total bytes, so I won't :)
>
> So uint32 seems like a fine compromise, I'll update now and we can
> reconsider later if necessary.
>
Ok, sgtm.

rossberg-chromium: No love for my "homogeneous leb128" proposal...? :)

lukewagner: Oh, hah, sorry, I didn't mean to ignore it.  It just sounds like more work to validate and then we'd need some irregular (compared to everything else) spec language to explain it.  `uint32` seems like the least-bad compromise between these competing constraints, imho.  Happy to revisit later in a separate issue, though, just hoping not to hang up this one.

BinaryEncoding.md
```diff
@@ -142,33 +133,38 @@ A module may contain at most one import table section.
 #### Import entry
```
commonlisp: Is there an ordering dependency that the Import Table must come after Signatures which it indexes?

lukewagner: My plan was, after this and the other two PRs merge, I'd make a new PR that specified the precise order of all sections (as a simple regexp: `signatures? import_table? function_signatures? ...`) and reordered the sections in BinaryEncoding.md.  This would conflict with changes in this PR so I was holding off until after they merged.

binji:
lgtm

MikeHolman:
lgtm

JSStats:
@lukewagner Sorry again, I forgot to add in the mandatory offset with this new encoding in my test result above. When paired with this change, using zero for natural alignment gives almost a 1% improvement in compressed size. I guess the common case of two zeros together helps compresses just a little better. Don't let this block this PR, but there might still be something to consider in future.

lukewagner:
(Forgot to add that opcodes are `varuint32`.)

BinaryEncoding.md
```diff
@@ -63,17 +57,14 @@ sequence, then followed recursively by any child nodes.
 
 * Examples
   * Given a simple AST node: `I32Add(left: AstNode, right: AstNode)`
-    * First write the opcode for `I32Add` (uint8)
+    * First write the opcode for `I32Add` (`varuint32`)
```
titzer: We already have opcodes > 127, so this isn't quite right.

lukewagner: I'm not sure what the alternative is.  We'll surely hit the `uint8` limit of 256 opcodes (just from SIMD).  `uint16` would probably be "enough for anyone" (... probably) but will seriously bloat binaries.  LEB128 seems like the best thing here which takes 1 byte in the common case.  Remember, with opcode tables, ops can be reordered so that the hot ones which consume 99.9% of the bytes get the 1-byte encodings.

rossberg-chromium: For opcodes there could be more special tailored schemes, with secondary opcodes just in selected cases. For example, for SIMD in particular, you could imagine that there is just one primary opcode SIMD, and the zoo of actual operations is distinguished by a secondary byte. (Or alternatively, there could be one primary opcode per SIMD type.)

I think the advantage is that this provides a larger short opcode space (up to almost 2x), and a bit more flexibility in reserving/prioritising short opcodes for selected future operators -- opcode table notwithstanding.

lukewagner:
Rebased over some merge conflicts and leaving opcode as `uint8` since it sounds like that will need a separate PR and discussion over the best variable-length opcode encoding.

lukewagner:
@titzer PTAL

BinaryEncoding.md
lukewagner: Yeah, I had been thinking along these lines too (as a future binary iteration); I didn't realize the switch to LEB128 would be such a pain for V8 as an intermediate step.

On the subject: the opcode table would mean that the actual choice of indices was up to the module.  I had been thinking that the way to achieve what you're saying (255 ops fit in one byte instead of just 128) is just to use a variable-length scheme where 0xff was the two-byte prefix (and 0xffff mean 3 bytes, etc).  That's what I'll probably do in SM right now to workaround the fact that, for our internal asm.js->wasm conversion, we have >256 opcodes due to SIMD.

titzer:
lgtm thanks

BinaryEncoding.md
```diff
@@ -216,22 +213,21 @@ a `data_segment` is:
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
-| offset | `uint32` | the offset in linear memory at which to store the data |
-| size | `uint32` | the size of the data segment (in bytes) |
-| data | `bytes` | a sequence of `size` bytes |
+| offset | `varuint32` | the offset in linear memory at which to store the data |
+| size | `uint32` | size of `data` (in bytes) |
```
commonlisp: Should data size also be a varuint32?

lukewagner: Oops, I lost that change while rebasing, good catch!

PR #597, opened 2016-03-09
Define precise order of sections

lukewagner:
This PR defines the order of sections so that an engine can walk down an optional list of sections and easily match section ids when decoding a module.

A rationale for the particular order in this PR:
* Signatures section needed by everything so it goes first.
* Imports section as early as possible to allow a streaming module loader to fire off parallel fetches asap.
* Indirect Function Table must go after Function Signatures
* Export Table goes after all declarations that may possibly get exported one day (e.g., the Indirect Function Table)
* The Start Function doesn't declare anything, so it goes after exports.
* Function bodies go before Data segments so that parallel compilation may occur while data is being streamed, but after everything else since this is the meat of the module.
* Names at the end since they don't semantically block anything.

Two other details in this PR to note/discuss:
* With inline everything, there isn't a need for an end section.  To be symmetric, this PR mirrors the current choice in function bodies to have byte length, not an explicit end marker, end the module.
* Unknown sections are said to be allowed between any two sections and there isn't any caveat that tries to catch out-of-order sections: if the order of two sections is flipped, the first one will just get ignored.  Happy to tighten this down if people wanted but it seemed like more work which might not actually help anyone.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -89,10 +89,8 @@ for all sections. The encoding of all sections begins as follows:
 | id_len | `varuint32` | section identifier string length |
 | id_str | `bytes` | section identifier string of id_len bytes |
 
-Each section other than the End section is optional and may appear at most once.
-The End section must appear exactly once. If present, sections must occur in
-this precise order (interleaved or followed by unknown sections, as noted
-above):
+Each section is optional and may appear at most once.
+Known sections (from this list) may not appear out of order.
 
 * [Signatures](#signatures-section) section
 * [Import Table](#import-table-section) section
@@ -103,10 +101,7 @@ above):
 * [Start Function](#start-function-section) section
 * [Function Bodies](#function-bodies-section) section
 * [Data Segments](#data-segments-section) section
-* [End](#end-section) section
 * [Names](#names-section) section
-
-Known sections (from this list) may not appear out of order.
 
 The end of the last present section must coincide with the last byte of the
 module. The shortest valid module is 8 bytes (`magic number`, `version`,
@@ -247,17 +242,6 @@ a `data_segment` is:
 | offset | `varuint32` | the offset in linear memory at which to store the data |
 | size | `varuint32` | size of `data` (in bytes) |
 | data | `bytes` | sequence of `size` bytes |
-
-### End section
-
-ID: `end`
-
-This section is mandatory and indicates the end of the sections that affect
-semantics. Subsequent sections may be skipped or streamed lazily without
-affecting execution.
-
-| Field | Type | Description |
-| ----- |  ----- | ----- |
 
 ### Names section
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -89,9 +89,10 @@ for all sections. The encoding of all sections begins as follows:
 | id_len | `varuint32` | section identifier string length |
 | id_str | `bytes` | section identifier string of id_len bytes |
 
-Each section other than the End section is optional and may appear at most once
-and only in the precise order defined below (out-of-order sections are a
-validation error).
+Each section other than the End section is optional and may appear at most once.
+The End section must appear exactly once. If present, sections must occur in
+this precise order (interleaved or followed by unknown sections, as noted
+above):
 
 * [Signatures](#signatures-section) section
 * [Import Table](#import-table-section) section
@@ -105,13 +106,12 @@ validation error).
 * [End](#end-section) section
 * [Names](#names-section) section
 
-The End section must appear exactly once and is the last section that affects
-execution, including whether the module validates. Any validation error after
-the End section causes the containing section and all subsequent bytes to be
-ignored.
+Thus, the shortest valid module is 13 bytes (`magic number`, `version`,
+`size` = 4, `id_len` = 3, `id_str` = "end").
 
-The shortest valid module is thus 13 bytes: `magic number`, `version`,
-`size` = `4`, `id_len` = `3`, `id_str` = `end`.
+Additionally, known sections (from the above list) may not appear out of order
+and the end of the last present section must coincide with the last byte of the
+module. 
 
 ### Signatures section
 
@@ -264,9 +264,12 @@ affecting execution.
 
 ID: `names`
 
-When a binary WebAssembly module is viewed in a browser or other development
-environment, the names in this section, when present, are used as the names of
-functions and locals in the [text format](TextFormat.md).
+The names section does not change execution semantics and a validation error in
+this section does not cause validation for the whole module to fail and is
+instead treated as if the section was absent. The expectation is that, when a
+binary WebAssembly module is viewed in a browser or other development
+environment, the names in this section will be used as the names of functions
+and locals in the [text format](TextFormat.md).
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -71,14 +71,17 @@ The following documents the current prototype format. This format is based on an
 
 ## High-level structure
 
-The module starts with a magic number and version as follows.
+The module starts with a preamble of two fields:
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
 | magic number | `uint32` |  Magic number `0x6d736100` (i.e., '\0asm') |
 | version | `uint32` | Version number, currently 10. The version for MVP will be reset to 1. |
 
-This is followed by a sequence of sections. Sections can in general be repeated, but some can occur only once or have dependent sections that must preceed them but not immediately as unknown sections can occur in any order. Each section is identified by an immediate string. Sections whose identity is unknown to the WebAssembly implementation are ignored and this is supported by including the size in bytes for all sections. The encoding of all sections begins as follows:
+This preamble is followed by a sequence of sections. Each section is identified by an
+immediate string. Sections whose identity is unknown to the WebAssembly
+implementation are ignored and this is supported by including the size in bytes
+for all sections. The encoding of all sections begins as follows:
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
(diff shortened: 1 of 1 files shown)


Cellule:
In your PR Description you forgot about `Function Signatues` which I believe goes between `Import` and `Indirect Function` from looking at the change.
>  if the order of two sections is flipped, the first one will just get ignored

That means a silent failure for a known section. I would have preferred to trap if 2 known sections are swapped to help catch possible bugs.
Or maybe we could export a list of ignored/skip sections (obviously not needed in the short term however).

lukewagner:
> In your PR Description you forgot about Function Signatues which I believe goes between Import and 
> Indirect Function from looking at the change.

Correct, I was more just giving rationale; the proposed order is what's listed in the list in BinaryEncoding.md.

> That means a silent failure for a known section. I would have preferred to trap if 2 known sections are 
> swapped to help catch possible bugs.

Agreed, this is what I was on the fence about as well.  So we could define this by saying that any given section id is either "known" or "unknown" and only "unknown" sections can be skipped.  That's easy enough; sound good to everyone else?

lukewagner:
@titzer @binji comments?

binji:
Mostly a bikeshed, but is there any rationale for the location of the memory module? It seems more natural to me for that to be first.

JSStats:
@binji The memory limits can be used to optimize the compiled code so need to be before the function bodies to avoid blocking compilation, but that's the only 'rationale' that comes to mind.

lukewagner:
I liked the principle of "imports as early *as possible*" followed by all the exportable definitions (incl. memory), followed by the export table.  But this still allows the memory section to be in any one of 3 positions (relative to `function_signatures` and `function_table`), so you're right, that much is arbitrary.

JSStats:
Might there be some utility in retaining an 'end' style section market that indicates the end of the execution-semantic sections?  This might give the runtime more confidence to start execution before the entire file has been received, given that the tail might have a lot of meta information, and not increase the latency when adding meta information.

lukewagner:
Given that the engine will know itself which sections it understands and that unknown sections are ignored, I wouldn't think the engine would need any extra hints.  For example, regardless of the end section, an engine could start executing (assuming it's reading from a network stream) right after the data segments section finished (before the name section).

JSStats:
@lukewagner Might there ever be sections added that have a semantic effect on execution, so that the data segments is not the last etc? Also since unknown sections can come anywhere, how can it know that the optional data segments section is not still to come? If a sections directory could be added at the start then this would also resolve the matter.

lukewagner:
If new sections were added in the future, they would continue to be ignored by older engines (b/c they would be unknown to those engines).  Newer engines that recognized those sections would know to wait for them.  There's only one data section so with a precise ordering, after you've seen it, you can know there's nothing semantically-visible left.

JSStats:
@lukewagner But the data segments section is optional? So after reading all the function bodies, then seeing a big unknown section, and before a 'names' section, the runtime still does not know if it can start execution (there may be no data segments section to end the execution semantic sections)? The concern is that adding meta data in unknown sections will introduce latency that could be avoided.

The PR seems to define all sections as optional 'The shortest valid module is 8 bytes', but some depend on others. Could this is made clearer with a consistent text paragraph for each section stating the constraints under which it is required and that it is otherwise optional etc.

lukewagner:
I guess I'd like to see evidence of large non-semantic sections between spec-defined semantics-affecting sections (e.g., between bodies and data) to justify adding new "dividing" markers (which is a different proposal from an "end" marker which would literally be the last section).  Such markers could also be added in the future.

I think many of the constraints are already present in the validation criteria: e.g. the number of function bodies has to match function signatures, function table indices have to be in range (implying there must be a section), etc.  Are there any particular validation criteria that are missing?

JSStats:
@lukewagner There is the 'evidence':

Facts:

1. Projects such as the WebLL are adding source code meta information, which might be a significant size.

2. You have worked on the 'names' section adding similar meta information, but much more limited and have described it as being potentially large.

2. Placing such meta information after the execution-semantic information could reduce latency to starting execution.

3. The data-segments section is optional (I've asked a few times and you have not refused this).

4. With the data-segments section optional, the runtime will not know if it is still to come after unknown sections so would need to continue waiting on the unknown sections and the end of the file before execution.

There is a technical use case of reducing latency, somewhat of core goal for wasm, and I would like it accommodated in the presence of source code meta information that helps make the code more accessible which I personally think is important.

My preference would be to add a mandatory sections directory which would address this use case by letting the runtime know early what sections follow from which it could determine when it had received all the execution-semantic information. This would also address uses cases that require random access into the binary encoding, without requiring a side-cache. But a section marker would also be an option that could be added now until a sections directory is added.

lukewagner:
Ah, I see your point now: in the case where *no* data section is present, an engine would have to consume until the end of the stream just to know it wasn't yet still to come.  That makes sense and that's a good argument for a mandatory end section (which wasn't actually the last section and could be followed by any number of necessarily-unknown sections).  I'll rebase and add that back to the PR.

JSStats:
Thank you, very much appreciate your understanding.

lukewagner:
Rebased and pushed with that and above request for rejecting out-of-order known sections.

JSStats:
If a module needs to be valid to be accepted and 'the end of the last present section must coincide with the last byte of the module' then might this block early execution? Might the runtime need to wait until the end of the file just to check that it ends on a section boundary before starting execution?

Would it be possible to define content after the `end` section as not affecting the validity of the module, rather invalid content after the `end` would just be silently ignored?

lukewagner:
That makes sense: if the point here is to allow execution to begin after "End" (while, e.g., "Names" is still being downloaded), then nothing after "End" can invalidate the whole module.

lukewagner:
Thinking about the practical implications a bit more:
* The Names section would be used to name functions in `Error` callstacks and other reporting paths so to avoid a really unpleasant race condition for developers, an engine *would* want to block execution on use of the Name section, when present.
* In the MVP, there's actually no real stream optimization opportunity b/c there are no streams, only an `ArrayBuffer` of code.
* Post-MVP when [Streams](http://streams.spec.whatwg.org/) are added, it will be programmatically observable (w/o any timing dependence) whether validation completes at the End section or at the end of the module, so we'd need to specify and, given the above, specify the latter.
* With a Streams API, content libraries should be able to strip unneeded sections themselves (and go even further than proposed already, aborting streams when they are no longer necessary).

So all this adds up, I think, to not having a before-the-end End section.

binji:
I was thinking about this some. When a section has many potential orderings because it has few current constraints (like memory), we'll want to place it earlier if we guess that future sections will depend on it.

For example, imagine we add a new section that depends on the memory section (perhaps a dynamic linking section), it will necessarily go after the all the sections that precede the memory section, even if it doesn't depend on them.

I guess what I'm saying is that it seems like we'll always be able to produce a topological sort (because no section can depend on a section that will be added later), but we won't necessary have a very good one unless we can reasonably predict future dependencies.

Maybe this doesn't matter too much? Or am I missing something?

JSStats:
@lukewagner Perhaps reporting messages are not part of the spec, so a runtime has discretion over waiting for the `names` section or ignoring it. I seems possible to spec content after the `end` section as not causing a validation error and to still allow runtimes the discretion to wait?

Even for MVP the content will stream in and it seem possible for a runtime to choose to end the download after the `end` and to even not cache the content to save storage on a limited device, and a cloud content optimizer could also use it to strip a file without knowledge of which unknown sections affect execution.

JSStats:
@lukewagner Another point, but getting a little off topic, is that my understanding of the 'stream' support is that it is a JS API in code that the web developer supplies and thus not matters that the browser vendor or the user have discretion in? I think this is a very significant distinction. Depending on a JS API to meet the 'meta-info-download/latency' use case would meet distinctly different use cases to it being under the control of the browser/user.

Further, it might be possible to articulate a design principle to the effect that the user being able to 'view source' should also give the user control over the download of the meta information to view the source, rather than it being at the discretion of the web app developer to enable or disable it on the client side (developers can still decide to include meta info or not on the server side).

lukewagner:
Oops, I let this drop while on vacation.

@binji That makes sense, but in the future we can always choose to loosen ordering requirements.  FWIW, I'm expecting the section structure, and possibly order, to change again in one of our upcoming binary updates, but as this currently matches the precise format accepted by FF (and, though perhaps not exclusively, accepted by Chakra/V8), any objections to merging the current ordering and dealing with further changes in future PRs?  @titzer @Cellule?

@JSStats In the MVP, the only APIs for compiling wasm will take `ArrayBuffer`s which means that, from the wasm decoder/engine's pov, there is no streaming.  The same functions which take `ArrayBuffer`s in the MVP would be extended to take stream in the future (at least the `Promise`-returning APIs would).  The stream read by the engine could come either from the network (e.g., [`fetch`](https://fetch.spec.whatwg.org/)) or from user user code via a [writable stream](https://streams.spec.whatwg.org/#ws-model).

titzer:
V8 doesn't check the exact order of sections yet, but we will when the
order is fully specified.

JSStats:
@lukewagner Personally I think the higher level design is fundamentally too web developer centric and too anti-user. When making any design decision I would have asked: can this be designed to avoid exposing information about the runtime to the web, and can it be designed to put the user in control rather than the web code. This principle is absent in the design and it shows up here too, with the requirement that buffer or streaming be via software that the web developer supplies, software that the user has to just trust and that likely does not support features the user would want.

Adding an 'end' section would have given the user the option to reduce latency, and to reduce storage requirements, and to avoid exposing information about the validity of the following sections.

The higher level design of wasm could have been organised around privileged code, code that the community can scrutinised, just as a web browser is trusted, and this could have implemented the streaming options etc.

So as it stands a defensive web browser can just fail to report validation errors in meta information sections, as if they were not supported, but the web developers code might probe for a validation error to check for support and believing there is no support decide not to include the meta information and the code will be incomplete and the user loses out having not choice.

lukewagner:
@titzer so, good to merge?

@JSStats Non-web native environments have strictly more options for optimizing load time and can do what the web does and more.  E.g., they can mmap the wasm file all at once and seek randomly around, using the section lengths to easily skip over ignored sections.  If you want to continue this discussion, it seems like a new issue is the right place since this PR is focused on defining a precise order and what you're talking about is changing the rules for how and when things are validated.

JSStats:
@lukewagner You are removing the 'end' section and not giving it an ordering, but I guess you are doing it in a context in which the meta data sections are validated, so I guess the discussion on this could be moved to a separate issue, sorry.

Just for the record, many of my points are web related. A file can't be mmapped if it has not been downloaded and the download latency can be significant on the web. A web app can not know if the user will need the meta information unless it prompts them or the browser exposes a personal setting to the app, and this does not seem a good approach to me and this is a web issue.

titzer:
lgtm

PR #599, opened 2016-03-10
Encodings for i32.eqz and i64.eqz.

sunfishcode:
This renames bool.not to i32.eqz and adds an i64.eqz encoding.


rossberg-chromium:
lgtm (I wonder whether we should eventually do some opcode clean-up before 1.0)

titzer:
lgtm

PR #608, opened 2016-03-13
Revise, update, and expand several Rationale entries.

sunfishcode:
Note that this removes the "The history of ```nop``` instructions"
sentence, because the long history of ```nop``` instructions is mostly
about uses that don't apply to wasm: alignment padding, hazard avoidance,
vliw packing, branch delay slot stuffing, timing, and modifying code at
runtime. The one use that does apply to wasm is already described in the
preceding paragraph.

### DIFF
--- a/Rationale.md
+++ b/Rationale.md
@@ -232,7 +232,7 @@ register allocation algorithms, offloading some of the optimization work from
 the WebAssembly VM.
 
 
-## Variable-Length Argument Lists
+## Variable-Length Argument Lists ("varargs")
 
 C and C++ compilers are expected to implement variable-length argument lists by
 storing arguments in a buffer in linear memory and passing a pointer to the
--- a/Rationale.md
+++ b/Rationale.md
@@ -205,12 +205,13 @@ variables by creating a separate stack data structure within linear memory. This
 stack is sometimes called the "aliased" stack, since it is used for variables
 which may be pointed to by pointers.
 
-This obstructs WebAssembly from performing clever optimizations on the stack and
-computing precise liveness of such variables, but this loss isn't expected to be
-consequential. Common compiler optimizations such as LLVM's global value
-numbering effectively split address-taken variables into parts, shrinking the
-range where they actually need to have their address taken, and creating new
-ranges where they can be allocated as local variables.
+Since the aliased stack appears to the WebAssembly engine as normal memory,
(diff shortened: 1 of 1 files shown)


Rationale.md
```diff
@@ -162,9 +173,9 @@ variables by creating a separate stack data structure within linear memory. This
 stack is sometimes called the "aliased" stack, since it is used for variables
 which may be pointed to by pointers.
 
-This prevents WebAssembly from performing clever optimizations on the stack and
-liveness of such variables, but this loss isn't expected to be
-consequential. Common C compiler optimizations such as LLVM's global value
+This obstructs WebAssembly from performing clever optimizations on the stack and
```
titzer: Can you reword this paragraph to be more direct? Something like:

Since the aliased stack appears to the WebAssembly engine as normal memory, WebAssembly optimizations that would target the aliased stack need to be more general, and thus more complicated. We observe that common compiler optimizations on the producer side, such as LLVM's global value numbering, effectively split address-taken variables into many small ranges that can often be allocated as local variables. Thus our expectation that any loss of optimization potential here is minimal.

sunfishcode: I agree, it's much clearer that way. Updated.

Rationale.md
```diff
@@ -162,12 +173,13 @@ variables by creating a separate stack data structure within linear memory. This
...
-This prevents WebAssembly from performing clever optimizations on the stack and
-liveness of such variables, but this loss isn't expected to be
-consequential. Common C compiler optimizations such as LLVM's global value
-numbering effectively split address-taken variables into parts, shrinking the
-range where they actually need to have their address taken, and creating new
-ranges where they can be allocated as local variables.
+Since the aliased stack appears to the WebAssembly engine as normal memory,
+WebAssembly optimizations that would target the aliased stack need to be more
+general, and thus more complicated. We observe that common compiler
+optimizations on the producer side, such as LLVM's global value numbering,
```
naturaltransformation: Do we ever define "producer"? It is also mentioned in AstSemantics.md but I don't think it is ever defined. It is pedantic but these kind of docs usually are. Perhaps it should be defined in README.md or a separate Terminology doc.

sunfishcode:
I reworded the paragraph about "producers" to hopefully be more understandable.

titzer:
lgtm

kripken:
lgtm

PR #615, opened 2016-03-18
Fix feature testing

### LINKED ISSUE #613: has_feature is referenced but not defined
Dead link here: https://github.com/WebAssembly/design/blob/master/FeatureTest.md#feature-test.  Also referenced once or twice elsewhere in the repo but does not seem to be defined anywhere.

jfbastien:
It was removed in #523, but was still referenced. Leave it as a high-level sketch for now.

Fixes #613.


jfbastien:
Assuming this is good for now. We can iterate on further fixes, this PR at least makes it correct.

Issue #515, opened 2016-01-12
signature restriction for imports


MikeHolman:
With no way to map `i64` into JavaScript, I think we should restrict `call_import` to disallow signatures containing `i64` types, at least for the web platform.

Also, when calling an import with integer type, we should define how it is interpreted. In asm.js we force FFI calls to be signed, I think interpreting as signed should be fine for wasm as well?

titzer:
That sounds OK to me. V8 truncates int64 to 32-bits, but forbidding them is
even better.

kg:
> With no way to map i64 into JavaScript, I think we should restrict call_import to disallow signatures containing i64 types, at least for the web platform.

ECMAScript will have i64/u64 eventually though. Would we want this restriction to hold forever? Or just restrict it so that it's invalid when loading the module in a runtime without i64/u64?

sunfishcode:
Me too. JavaScript is very likely to gain real support for 64-bit integers in the future; when it does, we can allow wasm imports to use them, and then we'll be in a better place than we'd be if we implement some temporary workaround now that we'd be stuck supporting.

kripken:
I agree on both: sounds good to disallow i64 imports for now, and also to make i32s interpreted as signed.

jfbastien:
Can we disallow i64 until ECMAScript supports it, but agree that toolchains should just expose two i32s for now for user's convenience? That's easy for parameters, but return values are tricky.

sunfishcode:
Exposing something for convenience now would be setting ourselves up with an annoying compatibility constraint in the future. Unless there's a clear need otherwise, it seems better to keep things simple.

kg:
Upon more thought about this, I suspect if we disallow i64 for now, we will end up with some syscalls that pass pairs of i32s across the wire anyway. Maybe it's better to wait until it happens, but we should consider how we are going to address that when it does... I feel like that will become a de-facto ABI for wasm<->JS either way.

Once we *do* have i64 in JS, how much performance would we be leaving on the table if syscalls continued to split those i64s into pairs of i32's? Could we safely feature-detect at runtime and drop the splitting? Is it even worth being concerned about the performance of syscalls with i64 arguments?

lukewagner:
+1 for "temporarily disallow until JS has real int64s`".  ISTR this being discussed somewhere else with the same conclusion but I can't find it now.

[...]

PR #618, opened 2016-03-21
added note i64 values in web embedded

wanderer:


Web.md
```diff
@@ -57,6 +57,11 @@ would inhibit export. In general, `ArrayBuffer` could be viewed as an
 optimization of copy in/out that was only available when linear memory
 behaved like an `ArrayBuffer` (or `SharedArrayBuffer`).
 
+## 64-bit integers
+Currently Javascript doesn't support 64-bit itergers so they will be disallowed
```
jfbastien: "integers"

Web.md
```diff
@@ -57,6 +57,11 @@ would inhibit export. In general, `ArrayBuffer` could be viewed as an
 optimization of copy in/out that was only available when linear memory
 behaved like an `ArrayBuffer` (or `SharedArrayBuffer`).
 
+## 64-bit integers
+Currently Javascript doesn't support 64-bit integers so they will be disallowed
+as a return value in JS binding. Once JS has [Typed Objects](http://wiki.ecmascript.org/doku.php?id=harmony:typed_objects)
```
titzer: I think you should explicitly mention imports and exports and parameter and return types.

Web.md
```diff
@@ -57,6 +57,12 @@ would inhibit export. In general, `ArrayBuffer` could be viewed as an
 optimization of copy in/out that was only available when linear memory
 behaved like an `ArrayBuffer` (or `SharedArrayBuffer`).
 
+## 64-bit integers
+
+Currently Javascript doesn't support 64-bit integers so they will be disallowed
+as parameters in imports and exports and in return values in the JS bindings. Once JS has [Typed Objects](http://wiki.ecmascript.org/doku.php?id=harmony:typed_objects)
```
wanderer: @titzer is this accurate?

titzer: You can probably just drop the Typed Objects reference.

wanderer: ok updated

sunfishcode:
Also related: https://github.com/WebAssembly/design/issues/515

flagxor:
So I believe the intention was that i64 be allowed in imports + exports, but fail if called from JS. This allows hooking up cross module calls with i64 args.

This should be clarified in the docs. Will do so soon (updated #515 to reflect this).

PR #628, opened 2016-03-29
Add postorder opcodes

titzer:
Add opcodes for `else` and `end`. Remove the `if_else` opcode. Remove counts from blocks and loops.

Note also that this PR slightly rearranges the lower opcode space to put all the control operators in order and leave a little hole between `0x0a` and `0x0f` for adding a `catch` operator in the future.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -330,6 +330,7 @@ It is legal to have several entries with the same type.
 | `end` | `0x0f` | | end a block, loop, or if |
 
 The `br_table` operator has an immediate operand which is encoded as follows:
+Note that there is no explicit `if_else` opcode, as the else clause is encoded with the `else` bytecode.
 
 | Field | Type | Description |
 | ---- | ---- | ---- |
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -319,14 +319,15 @@ It is legal to have several entries with the same type.
 | `nop` | `0x00` | | no operation |
 | `block` | `0x01` | count = `varuint32` | a sequence of expressions, the last of which yields a value |
 | `loop` | `0x02` | count = `varuint32` | a block which can also form control flow loops |
-| `if` | `0x03` | | high-level one-armed if |
-| `if_else` | `0x04` | | high-level two-armed if |
+| `if` | `0x03` | | begin if expression |
+| `else` | `0x04` | | begin else expression of if |
 | `select` | `0x05` | | select one of two values based on condition |
 | `br` | `0x06` | relative_depth = `varuint32` | break that targets a outer nested block |
 | `br_if` | `0x07` | relative_depth = `varuint32` | conditional break that targets a outer nested block |
 | `br_table` | `0x08` | see below | branch table control flow construct |
 | `return` | `0x14` | | return zero or one value from this function |
 | `unreachable` | `0x15` | | trap immediately |
+| `end` | `0x17` | | end a block, loop, or if |
 
 The `br_table` operator has an immediate operand which is encoded as follows:
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -317,17 +317,17 @@ It is legal to have several entries with the same type.
 | Name | Opcode | Immediate | Description |
 | ---- | ---- | ---- | ---- |
 | `nop` | `0x00` | | no operation |
-| `block` | `0x01` | count = `varuint32` | a sequence of expressions, the last of which yields a value |
-| `loop` | `0x02` | count = `varuint32` | a block which can also form control flow loops |
+| `block` | `0x01` |  | begin a sequence of expressions, the last of which yields a value |
+| `loop` | `0x02` |  | begin a block which can also form control flow loops |
 | `if` | `0x03` | | begin if expression |
 | `else` | `0x04` | | begin else expression of if |
 | `select` | `0x05` | | select one of two values based on condition |
 | `br` | `0x06` | relative_depth = `varuint32` | break that targets a outer nested block |
 | `br_if` | `0x07` | relative_depth = `varuint32` | conditional break that targets a outer nested block |
 | `br_table` | `0x08` | see below | branch table control flow construct |
-| `return` | `0x14` | | return zero or one value from this function |
-| `unreachable` | `0x15` | | trap immediately |
-| `end` | `0x17` | | end a block, loop, or if |
+| `return` | `0x09` | | return zero or one value from this function |
+| `unreachable` | `0x0a` | | trap immediately |
+| `end` | `0x0f` | | end a block, loop, or if |
 
 The `br_table` operator has an immediate operand which is encoded as follows:
 
@@ -344,15 +344,15 @@ out of range, `br_table` branches to the default target.
 ## Basic operators ([described here](AstSemantics.md#constants))
 | Name | Opcode | Immediate | Description |
 | ---- | ---- | ---- | ---- |
-| `i32.const` | `0x0a` | value = `varint32` | a constant value interpreted as `i32` |
-| `i64.const` | `0x0b` | value = `varint64` | a constant value interpreted as `i64` |
-| `f64.const` | `0x0c` | value = `uint64` | a constant value interpreted as `f64` |
-| `f32.const` | `0x0d` | value = `uint32` | a constant value interpreted as `f32` |
-| `get_local` | `0x0e` | local_index = `varuint32` | read a local variable or parameter |
-| `set_local` | `0x0f` | local_index = `varuint32` | write a local variable or parameter |
-| `call` | `0x12` | function_index = `varuint32` | call a function by its index |
-| `call_indirect` | `0x13` | signature_index = `varuint32` | call a function indirect with an expected signature |
-| `call_import` | `0x1f` | import_index = `varuint32` | call an imported function by its index |
+| `i32.const` | `0x10` | value = `varint32` | a constant value interpreted as `i32` |
+| `i64.const` | `0x11` | value = `varint64` | a constant value interpreted as `i64` |
+| `f64.const` | `0x12` | value = `uint64` | a constant value interpreted as `f64` |
+| `f32.const` | `0x13` | value = `uint32` | a constant value interpreted as `f32` |
+| `get_local` | `0x14` | local_index = `varuint32` | read a local variable or parameter |
+| `set_local` | `0x15` | local_index = `varuint32` | write a local variable or parameter |
+| `call` | `0x16` | function_index = `varuint32` | call a function by its index |
+| `call_indirect` | `0x17` | signature_index = `varuint32` | call a function indirect with an expected signature |
+| `call_import` | `0x18` | import_index = `varuint32` | call an imported function by its index |
 
 ## Memory-related operators ([described here](AstSemantics.md#linear-memory-accesses))
 


BinaryEncoding.md
```diff
@@ -317,18 +317,20 @@ It is legal to have several entries with the same type.
...
+| `else` | `0x04` | | begin else expression of if |
 | `select` | `0x05` | | select one of two values based on condition |
 | `br` | `0x06` | relative_depth = `varuint32` | break that targets a outer nested block |
 | `br_if` | `0x07` | relative_depth = `varuint32` | conditional break that targets a outer nested block |
 | `br_table` | `0x08` | see below | branch table control flow construct |
-| `return` | `0x14` | | return zero or one value from this function |
-| `unreachable` | `0x15` | | trap immediately |
+| `return` | `0x09` | | return zero or one value from this function |
+| `unreachable` | `0x0a` | | trap immediately |
+| `end` | `0x0f` | | end a block, loop, or if |
```
sunfishcode: In my postorder design, `if` has its own `endif` opcode. `block` and `loop` ends have an arity immediate, while in `if`+`else`+`endif` the arity immediate goes on the `if` to avoid being redundant between the `else` and the `endif`. Does your design omit arity immediates on `end` markers, and if so, how do you determine what to leave on the stack after a block exit?

JSStats: A new local AST node stack is created with the start of each effective block (including the `if` branches) and at the end of the block all the remaining stack elements are the block top level expressions of the effective block so an immediate expression count is not needed.

BinaryEncoding.md
```diff
@@ -344,15 +344,15 @@ out of range, `br_table` branches to the default target.
...
-| `call_import` | `0x1f` | import_index = `varuint32` | call an imported function by its index |
+| `i32.const` | `0x10` | value = `varint32` | a constant value interpreted as `i32` |
+| `i64.const` | `0x11` | value = `varint64` | a constant value interpreted as `i64` |
+| `f64.const` | `0x12` | value = `uint64` | a constant value interpreted as `f64` |
+| `f32.const` | `0x13` | value = `uint32` | a constant value interpreted as `f32` |
+| `get_local` | `0x14` | local_index = `varuint32` | read a local variable or parameter |
+| `set_local` | `0x15` | local_index = `varuint32` | write a local variable or parameter |
+| `call` | `0x16` | function_index = `varuint32` | call a function by its index |
+| `call_indirect` | `0x17` | signature_index = `varuint32` | call a function indirect with an expected signature |
+| `call_import` | `0x18` | import_index = `varuint32` | call an imported function by its index |
```
JSStats: With the plan to add an operator table there seems little need to be re-organising now or plan for a `catch` operator, but I guess it's a small matter for the few tools to update these. Or are you anticipating mapping the operators back to indexes internally for implementations and planning a suggested numbering for these.

BinaryEncoding.md
sunfishcode: Are you saying that the end of the block unconditionally pushes the last AST node of the block onto the AST node stack? This seems inconsistent with the arity-immediate used in `br` and `br_if`. If the arity is 0, they have no result-value operands, rather than just unconditionally having one operand.

titzer: On Tue, Mar 29, 2016 at 4:33 PM, Dan Gohman <notifications@github.com>
wrote:

> In BinaryEncoding.md
> <https://github.com/WebAssembly/design/pull/628#discussion_r57733287>:
>
> >  | `select` | `0x05` | | select one of two values based on condition |
> >  | `br` | `0x06` | relative_depth = `varuint32` | break that targets a outer nested block |
> >  | `br_if` | `0x07` | relative_depth = `varuint32` | conditional break that targets a outer nested block |
> >  | `br_table` | `0x08` | see below | branch table control flow construct |
> > -| `return` | `0x14` | | return zero or one value from this function |
> > -| `unreachable` | `0x15` | | trap immediately |
> > +| `return` | `0x09` | | return zero or one value from this function |
> > +| `unreachable` | `0x0a` | | trap immediately |
> > +| `end` | `0x0f` | | end a block, loop, or if |
>
> In my postorder design, if has its own endif opcode. block and loop ends
> have an arity immediate, while in if+else+endif the arity immediate goes
> on the if to avoid being redundant between the else and the endif. Does
> your design omit arity immediates on end markers, and if so, how do you
> determine what to leave on the stack after a block exit?
>
Yes, I omit the arities, so that when you fall off the end of the block,
you can have at most one return value (i.e. the last value on the stack). I
think that's sufficient; if you want to use a multi-value block in the
future, you have to use a br with arity. (And I'm still not convinced that
multi-arity blocks and ifs will be a thing in the future).

JSStats: At the end of a block the last AST node on that blocks AST node stack is a result node of the block and the other AST nodes are the nodes with discarded expressions - if decoding into an AST tree then they are the AST nodes of the block and the block operators do not affect the number of child expressions that a block has.

The semantics of the arity-immediate used in the break operators is a matter I question! If interpreted as zero representing a single argument `nop` and 1 representing the values of a single expression then it works just fine. In the case of 0 there is an AST node representing zero values and in the case of 1 there is an AST node for the expression. But these do not affect the number of child nodes of a block which I presumed was what your 'arity immediate' encoded, or was it a type declaration for the block?

JSStats: @titzer I see it differently. We already have the case of zero-values for `nop` making it clear that it is not a values stack. If there were multiple value expressions and blocks passed through multiple values then each stack entry would represent an AST node and thus an expression and thus the results of that expression which could have multiple values or no values.

I think it would be subjectively nice for the following to be equivalent `(block (call $fn_returning_two_values)) == (call $fn_returning_two_values)` without having to do the following:

```
(block $l
  (destructuring_set_local (10 11) (call $fn_returning_two_values))
  (br_with_atomic_tuple $l (get_local 10) (get_local 11)))
```

Also to be able to write `(block ... (tuple (get_local 10) (get_local 11))` rather than having to use an atomic `br_with_atomic_tuple`.

Also for the following to be equivalent  `(call $fn_returning_two_values) == (block (call $fn_returning_two_values)) == (block $l (br $l (call $fn_returning_two_values)))`.

kripken: I think I agree with what @JSStats said about not needing the arity, but unless I'm missing something, it doesn't limit us to 0/1? Can't a block etc. push a "block starts here" element on the expression stack, and then when we hit the block end, everything in between belongs to the block? That is, if before the block the stack has

`A B C`

then the block begins, and at the end of the block we have

`A B C block-start D E F block-end`

at which point we create a block with `D E F` as its 3 elements, and the stack is once more `A B C` as it was before the block.

JSStats: @kripken It's possible we a talking about different aspects of the 'stack' because the arity of the break operators seems to have nothing to do with the block AST encoding, so it is not clear where this '0/1' issue is entering the discussion. In your example at the end of the block an AST node representing the block would be pushed on the node stack so it would be `A B C (block D E F)`. Consider the following example of an encoding and its decoding:

```
(block $l1
  (i32.const 1)
  (i32.const 2)
  (block $l2
    (i32.const 3)
    (i32.const 4)
    (if (i32.const 5)
	(then (br $l2))
	(else (br $l2 (i32.const 6))))
    (call $fn_returning_two_values)
    (nop))
  (nop))

block_start $l1
  i32.const 1      Stack $l1: (i32.const 1)
  i32.const 2      Stack $l1: (i32.const 1) (i32.const 2)
  block_start $l2
    i32.cont 3     Stack $l1: (i32.const 1) (i32.const 2) Stack $l2: (i32.const 3)
    i32.cont 4     ... Stack $l2: (i32.const 3) (i32.const 4)
    i32.cont 5     ... Stack $l2: (i32.const 3) (i32.const 4) (i32.const 5)
    if_start       ... Stack $l2: (i32.const 3) (i32.const 4) (if (i32.const 5) (then) (else)) Stack then:
      br $l2       ... Stack then: (br $l2 (nop))
    else           ... Stack $l2: (i32.const 3) (i32.const 4) (if (i32.const 5) (then (br $l2 (nop))) (else)) Stack else:
      i32.const 6  ... Stack else: (i32.const 6)
    end_if         ... Stack $l2: (i32.const 3) (i32.const 4) (if (i32.const 5) (then (br $l2 (nop))) (else (i32.const 6)))
    call $fn_returning_two_value  ... Stack $l2: (i32.const 3) (i32.const 4) (if (i32.const 5) (then (br $l2 (nop))) (else (i32.const 6))) (call $fn_returning_two_values)
    nop            ... Stack $l2: (i32.const 3) (i32.const 4) (if (i32.const 5) (then (br $l2 (nop))) (else (i32.const 6))) (call $fn_returning_two_values) (nop)
  block_end        Stack $l1 (i32.const 1) (i32.const 2) (block $l2 (i32.const 3) (i32.const 4) (if (i32.const 5) (then (br $l2 (nop))) (else (i32.const 6))) (call $fn_returning_two_values) (nop))
  nop              Stack $l1 (i32.const 1) (i32.const 2) (block $l2 (i32.const 3) (i32.const 4) (if (i32.const 5) (then (br $l2 (nop))) (else (i32.const 6))) (call $fn_returning_two_values) (nop)) (nop)
block_end          result: (block $l1 (i32.const 1) (i32.const 2) (block $l2 (i32.const 3) (i32.const 4) (if (i32.const 5) (then (br $l2 (nop))) (else (i32.const 6))) (call $fn_returning_two_values) (nop)) (nop))
```

The arity of the break operators do not seem to affect the decoding, nor did a call returning multiple values. AST nodes and expressions are always represented by a single AST node irrespective of the number of values the expression returns. If I have missed some concern then an example might help?

kripken: Sorry if I wasn't clear enough before. Yes, I agree with you, I also don't see an issue with the 0/1 arities that @sunfishcode mentioned.

JSStats: @kripken Given the lack of response from @sunfishcode to help us understand his perspective I can only speculate he is trying to shoe-horn the AST into a single value stack machine where zero-value nodes yield no stack element when interpreted. Wasm would be restricted to only being able to consume zero-value expressions in elements of a block that discard their result values thus any zero-value expression seen within a block's node stack is a top-level block node and since these are discarded a value need not be pushed on the stack when interpreted. There would a separate `block1` operator that returns the last expression value, and a separate `block0` operator that returns no values, and while they could be two separate operators @sunfishcode might be suggesting they be distinguished by an immediate 0/1 but I think it would be needed on the block_start marker. The break operators could no long accept a zero-value expression as it is no longer possible to represent this so there would need to be a separate `break0` and it would be a validation error to use these with `block1`. A `breakn` accepting a single non-zero-value expression would be used with `block1` discarding all but the first value, and could also be used with `block0` discarding all the expression values. These two break operators might be distinguished using the 0/1 immediate. The `if` operator already has implicit blocks, and @sunfishcode appears to be suggesting enforcing them both as either being `block0` or `block1`. The `select` operator would no longer be able to accept zero-value expressions, but this does not seem a big constraint.

It would seem to still be necessary to push a zero-value node on the stack when validating or decoding back to an AST, but perhaps once validated the code might be interpreted literally without pushing a value on the stack for a zero-value node.

Scaling to multiple-values might relax `block1` to pass through expression values with one or more values in which case an interpreter needs to deal with the value count anyway so there seems to be little gain restricting the zero-value case?

Also I see mention of plans for wasm to support more general types, I presume a struct that could be empty or have one slot or two slots etc, just as a zero-value or one-value or two-value etc expression would have, so an interpreted needs to deal with these variable sized stack elements anyway and would need to allocate variable size stack elements and track their types. In this context the zero-value restriction seems out of place to me.

Perhaps the intention is that `block1` consume a single value from the fall-through expression and the break expressions and only pass on this single value, but it could still accept a multiple value expression. Alternatively, a `block2` `block3` etc might be added to pass on more than one value, otherwise it would require multiple value expressions to be passed around in local variables and the loss of efficient coding would be the same as if blocks and `if` did not return any values.

CIL failed to plan ahead for multiple-value support and look where it has lead https://github.com/dotnet/roslyn/issues/347 using struct to represent them which is less efficient from an AST perspective. For example the first value can not be trivially consumed discarding the rest etc.

I think people need to give up on trying to shoe horn wasm into an interpreted single value stack machine (JVM style), and to focus on the efficient single pass validation and SSA transform after which it is represented by single value typed definitions. Actually I wonder if expressions were removed from wasm and the SSA transform were followed by a copy-propagation pass if the same SSA code would be obtained? If so then the AST might be view as just a efficient way to represent this.

sunfishcode: @JSStats I have an implementation of a postorder encoder and decoder using one-value-per-stack-entry, and it: works well, has an elegant path to supporting multi-value expressions (by pushing each value on the stack separately), and is indeed focused on single-pass validation and SSA construction (stack entries represent SSA definitions -- they literally hold an MDefinition* in SM).

Instead of having seperate `break0` and `break1`, my design uses the arity immediate proposal, but with a value-oriented interpretation: instead of specifying how many AST children are present, the arity field specifies how many values to expect. From an SSA construction perspective, actual values are what matter, and `void` "values" don't contribute to the task at hand.

JSStats: @sunfishcode You can have your stack of defs and avoid all the zero-value restrictions by simply pushing an integer count of the number of defs per AST node on the stack after the values - this would just be an implementation detail of how SM represents nodes during decoding. For example:

```
(nop) -> stack: 0
(i32.const 55) -> stack: (i32.const 55) 1
(block (nop) (i32.const 55)) -> (nop) 0 (i32.const 55) 1
```

This scales to more than one value
```
(tuple (i32.const 55) (i32.const 56)) -> stack: (i32.const 55) (i32.const 56) 2
```

And allows:
```
(i32.eqz (call $fn_returning_two_values_i32_f64))
```

That does not seem a big burden to avoid all the AST expressiveness restrictions and bloating the encoding with more annotations etc. Would you have any concerns with this solution?

sunfishcode: My concern would be that these extra counts would require extra logic to encode on the stack and to validate, and it's not clear to me what they contribute.

If your concern is dropping individual values of a multi-value expression, this can be implemented with or without these count entries.

JSStats: Here is an example of invalid code generation without the counts:
```
(block (i32.add (i32.const 55) (call $fn_returning_two_values_55_56)))

block_start
  (i32.const 55)     block_stack: 51
  call $fn_returning_two_values_55_56   block_stack: 51 56 55
  i32.add            block_stack: 51 111
block_end            parent_stack: 111
```

titzer:
This should target binary 11 here: #641

PR #630, opened 2016-03-31
if and if_else are implicit blocks

wanderer:
as found out here https://github.com/WebAssembly/spec/issues/270


titzer:
lgtm

sunfishcode:
The text of (what is now) Semantics.md has been significantly rewritten in this area, and already incorporates the change proposed here. Please re-open or file a new issue if there are any further clarifications desired.

PR #632, opened 2016-03-31
Clarify physical page exhaustion

jfbastien:
As discussed in https://github.com/WebAssembly/design/pull/629#discussion_r58138148

### DIFF
--- a/Nondeterminism.md
+++ b/Nondeterminism.md
@@ -45,6 +45,9 @@ currently admits nondeterminism:
      may vary between platforms.
  * Environment-dependent resource limits may be exhausted. A few examples:
    - Memory allocation may fail.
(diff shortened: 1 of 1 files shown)


sunfishcode:
This seems like a special case of something we should perhaps state more clearly: resources can become exhausted at literally any time, for any number of reasons.

jfbastien:
Indeed, but it's a non-obvious corner case that #629 makes yet-more confusing IMO by discussing virtual reservation guarantees but not physical reservation guarantees (there are no guarantees ;-).

Could we handle your suggestion in a separate PR?

sunfishcode:
Ah, I was missing some context for this PR at first. This lgtm.

PR #635, opened 2016-04-05
Prettify section names

rossberg-chromium:
As suggested in #623. Better suggestions for some of the names are welcome.

PR #638, opened 2016-04-05
Prettify section names

rossberg-chromium:
[Retargeted #635 to 0xB branch.]

As suggested in #623. Better suggestions for some of the names are welcome.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -92,42 +92,42 @@ for all sections. The encoding of all sections begins as follows:
 Each section is optional and may appear at most once.
 Known sections (from this list) may not appear out of order.
 
-* [Signatures](#signatures-section) section
-* [Import Table](#import-table-section) section
-* [Function Signatures](#function-signatures-section) section
-* [Indirect Function Table](#indirect-function-table-section) section
+* [Type](#type-section) section
+* [Import](#import-section) section
+* [Function](#function-section) section
+* [Indirection](#indirection-section) section
 * [Memory](#memory-section) section
-* [Export Table](#export-table-section) section
-* [Start Function](#start-function-section) section
-* [Function Bodies](#function-bodies-section) section
-* [Data Segments](#data-segments-section) section
+* [Export](#export-section) section
+* [Start](#start-section) section
+* [Code](#code-section) section
+* [Data](#data-section) section
 * [Names](#names-section) section
 
 The end of the last present section must coincide with the last byte of the
 module. The shortest valid module is 8 bytes (`magic number`, `version`,
 followed by zero sections).
 
-### Signatures section
+### Type section
 
-ID: `signatures`
+ID: `types`
 
-The signatures section declares all function signatures that will be used in the module.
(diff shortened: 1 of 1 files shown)


BinaryEncoding.md
```diff
@@ -145,29 +145,29 @@ The import section declares all imports that will be used in the module.
...
+module (their definitions appear in the [code section](#code-section)).
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
 | count | `varuint32` | count of signature indices to follow |
-| signatures | `varuint32*` | sequence of indices into the Signature section |
+| types | `varuint32*` | sequence of indices into the type section |
 
-### Indirect Function Table section
+### Indirection section
```
lukewagner: How about "table"?

sunfishcode:
"Type": Are there plans to use this section for types other than function types? If not, "types" is over-general and confusing, and leaving it as "Signatures" would be better.

"Function": How about "Declaration"?

"Code", "Data": I prefer "Function Bodies" and "Data Segments". Characters are cheap in this context, and being more descriptive helps newcomers as well as people who have focuses besides the wasm binary format.

lukewagner:
I'm not sure these strings, whatever their value, will appreciably help anyone stepping through the binary in a hex editor.  So I like the terseness (and lower case).

rossberg-chromium:
@sunfishcode, yes, we anticipate other uses of the types section, e.g. consider user-defined struct types in the future. This ties in with #640. "Declarations": hm, not sure, as most other sections declare things, too. Re descriptiveness: agreed with Luke.

@lukewagner, changed `indirect` to `table`. I thought that may be a bit too generic a name, but with your recent suggestions, it actually makes a lot of sense. (Although in that case, it should probably be `tables`. :)

lukewagner:
Heh, true; that PR would add an `s`.  lgtm thanks

rossberg-chromium:
@lukewagner, or perhaps we should just use singular for all section ids, like we do in prose?

lukewagner:
True; the 's' isn't really adding a lot of value.

rossberg-chromium:
Changed to singular.

rossberg-chromium:
Any objections landing this?

titzer:
lgtm

BinaryEncoding.md
```diff
@@ -92,42 +92,42 @@ for all sections. The encoding of all sections begins as follows:
...
+* [Table](#table-section) section
 * [Memory](#memory-section) section
-* [Export Table](#export-table-section) section
-* [Start Function](#start-function-section) section
-* [Function Bodies](#function-bodies-section) section
-* [Data Segments](#data-segments-section) section
-* [Names](#names-section) section
+* [Export](#export-section) section
+* [Start](#start-section) section
+* [Code](#code-section) section
```
jfbastien: Why not `text`? ;-)

jfbastien:
lgtm

lukewagner:
lgtm

PR #639, opened 2016-04-05
Reorder section size field

rossberg-chromium:
[Retargeted from #634.]

Put the size of a section after its name, as suggested in #623.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -81,16 +81,18 @@ The module starts with a preamble of two fields:
 This preamble is followed by a sequence of sections. Each section is identified by an
 immediate string. Sections whose identity is unknown to the WebAssembly
 implementation are ignored and this is supported by including the size in bytes
-for all sections. The encoding of all sections begins as follows:
+for all sections. The encoding of sections is structured as follows:
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
-| size  | `varuint32` | size of this section in bytes, excluding this size |
 | id_len | `varuint32` | section identifier string length |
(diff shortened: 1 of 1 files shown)


BinaryEncoding.md
```diff
@@ -81,16 +81,18 @@ The module starts with a preamble of two fields:
...
 | ----- |  ----- | ----- |
-| size  | `varuint32` | size of this section in bytes, excluding this size |
 | id_len | `varuint32` | section identifier string length |
 | id_str | `bytes` | section identifier string of id_len bytes |
+| payload_len  | `varuint32` | size of this section in bytes |
+| payload_str  | `bytes` | content of this section, of length payload_len |
 
 Each section is optional and may appear at most once.
 Known sections (from this list) may not appear out of order.
+The actual content of each section is encoded in its `payload_str`.
```
naturaltransformation: Nit: Instead of "actual content", can we just call it payload or "body"?

rossberg-chromium: Hm, "The payload is in the payload_str" would seem kind of vacuous; "body" would be okay. But what's wrong with (information) "content"?

naturaltransformation: The nit is more with "actual" than content.

rossberg-chromium: Fair enough, removed the "actual".

rossberg-chromium:
Objections to this?

jfbastien:
Do we want to do this change at the same time as postorder? Seems fine.

lukewagner:
lgtm.  Our assumption was indeed that this change was in 0xb.

PR #640, opened 2016-04-05
Extensible encoding of function signatures

rossberg-chromium:
[Retargeted #636.]

Better support future type system extensions with

- new forms of type definitions,
- functions with multiple return values,
- potential mixture of base type constructors (i32..f64) and structured types in common positions, by avoiding overlap in their encoding.

Motivation: when moving to a richer type language, you'll need a proper AST for type expressions. Currently, our grammar is simple:

value_type ::= i32 | i64 | f32 | f64
structural_type ::= value_type* -> value_type?
type_id ::= uint32

where structural types are the ones defined in the type section, and type_ids are references to these. In particular, there is no overlap between the different syntactic classes, they are always used in distinct places.

Now, when we grow the language, it becomes highly conceivable that (some of) this separation no longer applies, and that some of the phrases grow additional alternatives. For example, in a potential extension with both struct and function pointers, we might have the following:

value_type ::= i32 | i64 | f32 | f64 | type_id
structural_type ::= value_type* -> value_type?  | {value_type*}
type_id ::= int32

It's also conceivable that we will want to allow nested type expressions at some point, or not require naming every structural type, in which case we'd get something like

value_type ::= i32 | i64 | f32 | f64 | type_id | structural_type
structural_type ::= value_type* -> value_type?  | {value_type*}
type_id ::= int32

This PR simply ensures that the encoding of the current grammar is future-compatible with such potential extensions, by not baking in assumptions about the size or disjointness of syntactic classes: the production "value_type ::= type_id" could be encoded with another opcode for type references (followed by an index immediate), struct types with another opcode for the struct type constructor; the embedding of structural types into value types would require no extra opcode if we keep their index spaces disjoint (which is a design choice with zero cost).

A variation would be to overlay "type constructor" and "type id" opcode/index space e.g. by signedness, similar to what Luke suggests. That would save introducing an extra opcode for type references in the future. Regardless, I'd still propose to keep primitive and structural type constructor spaces disjoint.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -92,42 +92,42 @@ for all sections. The encoding of all sections begins as follows:
 Each section is optional and may appear at most once.
 Known sections (from this list) may not appear out of order.
 
-* [Signatures](#signatures-section) section
-* [Import Table](#import-table-section) section
-* [Function Signatures](#function-signatures-section) section
-* [Indirect Function Table](#indirect-function-table-section) section
+* [Type](#type-section) section
+* [Import](#import-section) section
+* [Function](#function-section) section
+* [Indirection](#indirection-section) section
 * [Memory](#memory-section) section
-* [Export Table](#export-table-section) section
-* [Start Function](#start-function-section) section
-* [Function Bodies](#function-bodies-section) section
-* [Data Segments](#data-segments-section) section
+* [Export](#export-section) section
+* [Start](#start-section) section
+* [Code](#code-section) section
+* [Data](#data-section) section
 * [Names](#names-section) section
 
 The end of the last present section must coincide with the last byte of the
 module. The shortest valid module is 8 bytes (`magic number`, `version`,
 followed by zero sections).
 
-### Signatures section
+### Type section
 
-ID: `signatures`
+ID: `types`
 
-The signatures section declares all function signatures that will be used in the module.
+The type section declares all function signatures that will be used in the module.
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
 | count | `varuint32` | count of signature entries to follow |
-| entries | `signature_entry*` | repeated signature entries as described below |
+| entries | `type_entry*` | repeated type entries as described below |
 
-#### Signature entry
+#### Type entry
 | Field | Type | Description |
 | ----- |  ----- | ----- |
 | param_count | `varuint32` | the number of parameters to the function |
 | return_type | `value_type?` | the return type of the function, with `0` indicating no return type |
 | param_types | `value_type*` | the parameter types of the function |
 
-### Import Table section
+### Import section
 
-ID: `import_table`
+ID: `imports`
 
 The import section declares all imports that will be used in the module.
 
@@ -145,29 +145,29 @@ The import section declares all imports that will be used in the module.
 | function_len | `varuint32` | function string length |
 | function_str | `bytes` | function string of `function_len` bytes |
 
-### Function Signatures section
+### Function section
 
-ID: `function_signatures`
+ID: `functions`
 
-The Function Signatures section declares the signatures of all functions in the
-module.
+The function section _declares_ the signatures of all functions in the
+module (their definitions appear in the [code section](#code-section)).
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
 | count | `varuint32` | count of signature indices to follow |
-| signatures | `varuint32*` | sequence of indices into the Signature section |
+| types | `varuint32*` | sequence of indices into the type section |
 
-### Indirect Function Table section
+### Indirection section
 
-ID: `function_table`
+ID: `indirect`
 
-The indirect function table section defines the module's 
+The indirection section defines the module's 
 [indirect function table](AstSemantics.md#calls).
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
 | count | `varuint32` | count of entries to follow |
-| entries | `varuint32*` | repeated indexes into the function table |
+| entries | `varuint32*` | repeated indexes into the function section |
 
 ### Memory section
 
@@ -182,11 +182,11 @@ associated with the module.
 | max_mem_pages | `varuint32` | maximum memory size in 64KiB pages |
 | exported | `uint8` | `1` if the memory is visible outside the module |
 
-### Export Table section
+### Export section
 
-ID: `export_table`
+ID: `exports`
 
-The export table section declares all exports from the module.
+The export section declares all exports from the module.
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
@@ -200,34 +200,35 @@ The export table section declares all exports from the module.
 | function_len | `varuint32` | function string length |
 | function_str | `bytes` | function string of `function_len` bytes |
 
-### Start Function section
+### Start section
 
-ID: `start_function`
+ID: `start`
 
-The start function section declares the [start function](Modules.md#module-start-function).
+The start section declares the [start function](Modules.md#module-start-function).
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
 | index | `varuint32` | start function index |
 
-### Function Bodies section
+### Code section
 
-ID: `function_bodies`
+ID: `code`
 
-The Function Bodies section assigns a body to every function in the module.
-The count of function signatures and function bodies must be the same and the `i`th
-signature corresponds to the `i`th function body.
+The code section assigns a body to every function in the module.
+The count of function declared in the [function section](#function-section)
+and function bodies defined in this section must be the same and the `i`th
+declaration corresponds to the `i`th function body.
 
 | Field | Type |  Description |
 | ----- |  ----- |  ----- |  ----- |
 | count | `varuint32` | count of function bodies to follow |
 | bodies | `function_body*` | sequence of [Function Bodies](#function-bodies) |
 
-### Data Segments section
+### Data section
 
-ID: `data_segments`
+ID: `data`
 
-The data segments section declares the initialized data that should be loaded
+The data section declares the initialized data that should be loaded
 into the linear memory.
 
 | Field | Type | Description |
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -121,7 +121,7 @@ The type section declares all function signatures that will be used in the modul
 #### Type entry
 | Field | Type/Value | Description |
 | ----- |  ----- | ----- |
-| constructor | `0x05` | the function type constructor |
+| constructor | `0x40` | the function type constructor |
 | param_count | `varuint32` | the number of parameters to the function |
 | param_types | `value_type*` | the parameter types of the function |
 | return_count | `uint8` | the number of results from the function (0 or 1) |
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -119,11 +119,15 @@ The type section declares all function signatures that will be used in the modul
 | entries | `type_entry*` | repeated type entries as described below |
 
 #### Type entry
-| Field | Type | Description |
+| Field | Type/Value | Description |
 | ----- |  ----- | ----- |
+| constructor | `0x05` | the function type constructor |
 | param_count | `varuint32` | the number of parameters to the function |
-| return_type | `value_type?` | the return type of the function, with `0` indicating no return type |
 | param_types | `value_type*` | the parameter types of the function |
+| return_count | `uint8` | the number of results from the function (0 or 1) |
+| return_type | `value_type?` | the result type of the function (if return_count is 1) |
+
+(Note: In the future, this section may contain other forms of type entries as well.)
 
 ### Import section
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -223,7 +223,7 @@ The start section declares the [start function](Modules.md#module-start-function
 
 ID: `code`
 
-The code section assigns a body to every function in the module.
+The code section contains a body for every function in the module.
 The count of function declared in the [function section](#function-section)
 and function bodies defined in this section must be the same and the `i`th
 declaration corresponds to the `i`th function body.
@@ -237,7 +237,7 @@ declaration corresponds to the `i`th function body.
 
 ID: `data`
 
-The data section declares the initialized data that should be loaded
+The data section declares the initialized data that is loaded
 into the linear memory.
 
 | Field | Type | Description |
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -35,6 +35,9 @@ A four-byte little endian unsigned integer.
 ### varint32
 A [Signed LEB128](https://en.wikipedia.org/wiki/LEB128#Signed_LEB128) variable-length integer, limited to int32 values.
 
+### varuint1
+A [LEB128](https://en.wikipedia.org/wiki/LEB128) variable-length integer, limited to the values 0 or 1. `varuint1` values may contain leading zeros. (This type is mainly used for compatibility with potential future extensions.)
+
 ### varuint32
 A [LEB128](https://en.wikipedia.org/wiki/LEB128) variable-length integer, limited to uint32 values. `varuint32` values may contain leading zeros.
 
@@ -115,19 +118,19 @@ The type section declares all function signatures that will be used in the modul
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
-| count | `varuint32` | count of signature entries to follow |
+| count | `varuint32` | count of type entries to follow |
 | entries | `type_entry*` | repeated type entries as described below |
 
 #### Type entry
-| Field | Type/Value | Description |
+| Field | Type | Description |
 | ----- |  ----- | ----- |
-| constructor | `0x40` | the function type constructor |
+| form | `uint8` | `0x40`, indicating a function type |
 | param_count | `varuint32` | the number of parameters to the function |
 | param_types | `value_type*` | the parameter types of the function |
-| return_count | `uint8` | the number of results from the function (0 or 1) |
+| return_count | `varuint1` | the number of results from the function |
 | return_type | `value_type?` | the result type of the function (if return_count is 1) |
 
-(Note: In the future, this section may contain other forms of type entries as well.)
+(Note: In the future, this section may contain other forms of type entries as well, which can be distinguished by the `form` field.)
 
 ### Import section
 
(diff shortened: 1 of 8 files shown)


BinaryEncoding.md
```diff
@@ -92,42 +92,46 @@ for all sections. The encoding of all sections begins as follows:
...
 | count | `varuint32` | count of signature entries to follow |
-| entries | `signature_entry*` | repeated signature entries as described below |
+| entries | `type_entry*` | repeated type entries as described below |
 
-#### Signature entry
-| Field | Type | Description |
+#### Type entry
+| Field | Type/Value | Description |
 | ----- |  ----- | ----- |
+| constructor | `0x40` | the function type constructor |
```
lukewagner: [Re-asking my [question](https://github.com/WebAssembly/design/pull/636/files#r58563196) in the closed PR]

Instead of making an arbitrary-feeling reservation, what if the nullary constructors were positive and the user-defined constructors were negative and we used (signed) `varint32`, starting at -1 and going down?

kripken:
I must have missed something, what is a "function type constructor"?

BinaryEncoding.md
```diff
@@ -92,42 +92,46 @@ for all sections. The encoding of all sections begins as follows:
...
 | param_count | `varuint32` | the number of parameters to the function |
-| return_type | `value_type?` | the return type of the function, with `0` indicating no return type |
 | param_types | `value_type*` | the parameter types of the function |
+| return_count | `uint8` | the number of results from the function (0 or 1) |
+| return_type | `value_type?` | the result type of the function (if return_count is 1) |
+
+(Note: In the future, this section may contain other forms of type entries as well.)
```
naturaltransformation: So will the set of acceptable type entry kinds be tied to specific versions of the binary format (and WASM)? Specifically, WASM consumer implementations cannot just implement a subset of the constructors and claim to support a specific version of WASM. I am assuming that is the case, so that it is impossible to reach a type entry with an unknown constructor value and "type entries" can never be safely skipped.

naturaltransformation:
In this context, a constructor is essentially a tag (from a tagged union) so function type constructor is the tag and also a constructor for building function types (i.e., apply 0x40 to a record to build a function type). This is in contrast to type constructor and data constructor which construct types (and possibly type definitions I'm assuming) and values respectively. Would the term "tag" or "type entry kind" be slightly friendlier here?

kripken:
Thanks @naturaltransformation, but I'm afraid I understood maybe 5% of that :) Am I stupid? Or does this PR suddenly and unexpectedly add a completely new concept to the design repo? Neither the term "constructor" nor "tag" seem to appear previously in this repo?

I don't see what tagged unions have to do with wasm function types, which are simple "this is the return type, these are the argument types" type things?

What is a "record" here?

Finally, what does "build a function type" mean? And why do we need to "build" ones, instead of just the simple way of defining them we had so far (like `(param f64) (ret i32)`)?

Again, I might be stupid here, sorry if so.

naturaltransformation:
Heh, no, sorry about that. It is my fault. I guess I tried to explain jargon with other jargon. The change is to leave room for other kinds of type_entries. "Tagged unions" is in the sense of an implementation detail. The bottom line is that the "function type constructor" is just a tag to indicate that a given type_entry is a function type and not a typedef or some other unforeseen kind of type_entry. As least that is how I'm understanding it.

naturaltransformation:
As I understand it, this change does not impact how WASM types are used in WASM modules at all, only that the the binary format decoder now needs to accommodate potential additional kinds of type_entries.

kripken:
Oh, I see, so this is saying "this type is a simple function type", and eventually this field will be used to indicate whether a type is something else, when we have such things? Cool, thanks.

Perhaps "constructor", "tag", etc. would be surprising for other people as well? I am obviously the farthest thing from a type theorist, but probably other non-type theorists will read this document too ;)

BinaryEncoding.md
```diff
@@ -92,42 +92,46 @@ for all sections. The encoding of all sections begins as follows:
...
 | Field | Type | Description |
 | ----- |  ----- | ----- |
 | count | `varuint32` | count of signature entries to follow |
-| entries | `signature_entry*` | repeated signature entries as described below |
+| entries | `type_entry*` | repeated type entries as described below |
 
-#### Signature entry
-| Field | Type | Description |
+#### Type entry
+| Field | Type/Value | Description |
```
naturaltransformation: Instead of a Type/Value column, maybe we can leave it as a Type column and move the 0x40 value to Description to be consistent with the other fixed value fields such as the module header.

naturaltransformation:
Yeah, it would be nice to have more accessible terminology. I suppose it is a "type_entry type", but that has its own problems.

JSStats:
This still seems far from extensible, rather it requires binary format version changes when adding new type declaration formats. Just a thought but could the type definitions section be abstracted into their own AST, following a similar format to the function bodies, and use the planned operator table to define future type definition opcodes?

rossberg-chromium:
I extended the PR description with more detail about what this PR is trying to achieve and how, hopefully answering most of the questions that came up.

Also addressed some comments:

@kripken, renamed `constructor` field to `form` (of type), lacking a better name.

@lukewagner, changed the return_count to be a varuint1. ;)

@lukewagner, @titzer, keeping the index space of primitive and structural type constructors disjoint keeps the door open to perhaps allow inline use of (some?) structural types in the future (avoiding to have to name _every_ type). Not sure we ever want to do that, but given the size of the index space, I see no reason to preclude it prematurely (I can't imagine a universe where we will ever have more than 128 different type constructors, or even get close to that number).

@JSStats, laying the ground works for being able to grow type encodings into proper AST encodings is exactly what this PR tries to achieve, see the updated description.

BinaryEncoding.md
```diff
@@ -92,42 +95,46 @@ for all sections. The encoding of all sections begins as follows:
...
-| count | `varuint32` | count of signature entries to follow |
-| entries | `signature_entry*` | repeated signature entries as described below |
+| count | `varuint32` | count of type entries to follow |
+| entries | `type_entry*` | repeated type entries as described below |
 
-#### Signature entry
+#### Type entry
 | Field | Type | Description |
 | ----- |  ----- | ----- |
+| form | `uint8` | `0x40`, indicating a function type |
```
lukewagner: We could keep the disjointness of primitive and structural type constructors by giving the former monotonically increasing integers and the latter monotonically decreasing integers (starting by giving functions a `form` of `-1`).  So same aesthetic preference for avoiding arbitrary-feeling statements like "noone will ever need more than `0x3f` primitive type constructors", but different meaning for the negative index than in the previous comment.  If so then, the type would be a `varint32`.

BinaryEncoding.md
```diff
@@ -92,42 +95,46 @@ for all sections. The encoding of all sections begins as follows:
...
 | param_count | `varuint32` | the number of parameters to the function |
-| return_type | `value_type?` | the return type of the function, with `0` indicating no return type |
 | param_types | `value_type*` | the parameter types of the function |
+| return_count | `varuint1` | the number of results from the function |
```
lukewagner: hah, that works

BinaryEncoding.md
rossberg-chromium: I'm fine with using var(u)int. But the type space potentially indexes 3 sorts of things: primitive constructors, structural constructors, type ids. If we want to use the signedness overlay trick, than it seems much more beneficial to reserve that for distinguishing between ids and constructors in the future, so that type id references could be single byte (in which case we probably need to assign negative numbers to all constructors now). WDYT?

lukewagner: > But the type space potentially indexes 3 sorts of things: primitive constructors, structural constructors, type ids

I didn't understand this part of the OP: why would we want to add the complication of "inlining" certain structural constructors instead of just using a type id?

> If we want to use the signedness overlay trick, than it seems much more beneficial to reserve that for distinguishing
> between ids and constructors in the future

Yes, if we can rule out the third case as I'm asking above then the encoding could be pretty simple: if positive, it's a pritimive, if negative, it's a (negated) type-id.  But that'd be the encoding of a *value type*.  `form` is the encoding of a different set: the set of compound constructors.  So as observed earlier, it could completely overlap with both primitives *and* type-ids.  I can see the argument for keeping the encoding of compound ctors disjoint from that of primitive ctors (so that you can represent the set of all constructors with an int), but that seems to work just fine with giving the primitive ctors negative indices.

lukewagner: Oops, edit: s/primitive/compound/ in last sentence.

rossberg-chromium: >
> I didn't understand this part of the OP: why would we want to add the
> complication of "inlining" certain structural constructors instead of just
> using a type id?
>

E.g. to reduce the overhead of one-off uses of structural types. Or imagine
you want to emit a more complicated nested struct, then you wouldn't need
to separate out and name each level.

Perhaps the other way round is a more compelling scenario: you may want to
allow primitive constructors in type definitions. Or they'll start to mix
when we introduce type import/exports some day. Also, it's hard to predict
what other thing might come along and change the story (generics? who
knows).

I'm not suggesting that there currently is a concrete reason to join the
spaces. But it doesn't seem completely unlikely to arise later, and given
that the cost of keeping it an option is zero, why preclude it?

Yes, if we can rule out the third case as I'm asking above then the
> encoding could be pretty simple: if positive, it's a pritimive, if
> negative, it's a (negated) type-id.
>

I'd actually invert that scheme, to avoid the negation.

lukewagner: What I don't understand is why all those future new things can't just be new `form`s of entries in the types section such that a type-id is all you need?

> I'd actually invert that scheme, to avoid the negation.

Wouldn't that give all the primitive types today (`i32`, etc) negative indices?

titzer: On Wed, Apr 6, 2016 at 6:14 PM, rossberg-chromium <notifications@github.com>
wrote:

> In BinaryEncoding.md
> <https://github.com/WebAssembly/design/pull/640#discussion_r58734661>:
>
> >  | Field | Type | Description |
> >  | ----- |  ----- | ----- |
> > +| form | `uint8` | `0x40`, indicating a function type |
>
> I didn't understand this part of the OP: why would we want to add the
> complication of "inlining" certain structural constructors instead of just
> using a type id?
> E.g. to reduce the overhead of one-off uses of structural types. Or
> imagine you want to emit a more complicated nested struct, then you
> wouldn't need to separate out and name each level. Perhaps the other way
> round is a more compelling scenario: you may want to allow primitive
> constructors in type definitions. Or they'll start to mix when we introduce
> type import/exports some day. Also, it's hard to predict what other thing
> might come along and change the story (generics? who knows). I'm not
> suggesting that there currently is a concrete reason to join the spaces.
> But it doesn't seem completely unlikely to arise later, and given that the
> cost of keeping it an option is zero, why preclude it? Yes, if we can rule
> out the third case as I'm asking above then the
> encoding could be pretty simple: if positive, it's a pritimive, if
> negative, it's a (negated) type-id.
> I'd actually invert that scheme, to avoid the negation.
>

Oh, you went and mentioned generics did you!

If we do indeed end up having generics (let's please call them by their
right name, type parameters :-)), then we have type parameter declaration
sites at function declarations and struct type declarations, then type
arguments that occur at usage sites: calls and instantiations in the
bytecode.

One way to handle type parameters (i.e. declarations) is to have another
type constructor which essentially corresponds to the "forall" quantifier.

But that doesn't solve the type arguments problem. There will be a scoping
problem since type arguments are essentially type expressions with free
variables. That means that type arguments that occur in a function body
need to be considered to be within a "scope" that binds the type parameters.

But all of that is kind of yucky, let's please not have generics or type
parameters in WebAssembly. I think that it's entirely reasonable to expect
source language implementations to employ their own specialization/boxing
schemes to deal with a WebAssembly engine that has no concept of type
parameters.

lukewagner: You had us scared until the last para ;-)  Agreed on that final sentiment: just like we wouldn't want to implement a particular class system, but provide the underlying GC building blocks, I'd hope that generics would be something that exist at the source language and are "compiled away" (in one way or another, I can imagine 3 distinct schemes for C++, Java and C#) by the time we get wasm.

rossberg-chromium: >
> What I don't understand is why all those future new things can't just be
> new forms of entries in the types section such that a type-id is all you
> need?
>

It actually bakes in a two-way restriction: all _non-nullary_ constructors
_must_ be named, and all _nullary_ constructors _cannot_ be named. My
intuition is that both these restrictions may turn out to be cumbersome or
even problematic in the future. At least it's not obvious to me that they
won't. For example, when such types need to cross module boundaries: either
as type imports/exports (you may need to export both kinds), or occurring
in the signature of some function (you may want to avoid exporting an
auxiliary type definition for every type you mention somewhere).

In contrast, where do you see the downside of avoiding conflicts between
the spaces until we understand the future better?

I'd actually invert that scheme, to avoid the negation.
>
> Wouldn't that give all the primitive types today (i32, etc) negative
> indices?
>

Yes, but would there be a problem with that?

rossberg-chromium: @titzer, ha, I knew I shouldn't have mentioned generics. This is getting OT, but let me just say that I would like to avoid their complexity as much as the next guy, while I'm also aware that "our language doesn't need generics" has become famous last words of language/VM designers. Static compilation only works for fairly weak, second-class polymorphism; in more expressive cases (which all big languages but C++ support) you'd be forced to introduce unions and lots of expensive runtime checks. Maybe that's okay for Wasm.

lukewagner: > In contrast, where do you see the downside of avoiding conflicts between the spaces until we 
> understand the future better?

I also don't understand what is being proposed in this PR to address this: superficially this is just a question of `0x40` vs. `-1`.  What are you proposing happens for these new-kinds-of-types?

> in more expressive cases (which all big languages but C++ support) you'd be forced to introduce unions 
> and lots of expensive runtime checks

Still OT, but: yes, for Java-style.  For C#-style, though, I was assuming that a C#-on-wasm runtime would actually need to ship with its own runtime machinery to do runtime generation of wasm for instantiations that only show up at runtime since I'd be surprised if we could design a feature in wasm that wasn't overly specialized to C# but that could still handle the C# use case.

rossberg-chromium: >
> In contrast, where do you see the downside of avoiding conflicts between
> the spaces until we
> understand the future better?
>
> I also don't understand what is being proposed in this PR to address this:
> superficially this is just a question of 0x40 vs. -1. What are you
> proposing happens for these new-kinds-of-types?
>
Hm, I thought we just agreed that signedness is best reserved for
distinguishing type ids. So this PR avoids clobbering the opposite sign
space, and instead just picks an arbitrary opcode for the function type
that doesn't collide with the primitive ones. 0x40 just because it
partitions the positive 1-byte signed LEB value range into two equal
halves, reserving one side for nullary, the other for non-nullary
constructors (which may or may not make sense, I'm open to better
suggestions; there are probably going to be far more nullary constructors
than others, but either way the space is comfortably large AFAICT).

lukewagner: > Hm, I thought we just agreed that signedness is best reserved for distinguishing type ids.

If non-nullary constructors don't show up in value types (only their type-ids), then both could have negative indices.  But I guess the counterargument is: maybe not the 3 non-nullary ctors we're thinking about *now* (func, struct, array), but perhaps *some* new thing in the the future and if positive indices are "reserved" for nullary and negative is "reserved" for type-ids, then we're out of luck (or, at the very least, we'd have to do break the pattern).  I guess I buy that, so `0x40` is fine.  More than just the aesthetic `-1` vs. `0x40` has been question of where this is going which I think I now have a better understanding of.

rossberg-chromium:
Ping on this one. Is there desire for more discussion, or should we land and leave potential opcode details for follow-ups?

JSStats:
From what I understand any change to the type section will require a wasm binary version bump anyway? Even adding functions with multiple return values would appear to not validate without a version bump due to the tight definition of the fields. So what would this change achieve?

rossberg-chromium:
Forwards compatibility. Even with version bumps, old versions should still be subsets of new versions.

lukewagner:
lgtm

BinaryEncoding.md
```diff
@@ -145,29 +152,29 @@ The import section declares all imports that will be used in the module.
...
+| types | `varuint32*` | sequence of indices into the type section |
 
-### Indirect Function Table section
+### Indirection section
 
-ID: `function_table`
+ID: `indirect`
 
-The indirect function table section defines the module's 
+The indirection section defines the module's
```
titzer: Worth spelling out "indirect function table" as per original text?

BinaryEncoding.md
```diff
@@ -200,34 +207,35 @@ The export table section declares all exports from the module.
...
-### Function Bodies section
+### Code section
 
-ID: `function_bodies`
+ID: `code`
 
-The Function Bodies section assigns a body to every function in the module.
-The count of function signatures and function bodies must be the same and the `i`th
-signature corresponds to the `i`th function body.
+The code section assigns a body to every function in the module.
```
titzer: "contains a function body for every function in the module"

Assign is just so...imperative :-)

BinaryEncoding.md
```diff
@@ -200,34 +207,35 @@ The export table section declares all exports from the module.
...
 | bodies | `function_body*` | sequence of [Function Bodies](#function-bodies) |
 
-### Data Segments section
+### Data section
 
-ID: `data_segments`
+ID: `data`
 
-The data segments section declares the initialized data that should be loaded
+The data section declares the initialized data that should be loaded
```
titzer: "should be" -> "is"

BinaryEncoding.md
rossberg-chromium: This one got renamed to table section on main branch.

BinaryEncoding.md
rossberg-chromium: Done

BinaryEncoding.md
rossberg-chromium: Done

BinaryEncoding.md
lukewagner: Incidentally, I came across a situation which supports the current design: in Wasm.Table we want to be able to declare the types of elements in the table definition/import.  The abovementioned scheme for local types seems to fit (allowing table elements to have any type that you can put in a local, or some restriction thereof), but the question is how to say "any function".  Well, since we already have this 0x40 "Function" constructor in the index space, that seems to be a good candidate (even if it's a slight abuse of logical category).  Similarly, the "Struct" and "Array" constructors could mean "any struct type" / "any array type" which could make sense one day.

PR #648, opened 2016-04-08
Add current_memory operator

rossberg-chromium:
After #629, the `memory_size` operator (which actually was missing from AstSemantics) is the last one not counting memory in page units. This PR species that, and also renames it to `current_memory` (as suggested [here](https://github.com/WebAssembly/spec/pull/273#issuecomment-206956351)).

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -82,18 +82,19 @@ operators.
 
 The main storage of a WebAssembly instance, called the *linear memory*, is a
 contiguous, byte-addressable range of memory spanning from offset `0` and
-extending for `memory_size` bytes which can be dynamically grown by
-[`grow_memory`](AstSemantics.md#resizing). The linear memory can be considered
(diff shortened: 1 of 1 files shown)


lukewagner:
Looks good, but could you also s/memory_size/current_memory/ in the rest of the design repo (there are a few).

rossberg-chromium:
Done, but I had to do a bit of rewording to make it fit.

lukewagner:
lgtm, thanks

jfbastien:
lgtm

This reminds me (not for this PR): didn't we discuss getting rid of this operator because of the race it may have (get current_memory and it's potentially already been changed by another thread)?

rossberg-chromium:
Merging with LGTMs and no objection

drom:
Shall we change BinaryEncoding.md with this new fix?

PR #656, opened 2016-04-14
Access to proprietary APIs apart from HTML5

loxal:


jfbastien:
As written this isn't accurate: when embedded by a JavaScript engine then you only get access to JavaScript APIs, but in a non-JavaScript embedder you may be able to access any API the embedder decides to expose. Could you update to clarify?

loxal:
This is great news, thanks for clarification!

FAQ.md
```diff
@@ -389,3 +389,7 @@ those that motivated the development of the
 
 Even Knuth found it worthwhile to give us his opinion on this issue at point,
 [a flame about 64-bit pointers](http://www-cs-faculty.stanford.edu/~uno/news08.html).
+
+## Will I be able to access proprietary platform APIs (e.g. Android / iOS)?
+
+Yes but it will depend on the _environment where a wasm VM is embedded_. Inside a browser it is most likely that you'll get access to the same HTML5 and other browser-specific APIs which are also accessible through regular JavaScript. However, if a wasm VM is provided as an “app execution platform” by a specific vendor, it might provide access to proprietary platform-specific APIs of e.g. Android / iOS.
```
jfbastien: Change "environment where a wasm VM is embedded" to "WebAssembly embedder". I think that's more consistent.

"most likely" isn't strong enough: you do get access to any existing web APIs, the same way JavaScript does.

I'd refer to NonWeb.md in the last sentence.

Also, Portability.md#api somewhere. I think this question is expanding on Portability.md#api.

loxal:
I’ve just squashed a fix.

jfbastien:
Looks good, thanks!

Issue #657, opened 2016-04-15
consider new "coercion" field of export/import declarations


lukewagner:
Today in asm.js, if you compile
`uint32_t f() { return UINT32_MAX; }`
in the obvious way, you'll get
`function asmModule() { 'use asm'; function f() { return -1 } return f }`
which, when called, will return `-1` to JS (which is `!=` `UINT32_MAX`) due to the fact that asm.js hard-codes the conversion from `i32` to JS `number` to interpret the `i32` as signed.

This trips up asm.js users in practice (who would naturally expect, if they wrote `UINT32_MAX` in their C, that they should test against `UINT32_MAX` in their JS) and it will surely trip up wasm users as well.  The same problem applies to arguments passed to FFI/import calls from wasm.

In asm.js, this could have been fixed by allowing functions to return either signed or unsigned integers (although there are tradeoffs), but in wasm, there is only `i32`.  Another fix is to require the toolchain, when compiling the above C code, to provide a light JS shim which explicity coerces to unsigned (`>>>0`) when the declared C return type is unsigned.

I'd like to consider in this issue a better and more general fix: What if export and import statements provided a new "coercion" field which was a sequence of bytes whose interpretation was, like the other byte-sequence fields of imports/exports, left up to the host environment.  When binding to JS, this is where we'd explain how to interpret integer arguments/returns.  However, I think there are many other potential uses, so I think we'd want to keep the contents of this field extensible, perhaps starting with a JSON blob like `{ret:'u'}`.  Of course, each wasm type would have a "default" coercion (what we're doing today) so this field could always be left empty.

Some other potential uses of the coercion field:
- We could allow `i64` to be passed/returned from JS _today_ using `{low, high}` objects and _later_ using `int64` value types by using the coercion field specify which one.
- By default, I think the highly-coercive `ToInt32`/`ToNumber` coercions would be applied when converting JS arguments in calls to wasm exports, but it might be nice to have a non-coercive option that threw if, e.g., the given JS value wasn't already a `number`.  This could make more sense when we start talking about GC/reference types.
- It's really common to want to pass/return strings from wasm linear memory to JS (that produce real JS strings).  There's actually not a super-efficient way to do this in JS atm (I see Emscripten's current `UTF8ArrayToString` does one `fromCharCode` and concat per character!).  A string coercion would thus be both a major usability _and_ performance improvement.
[...]

PR #659, opened 2016-04-18
add initial JS API spec

lukewagner:
The current demos and prototypes use a single temporary `Wasm.instantiateModule` JS API.  This PR breaks this function API into two separate steps: compilation and instantiation, based on some early discussion and consensus.  The PR also fleshes out some of the details of the interface, tying together ecma262 and ml-proto definitions.  This PR doesn't mention memory aliasing at all, that would be in a follow-up PR.


JS.md
```diff
@@ -0,0 +1,235 @@
+# JS API
```
jfbastien: I'd rather spell out "JavaScript", at least in titles.

JS.md
```diff
@@ -0,0 +1,235 @@
...
+
+In the MVP, the only way to access WebAssembly on the Web is through an explicit
```
jfbastien: Link to MVP.md.

JS.md
```diff
@@ -0,0 +1,235 @@
...
+JS API which is defined below. (In the future, WebAssembly may also be
+loaded and run directly from an HTML `<script type='module'>` tag&mdash;and any other
+Web API that loads ES6 modules via URL&mdash;as part of
```
jfbastien: You can use unicode directly instead of html-encoding.

JS.md
```diff
@@ -0,0 +1,235 @@
...
+
+#### `Wasm.compile`
+
+The `compile` function has the signature:
+```
+Promise<Wasm.Module> compile(TypedArray bytes)
+```
+If the given `bytes` argument is not a
+[TypedArray object](http://tc39.github.io/ecma262/#sec-typedarray-objects),
+a `TypeError` exception is thrown.
```
jfbastien: Does this allow streaming compilation?

JS.md
```diff
@@ -0,0 +1,235 @@
...
+following properties:
+
+### Constructor Properties of the `Wasm` object
+
+* `Wasm.Module` : see [`Wasm.Module` objects](#wasmmodule-objects)
+* `Wasm.Instance` : see [`Wasm.Instance` objects](#wasminstance-objects)
+* `Wasm.CompileError` : a [NativeError](http://tc39.github.io/ecma262/#sec-nativeerror-object-structure)
+   which indicates an error during WebAssembly decoding or validation
+* `Wasm.RuntimeError` : a [NativeError](http://tc39.github.io/ecma262/#sec-nativeerror-object-structure)
+   which indicates an error while running WebAssembly code
```
jfbastien: Do we want to further specify runtime errors? V8 currently [has the following](https://code.google.com/p/chromium/codesearch#chromium/src/v8/src/wasm/wasm-opcodes.h&l=332):
```
  kTrapUnreachable,
  kTrapMemOutOfBounds,
  kTrapDivByZero,
  kTrapDivUnrepresentable,
  kTrapRemByZero,
  kTrapFloatUnrepresentable,
  kTrapFuncInvalid,
  kTrapFuncSigMismatch,
  kTrapCount
```

JS.md
```diff
@@ -0,0 +1,235 @@
...
+```
+new Module(TypedArray bytes)
+```
+If the given `bytes` argument is not a
+[TypedArray object](http://tc39.github.io/ecma262/#sec-typedarray-objects),
+a `TypeError` exception is thrown.
+
+Otherwise, this function performs synchronous compilation of the TypedArray
+object:
+* The range of bytes compiled is the byte range subtended by the typed
```
jfbastien: "subtended"?

JS.md
```diff
@@ -0,0 +1,235 @@
...
+of *Exported Function* 
+[Exotic Object](http://tc39.github.io/ecma262/#sec-built-in-exotic-object-internal-methods-and-slots).
+Like [Bound Function](http://tc39.github.io/ecma262/#sec-bound-function-exotic-objects) Exotic Object,
+Exported Functions do not have the normal function internal slots but instead have:
+ * [[WasmInstance]] : the [`Eval.instance`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/eval.ml#L15)
+   containing the exported function
+ * [[WasmFunctionIndex]] : the index of the function inside the module
+
+Exported Functions also have the following data properties:
+* the `length` property is set to the exported function's signature's arity
```
jfbastien: That seems a bit odd. Is it how other exotic objects work?

JS.md
```diff
@@ -0,0 +1,235 @@
...
+Exported Functions do not have the normal function internal slots but instead have:
+ * [[WasmInstance]] : the [`Eval.instance`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/eval.ml#L15)
+   containing the exported function
+ * [[WasmFunctionIndex]] : the index of the function inside the module
+
+Exported Functions also have the following data properties:
+* the `length` property is set to the exported function's signature's arity 
+* the `name` is set to either:
+  * the generated string `"wasm-function[index]"`
+  * the function's name as defined in the [Names Section](BinaryEncoding.md#names-section)
```
jfbastien: Which takes precedent?

JS.md
lukewagner: That's a good question.  I was thinking that, if the list grew long over time (and if we're introducing a new constructor for each specific kind of error, this seems like a realistic possibility) this could end up taking a few KB of memory (for constructor objects and prototype objects).  JS seems to have a much [coarser](http://tc39.github.io/ecma262/#sec-properties-of-error-instances) set of error constructors which is a precedent we could follow.  We could instead have a `code` field of `RuntimeError` that provides a well-defined string literal for each of these specific cases (in addition to the impl-specific `message` field).  Any TC39 experience or direction here @rossberg-chromium ?

JS.md
lukewagner: That's definitely the goal: I was expecting that we'd simply add a stream overload to `Wasm.compile` and it still just returns a promise.

JS.md
lukewagner: I'll try "delimited"

JS.md
lukewagner: E.g., bound function objects (what you get from `f.bind()`) have a length (reflecting the new arity after binding) and name which is "bound <original function name>".  So yes, I think this is the precedent.

JS.md
lukewagner: Hmm, good question, and now that I write out an answer, I'm changing my mind :)
We have an option here: we've said "you can always ignore the names section", but if we define that the names *must* be used here, then that's no longer the case.  But that means `name` is nondeterministic which could break applications.  Now `Error.stack` (and `Error.message`) are already rather nondeterministic and that's just something we say is "OK", but `f.name` isn't, so maybe the right thing to do here is just to use `index` (as a Number as the "name") and leave the auto-generated vs. pretty-name question up to this already-nondeterministic feature.

kripken:
Looks like `Wasm.compile` is the async version, while `new Wasm.Module` is the sync version? That seems a little asymmetrical. Perhaps another option is `Wasm.compile` and `Wasm.compileSync`? That seems more consistent with apis like FileReader/FileReaderSync, node.js read/readSync, IndexedDB, etc.

JS.md
littledan: Yes, length seems good here. Nothing exotic here, move along folks.

What make functions exotic is their \[\[Call]] behavior. Wasm exported functions also have exotic call behavior. Could you spell out those semantics here?

JS.md
littledan: ECMAScript doesn't differentiate within broad classes of errors, but it is a permitted extension of implementations. Wasm can differentiate whatever it wants.

My question would be, what benefit do we hope to get from standardizing the distinction? Do we expect users to switch on the different errors? Or, do we expect errors to get better over time with more detailed implementation work (which standardizing too early might inhibit)?

JS.md
littledan: Oh sorry I see below; was using the wrong view in GitHub

JS.md
littledan: The convention with Promise-returning functions is to never ever throw, just return a rejected promise. That way, calling code doesn't have to worry about multiple kinds of failure.

JS.md
```diff
@@ -0,0 +1,233 @@
...
+to compile a `Wasm.Module` as described in the 
+[`Wasm.Module` constructor](#wasmmodule-constructor). On success, the
+`Promise` is [fulfilled](http://tc39.github.io/ecma262/#sec-fulfillpromise)
+with the resulting `Wasm.Module` instance. On failure, the `Promise` is 
+[rejected](http://tc39.github.io/ecma262/#sec-rejectpromise) with the 
+`Wasm.CompileError`.
+
+The asynchronous compilation is performed on the state of the given
+TypedArray when `compile` is first called; subsequent mutations of the
+TypedArray does not affect ongoing compilations.
```
lars-t-hansen: FWIW, shared memory does not allow you to guarantee that, since a TA can map shared memory and it can be mutated racily.  (Pedantic but true.)

JS.md
lukewagner: Ah, that makes sense, thanks for explaining, I'll update accordingly.

JS.md
binji: Yes, but if we add the "opt-in to shared memory" attribute to WebIDL, then this shouldn't be an issue, right?

lars-t-hansen: Not saying you couldn't restrict it, just that bare "TypedArray" does not express that restriction.

JS.md
lukewagner: > Do we expect users to switch on the different errors?

Yes, this seems like the key question.  Now, if we one day add some way to do fault/trap handling, the precise reason (and pc, and arguments, etc) would matter, but that's quite different than exceptions; with exceptions, I don't see there being that much differentiation you want to do other than reporting/debugging (for which a detailed `message` is what you want).

> Or, do we expect errors to get better over time with more detailed implementation work (which 
> standardizing too early might inhibit)?

Right, there is a risk we over-specify before we even know what these are good for and regret it later.  I think we can always add more precise information in the future (as fields of the `Wasm.RuntimeError` type), so perhaps that suggests having nothing more than the standard `message` field for now.

JS.md
lukewagner: Heh, yes, I thought about this while writing.  However, I still think there is an important semantic difference between "racy copy somewhere within `Wasm.compile`" vs. "continued racy access until the resolution of the promise".  I'll try to update to be more SAB-correct, though.

lukewagner:
@kripken Yes, agreed there is an asymmetry but there is also a symmetry with `new Function(code)` to consider as well.  As a tie-breaker, it seems like one should use the languages' built-in syntax for making new objects when possible.  I guess we could go the other way and try to have a `new WasmPromise(code)` (symmetric to [`FunctionPromise`](https://bugzilla.mozilla.org/show_bug.cgi?id=854627)).  @littledan Any precedent or opinions here?

JS.md
```diff
@@ -0,0 +1,233 @@
...
+Otherwise, this function starts an asychronous task to compile a `Wasm.Module`
+as described in the [`Wasm.Module` constructor](#wasmmodule-constructor).
+On success, the `Promise` is [fulfilled](http://tc39.github.io/ecma262/#sec-fulfillpromise)
+with the resulting `Wasm.Module` instance. On failure, the `Promise` is 
+[rejected](http://tc39.github.io/ecma262/#sec-rejectpromise) with the 
+`Wasm.CompileError`.
+
+The asynchronous compilation is logically performed on a copy of the state of
+the given TypedArray captured during the call to `compile`; subsequent mutations
+of the TypedArray after `compile` return do not affect ongoing compilations.
```
jfbastien: Could you mention the streaming compile, either here or in the todo section?

JS.md
```diff
@@ -0,0 +1,233 @@
...
+
+### `Wasm.Instance` `exports` property
+
+`Wasm.Instance` objects are created with an initial (writable, configurable,
+non-enumerable) `exports` data property whose value is the JS Array of exported
+JS Functions created by [`instantiate`](JS.md#wasmmoduleprototypeinstantiate).
+
+## Sample API Usage
+
+Given `demo.was` (encoded to `demo.wasm`):
```
jfbastien: `demo.wast`?

JS.md
lukewagner: Good idea, done.

JS.md
lukewagner: I'm trying to be more regular about using `.was` for "a single wasm module in the text format" and `.wast` for "the webassembly test file format, which can contain multiple modules as well as tests run on those modules".  In this case, I mean `.was` since we're talking about a single module.

JS.md
```diff
@@ -0,0 +1,237 @@
...
+* `Wasm.RuntimeError` : a [NativeError](http://tc39.github.io/ecma262/#sec-nativeerror-object-structure)
+   which indicates an error while running WebAssembly code
+
+### Function Properties of the `Wasm` object
+
+#### `Wasm.compile`
+
+The `compile` function has the signature:
+```
+Promise<Wasm.Module> compile(TypedArray bytes)
```
littledan: Minor: Maybe WebGL does it, but it still seems strange to me to accept *any* TypedArray, as opposed to either a specific kind (e.g., Uint8Array) or an ArrayBuffer.

JS.md
```diff
@@ -0,0 +1,237 @@
...
+
+The asynchronous compilation is logically performed on a copy of the state of
+the given TypedArray captured during the call to `compile`; subsequent mutations
+of the TypedArray after `compile` return do not affect ongoing compilations.
+
+In the [future](FutureFeatures.md#streaming-compilation), this function can be
+extended to accept a [stream](https://streams.spec.whatwg.org), thereby enabling
+asynchronous, background, streaming compilation.
+
+## `Wasm.Module` Objects
```
littledan: Any chance Wasm Modules could be ES2015 Module namespace exotic objects? This could make later integration with the loader spec more likely to work out smoothly.

JS.md
```diff
@@ -0,0 +1,237 @@
...
+   passing [[WasmInstance]], [[WasmFunctionIndex]], and `args`.
+ * Coerce the result of `Eval.invoke` as follows:
+   * if the return value is `None`, return `undefined`
+   * interpret `i32` as a signed integer and convert that integer to a Number value
+   * throw a `TypeError` if returning an `i64`
+   * return `f32`/`f64` as Number values, possibly performing
+     [canonicalization of NaNs](http://tc39.github.io/ecma262/#sec-setvalueinbuffer)
+
+Exported Functions do not have a [[Construct]] method and thus it is not possible to 
+call one with the `new` operator.
```
littledan: To be pedantic, you might want to make Wasm functions a little less exotic by giving them a [[Construct]] method which is the same as their [[Call]] method, and throwing when new.target is not undefined. Otherwise, they will have subtly different behavior from ordinary JavaScript functions in a way which is an extra burden to implement but provides no real value.

JS.md
```diff
@@ -0,0 +1,237 @@
...
+* the `length` property is set to the exported function's signature's arity 
+* the `name` is set to `index` as a Number value
+
+WebAssembly Exported Functions have a `[[Call]](this, argValues)` method defined as:
+ * Let `argTypes` be the list of value types defined by the signature of [[WasmFunctionIndex]].
+ * Let `args` be an empty list of coerced values.
+ * For each value type `t` in `argTypes` and value `v` in `argValues`:
+   * Append to `args` `v` coerced to `t` as follows:
+     * coerce `v` to `i32` via [`ToInt32(v)`](http://tc39.github.io/ecma262/#sec-toint32)
+     * throw a `TypeError` if `t` is `i64`
```
littledan: I really like this decision!

JS.md
lukewagner: Are there any other examples you can think of that accept a raw array of bytes but specify only a single view type?  An alternative which would, in common case, avoid creating a useless view, have the argument be an `ArrayBuffer` with optional `offset` and `length` arguments (which, if omitted defaulted to the begin and end, resp).  Sounds like we're [considering](https://bugzilla.mozilla.org/show_bug.cgi?id=973810) this in WebGL for the same reason.  WDYT?

JS.md
lukewagner: /me goes to see what a Module Namespace Exotic Object is...

I think you mean `Wasm.Instance`; a `Wasm.Module` doesn't have a direct correspondence to anything in ES6, afaics.

So for `Instance`, that is a good question and I like the symmetry.  We had wanted to avoid injecting all the export names directly into the instance object since these could shadow prototype methods.  Looks like Module Namespace Objects don't have that problem; there is not even a prototype.  There is currently only one method on `Wasm.Instance.prototype`: `instantiate`, so I see two options that avoid accidental shadowing:
* make the `exports` field of `Wasm.Instance` objects hold a Module Namespace Exotic Object
* convert `instantiate` into the `Wasm.Instance` constructor

I had considered the latter when initially writing this PR, but I thought that `m.instantiate()` looked a bit nicer.  But maybe it's more explicit to see the `new` and know that there are two distinct objects here.  WDYT?

JS.md
lukewagner: On the subject, I'd be curious to hear your thoughts on #657 which end up giving a reasonable alternative option here.

JS.md
lukewagner: Hm, based on various spec wording, I had thought that the "normal" way to say "I'm not a constructor" is to not have a [[Construct]] method.  E.g., from 9.3:
```
Built-in function objects that are not identified as constructors do not implement the [[Construct]] internal method unless otherwise specified in the description of a particular function.
```
In SM, this is as easy as setting or not setting a bit in the `JSFunction` object.  I thought this would actually make the implementation a bit easier since, for constructor calls, I thought it was necessary to return an object if the return value was primitive or something like that...

How does one cause `newTarget` to be undefined when calling [[Constructor]]?

s3ththompson:
Awesome! Thanks for writing this up Luke.
(Heads up to @domenic, with whom I was chatting about this API a while back.)

JS.md
```diff
@@ -0,0 +1,237 @@
+# JavaScript API
+
+In the [MVP](MVP.md), the only way to access WebAssembly on the Web is through
+an explicit JS API which is defined below. (In the future, WebAssembly may also
+be loaded and run directly from an HTML `<script type='module'>` tag—and
+any other Web API that loads ES6 modules via URL—as part of 
+[ES6 Module integration](Modules.md#integration-with-es6-modules).)
+
+## The `Wasm` object
```
domenic: Per https://w3ctag.github.io/design-principles/#casing-rules this should be `WASM`, not `Wasm`.

JS.md
domenic: The standard pattern across the web is to be "liberal in what you accept" and accept any BufferSource:

```
typedef (Int8Array or Int16Array or Int32Array or
         Uint8Array or Uint16Array or Uint32Array or Uint8ClampedArray or
         Float32Array or Float64Array or DataView) ArrayBufferView;

typedef (ArrayBufferView or ArrayBuffer) BufferSource;
```

JS.md
domenic: Definite +1 for the latter option

JS.md
```diff
@@ -0,0 +1,237 @@
...
+explicit control over both compiled-code caching and cross-window/worker code
+sharing.
+
+### `Wasm.Module.prototype.instantiate`
+
+The `instantiate` method has the signature:
+```
+Wasm.Instance instantiate([importObject])
+```
+If the receiver of `instantiate` is not a `Wasm.Module` instance, a `TypeError`
```
domenic: Nit: "receiver" is not a thing. "this value" is what you're looking for, I think

JS.md
```diff
@@ -0,0 +1,237 @@
...
+the [`exports`](#wasminstance-exports-property) data property is
+`exportObject`.
+
+## Exported Function Exotic Objects
+
+Functions exported by WebAssembly modules are reflected in JS via a new kind
+of *Exported Function* 
+[Exotic Object](http://tc39.github.io/ecma262/#sec-built-in-exotic-object-internal-methods-and-slots).
+Like [Bound Function](http://tc39.github.io/ecma262/#sec-bound-function-exotic-objects) Exotic Object,
+Exported Functions do not have the normal function internal slots but instead have:
```
domenic: Bound functions get a lot of important internal slots "for free" by delegating to their [[BoundFunctionTarget]]. Since you're not doing something like that, you need to at least include [[Prototype]], [[Extensible]], [[Realm]], and [[ScriptOrModule]].

JS.md
domenic: Yeah, I think it is more normal to not have a [[Construct]] (similar to method definitions and all built-in non-constructor functions).

JS.md
jfbastien: Isn't this an abbreviation, though? Seems like the proper abbreviation of **W**eb**As**se**m**bly would be `WAsm`. Ew.

lukewagner: Hmm, ISTR someone pointing to another precedent the last time everyone discussed.  But your link looks pretty authoritative and there's also SIMD/JSON, so I'll change it.

JS.md
lukewagner: Ah, `BufferSource`, that makes everyone happy, I like that.

JS.md
lukewagner: (comment race) hah, right, so it's half initialism, half abbreviation, defying all classification.

JS.md
lukewagner: fixed

JS.md
jfbastien: So Unicode it is! `﻿ＷＡＳＭ`

s3ththompson: < unhelpful>
`W̝̉͒̐͗ͨ̀A̯̳̘͇̤̟̤̓S̹ͪ́_m`
</ unhelpful>

JS.md
lukewagner: Ah, I see that now in the second para of 9.3.  Updated.

JS.md
littledan: My mistake. Class constructors define both \[\[Call]] and \[\[Construct]] and simply throw a **TypeError** when called, but methods and arrow functions define only \[\[Call]] and don't define \[\[Construct]].

JS.md
littledan: You're right, `Wasm.Instance` is more relevant. A JavaScript constructor can return whatever it wants, so it would be fine to use `new` and return a module. However, modules (even just exotic module namespace objects) have not shipped yet, so I could see the argument decouple ES2015 modules from Wasm's API.

I am not quite clear about the motivation for `Wasm.Module`. Why not just have the TypedArray-based constructor return a `Wasm.Instance`/module? If you pass it to another worker with structured clone, it'll get a different copy. If you share it with a same-origin iframe, it should be fine that they have the same function bodies. For a cross-origin iframe, it would be hard to call methods on it anyway; seems like you might as well postMessage the instance.

JS.md
littledan: Sorry, I had missed the case where the point was to make a view on part of a larger ArrayBuffer. I guess the WebGL precedent makes it reasonable to not check types. I hope it's not a very strong need to avoid allocating the TypedArray; that shouldn't be too expensive for this kind of case, I'd imagine, given how much work it'll take to generate machine code as output from the compilation.

littledan:
Re `Wasm.compile`/`new Wasm.Module`: I don't know of much precedent for something like this, and this is the first I've heard of `FunctionPromise` (though it does sound like a potentially good idea at first glance). The general guidance I've heard is, for new APIs, go with async-only and don't provide a blocking version. What's not quite clear to me about how that would apply is, should the amount of computation needed here constitute something "blocking"? And, on the flipside, how bad would it be for users to always have to deal with a Promise when executing wasm code? A constructor can theoretically return a Promise of what it's constructing, but that may be subclass-hostile (given that we don't have async constructors proposed, or a way to await the super call).

FutureFeatures.md
```diff
@@ -402,3 +402,27 @@ separate WebAssembly [instances](Modules.md). Another use case is allowing, on
 the Web platform, importing a JS `ArrayBuffer` as a linear memory. This would
 allow highly efficient, specialized code to be generated for accessing the
 `ArrayBuffer`.
+
+## Streaming Compilation
+
+The WebAssembly binary format is designed to allow streaming decoding,
+validation and compilation so that these tasks to happen, e.g., while the binary
```
rfk: typo? "so that these tasks to happen"

lukewagner: Indeed, thanks.  Sentence was awkward and redundant with below, so removed.

JS.md
lukewagner: (With `BufferSource`, if all you want is to compile a whole buffer, you can just pass the `ArrayBuffer`.)

JS.md
lukewagner: Ok, @jonco3 helped me understand these concepts a bit more.  It seems like `Wasm.Instance` directly corresponds to a [Module Record](http://tc39.github.io/ecma262/#sec-abstract-module-records) which is a separate-but-related thing from the Module Namespace object.  Since the Module Namespace object's structure is super-fixed (it's only exports), if Module Records were to be one day reflected as first-class JS objects (they are already represented this way internally in SM which is useful for self-hosting and unit testing), then they'd be a distinct JS object from the Namespace object.

So that suggests keeping exports and instance separate as it is now and changing `exports`s value from a POJO to a Module Namespace Exotic Object.  Talking to @jonco3, this would be pretty easy to implement in SM, we think.  In the spec, Module Namespace Objects have a [[Module]] slot, so we'd need to define a Module Record for each `Wasm.Instance` which seems doable...

That reopens the question is whether to keep `instantiate` a method or morph it into the `Wasm.Instantiate` constructor.

JS.md
lukewagner: Ok, so `WASM` (boring ascii)?

domenic: I'm not sure anymore. Either `WASM` or `WAsm` seem correct-ish :-/

jfbastien:
> Re `Wasm.compile`/`new Wasm.Module`: I don't know of much precedent for something like this, and this is the first I've heard of `FunctionPromise` (though it does sound like a potentially good idea at first glance). The general guidance I've heard is, for new APIs, go with async-only and don't provide a blocking version. What's not quite clear to me about how that would apply is, should the amount of computation needed here constitute something "blocking"? And, on the flipside, how bad would it be for users to always have to deal with a Promise when executing wasm code? A constructor can theoretically return a Promise of what it's constructing, but that may be subclass-hostile (given that we don't have async constructors proposed, or a way to await the super call).

One more thing to consider: the `.wasm` module is already being downloaded over the Intertubes, which is fundamentally async. I'm wary of adding async where it's not "natural" but download already being async means async compilation isn't out of place. I therefore agree that synchronous compilation could be removed, at least for now.

lukewagner:
I think synchronous compilation enables fundamental categories of uses of wasm even in the MVP: it enables wasm apps to generate wasm bytes (in linear memory), compile those bits, then call those bits.  This is something you'd want when implementing a wasm JIT (say, for [Mono](https://www.reddit.com/r/programmerchat/comments/4dxpcp/i_am_miguel_de_icaza_i_started_xamarin_mono_gnome/d1vdmsm)).  With the forthcoming (after this PR is done) first-class-`Wasm.Memory`/`Wasm.Table` PR, you'll even be able to do somewhat fine-grained JIT compilation/linking so people could start doing this with the MVP.  (Not having synchronous compilation would also technically be a regression from asm.js, which you can `eval`/`new Function`.)

JS.md
lukewagner: Well of the two, one makes me barf :) updated

lukewagner:
Ok, I updated the PR to create instances via the `WASM.Instance` constructor and to define the `.exports` of a `WASM.Instance` to be a Module Namespace Object.  While there are technically no properties on the `Instance` (or `Instance.prototype`) that could be shadowed, since Namespace Objects have a super-fixed structure, it seems prudent to keep them separate so that methods/properties could be added to `Instance` in the future.

Also, the updated PR is pretty handwavy about precisely how the Module Namespace Object spec language "plugs in" to wasm modules.  (I realized that, when we have re-exports of functions and exports of mutable globals, we'll need the proxy-like generality of Module Namespace Object; the plain old "create a POJO with these properties" definition won't cut it.)  It'll take a bit of spec-y plumbing to connect all the dots which I'm not super eager to do at this point; I'm more interested atm if the rough API here looks reasonable.

lukewagner:
Rebased and added a note at the top about `Wasm.instantiateModule`.

Any more comments?  I fully expect there are a number of changes to be made between this and the final API, but is everyone comfortable with this as a starting point to iterate on in follow-up PRs?  There's some more significant additions to be made on top of this PR (`WASM.Memory`, `WASM.Table`, etc) that I'd like to get to that I've been holding off on to see how these basics play out.

rossberg-chromium:
LGTM. The one nit I still have: Didn't we want to call it `Wasm`? It's not a proper acronym, so upper-case spelling makes no real sense to me.

lukewagner:
See @domenic's comment [above](https://github.com/WebAssembly/design/pull/659#discussion-diff-60498760).  Seems like we're actually in no-man's land with half initialism, half abbreviation.

rossberg-chromium:
I agree that none of the rules that @domenic links to applies. Even in prose you'd write Wasm -- it's a contraction, not an acronym.

Just saying. I don't really care too strongly. Just fear that people gonna start shouting WASM in prose all the time now, too. :)

lukewagner:
Totally agreed on not wanting to read "WASM" in prose.  For the last year, though, I've stuck to the convention of either abbreviating to "wasm" (in informal contexts) or expanding to "WebAssembly"; to me "Wasm" looks just as goofy (although we do use it in SM when implied by naming convention).

domenic:
WAsm is still an option!

titzer:
Screw this, V8 is going back to __wasm

rossberg-chromium:
Snake casing: `__w_asm__`. Ah, why the parsimony, underscores are cheap, `__w__asm__`.

@lukewagner, why would you say "wasm" in prose, it's a name, names ought to be capitalised!

lukewagner:
¯\_(ツ)_/¯

binji:
Think of the saved bytes if we go with `WA`! On the other hand, it's 2016; do we really want to go down the path of `creat`? So how about `WebAssembly`?

lukewagner:
I could dig it.  It also lowers probability of global-namespace collision ;)

stoklund:
+1 for WebAssembly

lukewagner:
Righto, based on +1s/:tada:, switched to `WebAssembly`.  Based on lgtm above, I'll merge in a bit unless any objections.

PR #660, opened 2016-04-19
Permit implementations to return canonical NaNs from arithmetic.

sunfishcode:
RISC-V and ARM in "default NaN" mode do not propagate NaN values in
arithmetic operations. Instead, they return a "canonical" NaN.

IEEE 754 doesn't require NaN propagation; it's merely a "should".

In order to ensure that wasm doesn't accidentally over-burden support
for the above and other potential future IEEE-754-conforming platforms,
this patch changes the NaN rules to permit either behavior.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -426,8 +426,8 @@ The same operators are available on 64-bit integers as the those available for
 ## Floating point operators
 
 Floating point arithmetic follows the IEEE 754-2008 standard, except that:
- - The IEEE 754-2008 recommendation that operations propagate NaN bits from
-   their operands is permitted but not required.
+ - The IEEE 754-2008 section 6.2 recommendation that operations propagate NaN
+   bits from their operands is permitted but not required.
  - WebAssembly uses "non-stop" mode, and floating point exceptions are not
    otherwise observable. In particular, neither alternate floating point
    exception handling attributes nor the non-computational operators on status
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -550,15 +550,16 @@ between the following:
    most significant bit and 0 in the remaining bits.
  - a NaN with the sign bit of the operand and a fraction field consisting of
    1 in the most significant bit, followed by all but the most significant
-   bits of the fraction field of the operand, followed by all 0s.
+   bit of the fraction field of the operand, followed by all 0s.
 
 If the operand of demotion is a NaN, the result is nondeterministically chosen
 between the following:
  - a NaN with a nondeterministic sign bit and a fraction field with 1 in the
    most significant bit and 0 in the remaining bits.
  - a NaN with the sign bit of the operand and a fraction field consisting of
-   1 in the most significant bit, followed by as many of all but the most
-   significant bit of the fraction field of the operand as fit.
+   1 in the most significant bit, followed by all but the most significant bit
+   of the fraction field of the operand, discarding the least significant bits
+   that don't fit.
 
 Reinterpretations always succeed.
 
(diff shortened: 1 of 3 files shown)


AstSemantics.md
```diff
@@ -426,9 +426,8 @@ The same operators are available on 64-bit integers as the those available for
 ## Floating point operators
 
 Floating point arithmetic follows the IEEE 754-2008 standard, except that:
- - The sign bit and fraction field of any NaN value returned from a floating
-   point arithmetic operator are deterministic under more circumstances than
-   required by IEEE 754-2008.
+ - The IEEE 754-2008 recommendation that operations propagate NaN bits from
+   their operands is permitted but not required.
```
jfbastien: This section details where wasm diverges from IEEE754, so the bullet should be in another list. Also, is there a reference paragraph in IEEE754 for this?

AstSemantics.md
```diff
@@ -546,15 +544,21 @@ Promotion and demotion of floating point values always succeed.
...
+   1 in the most significant bit, followed by all but the most significant
+   bits of the fraction field of the operand, followed by all 0s.
+
+If the operand of demotion is a NaN, the result is nondeterministically chosen
+between the following:
+ - a NaN with a nondeterministic sign bit and a fraction field with 1 in the
+   most significant bit and 0 in the remaining bits.
+ - a NaN with the sign bit of the operand and a fraction field consisting of
+   1 in the most significant bit, followed by as many of all but the most
+   significant bit of the fraction field of the operand as fit.
```
jfbastien: define "as fit"?

Nondeterminism.md
```diff
@@ -30,11 +30,12 @@ currently admits nondeterminism:
...
-   floating point operand types and a floating point result type receives
-   multiple NaN input values with differing bit patterns, it is nondeterministic
-   which bit pattern is used as the basis for the result (as it is in
-   IEEE 754-2008).
+ * Except when otherwise specified, when an arithmetic operator returns NaN,
+   there is nondeterminism in determining the specific bits of the NaN. However,
+   wasm does still provide the guarantee that NaN values returned from an operation
+   will not have 1 bits in their fraction field that aren't set in any NaN values
+   in the input operands, except for the most significant bit of the fraction field
+   which is typically set to 1.
```
jfbastien: "provides the guarantee...typically set to 1" hard to parse IMO. Guarantee / typically don't go well together :-s

jfbastien:
Can we change "nondeterministically" to "implementation-defined"? I'm specifically trying to guarantee that the behavior from one execution to another is the same if the compiler and hardware haven't changed, I think that's a property we can keep.

AstSemantics.md
sunfishcode: It's an area where wasm doesn't "follow" IEEE's recommendation.

I added a reference to the relevant section.

AstSemantics.md
sunfishcode: I'm open to suggestions.

Nondeterminism.md
sunfishcode: I'm open to better wordings.

sunfishcode:
I'm not sure that the relationship between one run and another is something we can meaningfully say anything about. I'd encourage you to submit a separate PR to propose specific wording that can be discussed.

sunfishcode:
I've now made an attempt at some revised wordings. In general though, while I'm open to suggestions, I want to avoid getting significantly more verbose, because the vast majority of people reading this document won't care about these details, and the remaining people who really need to know precisely what happens here should consult the spec instead anyway (which I'll update once this PR is resolved.)

sunfishcode:
Ping :-). Another advantage of this change is that it simplifies implementations that do wish to canonicalize after arithmetic, to avoid propagating unnecessary information or implementation details.

titzer:
lgtm

jfbastien:
lgtm. Sorry for the delay, I hadn't seen the reworded version.

sunfishcode:
Sounds like lgtms!

PR #661, opened 2016-04-20
Rename fields in memory section

lukewagner:
Does not affect the spec, but I think `initial` is a better name than `minimum` since the `initial` isn't necessarily the `minimum` if, in the future, we were to add a `shrink_memory` + `minimum` (which could be <= `initial`).  The PR also renames `max` to be more consistent with other names in this file.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -178,8 +178,8 @@ associated with the module.
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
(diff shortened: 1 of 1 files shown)


jfbastien:
Should we do "maximum" since the others are also spelled out? lgtm otherwise. Should this be in 0xb?

lukewagner:
Yes, "maximum" is more consistent.  The field names don't show up in the actual bits, so no need to care about version.

lukewagner:
Merging with one lgtm and no objections.

Issue #601, opened 2016-03-10
measure performance of other variable-length integer encoding schemes


lukewagner:
As suggested by [this HN comment](https://news.ycombinator.com/item?id=11263378) we should measure the comparative decode-time differences between LEB128 and the described PrefixVarint.  Some of the advantages won't apply since we're mostly dealing with 1- or 2-byte varints, but, still, having the tag bits in the LSB might allow an optimized decoder to reduce branching.

mbebenita:
@tterribe points out that we could reverse the prefix bits so that `bsr` could be used to quickly find the length of the remaining bytes.

qwertie:
@lukewagner Now that you mention it, wouldn't this encoding be nice?

~~~
0xxxxxxx 7 bits for the number
10xxxxxx xxxxxxxx 14 bits for the number
110xxxxx xxxxxxxx xxxxxxxx 21 bits for the number
1110xxxx xxxxxxxx xxxxxxxx xxxxxxxx 28 bits for the number
1111nnnn xxxxxxxx .... n+2 total bytes, with (n+1)*8 bits for the number
~~~
This is a combination of two different encoding styles, depending on the top 4 bits, and I'm just proposing it on the _guess_ that an _optimized_ implementation of this would be simpler than the `PrefixVarInt`. It doesn't have the nice property of "uses exactly the same # of bits as LEB128" - instead it uses

- Same # of bytes as LEB128 for integers <= 32 bits
- One more byte than LEB128 for some (not all) integers between 33 and 49 bits
- Same # of bytes as LEB128 for integers between 50 and 63 bits
- One fewer byte than LEB128 for some (not all) integers that are 64 bits or larger

| LEB128 holds up to   | This encoding holds up to |
|----------------------|---------------------------|
| 28 bits in 4 bytes   | 28 bits in 4 bytes        |
| 35 bits in 5 bytes   | 32 bits in 5 bytes        |
| 42 bits in 6 bytes   | 40 bits in 6 bytes        |
| 49 bits in 7 bytes   | 48 bits in 7 bytes        |
| 56 bits in 8 bytes   | 56 bits in 8 bytes        |
| 63 bits in 9 bytes   | 64 bits in 9 bytes        |
| 70 bits in 10 bytes  | 72 bits in 10 bytes       |
| ...                  | ...                       |
| 119 bits in 17 bytes | 128 bits in 17 bytes     |
| 126 bits in 18 bytes | N/A                      |
[...]

PR #662, opened 2016-04-20
Replace uint8 with varint7 in form field

lukewagner:
This needs to be variable-length.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -38,6 +38,9 @@ A [Signed LEB128](https://en.wikipedia.org/wiki/LEB128#Signed_LEB128) variable-l
 ### varuint1
 A [LEB128](https://en.wikipedia.org/wiki/LEB128) variable-length integer, limited to the values 0 or 1. `varuint1` values may contain leading zeros. (This type is mainly used for compatibility with potential future extensions.)
 
+### varuint7
+A [LEB128](https://en.wikipedia.org/wiki/LEB128) variable-length integer, limited to the values [0, 127]. `varuint7` values may contain leading zeros. (This type is mainly used for compatibility with potential future extensions.)
+
 ### varuint32
 A [LEB128](https://en.wikipedia.org/wiki/LEB128) variable-length integer, limited to uint32 values. `varuint32` values may contain leading zeros.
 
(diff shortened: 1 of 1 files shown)


kripken:
lgtm

lukewagner:
Merging based on lgtm and since it doesn't actually change the binary format at this time.

BinaryEncoding.md
```diff
@@ -126,7 +129,7 @@ The type section declares all function signatures that will be used in the modul
 #### Type entry
 | Field | Type | Description |
 | ----- |  ----- | ----- |
-| form | `uint8` | `0x40`, indicating a function type |
+| form | `varuint7` | `0x40`, indicating a function type |
```
JSStats: Just a suggestion but could this field eventually use the same encoding as for the opcodes.

lukewagner: True, that might make sense.  My hope is that we can use `varuint*` as the go-to default for everything and that only b/c of the extreme prevalence of opcodes (assuming the opcode-with-immediates table) *and* the consequent win of doubling the number of 1-byte opcodes do we make an exception for opcodes.

rossberg-chromium:
Not quite true that it doesn't change the binary format. You allow more now. ;)

FWIW, I agree with making all encodings of actual _numbers_ var[u]ints, but I'm less convinced that it is necessary for non-numbers. Opcodes are just discrete constants, and the opcode space for types in particular is super-sparse. Not sure why we need to allow multi-byte encodings for them before we have to.

lukewagner:
> Not quite true that it doesn't change the binary format. You allow more now. ;)

There's still only one valid constant.  I guess LEB128 allows leading zeroes, but other schemes in #601 do not.

But I'd strongly disagree that variable-length schemes should be used only for numbers: it seems quite unlikely we've fully thought through all possible uses of this (or any other fixed-width-field) candidate.  We're thinking today in terms of a small set of primitive and compound constructors, but it's hard to know there won't be, e.g., a user-defined set in the future, or some product-space we need to enumerate.  At a very low cost, we avoid having to make this bet based on limited understanding.

rossberg-chromium:
Not sure I follow. That's still no reason to extend the format now. Just allowing single bytes today is forward-compatible with generalising to LEB128 later, once that actually turns out to be needed. You can safely defer that generalisation until you know that you want to represent more than 128 cases, and that numbers are the best encoding. Restricting to single byte is conservative.

lukewagner:
It's more of a signal indicating that this won't always be a fixed-width field and the high bit of the byte is "reserved".

PR #641, opened 2016-04-05
Postorder opcodes

titzer:
This is a carry-over of #628 to binary version 11.

PR #666, opened 2016-04-21
Add end opcode to functions

rossberg-chromium:
Now that #641 has landed, here's the PR for the last of the changes I suggested in #623.

Function bodies are implicitly delimited by their byte size in the current encoding. This is unfortunate for a couple of reasons:

- With the current format, any stream decoder needs to perform two end checks on every opcode: against end-of-stream and against body size. (When streaming, you generally don't know the length of the stream up-front, nor whether it's well-formed, so neither check subsumes the other.) With this PR, only the end-of-stream check is needed.

- This is the only piece of the binary format that stands in the way of formulating the entire format unambiguously as a grammar. Because it depends on non-local (and lower-level) context information. It would be rather sad to get 99% there but lose it on the last meter.

This PR hence adds an explicit `end` opcode to functions. Pros:

- The binary format is fully structured and can be entirely parsed linearly from an abstract byte stream, without paying attention to any size information.

- Sizes would only be needed to (a) seek through the stream if desired, (b) validating that they are consistent. That decouples concerns nicely.

Cons:

- One extra redundant byte per function. Previous measurements suggest e.g. a 0.3% size increase on AngryBots.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -82,18 +82,19 @@ operators.
 
 The main storage of a WebAssembly instance, called the *linear memory*, is a
 contiguous, byte-addressable range of memory spanning from offset `0` and
-extending for `memory_size` bytes which can be dynamically grown by
-[`grow_memory`](AstSemantics.md#resizing). The linear memory can be considered
+extending up to a varying *memory size*.
+This size always is a multiple of the WebAssembly page size,
+which is 64KiB on all engines (though large page support may be added in the
+[future](FutureFeatures.md#large-page-support)).
+The initial state of linear memory is specified by the
+[module](Modules.md#linear-memory-section), and it can be dynamically grown by
+the [`grow_memory`](AstSemantics.md#resizing) operator.
+
+The linear memory can be considered
 to be an untyped array of bytes, and it is unspecified how embedders map this
 array into their process' own [virtual memory][]. The linear memory is
 sandboxed; it does not alias the execution engine's internal data structures,
 the execution stack, local variables, or other process memory.
-
-The initial state of linear memory is specified by the
-[module](Modules.md#linear-memory-section). The initial and maximum memory size
-are required to be a multiple of the WebAssembly page size, which is 64KiB on
-all engines (though large page support may be added in the
-[future](FutureFeatures.md#large-page-support)).
 
   [virtual memory]: https://en.wikipedia.org/wiki/Virtual_memory
 
@@ -152,7 +153,7 @@ interpreted as an unsigned byte index.
 
 Linear memory operators access the bytes starting at the effective address and
 extend for the number of bytes implied by the storage size. If any of the
-accessed bytes are beyond `memory_size`, the access is considered
+accessed bytes are beyond the current memory size, the access is considered
 *out-of-bounds*.
 
 The use of infinite-precision in the effective address computation means that
@@ -214,6 +215,10 @@ When a maximum memory size is declared in the [memory section](Module.md#linear-
 reserve the space up front or if enabling the reserved memory fails.
 When there is no maximum memory size declared, `grow_memory` is expected
 to perform a system allocation which may fail.
+
+The current size of the linear memory can be queried by the following operator:
+
+  * `current_memory` : return the current memory size in units of pages.
 
 As stated [above](AstSemantics.md#linear-memory), linear memory is contiguous,
 meaning there are no "holes" in the linear address space. After the
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -81,16 +81,18 @@ The module starts with a preamble of two fields:
 This preamble is followed by a sequence of sections. Each section is identified by an
 immediate string. Sections whose identity is unknown to the WebAssembly
 implementation are ignored and this is supported by including the size in bytes
-for all sections. The encoding of all sections begins as follows:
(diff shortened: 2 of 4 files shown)


yurydelendik:
Some information can be stored after end of the function. For example, it can be used to add padding to each function for patching the binary data in the future. There shall be a statement about what to do with it (e.g. forbid any extra data after first function end).

titzer:
Another con is that it adds a redundant check in the decoder, since a decoder must always check the bounds of the bytes, anyway.

rossberg-chromium:
@yurydelendik, I don't see how the current format allows that. If so, it's a bug.

@titzer, see the first bullet point of the description. ;)  At least for streaming decoders, I claim this change actually _removes_ tons of checks.

lukewagner:
@rossberg-chromium I was excited that it might be possible to remove the end-of-body branch when validating unknown wasm too (you start by checking that body_length is in-bounds then that the last opcode is `end`), but I don't think that scheme is valid: the end opcode could be slurped up as part of an immediate of a preceding opcode so you can't rely on hitting it.

Now, when you are decoding *known-valid* wasm (as is in the case in SM for our parallel compiler threads), there is no body_end check to begin with and so this proposal *would* remove a *per-top-level-iteration* end-of-body check: you simply have the `end` case in the switch be responsible for breaking the loop so the loop becomes `while(true)`.

So if we're doing bean-counting (none of this is remotely significant in overall perf), I think the latter saves more branches than the former.

Anyhow, +1 from me.

rossberg-chromium:
@lukewagner, sure, you still have to verify the size at the end of the body, and that the `end` opcode is there. You simply check that once you've finished decoding the body.

But _only_ there. You don't have to check for the size limit _at every operator_ anymore. So it's one check per function vs one check per operator in a streaming parser.

lukewagner:
@rossberg-chromium What I was trying to explain is that, if you are validating unknown wasm bytes, you still need to check *some* size limit (body or end of whole module) on every operator, just to make sure you're not running off the end.

rossberg-chromium:
@lukewagner, yes, you always need to check against end-of-stream. But without explicit `end` opcode you _additionally_ need to check against the size limit. When decoding from a stream, you can't know which condition becomes true first, so you always have to check both, for each opcode. With this change, however, the latter check is no longer necessary.

lukewagner:
Ah hah, I see what you're getting at.  I had been assuming the whole module was downloaded such that body-end always came first (MVP).  Now technically, with streaming, you could check once against `min(stream-end, body-end)` etc etc, but that's a lot more complexity.  Even our existing postorder code, now that I look at it, could get simpler/faster with the end opcode, so yes, double +1.

titzer:
I don't see why it's more complexity to check against a stream end. In the V8 decoder, it just uses one limit, and if there's an error, it sets the limit to the start, which terminates decoding. One check in the loop. Adding end only adds additional checks after decoding is finished; e.g. that one and only one end occurred. That seems totally redundant to me.

JSStats:
Just a thought, but might some of the loss be gained back by making this an end operator rather than an end market and giving it `return` semantics. There are a lot of small functions in AngryBots that have a return as the last statement. It could even accept an expression with a value to return in the case that the function was defined to return a value. This might remove the encoding efficiency issue which seems to be the only objection?

lukewagner:
@titzer We're considering the streaming case where stream-end might be less than body-end (because you need to wait for the next chunk).  Yes, you can still use 1 check in this case (taking the `min` of the two), but it seems simpler (and no more expensive) to just iterate until `end`.  Anyhow, these are super-predictable branches so I doubt this is worth discussing from a perf angle.  I do see how `end` would allow simpler code, though, in this streaming case.

lukewagner:
@JSStats Because wasm doesn't require a final return statement&mdash;it just uses the last expression if non-void&mdash;in some sense that is already the case for the `end` proposal: it's like a special "final return" opcode.  Actually, I kindof like it more for that reason; it explains why it is natural to terminate your iterative loop from the `end` case of the switch.

JSStats:
@lukewagner Yes, thank you, sorry for the noise. Making the end an explicit return probably does not address the size issue because the trailing `return` statements in AngryBots are redundant and the value of the last expression could be returned. I was pondering it from the perspective of a possible expressionless encoding where there is no last expression and it would need to interpret the end as an implicit `return` accepting values from registers.

titzer:
@lukewagner 

There are a billion gotchas with a streaming decoder, like what happens when the stream ends in the middle of a LEB or between an opcode and its immediate, in the middle of a br_table, etc. I am not convinced end does anything to help here.

rossberg-chromium:
@titzer, the nice thing about a streaming decoder is that it keeps you honest: you have to stick to the stream abstraction, no shortcuts allowed. That actually makes it very simple. The spec decoder uses a stream for that reason.

yurydelendik:
Looks like vendors might choose offset in the binary format as pc. This might be a small pro to have end opcode at the end of function -- it will provide additional point for instrumentation/debug operations at the end of a function (e.g. play role of a physical offset of the function epilogue) or just set non-empty ranges/boundaries for empty functions.

flagxor:
I think this should get merged. Andreas, can you resolve the conflicts and land?

rossberg-chromium:
Merged manually to 0xC branch.

PR #670, opened 2016-04-22
Text fix in BinaryEncoding

kripken:
Two lines seem to be in the wrong order. The br_table comment talks about the table below it, so the if-else statement should mind its own business.


rossberg-chromium:
lgtm. Might have been a merge fail?

kripken:
Yeah, looks like a merge fail.

PR #672, opened 2016-04-26
Add explicit argument counts

rossberg-chromium:
Rebased version of #595, with tweaks.


rossberg-chromium:
Merging with prior LGTMs, and given that implementations have already moved ahead.

PR #674, opened 2016-04-28
Minor wording fix: function bodies contain a list of expressions

sunfishcode:
Functions contain a list of expressions, rather than a single expression. In ml-proto's words, "`body : expr list`".


kripken:
lgtm

PR #680, opened 2016-05-02
Clarify type of the immediate offset operand

lars-t-hansen:


JSStats:
My reading of the original version is that if the address were i64 then the offset would be the same type so i64, so it might still be useful to not.

lars-t-hansen:
@JSStats, I don't understand.  I'm not trying to remove any constraints, I'm just trying to make the language make sense.

JSStats:
This still seems relevant and seems to have been removed: 'The immediate is the same type as the address-index.'? Perhaps I've missed something, but have you explained what the type of the immediate offset is and how it might change with the type of the index expression?

Also 'The alignment has same type as the address and offset operands.' might not be correct.

lars-t-hansen:
@JSStats, please read my initial comment.  "index" is not defined anywhere, the language in the current spec is therefore meaningless.

Even if the language were to made meaningful (instead of removing it, as I have done) it would be redundant, because the type of the address operand and the offset immediate are defined in a subsequent paragraph in the existing prose (which is why I propose removing it).

The phrase "The alignment has the same type as the address and offset operands" is an implication of the sentence removed from the second paragraph.  Of course, that language could have been in error.

JSStats:
@lars-t-hansen The 'address index' seem to be the 'address operand', but yes it seems redundant as below these is "In wasm32, address operands and offset attributes have type `i32` ...".

Perhaps someone else can comment on "The alignment has the same type as the address and offset operands". They are both non-negative integers, but the range of the alignment is much more limited and the alignment is a power of two so I would not call it an `i32` or `i64`?

AstSemantics.md
```diff
@@ -171,11 +170,12 @@ will add support for wasm64 and thus
...
-alignment value which is the same as the memory attribute size is considered
-to be a *natural* alignment. The alignment applies to the effective address and
-not merely the address operand, i.e. the immediate offset is taken into account
-when considering alignment.
+of 2 alignment attribute. An alignment value which is the same as the memory
+attribute size is considered to be a *natural* alignment. The alignment applies
+to the effective address and not merely the address operand, i.e. the immediate
+offset is taken into account when considering alignment.
+
+The alignment has same type as the address and offset operands.
```
lukewagner: Perhaps to be even more clear, you could have "The alignment has the same type (determined by wasm32/wasm64, as described above) as the address and offset operands."

lars-t-hansen: OK.  Will fix.

lukewagner:
lgtm, thanks

Issue #677, opened 2016-04-29
Understanding the type system for `br`s


kripken:
I can't figure out why these three should pass (all from spec test suite):

````
(module (func              (block $l (br $l (i32.const 1)))))

(module (func (result i32) (block $l (br $l (i32.const 1)))))

(assert_invalid
  (module (func (result f32) (block $l (br_if $l (f32.const 0) (i32.const 1)))))
  "type mismatch"
)
````

Is there documentation for the type system somewhere that I should read, that would explain why the last is bad but not the first two? Basically, documentation for the types of `br`s in particular?

I've mostly just looked at the spec tests and gotten things working, but I'm trying to enable complete validation testing in binaryen, and I see failures that suggest I have no idea how this type system actually works :)

**Comment by JSStats** [comment]
For the first two, consider their equivalent simplifications which are valid. The consumer is not expecting values to the excess are discarded.

```
(func (block $l (br $l (i32.const 1)))) == (func (i32.const1))

(func (result i32) (block $l (br $l (i32.const 1)))) == (func (result i32) (i32.const 1))
```

The last case is invalid because the break operators accept only one expression, which might be a multiple-value expression in future. It is specifically not an implicit tuple constructor.

```
(func (result f32) (block $l (br_if $l (f32.const 0) (i32.const 1)))) != (func (result f32) (tuple (f32.const 0) (i32.const 1)))
```

I demonstrated the multiple-value type system in https://github.com/WebAssembly/sexpr-wasm-prototype/pull/66 and proved it fits a single pass compiler, and the break operators accept a single argument expression. If you want to understand how the current type system could extrapolates to multiple value then this might help.

It might be a little confusing that an arity count has just been added to the very common break operators. I argued against it. Go figure.

I no longer think it is productive to explore these issues. There is discussion of removing the void type, in which case no value could be implicitly discarded, and the issue goes away. An expressionless encoding would achieve the same.

**Comment by kripken** [comment]
Sorry, that just makes me more confused. What is a "simplification"? Is it just to explain the concept, or is that how a checker would actually work?

All I'm looking for here is the concrete rules of how I assign a type to `br`, `br_if`, `br_table` and `block`, and what are the error conditions there. For example, the type of `i32.const` is clearly `i32` - I just want simple rules for those other nodes. Or are there no simple rules?

**Comment by JSStats** [comment]
```
unreachable - type none
nop - type void
br - type none
br_if - type void
br_table - type none
[...]

Issue #669, opened 2016-04-22
Expressionless encoding


JSStats:
For the discussion of the design and pros and cons of an expressionless encoding. I'll update this opening comment with the current proposal and rationale.

The proposal:

* A set of operators (aka instructions) that read their arguments from local variables (aka registers) and write their results to local variables so that wasm code can be expressed without expressions.  These operators would return no values if used in an expression, e.g the same result as `nop`. For example: `(set_local $r1 (i32.add (get_local $v1) (get_local $v2)))` becomes `(i32.add $r1 $v1 $v2)`.

* Some opcodes would have have variations that accept immediate constants. This is noted to significantly improve the encoding efficiency. For example `(i32.add/lc $r1 $v1 10)`.

* The existing parametric operators `select` and `set_local` become type specific. Thus all local variable indexes are type specific and use a type specific index space which helps compression and simplifies a prediction algorithm by avoiding special consider for these operators.

* The exiting `if` operator might no longer be as useful and might be removed, or kept for efficiently encoding diamond control flow patterns, but might be subsumed by an alternative more efficient encoding of `block`, and one suggestion is to replace `if` by `unless` https://github.com/WebAssembly/design/issues/676.

* The break operators would have no expression value. Encoding the number of values would not be necessary, a small encoding advantage.

* Functions could only return values via the `return` operator. The proposed `end` operator might want to accept arguments.

* Function multiple return values are supported by the call operators writing to multiple local variables, and the `return` operator reading a local variable for each returned value.

* A predictor is defined that assumes roughly an expression stack allocation pattern in the local variables. A simple proposal is to track the index of the last local variable written and for each type. The arguments of an operator are predicted to access the last value written, with an assumption of single use and a stack allocation pattern, so each access hit decreases the predicted index of that type, and the arguments are assumed to be read last-to-first. The result is predicted to be the next index. This proposal avoids a stack-top register.

* The type of the last variable written might also be a good predictor of the type of the next operator. Where special opcodes are defined that omit arguments that are prediction hits, these might also usefully be selected based on type of the last variable written to handle the frequent case by using a smaller set of opcodes. For example, to select between `i32.add` `i64.add` `f32.add` and `f64.add` based on the last variable written.

Rationale:

[...]

Issue #667, opened 2016-04-21
block/br return value optimizations


kripken:
There was some discussion about block and br return values, and in particular how much benefit we can get from them, as both our current compiler options (asm2wasm and wasm-backend) don't use them or barely use them (asm2wasm uses block return values in rare cases, for asm.js sequences/commas, but never uses br values). I investigated this topic in the binaryen optimizer (`block-returns` branch; note code is not cleaned up yet).

My approach is to process control flow and see where it joins up, at the end of an if-else or a block, and then see if there is a set_local that flows through all those arriving branches (this would occur e.g. if an optimizing compiler emitted a phi node there). If so, we can use block and br return values to pass the result, instead of set_locals, and put a single set_local on the *outside* of the block or if. In other words, it unifies the set_locals, like this:

````
(block $out
  (if
    (..)
    (block
      (set_local $x (i32.const 10))
      (br $out)
    )
  )
  ..
  (set_local $x (i32.const 20))
)

==>

(set_local $x
  (block $out
    (if
      (..)
      (br $out (i32.const 10))
    )
    ..
    (i32.const 20)
  )
)
````

This removes set_locals, and can then allow further optimizations, e.g. the new set_local is now on the outside where it might be sunk into a later get_local, etc., and other stuff.

On BananaBread, this removes around 4% of set_locals. (This was better than I expected, so I ran it through a bunch of tests to verify it's working properly.)

Note that this is an underestimate of the opportunity here, as

1. It operates on the non-SSA form binaryen AST. Probably in SSA form more could be done (operating directly on phis).
2. It doesn't track variables through basic blocks, each is only considered in its own basic block.
3. This removes set_locals and get_locals, which might then allow reducing the total number of locals. But we can't test that since the binaryen optimizer doesn't have a register allocator (yet).

Overall, then, it seems like block/br return value optimizations can be pretty useful.

I also did some measurements for the topic of multiple return values. I can't optimize them in the current AST, so I don't have full and tested data. But I counted how many locations have a set_local that can be optimized, and how many there are at each - when there is more than one, multiple return values might be employed. Looks like 15.4% of potential locations have at least one set_local that can be optimized, and they have an average of 1.12 per location. In other words, it's common to have a single optimizable set_local, and while sometimes there is more than one, that is rare.

[...]

PR #681, opened 2016-05-02
Simplify br_if by removing its value operand.

sunfishcode:
#677 is the latest in a long and complex discussion (https://github.com/WebAssembly/spec/issues/179, https://github.com/WebAssembly/spec/pull/180, https://github.com/WebAssembly/spec/pull/271, and others) about how to impose high-level typing rules on an assembly-style branch instruction. I and several other people actively involved in WebAssembly's design have been confused about what the rules are, which is not auspicious for its broader adventures.

The current state, where `br_if` has a value operand, but no return value, may satisfy certain properties relevant to the high-level typing system, but it is awkward from an assembly perspective, which is the perspective that motivated the instruction in the first place. It's essentially creating an optimized path for the case where a value is live on the branch-taken side of the branch but not the fallthrough side.

The bigger picture is that I don't think we've had the appropriate chance to figure out how we really want this instruction to work. I propose that `br_if`'s arity be limited to 0 in the MVP. We can change it to permit 1 in the future when we work out how we want it to work.

Addressing the concern of code size, the numbers being quoted for block return values in general are in the neighborhood of 1%. I don't know the specific fraction of that which involves `br_if`, but even in the worst case, I claim this is rounding error when compared to the broader concern of designing a clean language.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -278,9 +278,8 @@ further see the parallel, note that a `br` to a `block`'s label is functionally
 equivalent to a labeled `break` in high-level languages in that a `br` simply
 breaks out of a `block`.
 
-Branches that exit a `block`, `loop`, or `br_table` may take a subexpression
-that yields a value for the exited construct. If present, it is the first operand
-before any others.
+`br` and `br_table` may take a subexpression that yields a value for the exited
+construct. If present, it is the first operand before any others.
 
 ### Yielding values from control constructs
 


rossberg-chromium:
I don't quite follow. The design discussion you cite is not specific to `br_if`, so what is gained by removing the argument just there?

Also, this question will be pretty much moot when we eliminate implicit dropping, which probably is the way to go.

sunfishcode:
If you think it would simplify things even more, then perhaps we should also remove the value operand from all branches, in the MVP. The data in https://github.com/WebAssembly/design/issues/667 shows that it wouldn't have a big impact on overall code size. Given time to properly design how this feature should work, and data demonstrating that it has significant value, we can reintroduce it.

Also, I don't see how implicit dropping affects this. The `br_if` problem I described above would still apply, as I understand it.

rossberg-chromium:
Not simplify "more", but at all. ;) FWIW, I mentioned that very option on #677 a couple of days ago, where Alon replied "I certainly hope not".

That other `br_if` "problem" you mention could as well be solved by adding a return value, like you proposed a while ago. On the other hand, the discussion about removing implicit dropping also involved making more operators void, including sets and stores. That means that this "problematic" case would actually become the rule. Are you opposed to that as well?

lukewagner:
What's the problem with having `br_if` return its value operand (if branch not taken)?  Is this a problem that would be easier to solve with the explicit-drop rules?  That seems to me to be the most regular design; I had to double-check `check.ml` to see that wasn't already the case...

JSStats:
Even if there were some small utility for returning the value argument from `br_if`, the producer will likely have many live-out definitions so without multiple-value support this only helps for one of these values and the rest need to use the local variable support anyway. This was part of the rationale for the expressionless proposal https://github.com/WebAssembly/design/issues/669, that multiple-value expression support will never make it into wasm, and even if it did it's still limited to single use cases or would require destructuring back to local variables anyway.

What's the harm in just making the current `br_if` a special `br0_if` opcode and leaving people to add new opcodes in future if they really want to explore returning a value? Then the immediate arities can be removed. The zero value case is so very common that developers will be using the operator table to create it anyway, and runtimes will likely have specialized handling for these variations in the operator table, so why not just add opcodes for it now. Another way to look at these is special drop/break operators, which developers would be creating in the operator table too as these are the common case.

With `block` and `br` still returning values, `br_if` can still support diamond patterns returning values, but `br` has the same type-system issues, so why not just have these also return no values.

For the predictor proposed to optimize the expressionless encoding, if a 'value' is written in order before the condition is written to a local variable then this value will be predicted to be the next value used, as if it were returned from `br_if`, but the cost of a prediction miss is now lower.

lukewagner:
Hah, actually looking again, it appears we've actually *already* implemented `br_if`-returning-it's-value in SM (it's the obvious/symmetric thing to do if branches have values).

JSStats:
@lukewagner Re 'obvious/symmetric thing to do', it's like branching in a stack machine in which case the target and fall-through both have the same number of values on the stack, and accepting value(s) to the break operators gives something similar but on the implicit expression value stack. But in wasm without multiple-value expressions we can not have two or more values on this stack, and for example have the target and fall-through push another value and then make a call consuming these. The predictor for the expressionless encoding assumes a stack-like allocation so could achieve good prediction hits for such cases, and if there's a miss it is a small cost as operators accept immediate local variables, and it can be easily decoded into uses of local variables rather than a stack machine.

JSStats:
@lukewagner Re 'obvious/symmetric thing to do', it's like branching in a stack machine in which case the target and fall-through both have the same number of values on the stack, and accepting value(s) to the break operators gives something similar but on the implicit expression value stack. But in wasm without multiple-value expressions we can not have two or more values on this stack, and for example have the target and fall-through push another value and then make a call consuming these. The predictor for the expressionless encoding assumes a stack-like allocation so could achieve good prediction hits for such cases, and if there's a miss it is a small cost as operators accept immediate local variables, and it can be easily decoded into uses of local variables rather than a stack machine.

sunfishcode:
The data in #677 seems likely to end up in the neighborhood of 1% overall code size. And for that, we've ventured into uncharted territory here -- no other language or bytecode or ISA I know of has conditional branches with value operands as wasm does.

Making br_if return its operand sounds like an improvement. However, if it interacts with explicit dropping, then perhaps there could be a different problem. The safe thing would be to defer it until we can learn more, as it would seem the cost is low.

JSStats:
@sunfishcode Re 'no other language or bytecode or ISA I know of has conditional branches with value operands as wasm does': viewing the wasm expressions as a stack with restricted use, then CIL is a bytecode example that allows a conditional branch with values on the stack which could encode a wasm `br_if` passing values to the target and the pass-through.

JSStats:
Just to give a concrete example:

`(block $l (i32.add (br_if $l <v3> <cond>) <c>))`

could be expressed in a stack machine as:

```
<v3>
<cond>
br_if $l
<c>
add
$l:
```

kripken:
> we've ventured into uncharted territory here -- no other language or bytecode or ISA I know of has conditional branches with value operands as wasm does.

I think you can look at branches with values as phis. More specifically, if A branches either to B with a value, or otherwise continues to C without a value, then at B we would have a phi (consisting of all the other paths branching into B), and C would not have any phis.

In that perspective, I don't think there's anything odd about a conditional br with a value. The value "belongs" to the phi, which belongs to the target, not to the brancher. And everything branching to the target will provide a value, and so we have proper symmetry.

Instead of SSA form and phis, though, we have non-SSA locals and branches. But the parallel is there. And there is precedent even without the parallel, in Swift, which has SSA form but no phis, instead it has basic block arguments - branches are like calls in that they provide values that are received on the other side. A br with a value is pretty much that. And that's the same with or without a condition on the br.

JSStats:
@kripken The Swift blocks sound similar to https://github.com/stoklund/cretonne/blob/master/docs/langref.rst  But this has separate SSA variables, plus the locals (local stack), and the wasm expression results seem to match the SSA variables. It also claims to be able to convert uses of the locals into SSA form which is what the expressionless wasm encoding depends on and thus has no need for the SSA variables. The question is what is the most compact encoding to get the job done here, and with a low complexity for the runtime, and while not being too complicated to decode to something readable?

rossberg-chromium:
On 2 May 2016 at 21:01, Luke Wagner <notifications@github.com> wrote:

> What's the problem with having br_if return its value operand (if branch
> not taken)? Is this a problem that would be easier to solve with the
> explicit-drop rules? That seems to me to be the most regular design; I had
> to double-check check.ml to see that wasn't already the case...
>

That was #237. We deferred it because of technicalities with break typing
that needed resolving first. But with #271 landed, that has happened, so
it's time to revisit it.

However, isn't this feature obsolete if we want to go down the explicit
dropping road?

sunfishcode:
@kripken I agree that phis and wasm's `br_if` have a relationship, but there are enough conceptual differences that treating them as the same is confusing to me. A consumer can construct SSA form in a simple manner from `br_if`, though it may not be minimal. A producer with SSA form can't always translate a phi into a `br_if` operand.

@JSStats That's a good example, and if `br_if` is made to return its value (in the AST sense) or leave 1 value on the stack (in a stack machine sense), it would be close to that.

@rossberg-chromium My understanding of the explicit dropping idea is that it's fairly orthogonal to this issue. Having more values in more places might require more explicit drops, but are there other interactions?

rossberg-chromium:
@sunfishcode, well, our assumption so far was that explicit dropping is only feasible if we reduce the number of "spurious" non-void operators, particularly those whose result isn't providing any information. That includes sets and stores, but clearly, `br_if` is in the same class (you'd probably need a drop around the majority of `br_if`s with argument if you make them have a result).

lukewagner:
@rossberg-chromium You won't need to explicitly drop `br_if` with arity 0 which would, it seems, be the 99% majority case.  Otherwise, if we're in this 1% case where there actually is a live value across the branch, it stands to reason that it could be used by both successors of the branch (otherwise, you'd expect it to be in the branch in which it was used).  So this seems like `call`: if the drop-requiring case were more common, we'd add a drop form, but it doesn't seem to be so we can hold off.

rossberg-chromium:
@lukewagner, I don't think that is the relevant metric. The interesting metric is: the percentage of br_if uses of arity 1 whose return value is actually used. If that is way below, say, 30% (implying that way above 70% would require a drop), then it doesn't make sense to return the value by default. Rather, have those few cases fall back to whatever more general mechanism we'll provide to consume a value multiple times.

lukewagner:
@rossberg-chromium Yes, but I think we'll have a hard time convincing ourselves that we have a reliable measurement of that metric.  I feel much more confident that branch-with-value is going to be rare which, if anything, supports the original point of this issue that maybe we should, as with multi-value, hold off committing to anything (by fixing arity=0) until we have a better understanding of how code generators can actually use these features.

kripken:
@rossberg-chromium 

> well, our assumption so far was that explicit dropping is only feasible if we reduce the number of "spurious" non-void operators

Sounds interesting, where is that being discussed?

sunfishcode:
There are interesting discussions here about fixing the underlying issue here by having `br_if` return its value operand. However, there are open questions about how that would interacts with explicit dropping, and idea which is still being explored, and there are open questions about how producers are going to want to use it.

`br_if`'s value operand isn't necessary for Minimum Viability, and there's significant uncertainty about how best to design it. The safe thing to do here would be to semantically limiting its arity to 0, so that we don't end up spec'ing something that will conflict with future answers to the above questions.

kripken:
If `br_if` values can interact with explicit dropping, then isn't it best to first have the explicit dropping discussion, then get back to this?

Otherwise making a change now seems premature, especially given dropping might be a discussion coming up soon. And removing values from `br_if` introduces an odd asymmetry, so we should only do it if it's properly justified, and not preemptively.

sunfishcode:
It's a question of priorities, I suppose. Closing this, to wait for drop operators, at which point I'll re-evaluate where we are and consider a new proposal.

sunfishcode:
With @kripken indicating being "[ok](https://github.com/WebAssembly/design/pull/703#issuecomment-226807354)" with this proposal, I am re-opening this PR.

As a quick summary, this issue proposes removing br_if's value operand (by fixing its arity to 0) for the MVP. We'd retain the flexibility to reintroduce the value operand in the future should we choose to.

Issue #333, opened 2015-09-05
Imported global constants


JSStats:
asm.js supports global constants, and these can be initialized from imported value, but this seems to be lacking in wasm. For example, these can be used to pass in index masks that are based on the chosen linear memory length, and currently v8 and jsc can specialize on these and bake them into optimized code. For Odin these need to be initialized with literals to have them baked in. Without this support the MVP would not be viable for runtime chosen linear memory sizes using index masking.

sunfishcode:
While explicit address masking may have performance advantages for asm.js implementations on some platforms today, WebAssembly is a new platform being designed with long-term goals as well. Since explicit address masking is limited to power-of-two sizes, making it a mandatory part of the platform would be an onerous limitation, and making it an optional part of the platform would mean that implementations have to support multiple addressing modes, which would add complexity to the platform and constrain long-term flexibility.

JSStats:
I have been exploring this path and understand the matter well, and for those that do not know the history this has been a matter of dispute for some time, I had to split the the Emscripten support away from the main code base, and I was excluded from the early wasm development. I believe there is little technical merit in the opposition to this support and that the opposition will damage the wasm project.

The use of index masking is a matter for the application developer to choose. For example, a developer might choose to use this style to take advantage of pointer tagging, using the masking to both strip the tag and ensure the index is within bounds. This is a performance optimization that the developer chooses. This is an application specific optimization that requires some minimal support from wasm. If wasm is to have a good future then I believe it needs to support this code style well.

A wasm implementation that does not have the address space to allocate a 4G+ region and use page protection to avoid inline bounds check will need to use inline bounds checking. Masking an index before use is an easy way to prove to the compiler that the index is within bounds and that the index mask is not needed. I have implemented this on Odin, v8, and jsc, and the performance is competitive - an application developer may well want to use this style for balanced performance. Index masking is also relatively easy to hoist.

Being able to pass in constants, based on the chosen linear memory size is very important for the compiler to be able to optimize away bounds checking. It is important even without the use of index masking. For example, a function may want to compare an index to a constant based on the linear memory size on entry and with the propagation of constraints (such as ABCD style constraints) bounds checks might be optimized away.

[...]

PR #682, opened 2016-05-05
Add JS Memory and Table API, support dynamic linking

lukewagner:
This PR adds first-class `WebAssembly.Memory` and `WebAssembly.Table` objects along with the ability for modules to import and export any number of these from a given module.  This generalizes the current linear memory / table design wherein a module cannot import linear memory and is limited to 1 linear memory / table for internal use and export.  This PR also generalizes tables to be resizable (symmetric to memory) and parameterized by an element type which is, in the MVP, restricted to either "any callable function" or "any callable function with a given signature".  Since there can be multiple memories/tables, the memory and table opcodes are given an index immediate.

With this PR, it should be possible to implement dynamic linking in a toolchain so this PR also updates DynamicLinking.md to explain how.  See the note at the bottom of DynamicLinking.md for an obvious omission which we should consider adding (viz., constant-value imports/exports, which would be used to position data segments and function-table-elements at load time).

This PR also punts on the precise way to describe a signature in JS when calling the `WebAssembly.Table` constructor, saving that for its own PR where we can bikeshed; see TODO in `ToTypeDescriptor`.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -273,7 +273,10 @@ or to put all functions into the same table (with no signature). The former
 strategy allows the engine to eliminate dynamic signature-mismatch checks while
 the latter strategy gives each function a unique index (which may be necessary
 for C/C++ compatibility) and admits simpler and potentially more space-efficient
-[dynamic linking](DynamicLinking.md).
+[dynamic linking](DynamicLinking.md). Toolchains can also create many
+minimally-sized tables to implement a variety of [CFI schemes](http://clang.llvm.org/docs/ControlFlowIntegrity.html#forward-edge-cfi-for-virtual-calls)
+that constrain the set of valid call targets for a given `call_table`.
+
 
 ## Local variables
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -256,10 +256,12 @@ opaque handles/references at the cost of a bounds-checked table indirection.
 The table's element type dynamically constrains the type of elements stored 
 in the table and allows engines to avoid some type checks on table use. 
 When a WebAssembly value is stored in a table, the value's type must precisely
-match the element type. Depending on the operator/API used to store the value,
-this check may be static or dynamic. Host environments may also allow storing
-non-WebAssembly values in tables in which case, as with [imports](Modules.md#imports),
-the meaning of using the value is defined by the host environment.
+match the element type. Just like linear memory, updates to a table are
+observed immediately by all instances that reference the table. Depending on the
+operator/API used to store the value, this check may be static or dynamic. Host
+environments may also allow storing non-WebAssembly values in tables in which
+case, as with [imports](Modules.md#imports), the meaning of using the value is
+defined by the host environment.
 
 Every WebAssembly [instance](Modules.md) has one specially-designated *default*
 table which is indexed by [`call_indirect`](#calls) and other future
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -253,7 +253,7 @@ accessed by WebAssembly code indirectly through an integer index. This feature
 bridges the gap between low-level, untrusted linear memory and high-level
 opaque handles/references at the cost of a bounds-checked table indirection.
 
-The table's element type dynamically constrains the type of elements stored 
+The table's element type constrains the type of elements stored 
 in the table and allows engines to avoid some type checks on table use. 
 When a WebAssembly value is stored in a table, the value's type must precisely
 match the element type. Just like linear memory, updates to a table are
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -251,8 +251,9 @@ opaque handles/references at the cost of a bounds-checked table indirection.
 
 The table's element type dynamically constrains the type of elements stored 
 in the table and allows engines to avoid some type checks on table use. When
-tables are mutated, any stored value is necessarily coerced (possibly trapping)
-to the element type.
+tables are mutated, any stored value must match the element type (by one of
+static validation constraint, trapping dynamic type validation check, or dynamic
+coercion).
 
 In the MVP, the set of operations and types for tables is limited:
 * tables may only be accessed via [`call_table`](#calls);
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -413,8 +413,8 @@ different module which necessarily has a separate [types section](BinaryEncoding
 and thus index space of types, the signature match must compare the underlying 
 [`func_type`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/types.ml#L5).
 As noted [above](#table), table elements may also be host-environment-defined
-values in which case the meaning of a call is defined by the host-environment,
-much like calling an import.
+values in which case the meaning of a call (and how the signature is checked)
+is defined by the host-environment, much like calling an import.
 
 In the MVP, the single `call_indirect` operator accesses the [default table](#table).
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -307,7 +307,7 @@ cannot be arbitrarily aliased as bits.
 Global variables are accessed via an integer index into the module-defined 
 [global index space](Modules.md#global-index-space). Global variables can 
 either be [imported](Modules.md#imports) or [defined inside the module](Modules.md#global-section).
-After import or definition, there is no difference when calling into a table.
+After import or definition, there is no difference when accessing a global.
 
   * `get_global`: get the current value of a global variable
   * `set_global`: set the current value of a global variable
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -80,29 +80,30 @@ operators.
 
 ## Linear Memory
 
-The main storage of a WebAssembly instance, called the *linear memory*, is a
-contiguous, byte-addressable range of memory spanning from offset `0` and
-extending up to a varying *memory size*.
+A WebAssembly [instance](Modules.md) contains zero or more *linear memories*.
+A linear memory is is a contiguous, byte-addressable range of memory spanning
+from offset `0` and extending up to a varying *memory size*.
 This size always is a multiple of the WebAssembly page size,
-which is 64KiB on all engines (though large page support may be added in the
-[future](FutureFeatures.md#large-page-support)).
-The initial state of linear memory is specified by the
-[module](Modules.md#linear-memory-section), and it can be dynamically grown by
+which is fixed to 64KiB (though large page support may be added in an opt-in
+manner in the [future](FutureFeatures.md#large-page-support)).
+The initial state of a linear memory is specified by the
+[module](Modules.md#linear-memory-section), and can be dynamically grown by
 the [`grow_memory`](AstSemantics.md#resizing) operator.
 
-The linear memory can be considered
-to be an untyped array of bytes, and it is unspecified how embedders map this
-array into their process' own [virtual memory][]. The linear memory is
-sandboxed; it does not alias the execution engine's internal data structures,
-the execution stack, local variables, or other process memory.
+A module can define or import multiple linear memories and each memory access
+operator statically specifies which linear memory to operate on with a 
+[linear memory index](Modules.md#index-spaces) immediate.
+
+A linear memory can be considered to be an untyped array of bytes, and it is
+unspecified how embedders map this array into their process' own [virtual
+memory][]. Linear memory is sandboxed; it does not alias other linear memories,
+the execution engine's internal data structures, the execution stack, local
+variables, or other process memory.
 
   [virtual memory]: https://en.wikipedia.org/wiki/Virtual_memory
 
-In the MVP, linear memory is not shared between threads of execution. Separate
-instances can execute in separate threads but have their own linear memory and can
-only communicate through messaging, e.g. in browsers using `postMessage`. It
-will be possible to share linear memory between threads of execution when
-[threads](PostMVP.md#threads) are added.
+In the MVP, linear memory cannot be shared between threads of execution.
+The addition of [threads](PostMVP.md#threads) will allow this.
 
 ### Linear Memory Accesses
 
@@ -110,6 +111,9 @@ Linear memory access is accomplished with explicit `load` and `store` operators.
 Integer loads can specify a *storage size* which is smaller than the result type as
 well as a signedness which determines whether the bytes are sign- or zero-
 extended into the result type.
+
+Each `load` and `store` specifies the [linear memory index](Modules.md#index-spaces)
+to operate on with an unsigned integer immediate.
 
   * `i32.load8_s`: load 1 byte and sign-extend i8 to i32
   * `i32.load8_u`: load 1 byte and zero-extend i8 to i32
@@ -203,21 +207,25 @@ Out of bounds accesses trap.
 ### Resizing
 
 In the MVP, linear memory can be resized by a `grow_memory` operator. The
-operand to this operator is in units of the WebAssembly page size,
+`grow_memory` operator specifies which memory to resize with a
+[linear memory index](Modules.md#index-spaces) immediate. The dynamic
+operand to `grow_memory` is in units of the WebAssembly page size,
 which is defined to be 64KiB (though large page support may be added in 
 the [future](FutureFeatures.md#large-page-support)).
 
  * `grow_memory` : grow linear memory by a given unsigned delta of pages.
     Return the previous memory size in units of pages or -1 on failure.
 
-When a maximum memory size is declared in the [memory section](Module.md#linear-memory-section),
+When a maximum memory size is declared in the [memory section](Modules.md#linear-memory-sections),
 `grow_memory` must fail if it would grow past the maximum. However,
 `grow_memory` may still fail before the maximum if it was not possible to
 reserve the space up front or if enabling the reserved memory fails.
 When there is no maximum memory size declared, `grow_memory` is expected
 to perform a system allocation which may fail.
 
-The current size of the linear memory can be queried by the following operator:
+The current size of a given linear memory (specified by a
+[linear memory index](Modules.md#index-spaces)) can be queried by the following
+operator:
 
   * `current_memory` : return the current memory size in units of pages.
 
@@ -231,6 +239,40 @@ In the MVP, memory can only be grown. After the MVP, a memory shrinking
 operator may be added. However, due to normal fragmentation, applications are
 instead expected release unused physical pages from the working set using the
 [`discard`](FutureFeatures.md#finer-grained-control-over-memory) future feature.
+
+## Tables
+
+A *table* is similar to a linear memory whose elements, instead of being bytes,
+are opaque values of a particular *table element type*. This allows the table to
+contain values like GC references, raw OS handles, or native pointers that are
+accessed by WebAssembly code indirectly through an integer index. This feature
+bridges the gap between low-level, untrusted linear memory and high-level
+opaque handles/references at the cost of a bounds-checked table indirection.
+
+The table's element type dynamically constrains the type of elements stored 
+in the table and allows engines to avoid some type checks on table use. When
+tables are mutated, any stored value is necessarily coerced (possibly trapping)
+to the element type.
+
+In the MVP, the set of operations and types for tables is limited:
+* tables may only be accessed via [`call_table`](#calls);
+* the only allowed table element types are "function" and
+  "function with signature `X`";
+* tables may not be directly mutated or resized from WebAssembly code;
+  initially this must be done through the host environment (e.g., the
+  the `WebAssembly` [JavaScript API](JS.md#webassemblytable-objects).
+
+These restrictions should be relaxed in the 
+[future](FutureFeatures.md#more-table-operators-and-types). In the MVP,
+the primary purpose of tables is to implement indirect function calls
+in C/C++, using an integer index as the pointer-to-function and the table
+to hold the array of indirectly-callable functions. The compiler may either
+choose to group functions by signature (into N tables-with-signatures)
+or to put all functions into the same table (with no signature). The former
+strategy allows the engine to eliminate dynamic signature-mismatch checks while
+the latter strategy gives each function a unique index (which may be necessary
+for C/C++ compatibility) and admits simpler and potentially more space-efficient
+[dynamic linking](DynamicLinking.md).
 
 ## Local variables
 
@@ -314,43 +356,27 @@ explicit accesses to linear memory.
 In the MVP, the length of the return types sequence may only be 0 or 1. This
 restriction may be lifted in the future.
 
-Direct calls to a function specify the callee by index into a *main function table*.
+Direct calls to a function specify the callee by index into the 
+[function index space](Modules.md#index-spaces).
 
   * `call`: call function directly
 
 A direct call to a function with a mismatched signature is a module verification error.
 
-Like direct calls, calls to [imports](Modules.md#imports-and-exports) specify
-the callee by index into an *imported function table* defined by the sequence of import
-declarations in the module import section. A direct call to an imported function with a
-mismatched signature is a module verification error.
+Indirect calls to a function indicate the callee with an `i32` index into
+the [table](#tables) specified by a [table index](BinaryEncoding.md#index-spaces)
+immediate. The *expected* signature of the target function (specified by its
+index in the [types section](BinaryEncoding.md#type-section)) is given as a
+second immediate. 
 
-  * `call_import` : call imported function directly
+  * `call_table`: call function element in a table
 
-Indirect calls allow calling target functions that are unknown at compile time.
-The target function is an expression of value type `i32` and is always the first
-input into the indirect call.
-
-A `call_indirect` specifies the *expected* signature of the target function with
-an index into a *signature table* defined by the module. An indirect call to a
-function with a mismatched signature causes a trap.
-
-  * `call_indirect`: call function indirectly
-
-Functions from the main function table are made addressable by defining an
-*indirect function table* that consists of a sequence of indices into the
-module's main function table. A function from the main table may appear more
-than once in the indirect function table. Functions not appearing in the
-indirect function table cannot be called indirectly.
-
-In the MVP, indices into the indirect function table are local to a single
-module, so wasm modules may use `i32` constants to refer to entries in their own
-indirect function table. The [dynamic linking](DynamicLinking.md) feature is
-necessary for two modules to pass function pointers back and forth. This will
-mean concatenating indirect function tables and adding an operator `address_of`
-that computes the absolute index into the concatenated table from an index in a
-module's local indirect table. JITing may also mean appending more functions to
-the end of the indirect function table.
+The specified [table](#tables) must have either a "function" or "function with
+signature `X`" element type (in the MVP, these are the only two
+possibilities, but in the future, there may be other kinds of tables).
+If the table element type has a signature, it must match the signature of the
+call. Otherwise, the signature is checked at runtime and a signature mismatch
+causes a trap.
 
 Multiple return value calls will be possible, though possibly not in the
 MVP. The details of multiple-return-value calls needs clarification. Calling a
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -275,7 +275,7 @@ In the MVP, the primary purpose of tables is to implement indirect function
 calls in C/C++ using an integer index as the pointer-to-function and the table
 to hold the array of indirectly-callable functions. Thus, in the MVP:
 * tables may only be accessed from WebAssembly code via [`call_indirect`](#calls);
-* the only allowed table element type is "function" (no signature);
+* the only allowed table element type is `anyfunc` (function with any signature);
 * tables may not be directly mutated or resized from WebAssembly code;
   this can only be done through the host environment (e.g., the
   the `WebAssembly` [JavaScript API](JS.md#webassemblytable-objects)).
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -254,10 +254,12 @@ bridges the gap between low-level, untrusted linear memory and high-level
 opaque handles/references at the cost of a bounds-checked table indirection.
 
 The table's element type dynamically constrains the type of elements stored 
-in the table and allows engines to avoid some type checks on table use. When
-tables are mutated, any stored value must match the element type by either a
-static validation constraint, trapping dynamic type validation check, or dynamic
-coercion (depending on where the mutation occurs).
+in the table and allows engines to avoid some type checks on table use. 
+When a WebAssembly value is stored in a table, the value's type must precisely
+match the element type. Depending on the operator/API used to store the value,
+this check may be static or dynamic. Host environments may also allow storing
+non-WebAssembly values in tables in which case, as with [imports](Modules.md#imports),
+the meaning of using the value is defined by the host environment.
 
 Every WebAssembly [instance](Modules.md) has one specially-designated *default*
 table which is indexed by [`call_indirect`](#calls) and other future
@@ -408,6 +410,9 @@ signature and and trapping if there is a mismatch. Since the callee may be in a
 different module which necessarily has a separate [types section](BinaryEncoding.md#type-section),
 and thus index space of types, the signature match must compare the underlying 
 [`func_type`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/types.ml#L5).
+As noted [above](#table), table elements may also be host-environment-defined
+values in which case the meaning of a call is defined by the host-environment,
+much like calling an import.
 
 In the MVP, the single `call_indirect` operator accesses the [default table](#table).
 
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -80,19 +80,14 @@ operators.
 
 ## Linear Memory
 
-A WebAssembly [instance](Modules.md) contains zero or more *linear memories*.
-A linear memory is is a contiguous, byte-addressable range of memory spanning
-from offset `0` and extending up to a varying *memory size*.
-This size always is a multiple of the WebAssembly page size,
-which is fixed to 64KiB (though large page support may be added in an opt-in
-manner in the [future](FutureFeatures.md#large-page-support)).
-The initial state of a linear memory is specified by the
-[module](Modules.md#linear-memory-section), and can be dynamically grown by
-the [`grow_memory`](AstSemantics.md#resizing) operator.
-
-A module can define or import multiple linear memories and each memory access
-operator statically specifies which linear memory to operate on with a 
-[linear memory index](Modules.md#index-spaces) immediate.
+A *linear memory* is a contiguous, byte-addressable range of memory spanning
+from offset `0` and extending up to a varying *memory size*. This size is always 
+a multiple of the WebAssembly page size, which is fixed to 64KiB (though large
+page support may be added in an opt-in manner in the 
+[future](FutureFeatures.md#large-page-support)). The initial state of a linear
+memory is defined by the module's [linear memory](Modules.md#linear-memory-section) and
+[data](Modules.md#data-section) sections. The memory size can be dynamically
+increased by the [`grow_memory`](AstSemantics.md#resizing) operator.
 
 A linear memory can be considered to be an untyped array of bytes, and it is
 unspecified how embedders map this array into their process' own [virtual
@@ -101,6 +96,18 @@ the execution engine's internal data structures, the execution stack, local
 variables, or other process memory.
 
   [virtual memory]: https://en.wikipedia.org/wiki/Virtual_memory
+
+Every WebAssembly [instance](Modules.md) has one specially-designated *default* 
+linear memory which is the linear memory accessed by all the 
+[memory operators below](#linear-memory-access). In the MVP, there are *only*
+default linear memories but [new memory operators](FutureFeatures.md#multiple-tables-and-memories)
+will be added after the MVP which can also access non-default memories.
+
+Linear memories (default or otherwise) can either be [imported](Modules.md#imports)
+or [defined inside the module](Modules.md#linear-memory-section), with defaultness
+indicated by a flag on the import or definition. After import or definition,
+there is no difference when accessing a linear memory whether it was imported or
+defined internally.
 
 In the MVP, linear memory cannot be shared between threads of execution.
 The addition of [threads](PostMVP.md#threads) will allow this.
@@ -111,9 +118,6 @@ Linear memory access is accomplished with explicit `load` and `store` operators.
 Integer loads can specify a *storage size* which is smaller than the result type as
 well as a signedness which determines whether the bytes are sign- or zero-
 extended into the result type.
-
-Each `load` and `store` specifies the [linear memory index](Modules.md#index-spaces)
-to operate on with an unsigned integer immediate.
 
   * `i32.load8_s`: load 1 byte and sign-extend i8 to i32
   * `i32.load8_u`: load 1 byte and zero-extend i8 to i32
@@ -146,6 +150,8 @@ size in which case integer wrapping is implied.
 
 In addition to storing to memory, store instructions produce a value which is their 
 `value` input operand before wrapping.
+
+The above operators operate on the [default linear memory](#linear-memory).
 
 ### Addressing
 
@@ -207,25 +213,21 @@ Out of bounds accesses trap.
 ### Resizing
 
 In the MVP, linear memory can be resized by a `grow_memory` operator. The
-`grow_memory` operator specifies which memory to resize with a
-[linear memory index](Modules.md#index-spaces) immediate. The dynamic
-operand to `grow_memory` is in units of the WebAssembly page size,
+operand to this operator is in units of the WebAssembly page size,
 which is defined to be 64KiB (though large page support may be added in 
 the [future](FutureFeatures.md#large-page-support)).
 
  * `grow_memory` : grow linear memory by a given unsigned delta of pages.
     Return the previous memory size in units of pages or -1 on failure.
 
-When a maximum memory size is declared in the [memory section](Modules.md#linear-memory-sections),
+When a linear memory has a declared [maximum memory size](Modules.md#linear-memory-section),
 `grow_memory` must fail if it would grow past the maximum. However,
 `grow_memory` may still fail before the maximum if it was not possible to
 reserve the space up front or if enabling the reserved memory fails.
 When there is no maximum memory size declared, `grow_memory` is expected
 to perform a system allocation which may fail.
 
-The current size of a given linear memory (specified by a
-[linear memory index](Modules.md#index-spaces)) can be queried by the following
-operator:
+The current size of the linear memory can be queried by the following operator:
 
   * `current_memory` : return the current memory size in units of pages.
 
@@ -240,50 +242,49 @@ operator may be added. However, due to normal fragmentation, applications are
 instead expected release unused physical pages from the working set using the
 [`discard`](FutureFeatures.md#finer-grained-control-over-memory) future feature.
 
-## Tables
+The above operators operate on the [default linear memory](#linear-memory).
+
+## Table
 
 A *table* is similar to a linear memory whose elements, instead of being bytes,
 are opaque values of a particular *table element type*. This allows the table to
-contain values like GC references, raw OS handles, or native pointers that are
+contain values—like GC references, raw OS handles, or native pointers—that are
 accessed by WebAssembly code indirectly through an integer index. This feature
 bridges the gap between low-level, untrusted linear memory and high-level
 opaque handles/references at the cost of a bounds-checked table indirection.
 
 The table's element type dynamically constrains the type of elements stored 
 in the table and allows engines to avoid some type checks on table use. When
-tables are mutated, any stored value must match the element type (by one of
+tables are mutated, any stored value must match the element type by either a
 static validation constraint, trapping dynamic type validation check, or dynamic
-coercion).
+coercion (depending on where the mutation occurs).
 
-In the MVP, the set of operations and types for tables is limited:
-* tables may only be accessed via [`call_table`](#calls);
-* the only allowed table element types are "function" and
-  "function with signature `X`";
+Every WebAssembly [instance](Modules.md) has one specially-designated *default*
+table which is indexed by [`call_indirect`](#calls) and other future
+table operators. Tables can either be [imported](Modules.md#imports) or 
+[defined inside the module](Modules.md#table-section), with defaultness
+indicated by a flag on the import or definition. After import or definition,
+there is no difference when calling into a table whether it was imported or
+defined internally.
+
+In the MVP, the primary purpose of tables is to implement indirect function
+calls in C/C++ using an integer index as the pointer-to-function and the table
+to hold the array of indirectly-callable functions. Thus, in the MVP:
+* tables may only be accessed from WebAssembly code via [`call_indirect`](#calls);
+* the only allowed table element type is "function" (no signature);
 * tables may not be directly mutated or resized from WebAssembly code;
-  initially this must be done through the host environment (e.g., the
-  the `WebAssembly` [JavaScript API](JS.md#webassemblytable-objects).
+  this can only be done through the host environment (e.g., the
+  the `WebAssembly` [JavaScript API](JS.md#webassemblytable-objects)).
 
-These restrictions should be relaxed in the 
-[future](FutureFeatures.md#more-table-operators-and-types). In the MVP,
-the primary purpose of tables is to implement indirect function calls
-in C/C++, using an integer index as the pointer-to-function and the table
-to hold the array of indirectly-callable functions. The compiler may either
-choose to group functions by signature (into N tables-with-signatures)
-or to put all functions into the same table (with no signature). The former
-strategy allows the engine to eliminate dynamic signature-mismatch checks while
-the latter strategy gives each function a unique index (which may be necessary
-for C/C++ compatibility) and admits simpler and potentially more space-efficient
-[dynamic linking](DynamicLinking.md). Toolchains can also create many
-minimally-sized tables to implement a variety of [CFI schemes](http://clang.llvm.org/docs/ControlFlowIntegrity.html#forward-edge-cfi-for-virtual-calls)
-that constrain the set of valid call targets for a given `call_table`.
-
+These restrictions will be relaxed in the 
+[future](FutureFeatures.md#more-table-operators-and-types). 
 
 ## Local variables
 
-Each function has a fixed, pre-declared number of local variables which occupy a single
+Each function has a fixed, pre-declared number of *local variables* which occupy a single
 index space local to the function. Parameters are addressed as local variables. Local
 variables do not have addresses and are not aliased by linear memory. Local
-variables have value types and are initialized to the appropriate zero value for their
+variables have [value types](#types) and are initialized to the appropriate zero value for their
 type at the beginning of the function, except parameters which are initialized to the values
 of the arguments passed to the function.
 
@@ -293,6 +294,32 @@ of the arguments passed to the function.
 The details of index space for local variables and their types will be further clarified,
 e.g. whether locals with type `i32` and `i64` must be contiguous and separate from
 others, etc.
+
+## Global variables
+
+A *global variable* stores a single value of a fixed [value type](#types) and may be
+declared either *mutable* or *immutable*. This provides WebAssembly with memory
+locations that are disjoint from any [linear memory](#linear-memory) and thus
+cannot be arbitrarily aliased as bits.
+
+Global variables are accessed via an integer index into the module-defined 
+[global index space](Modules.md#global-index-space). Global variables can 
+either be [imported](Modules.md#imports) or [defined inside the module](Modules.md#global-section).
+After import or definition, there is no difference when calling into a table.
+
+  * `get_global`: get the current value of a global variable
+  * `set_global`: set the current value of a global variable
+
+It is a validation error for a `set_global` to index an immutable global variable.
+
+In the MVP, the primary use case of global variables is to represent
+instantiation-time immutable values as a necessary building block of
+[dynamic linking](DynamicLinking.md).
+
+After the MVP, when [reference types](GC.md) are added to the set of [value types](#types),
+global variables will be necessary to allow sharing reference types between
+[threads](PostMVP.md#threads) since shared linear memory cannot load or store
+references.
 
 ## Control flow structures
 
@@ -360,31 +387,29 @@ explicit accesses to linear memory.
 In the MVP, the length of the return types sequence may only be 0 or 1. This
 restriction may be lifted in the future.
 
-Direct calls to a function specify the callee by index into the 
-[function index space](Modules.md#index-spaces).
+Direct calls to a function specify the callee by an index into the 
+[function index space](Modules.md#function-index-space).
 
   * `call`: call function directly
 
 A direct call to a function with a mismatched signature is a module verification error.
 
 Indirect calls to a function indicate the callee with an `i32` index into
-the [table](#tables) specified by a [table index](BinaryEncoding.md#index-spaces)
-immediate. The *expected* signature of the target function (specified by its
-index in the [types section](BinaryEncoding.md#type-section)) is given as a
-second immediate. 
+a [table](#table). The *expected* signature of the target function (specified
+by its index in the [types section](BinaryEncoding.md#type-section)) is given as
+a second immediate.
 
-  * `call_table`: call function element in a table
+  * `call_indirect`: call function indirectly
 
-The specified [table](#tables) must have either a "function" or "function with
-signature `X`" element type (in the MVP, these are the only two
-possibilities, but in the future, there may be other kinds of tables).
-If the table element type has a signature, it must match the signature of the
-call. Otherwise, the signature is checked each time the `call_table` is executed,
-comparing the fixed caller signature and the dynamic callee signature and trapping
-if there is a mismatch. Since the callee may be in a different module which
-necessarily has a separate [types section](BinaryEncoding.md#type-section), and
-thus index space of types, the signature match must compare the underlying 
+Unlike `call`, which checks that the caller and callee signatures match
+statically as part of validation, `call_indirect` checks for signature match
+*dynamically*, comparing the caller's expected signature with the callee function's
+signature and and trapping if there is a mismatch. Since the callee may be in a
+different module which necessarily has a separate [types section](BinaryEncoding.md#type-section),
+and thus index space of types, the signature match must compare the underlying 
 [`func_type`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/types.ml#L5).
+
+In the MVP, the single `call_indirect` operator accesses the [default table](#table).
 
 Multiple return value calls will be possible, though possibly not in the
 MVP. The details of multiple-return-value calls needs clarification. Calling a
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -379,8 +379,12 @@ The specified [table](#tables) must have either a "function" or "function with
 signature `X`" element type (in the MVP, these are the only two
 possibilities, but in the future, there may be other kinds of tables).
 If the table element type has a signature, it must match the signature of the
-call. Otherwise, the signature is checked at runtime and a signature mismatch
-causes a trap.
+call. Otherwise, the signature is checked each time the `call_table` is executed,
+comparing the fixed caller signature and the dynamic callee signature and trapping
+if there is a mismatch. Since the callee may be in a different module which
+necessarily has a separate [types section](BinaryEncoding.md#type-section), and
+thus index space of types, the signature match must compare the underlying 
+[`func_type`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/types.ml#L5).
 
 Multiple return value calls will be possible, though possibly not in the
 MVP. The details of multiple-return-value calls needs clarification. Calling a
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -101,7 +101,7 @@ Every WebAssembly [instance](Modules.md) has one specially-designated *default*
 linear memory which is the linear memory accessed by all the 
 [memory operators below](#linear-memory-access). In the MVP, there are *only*
 default linear memories but [new memory operators](FutureFeatures.md#multiple-tables-and-memories)
-will be added after the MVP which can also access non-default memories.
+may be added after the MVP which can also access non-default memories.
 
 Linear memories (default or otherwise) can either be [imported](Modules.md#imports)
 or [defined inside the module](Modules.md#linear-memory-section), with defaultness
@@ -278,7 +278,7 @@ to hold the array of indirectly-callable functions. Thus, in the MVP:
   this can only be done through the host environment (e.g., the
   the `WebAssembly` [JavaScript API](JS.md#webassemblytable-objects)).
 
-These restrictions will be relaxed in the 
+These restrictions may be relaxed in the 
 [future](FutureFeatures.md#more-table-operators-and-types). 
 
 ## Local variables
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -315,7 +315,7 @@ After import or definition, there is no difference when calling into a table.
 It is a validation error for a `set_global` to index an immutable global variable.
 
 In the MVP, the primary use case of global variables is to represent
-instantiation-time immutable values as a necessary building block of
+instantiation-time immutable values as a useful building block for
 [dynamic linking](DynamicLinking.md).
 
 After the MVP, when [reference types](GC.md) are added to the set of [value types](#types),
--- a/DynamicLinking.md
+++ b/DynamicLinking.md
@@ -1,7 +1,7 @@
 # Dynamic linking
 
-WebAssembly supports load-time and run-time (`dlopen`) dynamic linking in the
-MVP. This is achieved by having multiple [instantiated modules](Modules.md)
+WebAssembly allows load-time and run-time (`dlopen`) dynamic linking in the
+MVP. This is enabled by having multiple [instantiated modules](Modules.md)
 share functions, [linear memory](AstSemantics.md#linear-memory) and
 [tables](AstSemantics.md#tables) using module [imports](Modules.md#imports)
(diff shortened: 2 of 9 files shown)


DynamicLinking.md
```diff
@@ -1,49 +1,102 @@
...
+Note that the representation of a C function-pointer in WebAssembly is an index
+into a function table, so the above scheme lines up perfectly with the
+function-pointer return value of `dlsym`.
+
+More complicated dynamic linking functionality (e.g., interposition, weak
+symbols, etc) can be simulated efficiently by assigning a function table
+index to each weak/mutable symbol, calling the symbol via `call_table` on that
+index, and mutating the underlying element as needed.
+
+After the MVP, we would like to standardize a single [ABI][] per source
```
kripken: I think there might be multiple ABIs per source language, and they might not be standardized, at least not formally. I think something like "it may be useful in the future to standardize ABIs". That phrasing also removes "we" so it's more general.

kripken: (I realize the text was there before, but the meaning changes with this PR.)

kripken:
I am very excited about Tables! :) That's all we need for some useful forms of dynamic linking.

But I am worried about Memories. First, a concern here is a binary code size increase due to adding an immediate to loads and stores, which are very common (those two are over 13% of opcodes, and adding one byte for each leads to a 6% increase in binary size). Second, imported memories are not strictly needed for dynamic linking; emscripten supports dlopen() and shared libraries right now without multiple memories, for example. The cost is copying the memory init during startup, but it's not that bad - is there some other benefit to multiple Memories that I am missing?

Overall, Tables and Memories seem like separate features. Perhaps it makes sense to discuss them separately?

lukewagner:
With partial specialization of immediates in the opcode table, then the 1 opcode entry (per load/store op) that would already be allocated to partially specialize natural alignment would also specify the heap index, so this doesn't necessarily impact code size.  I expect we'll discuss this PR for a while, during which time we can figure out opcode tables and have more concrete details, but I wouldn't argue against this feature just for presumed binary size reasons now.

Multiple memories are basically an obvious generalization that I think practically every person in the wasm group has independently proposed over the last year.  The use cases involve some of the ones given for having memory-imports at all in previous discussions: when implementing a VM, you could have separate host and guest memories.  Separate memories might also be a good tool to avoid OOMs due to 32-bit fragmentation (by splitting 1 huge contiguous range into multiple) and could be used by special subsystems like file i/o which can statically separate where different data lives.  Lastly, in the future, we might allow imports of other, non-`WebAssembly.Memory` objects (e.g., (`Shared`)`ArrayBuffer`, something to alias the bits of a `<canvas>`, ...), allowing fast and flexible access to these via memory operations.

AstSemantics.md
```diff
@@ -231,6 +239,40 @@ operator may be added. However, due to normal fragmentation, applications are
...
+to the element type.
+
+In the MVP, the set of operations and types for tables is limited:
+* tables may only be accessed via [`call_table`](#calls);
+* the only allowed table element types are "function" and
+  "function with signature `X`";
+* tables may not be directly mutated or resized from WebAssembly code;
+  initially this must be done through the host environment (e.g., the
+  the `WebAssembly` [JavaScript API](JS.md#webassemblytable-objects).
+
```
JSStats: Seems a small step to allow a table specialized to a particular function signature, thus allowing instances of `call_table` to do validation checks when compiling or loading and thus avoid dynamic signature checks. Could just generalize the current function table, adding a signature field that has a wild value when not specialized, and allow multiple function tables in the MVP.

AstSemantics.md
```diff
@@ -111,6 +112,9 @@ Integer loads can specify a *storage size* which is smaller than the result type
 well as a signedness which determines whether the bytes are sign- or zero-
 extended into the result type.
 
+Each `load` and `store` specifies the [linear memory index](Modules.md#index-spaces)
+to operate on with an unsigned integer immediate.
+
```
JSStats: It seems very unfortunate to move to a segmented memory model. Could this be re-imagined in a linear memory space?

The data segments could be removed from the MVP, and replaced by an interface to access data blobs in the wasm file, and this would appear sufficient to allow the wasm code to load these at dynamic locations at runtime.

The core issue seems to be that it would be useful to allow constant address references to be baked into the compiled code? For example, for a module to be able to reference data loaded in its own data segments at addresses known at compile time, or to be able to bake in the index of variables in the linear memory.

It would appear that this can be solved entirely by rewriting the wasm as it is loaded, and thus this segmented memory model it not necessary for the MVP?

For example, the code could use separate opcodes (macros) to encode linear memory accesses relative to a constant base and the loader would rewrite these adding the base that has been chosen at runtime by the loader.

This would be rather unfortunate for sharing compiled modules, and for caching of compiled modules. If the code is to be shared then these can not be baked into the code anyway, which just leaves the use case of caching which might allow a copy and patch rather than a re-compile which the `get_const` idea seems to support.

So the segmented memory model seems entirely unnecessary for a MVP with dynamic loading, for both the use cases of: top runtime performance by rewriting the segment bases into the offsets; or for the use case of sharing code in which case the offsets are dynamic anyway.

Modules.md
```diff
@@ -1,58 +1,95 @@
...
+
+## Imports
+
+A module can declare a sequence of **imports** which are provided, at
+instantiation time, by the host environment. There are several kinds of imports:
+* **function imports**, which can be called inside the module by the
+  [`call`](AstSemantics.md#calls) operator;
+* **linear memory imports**, which can be accessed inside the module by the
+  [memory operators](AstSemantics.md#linear-memory); and
+* **table imports**, which can be accessed inside the module by
```
JSStats: Perhaps these "linear memory imports" are the other use case for a segmented memory model? If the host wants to share a segment of memory with the wasm code then it can just alias part of the linear memory rather than putting the burden of supporting segments and checking segment bounds on the wasm code. If GC objects are supported in future then the host can just pass in a typed boxed reference to allow the wasm to operator on it, for example if it wants the wasm code to fill a mutable string it can just pass in this object, but it will not be accessed in wasm as linear memory.

jfbastien:
I'm quite a fan of memories, it'll also make it easier to separate stuff out of the current single heap, such as the user stack, reducing user-side bugs / security issues, and it'll potentially also reduce fragmentation if the compiler is clever about certain things (I think that'll be a longer term gain).

Modules.md
```diff
@@ -170,30 +217,45 @@ reserve up to the *maximum* is not. When a maximum memory size is *not* declared
...
+
+## Elements section
+
+For function tables, the intial contents of the tables' elements are sentinel
+values that throw if called. The *elements section* contains a possibly-empty
+array of *element segments* which define the initial contents of fixed
+`(offset, length)` ranges of a given table, specified by its 
+[table index](#index-spaces). The elements are specified by their 
+[function index](#index-spaces). This section is the table analogue of the
+linear memory [data section](#data-section).
```
JSStats: Sorry, I don't understand what the "elements section" contains? What are the "element segments"? Are they sub-ranges of table index spaces? Or is this a section with all the elements from all tables with the elements grouped into ranges as in the tables? Why is this needed and how is it to be used?

AstSemantics.md
jfbastien: I agree with the code sharing concern: different dynamic libraries have to agree what linear memory index values mean. They could say `0` is HEAP, `1` is user stack, `2` is rodata, `4` is globals, etc. They need to agree on it, though. Maybe we want to use strings here as well, and somehow make it compress well?

I think we may want to start with @lukewagner's proposal (stick with integers), and then as we get information on compression we can try out different approaches to resolve this problem.

AstSemantics.md
```diff
@@ -231,6 +239,40 @@ operator may be added. However, due to normal fragmentation, applications are
...
+are opaque values of a particular *table element type*. This allows the table to
+contain values like GC references, raw OS handles, or native pointers that are
+accessed by WebAssembly code indirectly through an integer index. This feature
+bridges the gap between low-level, untrusted linear memory and high-level
+opaque handles/references at the cost of a bounds-checked table indirection.
+
+The table's element type dynamically constrains the type of elements stored 
+in the table and allows engines to avoid some type checks on table use. When
+tables are mutated, any stored value is necessarily coerced (possibly trapping)
+to the element type.
```
jfbastien: Could you clarify "possibly trapping"?

AstSemantics.md
```diff
@@ -231,6 +239,40 @@ operator may be added. However, due to normal fragmentation, applications are
...
+[future](FutureFeatures.md#more-table-operators-and-types). In the MVP,
+the primary purpose of tables is to implement indirect function calls
+in C/C++, using an integer index as the pointer-to-function and the table
+to hold the array of indirectly-callable functions. The compiler may either
+choose to group functions by signature (into N tables-with-signatures)
+or to put all functions into the same table (with no signature). The former
+strategy allows the engine to eliminate dynamic signature-mismatch checks while
+the latter strategy gives each function a unique index (which may be necessary
+for C/C++ compatibility) and admits simpler and potentially more space-efficient
+[dynamic linking](DynamicLinking.md).
```
jfbastien: Splitting tables also allows some form of CFI based on classy hierarchy, similar to what [LLVM currently supports](http://clang.llvm.org/docs/ControlFlowIntegrity.html#forward-edge-cfi-for-virtual-calls).

AstSemantics.md
```diff
@@ -313,43 +355,27 @@ explicit accesses to linear memory.
...
+second immediate. 
+
+  * `call_table`: call function element in a table
+
+The specified [table](#tables) must have either a "function" or "function with
+signature `X`" element type (in the MVP, these are the only two
+possibilities, but in the future, there may be other kinds of tables).
+If the table element type has a signature, it must match the signature of the
+call. Otherwise, the signature is checked at runtime and a signature mismatch
+causes a trap.
```
jfbastien: Ha interesting. Could you dive into that runtime check usecase? What do you have in mind?

DynamicLinking.md
```diff
@@ -1,49 +1,102 @@
...
+* the function `exp`
+* the imported table(s) and linear memory (so that other modules may
+  import this one)
+
+The host-specific loader can then simply connect the imports
+and exports at instantantiation-time.
+
+One extra detail is what
+to use as the [module name](Modules.md#imports) for imports (since
+WebAssembly has a two-level namespace). One option is to have a single default
```
jfbastien: Link for two-level namespace.

JSStats:
@jfbastien The stack is a high frequency access area, and it's going to degrade the file size to be flipping between opcodes with different segments, so developers are not going be a 'fans' of code using that style. If developers want a bounds checked stack then they can emit explicit checks - this use case does not need explicit support by wasm.

Bounds checking lots of small segments would be hard to optimize. Each is going to need a large chunks of VM for a memory protection scheme, so they would need large base offsets so either use a lot of immediate code memory or increase register pressure.

I doubt the 'fragmentation' benefits would be realized in most cases anyway as the main use of these segments would likely be for small areas of memory per module and modules will be using global memory for general use.

Lastly, C code is generally not written in a segmented memory model these days. I recall the old days with all the different memory models, and I would not wish this on the next generation of web developers.

AstSemantics.md
JSStats: @jfbastien So if the table is specialized on function signature 1 and an attempt were made to write a function with signature 2 then it would trap because the types are not compatible - to guarantee that all entries in the table are of the tables declared type. Perhaps a runtime could coerce a function to match the table's signature, for example dropping result(s) or coercing some of the arguments to match, in which case it would do so.

AstSemantics.md
JSStats: Sorry, I missed this in my comment above and the inclusion of the inclusion of the "function with signature `X`" table type is welcomed.

@jfbastien If the signature of a function with a dynamic index is not know at runtime then a runtime check is needed to make sure that the caller and caller have the same function arguments and results. In the case that the `table` is specialized on a signature then this check can be avoided at runtime. There was a long issue discussing this some time ago and it looks like a nice resolution giving the developer the choice.

kripken:
@lukewagner: Agreed that we can keep code size issues for later in the discussion of this PR.

Regarding the use cases presented by you and @jfbastien:

1. They could also be done by a GC reference to an ArrayBuffer, which is a feature on the roadmap anyhow. There may be a perf difference, but for several of those use cases it probably wouldn't matter anyhow (e.g., for a separate file I/O space, emscripten has an option for that now with the filesystem stored out of main memory, and it works reasonably well). For others, we should measure.
2. Using multiple memory spaces in a C or C++ program is not going to directly work (since each pointer needs to be able to refer to anything), so I don't see how the user stack as suggested could be done in another space? This feature will be very limited in scope.

Given those two, I don't see Memories as well-motivated, certainly not for the MVP - in what way are we not viable otherwise?

AstSemantics.md
JSStats: @jfbastien The 'code sharing concern' I noted is that it is not possible to both patch or baking-in constant offsets to module linear memory bases, while also sharing the same code between different instances. I don't think the heap indexes would be compiled into indirections through a table of heap bases at runtime, which would be very costly, rather into access offsets?

lukewagner: Given that wasm is designed to allow multiple instances to coexist in the same process, there is already a "segmented" memory model: this PR simply removes the restriction that a single instance can only access 1 memory.  By default, the expectation is that all wasm modules would use a single memory.  From a C++ pov, multiple memories would correspond to the (also statically numbered) [address space](http://llvm.org/docs/LangRef.html#pointer-type) in LLVM (and address spaces would provide a direct way to access this functionality from C++).

I agree it's not strictly necessary in the MVP, but there is a distinct symmetry with tables and I think it's something we definitely want in the long-term, so at the very least, we should generalize the design so that N heaps is possible but constrain to 0..1 (like we've done in other cases).  But if we do multiple tables anyway and make the refactorings to allow better code sharing and less code-dirtying on cache hit (viz., getting away from patching), then I think multiple heaps isn't actually much work in practice.

AstSemantics.md
lukewagner: @jfbastien Expanded and clarified.  The idea is that, in the future if we have `set_table`, we could ensure the element types match statically.  But when mutating through the dynamic JS API (or other host APIs), some sort of dynamic check or coercion would be necessary.

AstSemantics.md
lukewagner: IIUC, that's what's already in the PR: if you `call_table` a table that has a fixed signature, you statically validate against that signature.

AstSemantics.md
lukewagner: Good point; added that use case.

AstSemantics.md
lukewagner: @jfbastien Pushed some clarifications, ptal.
@JSStats Agreed and great!  Feel free to delete (via 'X') "resolved" comments since I periodically scan open comments.

DynamicLinking.md
lukewagner: Happy to, but what do you think is a good canonical URL?

Modules.md
lukewagner: Yes, high-performance access to host-provided "memory-like things" is a definitely a use case here ((`Shared`)`ArrayBuffer`, something from a `<canvas>`, maybe a server wasm environment can have an efficient zero-copy network API, ...)

Modules.md
lukewagner: Basically, what this PR is doing is splitting 1 section (the table section) into two halves: two ways to "get" a table (import or define), and 1 way to stick your module's table elements into any table you "got".  The best way to think about this is as symmetric to memory imports/definitions and data segments.  I rewrote this section to be more clear, though.

lukewagner:
@kripken C++ could be given access to multiple memories through [address space](http://llvm.org/docs/LangRef.html#pointer-type) annotations.  This prohibits use of more than 1 *general purpose* memory (whose pointers could escape "into the wild", where static address space is lost), so unfortunately, I don't think this could be used for, e.g., the user stack.  But this *could* be used by self-contained subsystems, like the in-memory file-system and interfacing with APIs in the future.  Using GC references is a *further* generalization but would lose some optimization potential.

As I said in a comment above, if we thought this would be a major complexity burden, we could save ourselves some time in the MVP by constraining to 0..1.  But given the work I think we all will want to do to support better code sharing ([this](https://www.akkadia.org/drepper/dsohowto.pdf) type of stuff), I think the work required to support N heaps won't be too bad.  Even if we constrain to 0..1 in the MVP, though, I'd like to have the design for N understood/documented since it's very easy to make a design that bakes in assumptions of 1 and would then require duplicating a bunch of things.

lukewagner:
(Interested to hear what other engines think about the work required for multiple memories, of course; constraining to 0..1 is certainly an acceptable option if it looked prohibitive.)

AstSemantics.md
kripken: > From a C++ pov, multiple memories would correspond to the (also statically numbered) address space in LLVM (and address spaces would provide a direct way to access this functionality from C++).

Looking for what that feature is used for in LLVM, the best I can find is (1) thread-local storage on x86 and (2) gc things. However, both of those would be done very differently in wasm, wouldn't they?

Are there other uses of this feature in LLVM that would be relevant for us?

JSStats: @lukewagner I don't think segmented memory is 'something we definitely want in the long-term', quite the opposite I think it is a place wasm should never go. I'd like the wasm linear memory to be very fast to access, and I see segments compromising this goal. If you have a solution to the memory access performance challenge then that might change my mind?

The use case of accessing JS objects from wasm can use the GC object support, which can use the object pointer as the base and do inline bounds checking, and it's not clear that this would be any slower than accessing segments. A JS canvas etc is not going to be allocated in a large chunk of VM just to avoid bounds checks, so segments pointing to its raw data are going to be using a base and explicit bounds checks anyway?

What optimizations do imported segments support beyond what would be possible with GC object support anyway?

The only optimization opportunity that comes to mind is baking in an offset on the x86, but the base would probably be in a register on other architectures anyway.

I don't see any need to reserve a 0/1 field in operators now either. The common case will be using a single segment so an opcode dedicated to this is useful anyway now, and new opcodes with a segment field could be introduced in future to support segments.

jfbastien: > Looking for what that feature is used for in LLVM, the best I can find is (1) thread-local storage on x86 and (2) gc things. However, both of those would be done very differently in wasm, wouldn't they?
>
> Are there other uses of this feature in LLVM that would be relevant for us?

I would put statics, rodata and user stack in their own address space, and leave the user heap as just a user heap without anything else in there.

kripken: I don't understand, how can that work? A C or C++ pointer can point to either the stack or the heap, as @lukewagner and I said, so they can't be in separate address spaces, can they?

Are you suggesting fat pointers? Or some sophisticated type-based analysis like Cheerp does? Or am I missing something obvious? :)

DynamicLinking.md
jfbastien: Hmm, I guess we don't document two-level namespace anywhere? :confused:

AstSemantics.md
lukewagner: @JSStats I don't know what memory access performance challenge you're referring to.  If an engine wants to go the route of baking-in addresses/lengths, it can do that just as easily for multiple heaps.  If an engine wants to pin a register, it can still do that for the first (and usually only) memory.  Otherwise, the engine can hoist things into registers as needed locally.

I *can* actually see a GC `Memory`-reference type getting equivalent performance (if we stipulate things like a GC reference type that says not just `Memory`, but also "with `initial` >= `X`" etc).  I'd like wasm to be able to run with full functionality w/o requiring GC (so you could have wasm-only workers that simply don't create a GC heap or any of the JS-runtime data), but that could also be made to work when GC reference types are only used for locals (on the stack), never GC allocation.

@kripken In addition to "file system impl", "asm.js globals" is another use cases where you have a simple, static separation.  Another cool use case could be a *San tool that wants to maintain a whole shadow data structure to reflect memory; a separate memory is exactly what you want; with a single memory, you'd have to prematurely fix the heap size to split "shadow" from "real" and then "real" can also corrupt "shadow".  Over time, it's easy to imagine more use cases arising since this is an inherently artificial limitation.

Also, if you take a modular-web-app POV (instead of a monolithic-C++-app POV), there's not just 1 big heap; you have lots of modules in lots of heaps.  You don't want to force them all into the same heap (where they may clobber each other) or force them to do unnecessary copying if they form a data pipeline.  GC-`Memory` references handles the general case, but imports/exports would allow the pipeline to be declaratively composed using the plain old module system.  Again, not a killer feature, but (1) not hard (if we do GC Memories, multi-Memory imports are just a simple/special-case), (2) obvious symmetry in design given that we have multi function/table export (and probably other things too in the future); given the other things, someone would naturally ask "why *not* multiple memories?" and I don't think we have a strong concrete answer.  Again, if it's too much work as judged by engine authors, we could constrain 0..1 in MVP, but if we don't provision in the MVP, we'll have to duplicate a lot of stuff (more than just load/store ops) later which will be inelegant.

kripken: > In addition to "file system impl", "asm.js globals" is another use cases where you have a simple, static separation.

Yes, it would be nice for that. But we've been ok without it, and it's temporary anyhow.

> Another cool use case could be a *San tool that wants to maintain a whole shadow data structure to reflect memory; a separate memory is exactly what you want; with a single memory, you'd have to prematurely fix the heap size to split "shadow" from "real" and then "real" can also corrupt "shadow".

Yes, but that use case is well handled by a GC ref to a buffer, as the addition of bounds checks on the shadow data is not significant when doing *Sanning. There are also some *San use cases that would need a GC ref anyhow, consider a thread sanitizer that creates a buffer per thread, the number of threads may not be known at compile time.

> Also, if you take a modular-web-app POV (instead of a monolithic-C++-app POV), there's not just 1 big heap; you have lots of modules in lots of heaps. You don't want to force them all into the same heap (where they may clobber each other) or force them to do unnecessary copying if they form a data pipeline. GC-Memory references handles the general case, but imports/exports would allow the pipeline to be declaratively composed using the plain old module system.

First, if we're talking non-C++, would heaps even be used? Something like Java or C# might be compiled to use GC buffers anyhow. Or, if you meant C++, how would that work exactly - a pointer would only be able to refer to one of those heaps? Would the modules not communicate through shared pointers? (If so I don't think they need this feature?)

Second, in your proposal, Memories arrive at import time, i.e., when a module is instantiated? Then two modules can't refer to each other's Memories, can they? A further limitation is that this wouldn't be able to handle dlopen() style linking, that happens at runtime. (A GC ref to buffers, on the other hand, doesn't have any limitation and can handle arbitrary linking.)

Overall, I don't mean to pick at your specific examples. My general point is that, as you said, a GC ref to a buffer is a generalization of this, and we'll have it anyhow, so multiple Memories adds complexity to the spec and to VMs that is not strictly necessary. Perhaps an important use case that needs to avoid the sandboxing overhead of GC buffers could justify it, but the use cases so far seem kind of unusual. Also, we don't have data on that GC ref overhead either yet, so this seems premature.

> (2) obvious symmetry in design given that we have multi function/table export

I would argue that import/export of a Table, and having that Table be modifiable, is much more important than having multiple imported/exported Tables. A single Table shared between modules that can be changed is all dynamic linking needs to get good performance.

With that said, I agree that the symmetry argument is a good one.

rossberg-chromium:
I wonder whether a `call_table` operator still is the right thing to have at this point. If we plan to add other table entry types in the future, wouldn't it make more sense to decompose it into `get_table` and `call_ptr`? We want to add function pointers anyway, so maybe the logical step would be to do that now, and avoid a redundant special case?

lukewagner:
@rossberg-chromium That's an interesting idea, but if we add function pointers as a real value type, it seems like we'll end up needing full GC safe-point support which I've been enjoying not having to mess with.  I also wonder if we might accidentally rush the feature (b/c we only need this sliver of it) in a way that we might regret if we waited until we understood the whole GC problem domain.  Lastly, I'm interested in having this option that, even when we have first-class references, allowing wasm to only manipulate those references indirectly (through indices into tables) in a way that allows a wasm worker (or even webapp) to not need to create a GC heap or any of the related machinery; from this POV, you want the ability to pass in the index to the thing you want to operate on, which lines up nicely with `call_table`.

AstSemantics.md
lukewagner: > First, if we're talking non-C++, would heaps even be used?

Yes, with explicit (compiler-specific) address-space annotations on the pointer type.  You'd have to treat it specially and not let these pointers escape "into the wild".  The goal would be to avoid copies by working on this other data more directly.

> Then two modules can't refer to each other's Memories, can they?

You're right, this is the point of the "TODO: JS API for cyclic imports".  Basically, we can support this by a more powerful (but overkill, for the common single-module case) `Instance.instantiate` method that takes multiple modules that can import each other.

> My general point is that, as you said, a GC ref to a buffer is a generalization of this, and we'll have 
> it anyhow, so multiple Memories adds complexity to the spec and to VMs that is not strictly 
> necessary.

I don't think it will, assuming we end up supporting multiple heaps via GC refs in the future; you can think of it as importing a GC value and storing it in an immutable GC global.  If we have the general case, and support multiple imports/exports of the other first-class things, it will seem arbitrary to have this "zero or one" special case with memory forever.

rossberg-chromium:
@lukewagner, hm, I was referring to plain, C-style function pointers, not closures. How would that involve GC?

JSStats:
fwiw I think there should be a distinct variant of wasm that works without garbage collection, even after the support for GC objects has been added. It could be a variant that is declared and allows the runtime to optimize for this. Threads running such variants could run in parallel with garbage collection on other threads serving different instances.

@rossberg-chromium Some examples of `call_ptr` might help? If they are just C function pointers then this use case seems to be handled in wasm by `call_indirect` or `call_table`. If they are boxed opaque objects then are they a small set of pinned objects not subject to GC?

@lukewagner If external GC objects can be referenced by an index then the external GC needs to know about these references. It might be more generally useful to be able to import boxed objects defined to be pinned in memory and reference counted, so that the wasm thread need not be concerned about them being garbage collected yet still be able to pass them around as boxed objects.

rossberg-chromium:
@JSStats, very straightforward: introduce a new value type, say, `funcptr`; values can be created by an operator `(addressof <funcid>)`, and consumed by the call operator `(call_ptr <typeid> <ptr> <arg>*)`.

Moreover, you can introduce typed variants, i.e., value types `(funcptr <typeid>)`, where the call operator does not need to perform any runtime check.

In fact, I would prefer to avoid misleading terminology like "pointer" and "address" for this, since these are opaque references. That's how they differ from the table: the only reason we need that is to compile C, where function pointers have to be numeric, because (a) C allows stupid casts, and (b) you need to be able to store them in linear memory. For other purposes, this indirection is unnecessary.

JSStats:
@rossberg-chromium It appears that without involving garbage collection that these `funcptr`s would need to be interned objects anyway, so could be referenced by an index, so it's not clear what has been gained, and these objects could not be stored in the linear memory which is a loss, so for a wasm without garbage collection they appear to be a net loss?

This PR proposes typed function index spaces to avoid runtime signature checks.

Being able to store function references in linear memory seems a very compelling use case to me.

rossberg-chromium:
@JSStats, no interned objects needed, just plain code addresses.

DynamicLinking.md
```diff
@@ -1,49 +1,102 @@
...
-Although dynamic linking is not part of the MVP, it has significant implications
-on many aspects of the design that do impact the MVP, such as the way linear
-memory is managed, how module imports and exports are specified, and how globals
-and function pointers work.  Therefore we want to have some viable ideas to
-ensure we don't standardize a design that unnecessarily complicates the design
-or implementation of dynamic linking.
-
-  [hosted libraries]: https://developers.google.com/speed/libraries/
-  [service workers]: https://www.w3.org/TR/service-workers/
+WebAssembly supports load-time and run-time (`dlopen`) dynamic linking in the
```
titzer: Supports isn't really the right term here. It's more "enables" or "allows", since WebAssembly itself only provides mechanisms, and the rest is done in user space.

titzer:
I'm not sold on multiple memories, or why we actually want them quite yet. It's a tempting generalization, but I don't see the problem they actually solve. Actually, I think they create a far worse problem providing a statically segmented address space as the programming model. Would we want dynamically indexed memories in the future? Then why the restriction to static indexing now?

There are larger questions we need to discuss here, like what will the role of memory be in the future programming model, which has yet to really emerge. In particular, a future with managed data extensions make adding more generality with multiple memories less important, and tantalizingly suggest that memory or memories in the future will be relegated to IO duties rather than the main program storage.

JSStats:
@rossberg-chromium Thank you, I guess raw unboxed code addresses could be passed around and stored in local variables just as a floating point numbers are, and I guess these could be prohibited from escaping or could be boxed if they do escape and perhaps then qualified by the instance too. But the inability to store them in the linear memory is still a show stopper.

rossberg-chromium:
@JSStats, for that we continue to have tables. But they're only needed in some cases. And it's pretty clear that we will eventually get other opaque types that cannot be stored in linear memory anyway.

titzer:
One criteria that was considered important for the current design of function pointers was to avoid having local types for function pointers, because we weren't (and IMO still aren't) ready to cross that boundary for MVP.

lukewagner:
@titzer Assuming we get a first-class `Memory` references some time in the future (by which I mean: that you could store in a local and access dynamically), I don't see the fundamental difference/cost (performance, implementation, complexity) with allowing multiple memory imports.  It seems unnecessarily irregular if we have multiple imports/exports of functions, tables (and, in the future: values, global variables, thread-local variables, ...) but only single imports/exports of memories.  It begs the question: Why not memory too?

As for "why the static restriction now?"  I think it's the same answer for tables (assuming we get first-class `Table` references too): because it allows you to declaratively compose modules (via imports/exports) that share these things.  E.g., if one has a worker containing only modules that import/export memory/tables, and therefore don't pass around first-class GC references, you don't need a GC in that worker (and you can know this a priori).

Again, if we think it causes undue impl burden in the MVP, I'd be fine to simply constrain the number of memory imports to <=1, as we've done with the number of return types in signatures.  I'm not trying to create more MVP work; I'm trying to avoid long-term irregularity that isn't pragmatically necessary.

DynamicLinking.md
lukewagner: Agreed, changed.

JSStats:
@lukewagner But what is the benefit of being able 'to declaratively compose modules'?  The claimed reasons seems to be that 'you don't need a GC in that worker', but having boxed object and garbage collection seem somewhat orthogonal, and a worker could accept a boxed buffer reference without having to worry about garbage collection but would just need some guarantees that the object is not going to be moved or reclaimed which could be a burden for the supplier of this object and not the worker.

Practically most implementations will need to keep these buffer bases in memory and load them as needed whether they are defined declaratively or dynamically, so what has been achieved? Perhaps 32 bit x86 could be the exception where it might be possible to bake in the bases into the instruction immediate offsets.

If there is a benefit then perhaps it is something else? For example, knowing the size of a buffer at compile time might help optimize accesses. Knowing that these buffers are in an area of VM using memory protection to assist avoiding bounds checks might help, but then perhaps there could just be a dynamic buffer type that has these properties too, to have buffer types with constraints on their size and with different memory protection guard zones etc.

rossberg-chromium:
@lukewagner, seems to me that the issue isn't imports/exports so much as the need to add yet more immediates to memory operators (same for table). Between no selector, and truly dynamic selection, is there a strong enough need for an intermediate solution that penalises the base case we care about most, while not solving the general problem either?

lukewagner:
@JSStats First-class memories/tables could come from anywhere (created by JS, other modules) and have any lifetime (maybe they *aren't* supposed to be kept alive longer than the instance, to avoid leaks); so it's hard to know much a priori once you see a module using references at the very early stage of booting a worker.  I had been thinking maybe we'd ultimately have something explicit/declarative so that there were no heuristics or laziness required to create workers w/o GC heaps or JS runtimes.

lukewagner:
@rossberg-chromium That's a good point and some of the previous discussion had been leading me to think in that direction too.  So if we want to keep the ability to have (multiple) imports/exports of all these things, but not express the selection as an immediate operand, we could unify this with the "TODO" in this PR of having value imports of `i32`s, then say that all imports (of memories, tables, `i32`s) go into an index space of "constant values", and then you have a `get_const` operator with an index immediate.  In the case of memories/tables, `get_const` would return a first-class reference.  Then we add new operators taking memory/table references (post-MVP). Is that what you were thinking?

It seems like this could be nice and regular.  Some questions and potential answers:
* do we then say that functions are "just different"?  Part of what I liked about the current proposal is that function/table/memory are all symmetric, but then again functions are a different kind of thing so I guess it's ok if we import/define/call/export/addressof them totally differently.
* if we don't want first-class reference types in the MVP, then that rules out multiple tables (and multiple heaps, but I don't think anyone will miss those :).  Since homogeneous tables are only useful if you have multiple, then we might as well cut out those too, keeping only heterogeneous tables in the MVP (with the future option to have other table element types).  (Removing heterogeneous tables conveniently removes the big "TODO" in this PR of how to describe a signature in the JS API which I had been feeling should integrate with how we describe the type of a typed object in JS.)
* do we then have a notion of a "default" memory/table (where either an import or definition could be default) and the default is what is accessed by all of today's operators (thereby avoiding an immediate or `get_const`)?  It seems like this would be a flag on the import/def w/ the restriction that only one memory/table could set this flag.
* for the exports section, it seems then like each export is either a constant or a function (identified by resp index).

How does that sound to everyone?

lukewagner:
I updated the PR basically as described above, having a special "default" memory/table designation and thus no extra immediates in ops.  The PR also (re-)adds global variables (declared mutable or immutable) which addresses the dynamic-linking use case, paves the way for future use cases with references, and [matches asm.js](http://asmjs.org/spec/latest/#global-variable-type-annotations).  For the moment, *mutable* global variables cannot be imported/exported since this requires extra work (some new `WebAssembly.Global` API) and I'm not aware of a use case in the MVP.  As with `i64`, we can remove this limitation in the future.

PTAL everyone :)

AstSemantics.md
```diff
@@ -80,29 +80,37 @@ operators.
...
-instances can execute in separate threads but have their own linear memory and can
-only communicate through messaging, e.g. in browsers using `postMessage`. It
-will be possible to share linear memory between threads of execution when
-[threads](PostMVP.md#threads) are added.
+Every WebAssembly [instance](Modules.md) has one specially-designated *default* 
+linear memory which is the linear memory accessed by all the 
+[memory operators below](#linear-memory-access). In the MVP, there are *only*
+default linear memories but [new memory operators](FutureFeatures.md#multiple-tables-and-memories)
+will be added after the MVP which can also access non-default memories.
+
```
JSStats: If these non-default linear memories are not visible in the MVP design then perhaps just leave out mention of them.

AstSemantics.md
```diff
@@ -80,29 +80,37 @@ operators.
...
+Linear memories (default or otherwise) can either be [imported](Modules.md#imports)
+or [defined inside the module](Modules.md#linear-memory-section), with defaultness
+indicated by a flag on the import or definition. After import or definition,
+there is no difference when accessing a linear memory whether it was imported or
+defined internally.
+
```
JSStats: So linear memories can be shared between instances in the MVP? E.g. Created in one instance and then imported as the default in another.

Lets generalize this to allow a sub-range of one linear memory to be the linear memory of another child instance. Lets add an API to allow the wasm code to create this child instance and to pass it a sub-range of the parents linear memory. This would address many uses cases, including protecting the parent from the child instance accessing its memory outside the range allocated.

AstSemantics.md
```diff
@@ -210,7 +220,7 @@ the [future](FutureFeatures.md#large-page-support)).
  * `grow_memory` : grow linear memory by a given unsigned delta of pages.
     Return the previous memory size in units of pages or -1 on failure.
 
-When a maximum memory size is declared in the [memory section](Module.md#linear-memory-section),
+When a linear memory has a declared [maximum memory size](Modules.md#linear-memory-section),
 `grow_memory` must fail if it would grow past the maximum. However,
 `grow_memory` may still fail before the maximum if it was not possible to
 reserve the space up front or if enabling the reserved memory fails.
```
JSStats: Where would the linear memory maximum be declared for an imported memory? In the module in which it was created? Seems that they would need to be allowed to be created elsewhere for many use cases.

AstSemantics.md
```diff
@@ -232,12 +242,49 @@ operator may be added. However, due to normal fragmentation, applications are
...
+
+## Table
+
+A *table* is similar to a linear memory whose elements, instead of being bytes,
+are opaque values of a particular *table element type*. This allows the table to
+contain values—like GC references, raw OS handles, or native pointers—that are
+accessed by WebAssembly code indirectly through an integer index. This feature
+bridges the gap between low-level, untrusted linear memory and high-level
+opaque handles/references at the cost of a bounds-checked table indirection.
+
```
JSStats: Sounds like a good description of the use case and rationale.

AstSemantics.md
```diff
@@ -232,12 +242,49 @@ operator may be added. However, due to normal fragmentation, applications are
...
+The table's element type dynamically constrains the type of elements stored 
+in the table and allows engines to avoid some type checks on table use. When
+tables are mutated, any stored value must match the element type by either a
+static validation constraint, trapping dynamic type validation check, or dynamic
+coercion (depending on where the mutation occurs).
+
```
JSStats: Sounds good, but would the coercions need to be spelled out. For example are the arguments and results of imported functions coerced; are floats coerced to integers; strings to numbers; missing arguments to zero?

AstSemantics.md
```diff
@@ -232,12 +242,49 @@ operator may be added. However, due to normal fragmentation, applications are
...
+Every WebAssembly [instance](Modules.md) has one specially-designated *default*
+table which is indexed by [`call_indirect`](#calls) and other future
+table operators. Tables can either be [imported](Modules.md#imports) or 
+[defined inside the module](Modules.md#table-section), with defaultness
+indicated by a flag on the import or definition. After import or definition,
+there is no difference when calling into a table whether it was imported or
+defined internally.
+
```
JSStats: Tables of functions seems a use case worthy of special attention, and would warrant allowing multiple function tables in the MVP or well before tables of other objects.

Could this note that the table element type would be defined in the module instance. If this were not static then it would not be of user for optimize of AOT compiled code.

AstSemantics.md
```diff
@@ -248,6 +295,32 @@ The details of index space for local variables and their types will be further c
 e.g. whether locals with type `i32` and `i64` must be contiguous and separate from
 others, etc.
 
+## Global variables
+
+A *global variable* stores a single value of a fixed [value type](#types) and may be
+declared either *mutable* or *immutable*. This provides WebAssembly with memory
+locations that are disjoint from any [linear memory](#linear-memory) and thus
+cannot be arbitrarily aliased as bits.
+
```
JSStats: Great, see also https://github.com/WebAssembly/design/issues/333  These might also address some code path selection based on features.

But if only immutable globals are need then perhaps just implements these and not mutable globals?

Do we know if the compiler really need un-aliased locations, or is the linear memory just fine? 

Problem with mutable globals is that they need a pointer outside of the linear memory, increasing register pressure. The 32-bit x86 might be an exception where the offset can be baked in, but with code sharing even that would not be possible.

DynamicLinking.md
```diff
@@ -1,49 +1,87 @@
...
+To implement run-time dynamic linking (e.g., `dlopen` and `dlsym`):
+* `dlopen` would compile and instantiate a new module, storing the compiled
+  instance in a host-environment table, returning the index to the caller.
+* `dlsym` would be given this index, pull the instance out of the table,
+  search the instances's exports, append the found function to the function
+  table (using host-defined functionality in the MVP, but directly from
+  WebAssembly code in the
+  [future](FutureFeatures.md#more-table-operators-and-types)) and return the
+  table index of the appended element to the caller.
+
```
JSStats: Cool, and more than I expected for the MVP.

JS.md
```diff
@@ -224,24 +270,204 @@ WebAssembly Exported Functions have a `[[Call]](this, argValues)` method defined
...
+
+Let `buffer` be a new `ArrayBuffer` whose
+[[[ArrayBufferData]]](http://tc39.github.io/ecma262/#sec-properties-of-the-arraybuffer-prototype-object)
+aliases `m` and whose 
+[[[ArrayBufferByteLength]]](http://tc39.github.io/ecma262/#sec-properties-of-the-arraybuffer-prototype-object)
+is set to the byte length of `m`.
+
+Return a new `WebAssembly.Memory` instance with [[Memory]] set to `m` and
+[[BufferObject]] set to `buffer`.
+
```
JSStats: Can these constructors capture all the subtle variations on linear memories that a needed. For example, a linear memory page aligned, or with a trailing protected guard zone, or at absolute zero in the address space.

AstSemantics.md
JSStats: For these global variables to be dynamically importable, or 'instantiation-time immutable values', it would appear that they can not be baked into the code in AOT compilation. If so then it is not clear what utility they have, and they would not appear to really be necessary?

I see some potential for the compiler to optimize code knowing that they will not change, similar to a `const var`, but that does not appear to be a core part of this PR? Also `let` variables have this property too, so if these were added then these might help address some of these optimization opportunities.

I recall mention of patchable values in prior discussion, and are these global constant variables intended to meet this use case? This would appear to require each instance to have a separate copy of the code, but that would somewhat defeat the purpose of sharing code.

For immutable global variables defined in the module, could they be baked into the code AOT?

There might be some significant performance burdens for shared code, extra dereferences. Does the wasm model allow the web developer to make the choice on these trade offs. Should a module have a flag to indicate that it is likely to be shared code, so that only in this case the runtime might avoid re-compiling or copying the code for each instance.

AstSemantics.md
lukewagner: They are "left out", insofar as there is no extra impl burden.  But it's useful to explain the target state, though, since it motivates other aspects of the design.

AstSemantics.md
lukewagner: I'd rather consider that in a separate issue; there are non-trivial performance and implementation implications.

AstSemantics.md
lukewagner: See Modules.md#imports which describes that import statements specify an initial and maximum which constrain the imported memory.

AstSemantics.md
lukewagner: This is spelled out in `call_table`, which requires a precise signature match.  I realized that I forgot to spell out what happens if a non-wasm (viz., JS) function is assigned to a table element, though (it's host-defined and, in the case of JS, always succeeds).  I'll add something for that.

AstSemantics.md
lukewagner: There are new sections in FutureFeatures.md on supporting multiple tables and multiple element kinds.

AstSemantics.md
lukewagner: For immutable globals, a compiler can either perform instantiation-time patching of immediates or use indirect through the global table.  This has the same tradeoffs (dirtying pages vs. GOT-indirection) as with normal dynamic linking.

Mutable variables need to be supported for engines that compile asm.js to wasm anyway, so it' no big deal to support in the MVP and might actually be useful for a tool (e.g., asm.js2wasm) that needs to allocate memory without perturbing linear memory.  Importing/Exporting mutable global variables is an extra burden, though, and that is explicitly ruled out in the MVP if you look at JS.md.

JS.md
lukewagner: Yes: the goal is to have 100% symmetry between a `Memory` constructor and the linear memory section in modules: how a memory is created should be orthogonal from how it is used.

JSStats: Re: 'how a memory is created should be *orthogonal* from how it is used'. This statement does seem to be based on the technical realities as linear memories create for use by wasm instances might need properties that are constrained by their use. For example, it is only possible to have one linear memory at absolute zero in a process address space, so this property is hardly 'orthogonal' to the use, and the code in which it is used can be optimized for it, and the technical reality is that this code can run faster as I have already demonstrated. A development path abstract and 'orthogonal' to the technical realities is not one I support. The entire concept of multiple linear memories is also abstract from the technical reality of a common address space per process.

AstSemantics.md
JSStats: Since no use case for mutable global variables has been made I do not support them being added back here. This matter had already been considered, and they were removed. Implementations are welcome to support them internally for asm.js but this need not be part of wasm. No new performance number have been presented after the decision to remove the global variables that suggests they were needed?

AstSemantics.md
JSStats: How about considering it first as it meets many of the use cases for multiple linear memories. What are the 'non-trivial performance and implementation implications'?  From why I understand these would have better performance than multiple linear memories because there is just one base per instance?

So if the performance of this approach has 'non-trivial performance and implementation implications' then multiple linear memories is even more of a burden, plus the burden of managing and supporting these multiple linear memories. Please, lets look at the performance and implementation implications? Lets look at the plan.

JS.md
lukewagner: That particular problem exists independently: even without any memory imports at all you can't know, when compiling code, whether it will be possible to allocate the memory at absolute zero.  Potential solutions to this problem (recompilation; some new "wasm process" abstraction, etc) would work just as well with the changes in this PR.

AstSemantics.md
lukewagner: Importing a subrange means that you can't rely on any guard pages being at the end since you could well be looking into the middle of a bigger linear memory.  You could imagine a separate kind of import that says "this is an import of a subrange of a bigger linear memory" that then impeded guard-page-based optimizations, but that'd be an *additional* kind of import to be considered separately.

JSStats: Yes, I guess so. Fwiw, guard pages are not necessary for the index masking strategy, just a spill area, so this strategy could be viable using index masking.

AstSemantics.md
```diff
@@ -248,6 +297,32 @@ The details of index space for local variables and their types will be further c
...
+Global variables are accessed via an integer index into the module-defined 
+[global index space](Modules.md#global-index-space). Global variables can 
+either be [imported](Modules.md#imports) or [defined inside the module](Modules.md#global-section).
+After import or definition, there is no difference when calling into a table.
```
kripken: `table` copy-paste error?

lukewagner: Yup, fixed!

Modules.md
```diff
@@ -170,30 +231,154 @@ reserve up to the *maximum* is not. When a maximum memory size is *not* declared
...
+to their absolute order as defined in [BinaryEncoding.md](BinaryEncoding.md).
+
+The definition index space is used by:
+* [exports](#exports), to indicate which definition to export
+* [elements sections](#elements-section), to place definitions into tables
+
+In the future, an `address_of` operator could be added which, given a definition
+index immediate, returns a first-class [reference](GC.md) to the definition.
+
+Note: the definition index spaces is a validation/compile-time concept, not
```
kripken: type: `space[s]`

flagxor:
Some questions about things not clear from the PR:

* Do table values altered by get/set affect previously instantiated modules? Or are they "locked in" at instantiate time? (Same with grow?)

* call_indirect suggests signature mismatch between modules will throw at call time, but another location suggests cooercion to JS types and back. Which is it? (I would hope wasm exports that are imported drill the values all the way through?) I would assume imports of regular JS functions would coerce?

* BinaryEncoding.md isn't updated in this change. I assume this means that the indirect call table section is retained. How does this interact with Tables? When are the entries from the wasm module stored to the table?

Thanks!

lukewagner:
Good questions, I'll update PR to make them more clear, but, briefly:
* Tables are like linear memory in that updates are observed by all instances immediately.
* Types are checked when the caller/callee are both wasm; when you store a (non-exported-wasm) JS function in a table, it is given a special host-defined element value that accepts any caller signature (honey badger don't care).
* I've postponed updating BinaryEncoding.md so we can do the careful lockstep iteration dance (probably on a binary_0xd branch).  To reflect this PR, the import, table, memory, and data sections would change and a new elements section would be added.  Function bodies wouldn't, though.

pizlonator:
I think that I object to this proposal.  I have a hard time seeing the complete use case.  Probably for a feature of this sophistication, multiple convincing use cases would have to exist.  The fact that this is not directly useful from C++ AFAICT is a concern.  The fact that this complicates implementation is a concern.

Can someone convince me why I want this feature?

lukewagner:
The concrete motivation is *dynamic linking*; with this PR, wasm compilers have the building blocks to implement "real" (efficient, separately compilable/cacheable) Just Works dynamic linking; without it they can't.  The updates to DynamicLinking.md in this PR attempt to explain in more detail and I'm happy to expand on it to do a better job.

But at a high level:

The PR adds `WebAssembly.Memory` almost exactly as was previously discussed (which you were arguing (me) in favor of).  Given importable memory, you can implement almost-dynamic linking (which was my objection at the time); it mostly just breaks down at function-pointer calls.  My realization after that discussion is that the indirect function table is actually quite symmetric to memory and not only useful for functions; tables are just a special kind of memory that store opaque values of some table element type instead of transparent bytes.  So the symmetric treatment is given to allow `WebAssembly.Table` to be created, imported and exported.

The last piece of the PR is immutable global imports which provide wasm the building block to do what native dynamic linking does with, in ELF's case, the Global Offset Table (GOT).  If I'm dynamically linking, I don't know where, in the shared memory/table, my global data / function-table elements start.  Immutable imports provide a general way to let these values be determined at instantiation- (load-) time so they can either be patched in directly (at the cost of dirtying code pages, something native dynamic linking tries to avoid) or writing into special memory locations (analogous to the GOT).  Immutable global imports are also pretty easy to implement and already present in asm.js.

Once you have these three things, you can explicitly share all the state which modules need to share to implement dynamic linking; there is no need for the previously-discussed big magical/stateful instance; with this PR, an instance is just an immutable tuple binding code to memory + table + imported values.  Although it's not fully fleshed out, I believe this design will also naturally generalize to other stateful things we've already discussed that modules would need to share (e.g., thread-locals, mutable globals).

pizlonator:
@lukewagner Based on the above, I don't see how multiple memories helps dynamic linking.  @kripken seems to be saying that they are not needed.  Why add them, then?

lukewagner:
@pizlonator In the MVP, there are 0 or 1 memories (and tables), same as before the PR.  (Note: some of the early comments you may be reading are based on a previous iteration of the current PR.)

After the MVP *and* after adding GC support, we could then allow references to `Memory` and add `load`/`store` ops that operated on a `Memory` reference operand.  Then they'd be fully first class: you could take a `Memory` in as an argument to an exported function.  I think this could have pretty huge utility when you imagine webapps composed of multiple independent wasm instances that could use these first class memory references to avoid what would otherwise be copies.  You can also imagine future Web APIs that return things that could be used as memories, again to avoid copies.  This PR doesn't commit us to any of this, but it does leave a clear path.

AstSemantics.md
```diff
@@ -80,29 +80,37 @@ operators.
...
-[module](Modules.md#linear-memory-section), and it can be dynamically grown by
-the [`grow_memory`](AstSemantics.md#resizing) operator.
-
-The linear memory can be considered
-to be an untyped array of bytes, and it is unspecified how embedders map this
-array into their process' own [virtual memory][]. The linear memory is
-sandboxed; it does not alias the execution engine's internal data structures,
-the execution stack, local variables, or other process memory.
+A *linear memory* is a contiguous, byte-addressable range of memory spanning
+from offset `0` and extending up to a varying *memory size*. This size is always
```
jfbastien: We can split this into a follow-up: why not have a user-defined mapping for memory, mapping the numbers to strings? i.e. a module would say "I have a memory called `HEAP` which I map to `1`, and a memory called `SECRETS` which I map to `2`". That way different toolchains don't have to agree on magical numbers but just on strings. Two dynamic objects can use different numbering schemes if they want, and things will align fine if they agree on string names.

AstSemantics.md
```diff
@@ -232,12 +242,53 @@ operator may be added. However, due to normal fragmentation, applications are
...
 [`discard`](FutureFeatures.md#finer-grained-control-over-memory) future feature.
 
+The above operators operate on the [default linear memory](#linear-memory).
+
+## Table
+
+A *table* is similar to a linear memory whose elements, instead of being bytes,
+are opaque values of a particular *table element type*. This allows the table to
+contain values—like GC references, raw OS handles, or native pointers—that are
+accessed by WebAssembly code indirectly through an integer index. This feature
```
jfbastien: Same as above: should these integer indices be user-mapped using strings?

AstSemantics.md
lukewagner: In a sense, the PR already does this (if we're talking about imports): the string is the imported name, the index is the immutable global it is stored in.  (But there is this special "default" memory which avoids the need to give every load/store an immediate index which was the objection practically everyone had to the previous iteration of the PR.)

AstSemantics.md
lukewagner: Same answer as above :)

AstSemantics.md
```diff
@@ -248,6 +299,32 @@ The details of index space for local variables and their types will be further c
 e.g. whether locals with type `i32` and `i64` must be contiguous and separate from
 others, etc.
 
+## Global variables
+
+A *global variable* stores a single value of a fixed [value type](#types) and may be
+declared either *mutable* or *immutable*. This provides WebAssembly with memory
+locations that are disjoint from any [linear memory](#linear-memory) and thus
+cannot be arbitrarily aliased as bits.
```
jfbastien: Is the intent that when we support multiple memories those can also be loaded as immutable? That seems very useful for a bunch of usecases, and orthogonal to `mprotect`.

AstSemantics.md
```diff
@@ -248,6 +299,32 @@ The details of index space for local variables and their types will be further c
...
+
+Global variables are accessed via an integer index into the module-defined 
+[global index space](Modules.md#global-index-space). Global variables can 
+either be [imported](Modules.md#imports) or [defined inside the module](Modules.md#global-section).
+After import or definition, there is no difference when accessing a global.
```
jfbastien: Global imports also specify mutability?

AstSemantics.md
lukewagner: Correctamundo

AstSemantics.md
lukewagner: That's right; this is currently mentioned in Modules.md#imports for "global variable imports".

AstSemantics.md
```diff
@@ -232,12 +242,49 @@ operator may be added. However, due to normal fragmentation, applications are
...
+## Table
+
+A *table* is similar to a linear memory whose elements, instead of being bytes,
+are opaque values of a particular *table element type*. This allows the table to
+contain values—like GC references, raw OS handles, or native pointers—that are
+accessed by WebAssembly code indirectly through an integer index. This feature
+bridges the gap between low-level, untrusted linear memory and high-level
+opaque handles/references at the cost of a bounds-checked table indirection.
+
+The table's element type dynamically constrains the type of elements stored
```
titzer: statically constrains?

DynamicLinking.md
```diff
@@ -1,49 +1,87 @@
...
-Although dynamic linking is not part of the MVP, it has significant implications
-on many aspects of the design that do impact the MVP, such as the way linear
-memory is managed, how module imports and exports are specified, and how globals
-and function pointers work.  Therefore we want to have some viable ideas to
-ensure we don't standardize a design that unnecessarily complicates the design
-or implementation of dynamic linking.
-
-  [hosted libraries]: https://developers.google.com/speed/libraries/
-  [service workers]: https://www.w3.org/TR/service-workers/
+WebAssembly allows load-time and run-time (`dlopen`) dynamic linking in the
```
titzer: Wordsmithing here:

WebAssembly enables load-time and run-time ... by having ...

DynamicLinking.md
```diff
@@ -1,49 +1,87 @@
...
+MVP. This is enabled by having multiple [instantiated modules](Modules.md)
+share functions, [linear memories](AstSemantics.md#linear-memory),
+[tables](AstSemantics.md#table) and [constants](AstSemantics.md#constants)
+using module [imports](Modules.md#imports) and [exports](Modules.md#exports). In
+particular, since all (non-local) state that a module can access can be imported
+and exported and thus shared between separate modules' instances, toolchains
+have the building blocks to implement dynamic loaders.
+
+Since the manner in which modules are loaded and instantiated is defined by the
```
titzer: Should we settle on the term embedding environment?

FutureFeatures.md
```diff
@@ -425,3 +415,46 @@ of WebAssembly in browsers:
   would enable Web apps to perform their own (["layer 1"](BinaryEncoding.md))
   custom compression (on top of the spec-defined binary format, under generic
   HTTP `Content-Encoding` compression).
+
+## Multiple Tables and Memories
+
+In the MVP, there can only be (at most one) default table/memory and there are
```
titzer: Wordsmith:

The MVP limits modules to at most one memory and at most one table (the default ones).

AstSemantics.md
lukewagner: Agreed, fixed.

DynamicLinking.md
lukewagner: Grepping, we use "host environment" a bunch, "embedding environment" not at all, although we do refer to Web/non-Web/browser embeddings in 6 cases.  Maybe we could normalize in a separate PR?

DynamicLinking.md
lukewagner: Fixed

FutureFeatures.md
lukewagner: Fixed, but re-added a bit at the end for "and there are only operators for accessing the default table and memory".

Modules.md
```diff
@@ -170,30 +231,154 @@ reserve up to the *maximum* is not. When a maximum memory size is *not* declared
...
+non-table definitions.
+
+The table index space is only used by the [elements section](#elements-section).
+In the MVP, there is at most one table so this index space is just
+a placeholder for when there can be 
+[multiple tables](FutureFeatures.md#multiple-tables-and-memories).
+
+## Initializer Expression
+
+Initializer expressions are evaluated at instantiation time and are currently
```
titzer: Do we really need to have this extra section? Couldn't the start function simply set these up?

lukewagner: By the time the start function has run, the data segments must have already been copied into linear memory; the key is being able to, one way or another, use an instantiation-time-provided value to determine the offset.  For mutable global variables of course we could do zero-initialization, but for *immutable* global variables we need *some* initializer.  We could simply use a type-appropriate numeric immediate (and then we'll have to figure out something for globals with GC reference type), but it felt slightly more regular to use this initializer-expression concept if it was already present for data segment offsets since this affords us greater flexibility in the future.

DynamicLinking.md
jfbastien: We used to have "embedder" in a few places, looks like only 2 are left now.

DynamicLinking.md
```diff
@@ -1,49 +1,87 @@
...
+A more realistic module using libc would have more imports including:
+* an immutable `i32` global import for the offset in linear memory to place
+  global [data segments](Modules.md#data-section) and later use as a constant
+  base address when loading and storing from globals
+* an immutable `i32` global import for the offset into the indirect function
+  table at which to place the modules' indirectly called functions and later
+  compute their indices for address-of
+
+One extra detail is what to use as the [module name](Modules.md#imports) for
+imports (since WebAssembly has a two-level namespace). One option is to have a
```
jfbastien: Is there a reference for the two-level namespace stuff? I can't seem to find it, maybe just file a bug to add it later so the PR doesn't get bogged down?

FutureFeatures.md
```diff
@@ -425,3 +415,46 @@ of WebAssembly in browsers:
...
+prerequisite, WebAssembly would need first-class support for 
+[GC references](GC.md) in expressions and locals. Given that, the following
+could be added:
+* `get_table`/`set_table`: get or set the table element at a given dynamic
+  index; the got/set value would have a GC reference type
+* `grow_table`: grow the current table (up to the optional maximum), similar to
+  `grow_memory`
+* `current_table_length`: like `current_memory`.
+
+Additionally, in the MVP, tables' an only store untyped functions. This could be
```
jfbastien: I don't get this sentence.

lukewagner: Pushed an updated attempted to clarify.

DynamicLinking.md
lukewagner: Righto: #706.

Modules.md
titzer: This is a tricky corner if we start allowing initialization expressions, since I assume you want to propose they run before data segments are copied? That would require imposing restrictions on what expressions can do so that they don't accidentally observe the state of the world before data segments are loaded. In either case, that means compiling + running expressions just to get initial values of imports. I think we want to avoid that cost (and also not require an engine have a full WASM interpreter to avoid that cost). It would be simpler to kick that initialization problem up to the embedding environment and thus require that immutable globals either have a module-specified numeric value or are imported (and thus the embedding environment must provide their values). Immutable GC reference type globals could be initialized by the embedder too, or if we have WASM managed data, given as serialized objects in an appropriate objects section.

JSStats: The need for initialization expressions here worries me took. I also have some doubts about the use case because practically runtimes on the web will probably need to decide if shared code is a single instance of code which can not be patched, or alternatively a copy that is patched? Developers will probably need some consistency here to know if creating a lot of instances of the same code is going to use a lot of memory or not. Lets say that the runtimes decide to reuse a single instance of code so can not patch it, then this support seems unnecessary?

lukewagner: @titzer The idea is that initializer expressions would never remotely encompass the full set of expressions.  Initially here we have constants and immutable globals.  Maybe later we add `i32.add` :)  But I have no interest in generalizing past that; I'd like `EvalInitializerExpression` to be a 10 line function.  Maybe you're worried it's a slippery slope or it's too cute?  In that case, we could just do something custom for these two cases.  But I think we'll be able to restrain ourselves from slipping.

AstSemantics.md
```diff
@@ -232,12 +242,53 @@ operator may be added. However, due to normal fragmentation, applications are
...
+[defined inside the module](Modules.md#table-section), with defaultness
+indicated by a flag on the import or definition. After import or definition,
+there is no difference when calling into a table whether it was imported or
+defined internally.
+
+In the MVP, the primary purpose of tables is to implement indirect function
+calls in C/C++ using an integer index as the pointer-to-function and the table
+to hold the array of indirectly-callable functions. Thus, in the MVP:
+* tables may only be accessed from WebAssembly code via [`call_indirect`](#calls);
+* the only allowed table element type is "function" (no signature);
```
rossberg-chromium: Nit: can we name this `anyfunc`, or something like that? That makes it clearer that it's a special form of `any` type, and avoids confusion with the type `(func)` with an empty signature (as opposed to none).

lukewagner: Good point, fixed.

Modules.md
```diff
@@ -170,30 +231,154 @@ reserve up to the *maximum* is not. When a maximum memory size is *not* declared
...
+the declared maximum. As with linear memory, when a maximum is declared,
+implementations *should* (non-normative) attempt to reserve virtual memory up to
+the maximum size. While failure to allocate the *initial* memory size is a
+runtime error, failure to reserve up to the *maximum* is not. When a maximum
+memory size is *not* declared, on architectures with limited virtual address
+space, engines should allocate only the initial size and reallocate on demand.
+
+## Elements section
+
+The intial contents of a tables' elements are sentinel values (that would throw
```
rossberg-chromium: Nit: s/throw/trap/

Modules.md
```diff
@@ -170,30 +231,154 @@ reserve up to the *maximum* is not. When a maximum memory size is *not* declared
...
+This split aids in streaming compilation by putting the function bodies,
+which constitute most of the byte size of the module, near the end so that all
+metadata necessary for recursive module loading and parallel compilation is
+available before compilation begins.
+
+## Definition Index Space
+
+The *definition index space* represents the union of all definitions introduced
+by the [import](#imports), [global](#global-section), [memory](#linear-memory-section),
+[table](#table-section), and [code](#code-section) sections. Each of these
```
rossberg-chromium: Nit: function section instead of code section? It's equivalent, but function may be clearer.

Modules.md
```diff
@@ -170,30 +231,154 @@ reserve up to the *maximum* is not. When a maximum memory size is *not* declared
...
+   definition in the module;
+ * the *code* section contains the [function body](BinaryEncoding.md#function-bodies)
+   of each function declared by the function section.
+
+This split aids in streaming compilation by putting the function bodies,
+which constitute most of the byte size of the module, near the end so that all
+metadata necessary for recursive module loading and parallel compilation is
+available before compilation begins.
+
+## Definition Index Space
```
rossberg-chromium: Hm, can we somehow avoid this cumulative index space? It is weird that the same entity has different indices, depending on where it is referenced from. It also creates brittle forward compatibility -- imagine, for example, that we need to make type definitions exportable at some point (a rather likely scenario), what then?

Instead, we could e.g. distinguish the index/name spaces at the reference site. That will require an extra byte, but OTOH, the actual value will be smaller, which might give at least some space back in the LEB encoding (and compression).

lukewagner: Yes, I think we could do that and I see the reasoning.  The two uses (exports and table elements) constitute a negligible number of bytes.  In the MVP, when there is only one element type, it'll feel kindof contrived to have each table element say "it's a function", but this will make more sense when there can be more general element types.  Alternatively, we could have the element type determine the index space (so `anyfunc` table elements would use a function index) but this will lead to irregularity for other element types so I'm thinking it's not worth it.

Modules.md
lukewagner: Updated

Modules.md
lukewagner: Updated

Modules.md
lukewagner: Updated

Modules.md
JSStats: @lukewagner What about the use case? Will the browsers be implementing shared code or creating new copies for each instance, or will web developers be able to express a preference?

lukewagner: The use case has been given multiple times now in this issue.  I think browsers will do both sharing and copying in practice (e.g., on register-starved x86, for small modules, it may be a distinct win to be able to patch in heap base/length), with, I'd expect, a preference for sharing.  Certainly we're working towards sharing in FF.  I don't think this detail should be exposed to the user since the right choice depends on a number of device-specific parameters.

PR #683, opened 2016-05-06
Make opcode 0x00 `unreachable`.

sunfishcode:
Make opcode 0x00 `unreachable`, and move `nop` to a non-zero opcode.

All-zeros is one of the more common patterns of corrupted data. This
change makes it more likely that code that is accidentally zeroed, in
whole or in part, will be noticed when executed rather than silently
running through a nop slide.

Obviously, this doesn't matter when an opcode table is present, but
if there is a default opcode table, it would presumably use the
opcodes defined here.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -326,7 +326,7 @@ It is legal to have several entries with the same type.
 
 | Name | Opcode | Immediates | Description |
 | ---- | ---- | ---- | ---- |
-| `nop` | `0x00` | | no operation |
+| `unreachable` | `0x00` | | trap immediately |
 | `block` | `0x01` |  | begin a sequence of expressions, the last of which yields a value |
 | `loop` | `0x02` |  | begin a block which can also form control flow loops |
 | `if` | `0x03` | | begin if expression |
@@ -336,7 +336,7 @@ It is legal to have several entries with the same type.
 | `br_if` | `0x07` | argument_count : `varuint1`, relative_depth : `varuint32` | conditional break that targets an outer nested block |
 | `br_table` | `0x08` | see below | branch table control flow construct |
 | `return` | `0x09` | argument_count : `varuint1` | return zero or one value from this function |
-| `unreachable` | `0x0a` | | trap immediately |
+| `nop` | `0x0a` | | no operation |
 | `end` | `0x0f` | | end a block, loop, or if |
 
 Note that there is no explicit `if_else` opcode, as the else clause is encoded with the `else` bytecode.


kripken:
Nice!

jfbastien:
I like, A64 does the same :)

JSStats:
I see absolute no merit in this change. With an operator table the most common opcode will likely have index zero so in this case it does not help security, and without an operator table this will likely regress the compressed file size (assuming the default indexes are sorted by general frequency of use). The wasm is filtered through a decoder/validator that needs to be secure and well tested, and a zero is proportionally no more significant than any other value for security. The typical decoder converts directly to an SSA form an internal representation where this is not relevant again.

The only use case that comes to mind is an internal wasm code generator, such as asm.js-to-wasm, that might want to avoid validation assuming valid code generation, but this can just use a special operator table to harden it if it wishes. Lets not compromise the web wasm encoding just for this.

jfbastien:
Choosing to not have an operator table is choosing to compress poorly, I don't think we need to care about this. I don't understand your objection otherwise: why is one number better than another?

It seems easy to zero-init memory and fail to fill it properly, or mess up offsets. I agree @sunfishcode's suggestion doesn't fix many types of bugs, but it fixes a small class of them and I can't see a single downside.

JSStats:
@jfbastien Surely a default operator table would be chosen to compress well? If an operator table is required for good compression then it will be the default for the web anyway and web developers could assign zero to `unreachable` if they wished using this operator table, but since it is the developers decision it can't improve the runtime security.

The compression literal encoder will assign short codes to common byte symbols. It seems natural to use small values for high frequency bytes, and natural to use a zero base, but an alternative base could be chosen but it would be a pervasive decision that zero is the uncommon byte symbol. Access offsets of zero are most common; an alignment of zero could represent natural alignment; the local variable indexes are zero based and can be usefully sorted by frequency of use with the lowest first; a break target of zero is most common; etc. Some of these are naturally zero based, and having to offset them by one would just frustrate and complicate a decoder.

It just seems to me that the 'security' use case noted as rationale for this change has little merit, and that being able to use zero for the most common opcode might have much more merit.

titzer:
This needs to be rebased onto a branch for 0xc

lukewagner:
The change sgtm.  We were planning a big opcode-table-defragmentation at some point, should that be in 0xc too?  Is there an 0xc branch started yet?

sunfishcode:
Closing to rebase on binary_0xc.

I'll also note that this is not about security. This is just about being slightly friendlier in the presence of a rare-but-not-unheard-of bug or accident.

PR #684, opened 2016-05-06
Make opcode 0x00 `unreachable`.

sunfishcode:
Make opcode 0x00 `unreachable`, and move `nop` to a non-zero opcode.

All-zeros is one of the more common patterns of corrupted data. This
change makes it more likely that code that is accidentally zeroed, in
whole or in part, will be noticed when executed rather than silently
running through a nop slide.

Obviously, this doesn't matter when an opcode table is present, but
if there is a default opcode table, it would presumably use the
opcodes defined here.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -326,7 +326,7 @@ It is legal to have several entries with the same type.
 
 | Name | Opcode | Immediates | Description |
 | ---- | ---- | ---- | ---- |
-| `nop` | `0x00` | | no operation |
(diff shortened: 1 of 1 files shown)


rossberg-chromium:
LGTM

sunfishcode:
This has an LGTM; is there anything else to consider? This in 0xc, so if we do this now we should probably do the other opcode defragmentation in 0xc too, but that seems doable.

kripken:
Not just an lgtm, but also 4 thumbs up emojis! :)

PR #689, opened 2016-05-12
Remove link to wasm-e2e from README.md

binji:
It is not maintained, so we shouldn't call it out on the first page you
see.


lukewagner:
lgtm

PR #690, opened 2016-05-13
Remove redundant text on quad precision types

distransient:


binji:
I thought maybe this was a joke (but it was only repeated three times)

PR #695, opened 2016-05-23
Move select out of the "control flow opcodes" section.

sunfishcode:
And into the "Basic opcodes" section, renumbering the remainder of the
section accordingly.

This is for the binary_0xc branch, and is one step of a broader opcode reorganization.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -1,13 +1,18 @@
 # Abstract Syntax Tree Semantics
 
-WebAssembly code is represented as an Abstract Syntax Tree (AST) where each node
-represents an expression. Each function body consists of a list of expressions.
-All expressions and operators are typed, with no implicit conversions or overloading rules.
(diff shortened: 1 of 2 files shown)


kripken:
lgtm

titzer:
I think we should do a big one-step reorganization instead of small steps, and the reorganization needs to balance a couple of things: compressibility, extensibility, and logical consistency. In particular, some advice from the Brotli compression folks suggested that the opcodes should be organized to reduce entropy and allow the compression algorithm to do a better job.

sunfishcode:
If someone wants to compute a Brotli-friendly order, that's fine with me. If we wish to balance it with logical consistency, organizing the opcodes for logical consistency before we begin will simplify the process.

rossberg-chromium:
Should be subsumed by #826 now.

PR #696, opened 2016-05-23
Put f32.const before f64.const, for consistency.

sunfishcode:
This is for the binary_0xc branch, and is one step of a broader opcode reorganization.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -1,13 +1,18 @@
 # Abstract Syntax Tree Semantics
 
-WebAssembly code is represented as an Abstract Syntax Tree (AST) where each node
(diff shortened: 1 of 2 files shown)


kripken:
lgtm

rossberg-chromium:
I'm fine with all the changes you have suggested so far. But instead of making many tiny changes individually, how about a single comprehensive PR for reorganising the opcode space? That would help seeing the big picture and not fragmenting the discussion.

titzer:
Shall we do this in 0xD?

rossberg-chromium:
Should be subsumed by #826 now.

PR #702, opened 2016-06-06
Suggested Compression layer 1 document.

KarlSchimpf:
Here is a suggested design document for layer 1 compression WASM modules.

It is a pull request in the design repo. If we want, we create an issue for high level discussions instead of a pull request.

Also we may want to move some details into a document in https://github.com/WebAssembly/decompressor-prototype but I'm just posting this way for more visibility.


KarlSchimpf:
Closing this pull request. Replacing with this
[pull request](https://github.com/WebAssembly/design/pull/705).

Issue #698, opened 2016-05-27
Add a `get_and_zero_local` operator.


ghost:
The use case is to help inform the runtime compiler that local variables are not used (or constants) over code ranges.

This might be exploited by a baseline compiler to avoid flushing local variables cached in registers to memory. If a use of a local variable also zeros it then whatever value it had is no longer needed.

For decoding to SSA, if wasm code can use `get_and_zero_local` on the last use of a local then at control flow merge points the locals not live will be flagged as constant zero rather than having potentially different definitions, and this would avoid the need to emit a phi for them rather a single definition could be carried forward flagged as zero.

My expressionless encoding experiment noted excess phi's being generated for unused locals as a problem, but perhaps being able to zero a local on the last use of a definition offers a way to address this. Something to explore.

flagxor:
Don't think this is necessary for MVP.
This seems like a very particular flavor of trying to pre-compute liveness. That might be an interesting direction, but also seems very open ended.
Worth exploring in the future.

JSStats:
@flagxor It might not be necessary for the MVP, or it might be, we just don't have a MVP at this point and far from it I seem to be the only person articulating any design that might scale to handle multiple values. One of my proposals was the expressionless encoding but on exploration it caused more effort decoding and produced more phi's, however it later occurred that `get_and_zero_local` might address some of these shortcomings. If decoding efficiency is a show stopper for a MVP then this might be too.

How do we know that this can be explored later efficiently? Perhaps other design decisions will make it too inefficient to encode? I would like to see a path forward, a plan, and the expressionless encoding is at least a path and a plan.

Perhaps you would like to help move the design forward and articulating some proposal for handling multiple return values from functions might at least create some discussion?

Issue #697, opened 2016-05-24
LES as the text format


qwertie:
I would like to propose the next version of [LES](http://loyc.net/les/) to be used as the WebAssembly text format. More specifically:
- I propose that the Wasm text format be _compatible_ with the next version of [LES](http://loyc.net/les), in the sense of being a subset of it, such that a "full" LES parser would not reject anything that is valid Wasm text.
- For the MVP, those that want to parse Wasm text will be able to choose whether to use a custom parser dedicated to Wasm or a generic LES parser.
- My proposed syntax (link at bottom) _is not_ compatible with the current LES specification, but LES is in beta and can still be tweaked to Wasm's needs. Based on the Wasm text format, a third version of LES (LESv3) will be drafted before the end of 2016.

This gives the CG some freedom to make _some_ changes to LES and not others. Specifically, any elements that make sense _only_ in WebAssembly (e.g. keywords for wasm opcodes) would not be permitted, but changes such as tweaks to operator precedence, handling of semicolons, the grammar of LES "superexpressions", or the name used for "infinity", are fine.
## Why use LES for WebAssembly?
- **Fewer parsers will be needed in the future**: in today's world, every new language needs a new parser to be written, bikeshedded, and specified from scratch. While LES is not sufficient for all languages, it will be enough for some (and in particular it is sufficient for Wasm, which has weaker reasons than most languages to use a custom grammar); being part of Wasm encourages use of LES elsewhere (_edit_: consider DSLs and advanced search boxes. I wrote some GIS software that did formulas & searches, so used an ANTLR-based parser. But why are people still writing custom parsers for expressions? There should be a standard - so here it is.)
- More importanty, **learning curves decrease**: imagine a future world in which a user has _already_ used LES as a programming language or a data language. She wants to write Wasm assembly for the first time, and _it's easier to learn_, because the syntax is already familiar. She may have many concepts to learn regarding the _semantics_ of Wasm, but at least the syntax is easy. Conversely, she may learn Wasm first and another LES-based language later. Either way, the same benefit accrues.
  
[...]

PR #704, opened 2016-06-07
A JS-style Text Format

sunfishcode:
This proposes an official text format for WebAssembly, aimed at browsers to use in "View Source", debugging, and related tooling. It uses a JavaScript-like syntax for readability and familiarity on the Web, though it differs from JS in several respects, as it aims to reflect the underlying WebAssembly language.

This proposal is meant to serve as a beginning. We'd like to establish this as a concrete place to start, so that we can then iterate, as we did with BinaryFormat.md.

You can try out a prototype of this proposal yourself in [Firefox Nightly](https://nightly.mozilla.org/), for example by playing the [AngryBots demo](http://webassembly.github.io/demo/) with the debugger open and examining the wasm file in the debugger.

[Here's a screenshot of it in action.](http://people.mozilla.org/~dgohman/angrybots-text.png)

A more complete description of the grammar and a parser implementation are available [here](https://github.com/mbebenita/was). If this proposal is accepted, we'd like to move this repository under the WebAssembly GitHub organization to serve as the interim spec for the text format during the initial discussion.

### DIFF
--- a/TextFormat.md
+++ b/TextFormat.md
@@ -38,7 +38,7 @@ support more human-readable representations, but never at the cost of accurate r
 
 # Official Text Format
 
-## Philosophy:
+## Design Principles:
 
  - Use JS-style sensibilities when there aren't reasons to do otherwise.
  - It's a compiler target, not a programming language, but readability still counts.
--- a/TextFormat.md
+++ b/TextFormat.md
@@ -18,7 +18,7 @@ WebAssembly semantics.
 The text format is neither a subset or superset of JavaScript syntax, and it
 cannot be directly evaluated as JavaScript. Some parts of the syntax resemble
 JavaScript so that it's familiar and readable, however there are also
-substantive reasons to use notation that is different than JavaScript (for
+substantive reasons to use notation that is different from JavaScript (for
 example, WebAssembly has a 32-bit integer type, and it is represented
 in the text format, since that is the natural thing to do for WebAssembly,
 regardless of JavaScript not having such a type). On the other hand,
@@ -40,7 +40,7 @@ support more human-readable representations, but never at the cost of accurate r
 
 ## Philosophy:
 
- - Use JS-style sensibilities when there aren't reasons otherwise.
+ - Use JS-style sensibilities when there aren't reasons to do otherwise.
  - It's a compiler target, not a programming language, but readability still counts.
 
 
--- a/TextFormat.md
+++ b/TextFormat.md
@@ -106,7 +106,7 @@ of the function syntax, reflecting how function bodies in wasm are block-like.
 The last expression of the function body here acts as its return value. This
 works in all block-like constructs (`block`, function body, `if`, etc.)
 
-`>s` means *signed* greater-than. explicit unsigned or signed operators will be
+`>s` means *signed* greater-than. Explicit unsigned or signed operators will be
 suffixed with 'u' or 's', respectively.
 
 The `$` sigil on user names cleanly ensures that they never collide with wasm
--- a/TextFormat.md
+++ b/TextFormat.md
@@ -361,7 +361,7 @@ Signed and unsigned operators are suffixed with `s` and `u`, respectively,
 so `i32.div_s` looks like `/s`, and so on.
 
 These operators use C/JS-style operator precedence rules, except that in order
-to correct [an old mistake](http://www.lysator.liu.se/c/dmr-on-or.html)),
+to correct [an old mistake](http://www.lysator.liu.se/c/dmr-on-or.html),
 it is not valid to rely on precedence alone between a `&`, `|`, or `^` operator
 and another operator of similar precedence. Parentheses are required in such
 contexts in order to ensure that the code reads as intended.
--- a/TextFormat.md
+++ b/TextFormat.md
@@ -5,26 +5,21 @@ The purpose of this text format is to support:
   source can be viewed) in a natural way.
 * Presentation in browser development tools when source maps aren't present
   (which is necessarily the case with [the Minimum Viable Product (MVP)](MVP.md)).
-* Writing WebAssembly code directly for reasons including pedagogical,
-  experimental, debugging, optimization, and testing of the spec itself.
+* Working with WebAssembly code directly for reasons including pedagogical,
+  experimental, debugging, profiling, optimization, and testing of the spec
+  itself.
 
 The text format is equivalent and isomorphic to the [binary format](BinaryEncoding.md).
 
-The text format will be standardized, but only for tooling purposes:
-* Compilers will support this format for `.S` and inline assembly.
-* Debuggers and profilers will present binary code using this textual format.
-* Browsers will not parse the textual format on regular web content in order to
-  implement WebAssembly semantics.
+The text format will be standardized, but only for tooling purposes; browsers
+will not parse the textual format on regular web content in order to implement
+WebAssembly semantics.
 
-Given that the code representation is actually an
-[Abstract Syntax Tree](AstSemantics.md), the syntax would contain nested
-statements and expressions (instead of the linear list of instructions most
-assembly languages have).
-
-There is no requirement to use JavaScript syntax; this format is not intended to
-be evaluated or translated directly into JavaScript. There may also be
+The text format is neither a subset or superset of JavaScript syntax, and it
+cannot be directly evaluated as JavaScript. Some parts of the syntax resemble
+JavaScript so that it's familiar and readable, however there are also
 substantive reasons to use notation that is different than JavaScript (for
-example, WebAssembly has a 32-bit integer type, and it should be represented
+example, WebAssembly has a 32-bit integer type, and it is represented
 in the text format, since that is the natural thing to do for WebAssembly,
 regardless of JavaScript not having such a type). On the other hand,
 when there are no substantive reasons and the options are basically
@@ -43,39 +38,365 @@ support more human-readable representations, but never at the cost of accurate r
 
 # Official Text Format
 
-WebAssembly currently doesn't have a final, official, text format. As detailed above the
-main purpose of the text format will be for human consumption, feedback from humans on
-readability will therefore factor into standardizing a text format.
+## Philosophy:
 
-There are, however, prototype syntaxes which are used to bring up WebAssembly: it's easier
-to develop using a text format than it is with a binary format, even if the ultimate
-WebAssembly format will be binary. Most of these prototypes use [s-expressions][] because they
-can easily represent expression trees and [ASTs](AstSemantics.md) (as opposed to CFGs)
-and don't have much of a syntax to speak of (avoiding syntax bikeshed discussions).
+ - Use JS-style sensibilities when there aren't reasons otherwise.
+ - It's a compiler target, not a programming language, but readability still counts.
 
-  [s-expressions]: https://en.wikipedia.org/wiki/S-expression
 
-Here are some of these prototypes. Keep in mind that these *aren't* official, and the final
-official format may look entirely different:
+## High-level summary:
 
-* [Prototype specification][] consumes an s-expression syntax.
-* [WAVM backend][] consumes compatible s-expressions.
-* [sexpr-wasm prototype][] consumes compatible s-expressions, and works closely with the [V8 prototype][].
-* [LLVM backend][] (the `CHECK:` parts of these tests) emits compatible s-expressions.
-* [ilwasm][] emits compatible s-expressions.
-* [wassembler][] consumes a different syntax, and works closely with the [V8 prototype][].
-* [binaryen][] can consume compatible s-expressions.
+ - Curly braces for function bodies, blocks, etc., `/* */`-style and `//`-style
+   comments, and whitespace is not significant.
+   Nested `/* */`-style comments follow JavaScript's rules.
 
-  [prototype specification]: https://github.com/WebAssembly/spec/tree/master/ml-proto/test
-  [LLVM backend]: https://github.com/llvm-mirror/llvm/tree/master/test/CodeGen/WebAssembly
-  [WAVM backend]: https://github.com/AndrewScheidecker/WAVM/tree/master/Test
-  [wassembler]: https://github.com/ncbray/wassembler/tree/master/demos
-  [V8 prototype]: https://github.com/WebAssembly/v8-native-prototype
-  [ilwasm]: https://github.com/WebAssembly/ilwasm
-  [sexpr-wasm prototype]: https://github.com/WebAssembly/sexpr-wasm-prototype
-  [binaryen]: https://github.com/WebAssembly/binaryen
+ - Semicolons separate expressions inside blocks, loops, and so on. The
+   last statement in the sequence may omit its trailing semicolon to indicate
+   that its result is being used as the result value of the parent construct.
+
+ - `get_local` looks like a simple reference; `set_local` looks like an
+   assignment. Constants use a simple literal syntax. This makes wasm's most
+   frequent opcodes very concise.
+
+ - Infix syntax for arithmetic, with simple overloading. Explicit grouping via
+   parentheses. Concise and familiar with JS and others.
+
+ - Prefix syntax with comma-separated operands for all other operators. For less
+   frequent opcodes, prefer just presenting operator names, so that they're easy
+   to identify.
+
+ - Typescript-style `name: type` declarations.
+
+ - Parentheses around call arguments, eg. `call $functionname(arg, arg, arg)`,
+   and `if` conditions, eg. `if ($condition) { call $then() } else { call $else() }`,
+   because they're familiar to many people and not too intrusive.
+
+ - Put labels "where they go".
+
+
+## Examples:
+
+### Basics
+
+```
+  function $fac-opt ($a:i64) : (i64) {
+    var $x:i64;
+    $x = 1;
+    br_if ($a <s 2) $end;
+    loop $loop {
+      $x = $x * $a;
+      $a = $a + -1;
+      br_if ($a >s 1) $loop;
+    }
+  $end:
+    $x
+  }
+```
+
+(from [fac.wast](https://github.com/WebAssembly/spec/blob/master/ml-proto/test/fac.wast))
+
+The function return type has parentheses for symmetry with the parameter types,
+anticipating adding multiple return values to wasm in the future.
+
+The curly braces around the function body are not a `block` node; they are part
+of the function syntax, reflecting how function bodies in wasm are block-like.
+
+The last expression of the function body here acts as its return value. This
+works in all block-like constructs (`block`, function body, `if`, etc.)
+
+`>s` means *signed* greater-than. explicit unsigned or signed operators will be
+suffixed with 'u' or 's', respectively.
+
+The `$` sigil on user names cleanly ensures that they never collide with wasm
+keywords, present or future.
+
+The last statement has no trailing semicolon; this indicates that its result
+value is being used as the block result value.
+
+
+### Linear memory addresses
+
+Addresses are printed as `[base,+offset,align=val]`. The offset can be omitted
+when it is zero, and the alignment can be omitted for natural alignment.
+
+Addresses are prefixed by an identifier giving the type of a load, or the
+type of the value to store, optionally followed by a colon and a number of
+bits of memory to access (for truncating stores or extending loads).
+For loads, the number of bits is followed by `s` or `u` for signed and
+unsigned.
+
+Examples:
+
+```
+  function $loads_and_stores ($ptr : i32) : (i32) {
+    var $t : i32;
+
+    // Simple cases.
+    $t = i32[$ptr];           // plain i32.load from address $ptr
+    i32[$ptr] = $t;           // plain i32.store of $t to address $ptr
+
+    // Various modifiers.
+    f32[$ptr,+4] = -0x1.8p0;  // f32.store to address $ptr plus 4
+    i64:8s[$ptr,+17,align=1]; // i64.load8_s from address $ptr plus 17 with alignment 1
+  }
+```
+
+### A slightly larger example:
+
+Here's some C code:
+
+```
+  float Q_rsqrt(float number)
+  {
+      long i;
+      float x2, y;
+      const float threehalfs = 1.5F;
+
+      x2 = number * 0.5F;
+      y  = number;
+      i  = *(long *) &y;
+      i  = 0x5f3759df - (i >> 1);
+      y  = *(float *) &i;
+      y  = y * (threehalfs - (x2 * y * y));
+      y  = y * (threehalfs - (x2 * y * y));
+
+      return y;
+  }
+```
+
+Here's the corresponding LLVM wasm backend output + binaryen + slight tweaks:
+
+```
+  (func $Q_rsqrt (param $0 f32) (result f32)
+    (local $1 f32)
+    (set_local $1
+      (f32.reinterpret/i32
+        (i32.sub
+          (i32.const 1597463007)
+          (i32.shr_s
+            (i32.reinterpret/f32
+              (get_local $0))
+            (i32.const 1)))))
+    (set_local $1
+      (f32.mul
+        (get_local $1)
+        (f32.sub
+          (f32.const 0x1.8p+0)
+          (f32.mul
+            (get_local $1)
+            (f32.mul
+              (get_local $1)
+              (set_local $0
+                (f32.mul
+                  (get_local $0)
+                  (f32.const 0x1p-1))))))))
+    (f32.mul
+      (get_local $1)
+      (f32.sub
+        (f32.const 0x1.8p+0)
+        (f32.mul
+          (get_local $1)
+          (f32.mul
+            (get_local $0)
+            (get_local $1)))))
+   )
+```
+
+And here's the proposed text syntax:
+
+```
+   function $Q_rsqrt ($0:f32) : (f32) {
+     var $1:f32;
+     $1 = f32.reinterpret/i32(1597463007 - ((i32.reinterpret/f32($0)) >> 1));
+     $1 = $1 * (0x1.8p0 - $1 * ($0 = $0 * 0x1p-1) * $1);
+     $1 * (0x1.8p0 - $1 * $0 * $1)
+   }
+```
+
+This shows off the compactness of infix operators with overloading. In the
+s-expression syntax, these expressions are quite awkward to read, and this
+isn't even a very big example. But the text syntax here is very short.
+
+### Labels
+
+Excerpt from labels.wast:
+
+```
+  (func $loop3 (result i32)
+    (local $i i32)
+    (set_local $i (i32.const 0))
+    (loop $exit $cont
+      (set_local $i (i32.add (get_local $i) (i32.const 1)))
+      (if (i32.eq (get_local $i) (i32.const 5))
+        (br $exit (get_local $i))
+      )
+      (get_local $i)
+    )
+  )
+```
+
+Corresponding proposed text syntax:
+
+```
+  function $loop3 () : (i32) {
+    var $i:i32;
+    $i = 0;
+    loop $cont {
+      $i = $i + 1;
+      if ($i == 5) {
+        br ($i) $exit;
+      }
+    $exit:
+    }
+  }
+```
+
+Note that the curly braces are part of the `if`, rather than introducing a
+block. This reflects how `if` essentially provides `block`-like capabilities
+in the wasm binary format.
+
+### Nested blocks
+
+Label definitions, like the `$exit:` above, introduce additional blocks nested
+within the nearest `{`, without requiring their own `{`. This allows the deep
+nesting of `br_table` to be printed in a relatively flat manner:
+
+```
+  {
+    br_table ($index) [$red, $orange, $yellow, $green], $default;
+  $red:
+      // ...
+  $orange:
+      // ...
+  $yellow:
+      // ...
+  $green:
+      // ...
+  $default:
+  }
+```
+
+representing the following in nested form:
+
+```
+  (block $default
+    (block $green
+      (block $yellow
+        (block $orange
+          (block $red
+            (br_table (get_local $index), [$red, $orange, $yellow, $green] $default)
+          )
+          // ...
+        )
+        // ...
+      )
+      // ...
+    )
+    // ...
+  )
+```
+
+`br_table`s can have large numbers of labels, so this feature allows us to
+avoid very deep nesting in many cases.
+
+Note that when a label appears just before the closing `}`, it doesn't introduce
+a new block; it just provides a name for the enclosing block's label.
+
+## Operators with special syntax
+
+The default syntax for operators is their opcode name followed by a
+comma-separated list of operands. However, several operators have
+special syntax.
+
+
+### Control flow operators ([described here](https://github.com/WebAssembly/design/blob/master/AstSemantics.md))
+
+| Name | Syntax | Examples
+| ---- | ---- | ---- |
+| `block` | `{` … *label*: `}` | `{ br $a; a: }`
+| `loop` | `loop` *label* `{` … `}` | `loop $a { br $a; }`
+| `if` | `if` `(` *condition* `)` `{` … `}` | `if (0) { 1 }`
+| `if` | `if` `(` *condition* `) `{` … `} else `{` … `}` | `if (0) { 1 } else { 2 }`
+| `br` | `br` *label* | `br $where`
+| `br` | `br` `(` *expr* `)` *label* | `br ($v) $where`
+| `br_if` | `br_if` `(` *expr* `)` *label* | `br_if ($x < $y) $where`
+| `br_if` | `br_if` `(` *expr* `,` *condition* `)` *label* | `br_if ($v, $x < $y) $where`
+| `br_table` | `br_table` `(` *index-expr* `)` `[` *label* `,` … `]` `,` *default-label* | `br_table ($i) [$somewhere, $or_other], $default`
+| `br_table` | `br_table` `(` *expr* `,` *index-expr* `)` `[` *label* `,` … `]` `,` *default-label* | `br_table ($v, $i) [$somewhere, $or_other], $default`
+| `return` | `return` | `return`
+| `return` | `return` *expr* | `return $x`
+| `unreachable` | `unreachable` | `unreachable`
+
+
+### Basic operators ([described here](https://github.com/WebAssembly/design/blob/master/AstSemantics.md#constants))
+
+| Name | Syntax | Example
+| ---- | ---- | ---- |
+| `i32.const` | … | `234`, `0xfff7`
+| `i64.const` | … | `234`, `0xfff7`
+| `f64.const` | … | `0.1p2`, `infinity`, `nan:0x789`
+| `f32.const` | … | `0.1p2`, `infinity`, `nan:0x789`
+| `get_local` | *name* | `$x + 1`
+| `set_local` | *name* `=` *expr* | `$x = 1`
+| `select` | `select` *expr*, *expr*, *expr* | `select 1, 2, $x < $y`
+| `call` | `call` *name* `(`*expr* `,` … `)` | `call $min(0, 2)`
+| `call_import` | `call_import` *name* `(`*expr* `,` … `)` | `call_import $max(0, 2)`
+| `call_indirect` | `call_indirect` *signature-name* `[` *expr* `] (`*expr* `,` … `)` | `call_indirect $foo [1] (0, 2)`
+
+### Arithmetic operators ([described here](AstSemantics#32-bit-integer-operators))
+
+Arithmetic operators use C/JS-like infix and prefix notation.
+
+Add, sub, mul, div, rem, and, or, xor, shl, and shr operators use
+`+`, `-`, `*`, `/`, `%`, `&`, `|`, `^`, `<<`, and `>>`, respectively.
+
+`eq`, `ne`, `lt`, `le`, `gt`, `ge`, operators use
+`==`, `!=`, `<`, `<=`, `>`, `>=`, respectively.
+
+`eqz` and `neg` operators use prefix unary `!` and `-`, respectively.
+
+Signed and unsigned operators are suffixed with `s` and `u`, respectively,
+so `i32.div_s` looks like `/s`, and so on.
+
+These operators use C/JS-style operator precedence rules, except that in order
+to correct [an old mistake](http://www.lysator.liu.se/c/dmr-on-or.html)),
+it is not valid to rely on precedence alone between a `&`, `|`, or `^` operator
+and another operator of similar precedence. Parentheses are required in such
+contexts in order to ensure that the code reads as intended.
+
+All other arithmetic operators use their actual name in a prefix notation,
+such as `f32.sqrt(…)`, with comma-seperated operands.
+
+## Module-level syntax
+
+Type declarations:
+
+`type $ternary of function (i32,i32,i32) : (i32)`
+
+Imports:
+
+`import "print" as $print_int from "global.PrintingStuff" typeof function (f64) : ()`
+
+Function table:
+
+`table [$func0,$func1,$func2,…]`
+
+Exports:
+
+`export memory as "memory"`
+
+`export $func0 as "ThisIsFunc0"`
+
+Function definitions and local variables:
+
+```
+function $func0($x : i32, $y : i32, $y : i32) : (f64) {
+    var $a : i32, $b : i64, $c : f32
+```
 
 # Debug symbol integration
+
+## Debug symbol section
 
 The binary format inherently strips names from functions, locals, globals, etc,
 reducing each of these to dense indices. Without help, the text format must
@@ -83,3 +404,11 @@ therefore synthesize new names. However, as part of the [tooling](Tooling.md)
 story, a lightweight, optional "debug symbol" global section may be defined
 which associates names with each indexed entity and, when present, these names
 will be used in the text format projected from a binary WebAssembly module.
+
+## Code references
+
+In the binary format, references to locations in the code are done through
+byte offsets. In the text format, these references are represented symbolically
+with a special label syntax, prefixed by an `@` sign. These labels have no
+semantic effect, and merely serve to identify a location in the code that can
+be referred to from other sections.
--- a/TextFormat.md
+++ b/TextFormat.md
@@ -82,13 +82,15 @@ support more human-readable representations, but never at the cost of accurate r
   function $fac-opt ($a:i64) : (i64) {
     var $x:i64;
     $x = 1;
-    br_if ($a <s 2) $end;
-    loop $loop {
-      $x = $x * $a;
-      $a = $a + -1;
-      br_if ($a >s 1) $loop;
+    {
+      br_if ($a <s 2) $end;
+      loop $loop {
+        $x = $x * $a;
+        $a = $a + -1;
+        br_if ($a >s 1) $loop;
+      }
+    $end:
     }
-  $end:
     $x
   }
 ```
--- a/TextFormat.md
+++ b/TextFormat.md
@@ -334,16 +334,22 @@ special syntax.
 
 | Name | Syntax | Example
 | ---- | ---- | ---- |
-| `i32.const` | … | `234`, `0xfff7`
-| `i64.const` | … | `234`, `0xfff7`
-| `f64.const` | … | `0.1p2`, `infinity`, `nan:0x789`
-| `f32.const` | … | `0.1p2`, `infinity`, `nan:0x789`
+| `i32.const` | … | `234`, `0xfff7`, -47
+| `i64.const` | … | `234`, `0xfff7`, -47
+| `f64.const` | … | `0.1p+2`, `infinity`, `nan:0x789`, `-0x1p-2`
+| `f32.const` | … | `0.1p+2`, `infinity`, `nan:0x789`, `-0x1p-2`
 | `get_local` | *name* | `$x + 1`
 | `set_local` | *name* `=` *expr* | `$x = 1`
 | `select` | `select` *expr*, *expr*, *expr* | `select 1, 2, $x < $y`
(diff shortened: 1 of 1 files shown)


TextFormat.md
```diff
@@ -5,26 +5,21 @@ The purpose of this text format is to support:
   source can be viewed) in a natural way.
 * Presentation in browser development tools when source maps aren't present
   (which is necessarily the case with [the Minimum Viable Product (MVP)](MVP.md)).
-* Writing WebAssembly code directly for reasons including pedagogical,
-  experimental, debugging, optimization, and testing of the spec itself.
+* Working with WebAssembly code directly for reasons including pedagogical,
+  experimental, debugging, profiling, optimization, and testing of the spec
+  itself.
 
 The text format is equivalent and isomorphic to the [binary format](BinaryEncoding.md).
```
naturaltransformation: Is this statement still true given that function names may have to be converted from unrestricted byte sequences?

TextFormat.md
```diff
@@ -5,26 +5,21 @@ The purpose of this text format is to support:
...
-[Abstract Syntax Tree](AstSemantics.md), the syntax would contain nested
-statements and expressions (instead of the linear list of instructions most
-assembly languages have).
-
-There is no requirement to use JavaScript syntax; this format is not intended to
-be evaluated or translated directly into JavaScript. There may also be
+The text format is neither a subset or superset of JavaScript syntax, and it
+cannot be directly evaluated as JavaScript. Some parts of the syntax resemble
+JavaScript so that it's familiar and readable, however there are also
 substantive reasons to use notation that is different than JavaScript (for
```
naturaltransformation: Nit: different *from* JavaScript

TextFormat.md
```diff
@@ -43,43 +38,377 @@ support more human-readable representations, but never at the cost of accurate r
...
-  [LLVM backend]: https://github.com/llvm-mirror/llvm/tree/master/test/CodeGen/WebAssembly
-  [WAVM backend]: https://github.com/AndrewScheidecker/WAVM/tree/master/Test
-  [wassembler]: https://github.com/ncbray/wassembler/tree/master/demos
-  [V8 prototype]: https://github.com/WebAssembly/v8-native-prototype
-  [ilwasm]: https://github.com/WebAssembly/ilwasm
-  [sexpr-wasm prototype]: https://github.com/WebAssembly/sexpr-wasm-prototype
-  [binaryen]: https://github.com/WebAssembly/binaryen
+## Philosophy:
+
+ - Use JS-style sensibilities when there aren't reasons otherwise.
```
naturaltransformation: Nit: reasons to do otherwise

TextFormat.md
```diff
@@ -43,43 +38,377 @@ support more human-readable representations, but never at the cost of accurate r
...
+
+## High-level summary:
+
+ - Curly braces for function bodies, blocks, etc., `/* */`-style and `//`-style
+   comments, and whitespace is not significant.
+   Nested `/* */`-style comments follow JavaScript's rules.
+
+ - Semicolons separate expressions inside blocks, loops, and so on. The
+   last statement in the sequence may omit its trailing semicolon to indicate
+   that its result is being used as the result value of the parent construct.
```
naturaltransformation: parent construct meaning enclosing construct?

TextFormat.md
```diff
@@ -43,43 +38,377 @@ support more human-readable representations, but never at the cost of accurate r
...
+
+ - Put labels "where they go".
+
+
+## Examples:
+
+### Basics
+
+```
+  function $fac-opt ($a:i64) : (i64) {
```
naturaltransformation: Should there be a description about how names will be represented, possibly with a reference to the Function Names section in https://github.com/WebAssembly/design/blob/master/Web.md? This probably belongs with the description of $ prefix for names.

TextFormat.md
```diff
@@ -43,43 +38,377 @@ support more human-readable representations, but never at the cost of accurate r
...
+The function return type has parentheses for symmetry with the parameter types,
+anticipating adding multiple return values to wasm in the future.
+
+The curly braces around the function body are not a `block` node; they are part
+of the function syntax, reflecting how function bodies in wasm are block-like.
+
+The last expression of the function body here acts as its return value. This
+works in all block-like constructs (`block`, function body, `if`, etc.)
+
+`>s` means *signed* greater-than. explicit unsigned or signed operators will be
```
naturaltransformation: Nit: capitalization

TextFormat.md
```diff
@@ -43,43 +38,377 @@ support more human-readable representations, but never at the cost of accurate r
...
+suffixed with 'u' or 's', respectively.
+
+The `$` sigil on user names cleanly ensures that they never collide with wasm
```
naturaltransformation: Seems that $ isn't just used for explicit names (e.g., function names) but also implicit (generated) identifiers such as parameters, labels, and locals.

TextFormat.md
```diff
@@ -43,43 +38,377 @@ support more human-readable representations, but never at the cost of accurate r
...
+   )
+```
+
+And here's the proposed text syntax:
+
+```
+   function $Q_rsqrt ($0:f32) : (f32) {
+     var $1:f32;
+     $1 = f32.reinterpret/i32(1597463007 - ((i32.reinterpret/f32($0)) >> 1));
+     $1 = $1 * (0x1.8p0 - $1 * ($0 = $0 * 0x1p-1) * $1);
```
naturaltransformation: What will be required or admissible in terms of float representation?

TextFormat.md
```diff
@@ -43,43 +38,377 @@ support more human-readable representations, but never at the cost of accurate r
...
+```
+  function $loop3 () : (i32) {
+    var $i:i32;
+    $i = 0;
+    loop $cont {
+      $i = $i + 1;
+      if ($i == 5) {
+        br ($i) $exit;
+      }
+    $exit:
```
naturaltransformation: The notation may be construed as requiring function names, identifiers, and labels to share a namespace. Is this the case or are they disjoint? The `@` code point notation suggests disjoint.

TextFormat.md
```diff
@@ -43,43 +38,377 @@ support more human-readable representations, but never at the cost of accurate r
...
 therefore synthesize new names. However, as part of the [tooling](Tooling.md)
 story, a lightweight, optional "debug symbol" global section may be defined
 which associates names with each indexed entity and, when present, these names
 will be used in the text format projected from a binary WebAssembly module.
+
+## Code references
+
+In the binary format, references to locations in the code are done through
+byte offsets. In the text format, these references are represented symbolically
+with a special label syntax, prefixed by an `@` sign. These labels have no
```
naturaltransformation: Is there another term for this other than "label" which seems to overlap with "label" as in the label definitions and the earlier label section, unless that is intentional.

TextFormat.md
```diff
@@ -43,43 +38,377 @@ support more human-readable representations, but never at the cost of accurate r
...
+
+The curly braces around the function body are not a `block` node; they are part
+of the function syntax, reflecting how function bodies in wasm are block-like.
+
+The last expression of the function body here acts as its return value. This
+works in all block-like constructs (`block`, function body, `if`, etc.)
+
+`>s` means *signed* greater-than. explicit unsigned or signed operators will be
+suffixed with 'u' or 's', respectively.
+
```
JSStats: Could 'signed' be the default so that `>s` be just `>`? This would also fit the float comparisons which are signed.

TextFormat.md
```diff
@@ -43,43 +38,377 @@ support more human-readable representations, but never at the cost of accurate r
...
+   frequent opcodes, prefer just presenting operator names, so that they're easy
+   to identify.
+
+ - Typescript-style `name: type` declarations.
+
+ - Parentheses around call arguments, eg. `call $functionname(arg, arg, arg)`,
+   and `if` conditions, eg. `if ($condition) { call $then() } else { call $else() }`,
+   because they're familiar to many people and not too intrusive.
+
+ - Put labels "where they go".
```
JSStats: In the example below it looks like the block for a label is also implicit? This might be a problem as it would not encode where the block started.

TextFormat.md
```diff
@@ -43,43 +38,377 @@ support more human-readable representations, but never at the cost of accurate r
...
+            (get_local $1)))))
+   )
+```
+
+And here's the proposed text syntax:
+
+```
+   function $Q_rsqrt ($0:f32) : (f32) {
+     var $1:f32;
+     $1 = f32.reinterpret/i32(1597463007 - ((i32.reinterpret/f32($0)) >> 1));
```
JSStats: Could some of the cast operators also take advantage of the simple overloading, so this could be `$1 = f32(...)`

TextFormat.md
```diff
@@ -43,43 +38,377 @@ support more human-readable representations, but never at the cost of accurate r
...
+
+Note that the curly braces are part of the `if`, rather than introducing a
+block. This reflects how `if` essentially provides `block`-like capabilities
+in the wasm binary format.
+
+### Nested blocks
+
+Label definitions, like the `$exit:` above, introduce additional blocks nested
+within the nearest `{`, without requiring their own `{`. This allows the deep
+nesting of `br_table` to be printed in a relatively flat manner:
```
JSStats: Is there a loss of information here, a loss of the start of the block?

TextFormat.md
```diff
@@ -43,43 +38,377 @@ support more human-readable representations, but never at the cost of accurate r
...
+
+
+### Basic operators ([described here](https://github.com/WebAssembly/design/blob/master/AstSemantics.md#constants))
+
+| Name | Syntax | Example
+| ---- | ---- | ---- |
+| `i32.const` | … | `234`, `0xfff7`
+| `i64.const` | … | `234`, `0xfff7`
+| `f64.const` | … | `0.1p2`, `infinity`, `nan:0x789`
+| `f32.const` | … | `0.1p2`, `infinity`, `nan:0x789`
```
JSStats: Was there a resolution to the need to distinguish between i32 and i64, and between f64 and f32?

TextFormat.md
```diff
@@ -43,43 +38,377 @@ support more human-readable representations, but never at the cost of accurate r
...
+`+`, `-`, `*`, `/`, `%`, `&`, `|`, `^`, `<<`, and `>>`, respectively.
+
+`eq`, `ne`, `lt`, `le`, `gt`, `ge`, operators use
+`==`, `!=`, `<`, `<=`, `>`, `>=`, respectively.
+
+`eqz` and `neg` operators use prefix unary `!` and `-`, respectively.
+
+Signed and unsigned operators are suffixed with `s` and `u`, respectively,
+so `i32.div_s` looks like `/s`, and so on.
+
```
JSStats: Could the `s` be the default and dropped. It will make it just a little less verbose.

TextFormat.md
```diff
@@ -43,43 +38,377 @@ support more human-readable representations, but never at the cost of accurate r
...
+   last statement in the sequence may omit its trailing semicolon to indicate
+   that its result is being used as the result value of the parent construct.
+
+ - `get_local` looks like a simple reference; `set_local` looks like an
+   assignment. Constants use a simple literal syntax. This makes wasm's most
+   frequent opcodes very concise.
+
+ - Infix syntax for arithmetic, with simple overloading. Explicit grouping via
+   parentheses. Concise and familiar with JS and others.
+
```
JSStats: The 'simple overloading' is not well defined? For example is it only bottom-up, which I suggest?

TextFormat.md
sunfishcode: Yes. Arbitrary bytes are escaped so that they can be represented as valid unicode characters. The details are in the full grammar [here](https://github.com/mbebenita/was).

TextFormat.md
sunfishcode: Fixed.

TextFormat.md
sunfishcode: Fixed.

TextFormat.md
sunfishcode: Immediately-enclosing construct, yes. "Parent" seemed like a shorter way to say it that would still be clear.

TextFormat.md
sunfishcode: As mentioned above, the full details are in the grammar repo. The idea here was that TextFormat.md is just giving a basic overview of the design, and we'd have the actual spec elsewhere.

TextFormat.md
sunfishcode: Where?

TextFormat.md
sunfishcode: This is a detail that isn't yet resolved. One idea is to use `%` as a sigil for generated identifiers, to unambiguously distinguish it from `$` for explicit identifiers, though this isn't implemented yet.

TextFormat.md
sunfishcode: I expect C99-style hexadecimal will be required. The current grammar also includes decimal, though it's an open question whether values which require non-trivial rounding should be permitted.

TextFormat.md
sunfishcode: The basic proposal follows the same rules as the s-expression syntax currently.

TextFormat.md
sunfishcode: That's a good point. Some suggestions that have come up include "bookmark", "tag", or "debug label". What do you think?

TextFormat.md
sunfishcode: If this overall format is accepted, you're welcome to make a PR proposing this idea. My preference for now is to keep `s` and `u` which provide a nice parallel with the `_s` and `_u` in the opcode names in the spec.

TextFormat.md
sunfishcode: When multiple blocks are coalesced, they all start at the same place.

TextFormat.md
sunfishcode: You're welcome to propose this; for now I prefer to avoid ambiguity.

TextFormat.md
sunfishcode: The block starts at the same place as the block it's nested in.

TextFormat.md
sunfishcode: It's still an open issue. The current prototype implementation uses suffixes, however I'd ideally like to avoid them if feasible.

TextFormat.md
sunfishcode: As above, I prefer the symmetry between signed and unsigned, and the correspondence with the opcode names.

TextFormat.md
sunfishcode: There are multiple options here. I'm first looking to see if the overall approach will be accepted, before delving into the details of how overloading will work.

TextFormat.md
```diff
@@ -43,43 +38,377 @@ support more human-readable representations, but never at the cost of accurate r
...
+  function $fac-opt ($a:i64) : (i64) {
+    var $x:i64;
+    $x = 1;
+    br_if ($a <s 2) $end;
+    loop $loop {
+      $x = $x * $a;
+      $a = $a + -1;
+      br_if ($a >s 1) $loop;
+    }
+  $end:
```
JSStats: Just to make sure I understand, this `$end` label creates a block that starts just before the `$x = 1`?

If this block started after `$x = 1` then would it be presented with explicit curly-braces?

sunfishcode: This turned out to be a mistake. In the fac.wast file this file is translated from, the block starts after the set_local. I've now updated the example to accurately reflect where the block starts.

TextFormat.md
naturaltransformation: explicit unsigned... -> Explicit unsigned...

sunfishcode: Fixed; thanks!

TextFormat.md
```diff
@@ -43,43 +38,379 @@ support more human-readable representations, but never at the cost of accurate r
...
+The default syntax for operators is their opcode name followed by a
+comma-separated list of operands. However, several operators have
+special syntax.
+
+
+### Control flow operators ([described here](https://github.com/WebAssembly/design/blob/master/AstSemantics.md))
+
+| Name | Syntax | Examples
+| ---- | ---- | ---- |
+| `block` | `{` … *label*: `}` | `{ br $a; a: }`
```
binji: Having the label at the end is more sane, I agree. Any concern that it is differing from the equivalent JavaScript?
```
a: {
  break a;
}
```

TextFormat.md
```diff
@@ -43,43 +38,379 @@ support more human-readable representations, but never at the cost of accurate r
...
+| `loop` | `loop` *label* `{` … `}` | `loop $a { br $a; }`
+| `if` | `if` `(` *condition* `)` `{` … `}` | `if (0) { 1 }`
+| `if` | `if` `(` *condition* `) `{` … `} else `{` … `}` | `if (0) { 1 } else { 2 }`
+| `br` | `br` *label* | `br $where`
+| `br` | `br` `(` *expr* `)` *label* | `br ($v) $where`
```
binji: Seems weird that the value in parens is the br value here, even though it is typically the condition.

TextFormat.md
```diff
@@ -43,43 +38,379 @@ support more human-readable representations, but never at the cost of accurate r
...
+
+## Module-level syntax
+
+Type declarations:
+
+`type $ternary of function (i32,i32,i32) : (i32)`
+
+Imports:
+
+`import "print" as $print_int from "global.PrintingStuff" typeof function (f64) : ()`
```
binji: Not a big fan of this syntax, but I suppose there's plenty of time to bikeshed over the specifics. :)

TextFormat.md
sunfishcode: Yes, there is a concern. Accompanying it is the concern that large and deeply nested block structures, with extensive use of labels, are much more common in wasm than they are in typical JS. Under such conditions, having the label "where the branch goes" improves readability because one doesn't have to jump far away to find the top of the loop and then jump back down to find the bottom.

TextFormat.md
sunfishcode: I guess you have to remember that `br` doesn't have a condition, while `br_if` does. I grant that this may not be obvious, but `br` having a value operand doesn't have any kind of analog in JS, so there's no obvious precedent to follow here.

TextFormat.md
binji: Yeah, I think you're right, it's better to have labels where they go.

Though I just thought of another concern; shadowing a label name becomes a bit harder to detect because you have to look for the closest nested in either direction, e.g.
```
{
  loop $a {
    ...
    br $a;
  }
$a:
  ...
}
  
```

But one could argue that this is like going to the doctor and saying "it hurts when I do this"... :)

TextFormat.md
sunfishcode: No; this import syntax is something we added relatively late as we were going through making sure we had everything covered. Suggestions welcome :-).

TextFormat.md
```diff
@@ -43,43 +38,379 @@ support more human-readable representations, but never at the cost of accurate r
...
+| `return` | `return` | `return`
+| `return` | `return` *expr* | `return $x`
+| `unreachable` | `unreachable` | `unreachable`
+
+
+### Basic operators ([described here](https://github.com/WebAssembly/design/blob/master/AstSemantics.md#constants))
+
+| Name | Syntax | Example
+| ---- | ---- | ---- |
+| `i32.const` | … | `234`, `0xfff7`
```
binji: probably worth having an example of a signed value here, to show that it's valid (I assume).

TextFormat.md
sunfishcode: Indeed; "don't do that" ;-). Alternatively, one might imagine prohibiting shadowing loop labels with block labels or vice versa. My guess is that we can address this at a higher level.

TextFormat.md
binji: Well, it's kind of like `return` or `yield`: they both "send" a value somewhere else. And it's also not like those, because the branch is implicit. But if we use those as a starting point, you get:
```
br $where;
br 5 $where;
br_if ($x < $y) $where;
br_if 5 ($ < $y) $where;
```
But I'm guessing this would make parsing non-LALR(1).

TextFormat.md
```diff
@@ -43,43 +38,379 @@ support more human-readable representations, but never at the cost of accurate r
...
+| `call` | `call` *name* `(`*expr* `,` … `)` | `call $min(0, 2)`
+| `call_import` | `call_import` *name* `(`*expr* `,` … `)` | `call_import $max(0, 2)`
+| `call_indirect` | `call_indirect` *signature-name* `[` *expr* `] (`*expr* `,` … `)` | `call_indirect $foo [1] (0, 2)`
+
+### Arithmetic operators ([described here](AstSemantics#32-bit-integer-operators))
+
+Arithmetic operators use C/JS-like infix and prefix notation.
+
+Add, sub, mul, div, rem, and, or, xor, shl, and shr operators use
+`+`, `-`, `*`, `/`, `%`, `&`, `|`, `^`, `<<`, and `>>`, respectively.
```
distransient: Could `^` as `xor` cause confusion?

TextFormat.md
yurydelendik: > this would make parsing non-LALR(1).

It might be confusing for a reader to see `br $var $lab;` too

TextFormat.md
yurydelendik: Is there a reason why it would cause? (Assuming that majority of people familiar with C or JavaScript)

There is always an options to leave additional alternative for infix operators in form of function call similar to e.g. `i32.min`, so xor operator will have `i32.xor` in addition to `^`.

TextFormat.md
yurydelendik: > Any other variants you explored before settling on this?

The only closest option was specifying parameters of import (or export, type, etc.) via comma is similar style of s-expr. Currently we stopped at JavaScript-like syntax (in exception of `typeof` and `of` keywords). The latter sounds like a more better alternative than "positional" parameters of s-expr.

TextFormat.md
yurydelendik: We can add some keyword when value has to be "send" e.g. `br as 5 $where` or even `br yield 5 $where`.

rossberg-chromium: Br's with argument are invoking a continuation, optionally passing arguments to it. They are very much like calls, and I would expect the syntax to reflect that analogy by having the arguments go after the label. Also, the proposed syntax for `br_if` will not mesh well with a potential multi-value extension.

How about the following:
```
br $l
br $l(args)
br $l if (cond)
br $l(args) if (cond)
```
This structure is somewhat analogous to C's do-while.

Br_table is a bit more tricky, but we can probably find something analogous as well.

TextFormat.md
rossberg-chromium: I think introducing overloading here is a serious problem. First of all, it violates the stated design goal that the text format is a concrete syntax for the Wasm AST, and that it is isomorphic to the binary format:

- Parsing it into the AST would require full type checking and then a transformation, which obviously means that the result is not a direct AST anymore.
- Conversely, "isomorphic" implies that there is a bijection, but rendering the AST in text actually loses information, so is not injective.

Moreover, the overloading isn't even unambiguous. For example, consider the block
```
{ unreachable + unreachable; 6; }
```
What type does the addition resolve to? Sure this is useless code, but that's besides the point. It still needs to unambiguously map to a Wasm program.

I think we should avoid overloading in the textual format. It doesn't reflect Wasm adequately.

That doesn't mean that there can't be syntactic convenience, but it needs to be syntax-driven, not type-driven. To throw out just one idea, we could have something similar to OCaml's parenthesised-open expression:
```
i64.(4 + 6 * $x)
```
Any operator inside `i64.( expr )` is for i64 (unless nested in another such construct).

kripken: I think these are good points, however, if the code is isomorphic *except* for patterns like `unreachable + unreachable`, it might be better to lose information there, if that helps readability. I'm not saying that would be a pretty solution (e.g., "i32 if ambiguous"), but if it allows an overall better syntax, it might be worth it.

But on the other hand, I think `i32.(unreachable + unreachable)` is worth considering. Perhaps we could allow `i32.(unreachable + unreachable)` etc., and it would be necessary for ambiguous cases like this, but optional otherwise, so we can still write `$x + $y`. I could see the counterargument that the complexity is annoying though.

Side note, `i32.(operations)` might also be used to denote signed/unsigned, perhaps, `i32.(x / y)` vs `u32.(x / y)`?

TextFormat.md
distransient: Js users will be familiar with it as the exponentiation operator

yurydelendik: > Js users will be familiar with it as the exponentiation operator

'^' defined in JS as bitwise-xor (see http://www.ecma-international.org/ecma-262/6.0/#sec-binary-bitwise-operators)

distransient: my bad

Sorry, my brain's in the wrong gear

TextFormat.md
JSStats: @rossberg-chromium The binary encoding could be changed to better support the text format and address these issues. Some simple bottom-up specialization would also be expected to improve compression as it eliminates redundant information and would work with a single pass (no-look-ahead) decoder.

TextFormat.md
sunfishcode: Examples added. I also added a paragraph explaining a subtlety with respect to ambiguity between `(f32.const -1.0)` and (f32.neg (f32.const 1.0)).

TextFormat.md
sunfishcode: @rossberg-chromium It's true that the current proposal has a hole as you describe, however it is possible to characterize the example you give as a tricky corner case that we can address, rather than a fundamental problem with the approach.

I discussed it with @yurydelendik, and he observed that since we have a simple explicit-opcode-name syntax for all operators like `i32.reinterpret/f32`, one option would be to allow that syntax for all operators, even those operators that also have short-hand syntax like `+`. So we could write the example above as this:

`{ i32.add(unreachable, unreachable); 6; }`

Then we'd just say that `+` etc. can only be used when its type can be inferred (which, in practice, will be almost always), while the `i32.add` etc. syntax can always be used.

> Parsing it into the AST would require full type checking and then a transformation, which obviously means that the result is not a direct AST anymore.

It's not clear to me what this means. It is true that this proposal assigns JS-style readability a somewhat higher weight than parsing complexity, however, that's a reflection of the design principles.

TextFormat.md
sunfishcode: @rossberg-chromium If `br_if` is made to return its value, as several people in https://github.com/WebAssembly/design/pull/681 proposed, tightly associating the args with the label is less desirable.

TextFormat.md
titzer: @JSStats The constraints on the binary format come from performance and size considerations, and it would wrong to change it for textual considerations. We've discussed overloading before and the consensus has been against it for several reasons.

JSStats: @titzer I expect some specialization will reduce the file *size* as it removes redundant information. The performance burden would come from looking up a type specialized operator and this does not appear costly for a compiler. The bottom up types already need to be computed to validate the opcodes, so there is no added burden there. It might be a bigger performance burden for your interpreter if it needs to lookup the specialized operator for each instruction, but the operator table that has been discussed would also be a burden here, so it might be necessary for such an interpreter to do some rewriting while validating and in that case it would not appear a big burden. Lets get some data.

wanderer:
would basic macros be out of scope here?

JSStats:
@wanderer I think it has been conceded that the official wasm binary format will not encode macro meta information so they could not be presented. My understanding is that this text format is not a macro-assembler, rather the presentation text format. People could extend it to be a macro-assembler but that would not constrain the binary format or runtimes so it is not necessary to consider here.

qwertie:
I would just like to plug #697 because I haven't received many comments there. Also, I recently [reviewed LES to consider changes](http://loyc.net/2016/potential-new-les.html) that might make it more attractive as a WebAssembly text format.  I'm just starting to work on a parser for LES in JS and might later do one for C++ (for binaryen).

wanderer:
Will this all so display [filters](https://github.com/WebAssembly/decompressor-prototype/blob/master/CompressionLayer1.md) for layer 1 compression?

yurydelendik:
> Will this text format allso display filters for layer 1 compression?

Looks like filter language has its own syntax -- there is no common constructs between WASM code and filters code.

yurydelendik:
[Experimental parser](https://github.com/mbebenita/was) and [syntax highlight](https://bugzilla.mozilla.org/show_bug.cgi?id=1280471) for the new text format at http://people.mozilla.org/~ydelendik/tmp/waseditor/

TextFormat.md
drom: I think JS (C++14) style binary literals"  `0b010101` would be nice too.

flagxor:
After some communication between implementers, we've decided to focus on having all browsers be able to display linear opcodes for the MVP timeframe. Moving to Discussion.

drom:
@flagxor does "linear opcodes" means actual numbers, not S-expression? not JS-like language?

JSStats:
@drom I guess it means a display like an assembler code, one operator per line, etc.

I strongly expect that this community group expects an expression based language, that they might accept display as a linear list of opcodes for the MVP but only so long as there is a clear plan and a well articulated expression base text format and a proof-of-concept implementation would make this clear. People might be prepared to defer largely bike-shedding text format issues until after the MVP, but they know the difference between a linear stack machine code and an expression base language.

I don't think the core issues here can be deferred until after the MVP rather there a many issues that constrain the language design and are core issues.

JSStats:
If there are pop's in the language then it is not expression based. Typical expression blocks discard excess values of statements, so a special operator is needed to build values.

```
block
i32.const 4
f64.const 5.5
call $foo
i64.add
end

{
  const (i64 $c0, i64 $c1) = $foo(4, 5.5d0);
  $c0 + $c1; // but does not consume the stack values, might be used again
}

// So perhaps this:
{
  // Builds arguments from all the values of argument operators.
  // But doesn't the above read much better?
  call_with_values(i64.add, $foo(4, 5.5d0)).
}
```

```
block
i32.const 4
f64.const 5.5
get_local 0
br_if[arity=2] 0
i32.const 6
f64.const 7.5
end

{
  br_if $b0, values(4, 5.5d0), $l0
  values(6, 7.5)
} $b0
```

I don't think these examples demonstrate the real problems, but based on these I assume that this stack code allows operators that return no values to appear anywhere, so it removes that constraint. This would at least allow values to be stored in local variables if they need to be consumed in a pattern that does not fit the stack order.

I think `drop` is ok in an expression language. Assuming presentation blocks are inserted before void operators not at the head of a block:

```
i32.const 1
i32.const 2
drop
i32.const 3
i32.add

{
  1 + {drop(2); 3};
}

i32.const 1
i32.const 2
i32.const 3
drop
drop
i32.const 4
i32.add

{
  1 + {drop(2, 3); 4};
}
```

The problem occurs when values are consumed out of stack order:
```
{
  const (i32 $c0, i32 c1) = call ...;
  $c1 / $c0;
}

or

{
  const (i32 $c0, i32 c1, i32 c2) = call ...;
  $c2 / $c0;
}
```

For this out-of-order consumption we need `get_value`, and then need an efficient way to dispose of the values up the stack, need a way to inform the decoder on their last use `get_and_zero_value`, so are back to the old block semantics.

Can people think of any other good examples that demonstrate stack machine code that will not present well in an expression based language?

JSStats:
@drom Thank you, there are some ideas from stack machines such a Fourth that I have pondered. For example the Fourth `pick` operation I have been trying to promote as `get_value`.

There might be a number of ways to shuffle stack elements around to consume them in the order required with Fourth style operators, compared to simply being able to encode access to the desired element.

Being able to `swap` an element up the stack to the TOS and then `drop` it would provide a method to discard values up the stack. I have been promoting `get_and_zero_local` to gain a similar ability to inform the compiler that a value is discarded in an expressionless encoding.

I think there was a point in wasm development, with the AST, in which it was not necessary to compare the stack elements at control flow merge points, rather just the local variables. If we allow `swap` and other changes to the stack then this might increase the decoding burden.

We seem to have a good alternative in the prior approach, of just building up values within blocks on the values stack, and not allowing them to be changed. They could be referenced by index, rather than shuffling them around. Downside is a deeper decode-time type stack, but perhaps it would be worth it.

In any case we don't yet have `swap` and `pick` etc, so the current proposal just does not look tenable it just does not allow access to the values on the stack, rather seems to force definitions into local variables. If wasm is going to be so dependent on local variables then why not just simplify the decoder a lot and remove the stack, make it a register base code.

qwertie:
@JSStats I'm surprised you moved your comments here, as it was my impression that you were not talking mainly about the text format, but about the more fundamental idea (applicable to the binary format) of whether Wasm should stick with a more traditional AST. It certainly affects the text format, of course, but as I was saying in #697, it doesn't seem to require fundamental changes in the text syntax.

> the current proposal just does not look tenable it just does not allow access to the values on the stack, rather seems to force definitions into local variables

I'm sure the backers of the new stack machine (whoever they are) think it's "tenable", and so do I. I imagine they will add a new operator or two, if data shows that such operators would provide a large benefit. But I'll bet avoiding locals won't be found important enough for MVP, even if multi-value support were added, which itself seems unlikely. That said, I personally prefer an expression-based language and, if I had the opportunity, would have loved to look for alternate ways of achieving similar efficiency gains in the traditional AST regime.

drom:
@JSStats [Forth](https://en.wikipedia.org/wiki/Forth_(programming_language)) (not Fourth) language. Has a lot to offer to stack machine designers. In fact it influenced most of Virtual stack machines like Java Bytecode and .NET CIL. The syntax is a bit unusual, but the concept of stack manipulation words: `drop`, `dup`, `swap`, `over` ... is shown it's efficiency over the years.

drom:
Here is basic comparison table https://github.com/drom/wast-forth/blob/master/stack-manipulation.csv
of "Stack Manipulation" operations in [FCode](https://en.wikipedia.org/wiki/Open_Firmware), [Java Bytecode](https://en.wikipedia.org/wiki/Java_bytecode), and [CIL](https://en.wikipedia.org/wiki/Common_Intermediate_Language)

JSStats:
@drom I think wasm is still ok wrt 'structural control flow operations'. Perhaps it could do more to support loop analysis, to make the stride explicit etc. But what about structured local variables, these are important in languages too?

Your list ignores the `pick` operation, to be able to pick a value up the stack. The drop operator also frustrates structured local variables as it ends the scope of a variable at a point that is not structured in a block, and combined with `swap` completely breaks block scoping of variables. I just don't buy the claimed benefits of the stack-machine, they are not necessary for multiple value support, and being able to present structured code was a worthy goal for wasm in my opinion.

It's interesting that CIL is restricted to `pop` and `dup` and I wonder if these are for performance reasons, to make validation efficient. If so then perhaps wasm would hit the same barriers and not be able to introduce pick/swap in further, and if these are essential for good multiple value support in this design then the change in direction to the stack machine was never justified!

Can someone answer if wasm decoders will be required to record and compare the values stack for changes at every control flow path merge point, and what is the cost? In the expressionless encoding experiment this was a show stopper.

There will still remain the possibility of canonicalizing on a different code, to consider the stack-machine as just a compressed encoding of the canonical code, and to present the canonical code in a structured manner. But then it is not possible in general to specify the encoding in the canonical code, just as it is not expected that the choices of encoding that a compressor uses are not expressed in the source input to the compressor. e.g. use a functional form of SSA as the wasm language that developers actually work with.

drom:
@JSStats I have added `pick` operation to the table. In fact it is my favorite "stack manipulation" pokemon. I have even done RTL for the "PICKy: stack machine" with `pick` support: https://docs.google.com/document/d/1NeRwIxzSPB8-m9597fydIN5dD1CBYSjBZz8WXZJ82jU/edit?usp=sharing . You can do so many interesting tricks with `pick`, you can significantly reduce number of local variables, and number of `get_local` / `set_local` operations. Also, defectively `dup = 0 pick` and `over = 1 pick`. It gives you sort of "immutable" access deep into the value stack.

But: you have to limit `pick` power. Giving it only "constant-depth" access. Variable depth `pick` is call for disaster and can't be validated statically. With all that power only Forth/FCode make use of it. Both JVM and CIL not using it's power.

JSStats:
@drom Thank you, and good to hear. Here's an example that I am not sure how to handle?

```
block
i32.const 1  // let this be stack slot 0
i32.const 2  // let this be stack slot 1
pick 0
pick 1
i32.add
pick 0
pick 1
i32.mul
i32.add
// Gets messy here, need to keep the result but drop slots 0 and 1??
????
end
```

In structured code, with the prior block semantics, the unused definitions are just discarded, so this could have been presented in a structured format e.g.
```
{
 const $c0 = 1;
 const $c1 = 2;
 $c0 + $c1 + $c0 * $c1
}
```

Adding `pick` aka `get_value` alone does not appear to demand checking the stack at each control flow merge point, as it does not mutate the stack. It would have allow better multiple-value support as the multiple values could just be pushed on the stack and still accessed. It would allow the function arguments to be on the stack for the callee rather than in local variables.

drom:
@JSStats let me annotate your first example:

```js
                        [ stack notation ]
i32.const 1    [ c1 ]
i32.const 2    [ c1 c2 ]
pick 0             [ c1 c2 c2 ]
pick 1             [ c1 c2 c2 c2 ]
i32.add          [ c1 c2 (c2+c2) ]
pick 0            [ c1 c2 (c2+c2) (c2+c2) ]
pick 1            [ c1 c2 (c2+c2) (c2+c2) (c2+c2) ]
i32.mul         [ c1 c2 (c2+c2) ((c2+c2)*(c2+c2)) ]
i32.add         [ c1 c2 ((c2+c2)+((c2+c2)*(c2+c2))) ]
// Gets messy here, need to keep the result but drop slots 0 and 1??
????
```
why you were caring `c1`? and unused copy of `c2`?

drom:
@JSStats your second example:
```c
{
 const $c0 = 1;
 const $c1 = 2;
 $c0 + $c1 + $c0 * $c1
}
```
assuming correct operation precedence:

in stack machine:
```c
               [ stack notation ]
i32.const 1    [ c0 ]
i32.const 2    [ c0 c1 ]
pick 1         [ c0 c1 c0 ]
pick 1         [ c0 c1 c0 c1 ]
i32.mul        [ c0 c1 (c0*c1) ]
i32.add        [ c0 (c1+(c0*c1)) ]
i32.add        [ (c0+(c1+(c0*c1))) ]
```
You don't really need to drop data, just use it.

JSStats:
Tried to give $c0 and $c1 absolute indexes for the example. With relative offsets:

```
block
i32.const 1  [ c1 ]
i32.const 2  [ c1 c2 ]
pick 1       [ c1 c2 c1 ]
pick 1       [ c1 c2 c1 c2 ]
i32.add      [ c1 c2 (c1+c2) ]
pick 2       [ c1 c2 (c1+c2) c1 ]
pick 2       [ c1 c2 (c1+c2) c1 c2 ]
i32.mul      [ c1 c2 (c1+c2) (c1 * c2) ]
i32.add      [ c1 c2 ((c1+c2) + (c1 * c2)) ]
// Gets messy here, need to keep the result but drop slots 0 and 1??
????
end
```

So the challenge is how to keep it structured, while informing the decoder that $c0 and $c1 are no longer used (or even better at their last use), and can both of these constrains be met?? I guess `swap` could be used to consume $c0 and $c1 at the i32.mul rather than duplicating them, but where would that leave the code structure? Can we have our cake and eat it too?

JSStats:
@drom Interesting example, but it did require re-ordering the operations which broken the intent of the example. Lets assume that the i32.add and i32.mul are actually function calls and that they need to be called in the order given in the example.

drom:
@JSStats I understand the challenge. Here are 3 popular Forth solutions:

  1. Good stack scheduling / ordering algorithm to relax for the need to go deep into stack. Like my answer to your second example performs the same computations just in different order.
  2. the chess match above ends with: `swap drop swap drop` --> checkmate#. It is often works when function returns only one result. For this specific case Forth use `nip` operation. `: nip   swap drop ;`
  3. If you plan storing this result in the local variable here is solution: `(set_local $foo) drop drop`.

drom:
@JSStats if I compile following ASM.JS :+1: 
```js
function test2 (c1, c2) {
  c1 = c1 |0;
  c2 = c2 |0;
  return (c1 + (c2 + (c1 * c2)));
}
```
with binaryen today I am getting the following code (11 byte):
```lisp
(func $test2 (param $0 i32) (param $1 i32) (result i32)
    (i32.add
      (get_local $0)
      (i32.add
        (get_local $1)
        (i32.mul
          (get_local $0)
          (get_local $1)
        )
      )
    )
  )
```
good stack machine code would be (5 byte):
```c
(func $test
pick 1         [ c0 c1 c0 ]
pick 1         [ c0 c1 c0 c1 ]
i32.mul        [ c0 c1 (c0*c1) ]
i32.add        [ c0 (c1+(c0*c1)) ]
i32.add        [ (c0+(c1+(c0*c1))) ]
)
```

JSStats:
@drom Ok, thank you for the solutions. I also see potential changing the arguments to be on the stack and access by `pick`.

But it's not clear how you 'end game' could be represented in an expression. Also I think the following would inform the decoder earlier that c1 and c2 are consumed, which might make a difference to register pressure in a baseline compiler:

```
block
i32.const 1  [ c1 ]
i32.const 2  [ c1 c2 ]
pick 1       [ c1 c2 c1 ]
pick 1       [ c1 c2 c1 c2 ]
i32.add      [ c1 c2 (c1+c2) ]
swap 2       [ (c1+c2) c2 c1 ]  // But this messes up the structure.
swap 1       [ (c1+c2) c1 c2 ]
i32.mul      [ (c1+c2) (c1*c2) ]  // Lower register pressure here now.
i32.add      [ ((c1+c2) + (c1 * c2)) ]
end
```

Another 'move' in this 'chess game' is that we can invalidate a stack slot and define it to be a validation error to consume it. Let `pick_and_void` duplicate a stack element to the TOS and fill the source stack element with this `void` value. Now we can keep the structure (no stack shuffling) and inform the compiler at the last use:

```
block <arity=1>
i32.const 1      [ c1 ]
i32.const 2      [ c1 c2 ]
pick 1           [ c1 c2 c1 ]
pick 1           [ c1 c2 c1 c2 ]
i32.add          [ c1 c2 (c1+c2) ]
pick_and_void 2  [ void c2 (c1+c2) c1 ]
pick_and_void 2  [ void void (c1+c2) c1 c2 ]
i32.mul          [ void void (c1+c2) (c1 * c2) ]  // lower register pressure now
i32.add          [ void void ((c1+c2) + (c1 * c2)) ]
// Assume the block discards unused values.
end
```

drom:
@JSStats yes, I like the `pick_and_void` move! :+1:

JSStats:
Attempting to sketch a formatting rule with `pick_and_void` or `pick_and_zero` suggests that it might be necessary to made it a validation error to use `pick` when `pick_and_void` could have been used. This might be sufficient to uniquely define the code corresponding to structured uses of lexical constants which would be implemented via the pick operators.

drom:
@JSStats is the "sketch" you are doing, for validation phase or stack scheduling?

drom:
@JSStats in Forth `pick` operation has an evil twin brother -- `roll` that removes the element deep from the stack. [ANS94](http://www.forth.org/svfig/Win32Forth/DPANS94.txt) 6.2.2150 : `ROLL ( xu xu-1 ... x0 u -- xu-1 ... x0 xu )` but this behavior not considered to be nice.

JSStats:
@drom Just trying to understand if the structure can be preserved to keep a 'familiar' text format. If both `pick` and `pick_and_void` are valid where  `pick_and_void` could be used then this extra information would need to be explicit in the text format and that would not be a 'familiar' named constant reference in the text format. A validation rule might address this.

Problem with `roll` is that is messes up the stack, and can that fit in with a structure text format, I don't know?

Here are some more examples that illustrate some more challenging corner cases for the text formatter. I have not found any show stoppers yet, but some to consider would be welcomed.

```
block <arity=1>
call $fn_returning_i32_i32_i32
i32.add
i32.add
end

{
  const (i32 $c0, i32 $c1, i32 $c2) = $fn_returning_i32_i32_i32();
  const i32 $c3 = $c1 + $c2;
  $c0 + $c3;
}
```

```
block <arity=1>
call $fn_returning_i32_i32_i32    [ 0, 1, 2]
pick_and_void 2     [ void, 1, 2, 0]
pick_and_void 2     [ void, void, 2, 0, 1]
pick_and_void 2     [ void, void, void, 0, 1, 2]
i32.add             [ void, void, void, 0, 1 + 2]
i32.add             [ void, void, void, 0 + (1 + 2)]
end                 [ 0 + (1 + 2)]

{
  const (i32 $c0, i32 $c1, i32 $c2) = $fn_returning_i32_i32_i32();
  const i32 $c3 = $c0;
  const i32 $c4 = $c1;
  const i32 $c5 = $c2;
  const i32 $c6 = $c4 + $c5;
  $c3 + $c6;
}
```

drom:
@JSStats I was thinking about the name for the `pick_and_void` command, and came out with two candidates `take` and `pull` :smile:

drom:
@JSStats All this `pick / pick_and void` discussion bigger then the scope of this issue. We should probably move it to the different thread.

JSStats:
@drom Could move discussion specific to the `pick` proposal here https://github.com/WebAssembly/design/issues/685 also see https://github.com/WebAssembly/design/issues/698 although that was directed at the expressionless encoding proposal but same motivation. Would not worry about names, if you can find show-stoppers or inefficiencies etc that is what is needed at this point, we need to be able to put forward a counter proposal that plausibly retains the text format structure and addresses the efficiency use cases as best it can.

@flagxor did open this general issue for discussion here, and the use case being discussed is retaining the structure of the text format (a JS-style text format) rather than giving up as some implementers have proposed. No one has objected yet!

drom:
@sunfishcode any idea how your text format proposal would be affected by new "stack machine" trend?

sunfishcode:
@drom With the current stack-machine changes, our text format will require only a few additional features; possibly the addition of syntax for a "first" expression (similar to block, but returns the first value rather than the last) and possibly also syntax for a scoped or restricted "let" expression. I don't know what the actual syntax for these will be yet, but it's mostly just a matter of aesthetics :-).

As @flagxor mentioned above, this proposal has been removed from consideration for standardization in wasm, at least for now. The text format proposed here will continue to be developed at https://github.com/mbebenita/was and in Firefox.

PR #708, opened 2016-06-15
First version of DebuggingFramework.md.

dekimir:
@lukewagner @sunfishcode @yurydelendik @titzer @jfbastien PTAL


lukewagner:
Thanks for writing this up!  At a high-level, this matches my basic understanding of the idea.

DebuggingFramework.md
```diff
@@ -0,0 +1,139 @@
...
+What the design _does_ specify is two interaction protocols:
+
+1. between the devtools and the debugger, and
+2. between the debugger and the wasm execution engine
+
+## Action Flow
+
+For reference, the entities are charted in the figure below:
+
+![Entities making up the debug framework](wasm-debug-chart.png)
```
lukewagner: Nice chart!  I'd put the "Debugger" box inside the browser box since it's executing JS or wasm code.

DebuggingFramework.md
```diff
@@ -0,0 +1,139 @@
...
+Note that 3 and 5 are bidirectional arrows: sometimes the flow of action goes
+from the browser through the debugger to devtools.  For instance, this will
+happen when a breakpoint is hit and the wasm execution is paused.  The browser
+will then notify the debugger, who will in turn notify devtools after
+translating the wasm program counter to the corresponding source line.
+
+## About the Debugger
+
+We propose that the debugger be a JavaScript library providing the interface
+described in the next section.  This allows for easy access to the browser, as
```
lukewagner: I'd s/JavaScript library/JavaScript API/ and then note that in the future (perhaps linking to GC.md), wasm would be able to access the API directly so that the debugger could be compiled wasm.

lukewagner: It'd also be good to point out the option of having this same API debug JS (maybe not in the initial version of the API, but after some iteration).

DebuggingFramework.md
```diff
@@ -0,0 +1,139 @@
...
+
+This also makes it possible for the debuggee to contain a URL for the debugger
+to be used on it.  This URL can point to any server, allowing debugger reuse
+among many debuggees.
+
+## UI -> Debugger Protocol Operations
+
+First, we list the basic debugger operations that the UI can invoke:
+
+- get a source location given a mangled function name
```
lukewagner: To understand what are the right set of primitives/operations here, it'd be nice to outline some end-to-end flows from user action to UI response to the action.

DebuggingFramework.md
```diff
@@ -0,0 +1,139 @@
...
+
+1. When the page is first rendered, the debuggee is loaded into the browser and
+   sent to the wasm execution engine.
+2. The user interacts with the browser's devtools UI (eg, requesting a source
+   listing for a function to set a breakpoint).
+3. Devtools call on the debugger to execute operation(s) necessary to fulfilling
+   the user's request.  After going through steps 4 and/or 5, the debugger
+   returns a result for the devtools to display.
+4. The debugger consults the debug info and/or source files to obtain the proper
+   mapping between the source code and the wasm.
```
sunfishcode: Nit: the chart lacks an arrow from the source files to the debugger.

sunfishcode:
Overall this looks like a good start.

This specifies the browser devtools as the UI that the user will be interacting with. Another idea which has been discussed is that debuggers could provide their own UI, perhaps in a frame. The proposal here has the advantage that it should be easier to develop simple debuggers, because debugger writers need not implement their own UI, as the browser devools UI will often be sufficient. However, some debuggers may still want their own UI, for example to provide greater source-language-specific customization. Would it be difficult to accommodate this option in this proposal?

DebuggingFramework.md
dekimir: I felt that the wasm module and the debugger should be shown outside, because they don't come with the browser but are downloaded.  Then I try to explain in the text how they are used by the browser when obtained.

DebuggingFramework.md
dekimir: Yeah, I felt the extra arrow would clutter the picture without adding much information value.  I glommed the debug info and source files rectangles together to suggest that they act in unison.

DebuggingFramework.md
dekimir: I'm working on a Chrome prototype that does just that.  Armed with that experience, I'll summarize here some of the end-to-end flows.

dekimir:
@sunfishcode: agree that some debuggers will want to provide their own UI (eg, Brackets and IntelliJ, which already do so for JS).  That requires exposing the debugger<>debuggee interface as a remote-client protocol.  We'd probably want to standardize the protocol so the external debugger only need implement it once (and not once per supported browser).  I guess the question is how hard is it to standardize the remote-debugging protocols of multiple browsers?

DebuggingFramework.md
dekimir: Done.

DebuggingFramework.md
```diff
@@ -0,0 +1,145 @@
+#WebAssembly Debugging
+
+This is a design proposal for enabling source-level debugging of WebAssembly
+modules.
```
jfbastien: That's a bit redundant, given that we're in the design repo ;-)

DebuggingFramework.md
jfbastien: IIUC 4. and 5. are what you're specifying here, right? It sounds like @lukewagner wants to drop 3. from the diagram, but you'd rather keep it because it can be separate (not being specified in this doc)?

dekimir: Actually, this aims to specify 3 and 5, while leaving 4 unspecified (other than saying how the debugger can find the debug info and source files).  The way the debugger interprets the debug info will be hidden from other actors in this picture, freeing the debugger implementers to choose whatever suits them.

DebuggingFramework.md
```diff
@@ -0,0 +1,145 @@
...
+happen when a breakpoint is hit and the wasm execution is paused.  The browser
+will then notify the debugger, who will in turn notify devtools after
+translating the wasm program counter to the corresponding source line.
+
+## About the Debugger
+
+We propose that the debugger implement a JavaScript API (by, for example, being
+a JavaScript library) that provides the actions described in the next section.
+This allows for easy access to the browser, as well as ease of implementation
+because JavaScript is widely known and supported.
```
jfbastien: I'm worried about 64-bit integer support, and eventually other types such as SIMD. Would we just split them up into 32-bit values? That's clunky but doesn't see too bad.

dekimir: Are you saying it's a problem that some wasm types don't map directly to JS types?  That sounds like a minor issue, easily worked around.  Or is there more that I'm missing?

jfbastien: Yup, that's exactly what I mean. Agreed it's not a big issue, just icky.

DebuggingFramework.md
lukewagner: I see your point and just noticed that "wasm module" is also, symmetrically, outside the box.

DebuggingFramework.md
dekimir: Done.

AndrewScheidecker:
I'd love for there to be a low-level debugging API that is implemented by the WASM VM:
* A query to enumerate threads
* A query to get the current call stack from a thread: an array of frames, each frame containing the function (maybe as a module instance pointer + function index), bytecode offset, local variable and operand stack values (if known)
* Queries to get the value of non-exported module instance definitions: memories, tables, global and in the future thread-local variables
* Commands to pause/resume execution of a thread
* Commands to add/remove breakpoints: when execution reaches some bytecode address in a function, suspend the thread and enqueue a message to the debugger
* A command to register a message queue to receive notifications of breakpoints, traps, etc

The goal would be to factor source-level debugging out of the WASM VM, to allow it to be implemented in unprivileged code, and to allow the source-level debugging implementations to be portable between different WASM VM.

PR #710, opened 2016-06-17
Simplify loop by removing second label

sunfishcode:
This removes loop's top label and if/else's end label, simplifying them, according to @rossberg-chromium's suggestion in https://github.com/WebAssembly/design/pull/652.

With this PR, the only constructs that introduce labels are `block` (label at the end), `loop` (label at the beginning), and the function body itself (label at the end) (since the current semantics are that the function body acts like an implicit block).

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -370,7 +370,7 @@ The `nop`, `br`, `br_if`, `br_table`, and `return` constructs do not yield value
 Other control constructs may yield values if their subexpressions yield values:
 
 * `block`: yields either the value of the last expression in the block or the result of an inner branch that targeted the label of the block
-* `loop`: yields either the value of the last expression in the loop
+* `loop`: yields the value of the last expression in the loop
 * `if`: yields either the value of the last *then* expression or the last *else* expression or the result of an inner branch that targeted the label of one of these.
 
 In all constructs containing block-like sequences of expressions, all expressions but the last must not yield a value.
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -345,9 +345,12 @@ a value and may appear as children of other expressions.
 ### Branches and nesting
 
 The `br` and `br_if` constructs express low-level branching.
-Branches may only reference labels defined by an outer *enclosing construct*.
-This means that, for example, references to a `block`'s label can only occur 
-within the `block`'s body.
+Branches may only reference labels defined by an outer *enclosing construct*,
+which can be a `block` (with a label at the `end`), `loop` (with a label at the
+beginning), `if` (with a label at the `end` or `else`), `else` (with a label at
+the `end), or the function body (with a label at the `end`). This means that,
+for example, references to a `block`'s label can only occur within the
+`block`'s body.
 
 In practice, outer `block`s can be used to place labels for any given branching
 pattern, except for one restriction: one can't branch into the middle of a loop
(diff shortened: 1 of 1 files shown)


titzer:
As a simplification, this PR makes sense. However, we've gone back and forth on the label issue multiple times due to trying to serve different goals: either code density, performance, or implementation simplicity. While I'd be happy to make this change for simplification, I am not sure we should do that just yet without some data on the other goals. Specifically, implementing implicit blocks for if was quite a pain to get right, but was justified at the time that it would make sense for code size. It'd be a shame to remove them now for simplicity and then add them back later when we again want to optimize for code size.

sunfishcode:
Changing if wasn't in my original proposal, but in his oposition to that proposal, @rossberg-chromium requested we change if too because he is ["confident that is a super-rare case"](https://github.com/WebAssembly/design/pull/652#issuecomment-220658895).

AstSemantics.md
```diff
@@ -266,9 +266,11 @@ a value and may appear as children of other expressions.
 ### Branches and nesting
 
 The `br` and `br_if` constructs express low-level branching.
-Branches may only reference labels defined by an outer *enclosing construct*.
-This means that, for example, references to a `block`'s label can only occur 
-within the `block`'s body.
+Branches may only reference labels defined by an outer *enclosing construct*,
+which can be a `block` (with a label at the end), `loop` (with a label at the
+beginning), or the function body (with a label at the end). This means that,
```
JSStats: 'or the function body (with a label at the end)' - this seems a significant new change too? Could this be deferred to discussion in a separate issue. One concern is that this might require an interpretation of the break arguments as the return values when the multi-value support has not been determined yet - this interpretation might be ok but there are different ways this could go.

AstSemantics.md
```diff
@@ -289,7 +291,7 @@ Other control constructs may yield values if their subexpressions yield values:
 
 * `block`: yields either the value of the last expression in the block or the result of an inner branch that targeted the label of the block
 * `loop`: yields either the value of the last expression in the loop or the result of an inner branch that targeted the end label of the loop
```
JSStats: Would the result of the loop also need to be changed above? There could no longer be 'an inner branch that targeted the end label of the loop'?

Might it be a complementary simplification to also have `loop` return the `void` type, so return no values? Would all the phi's be in local variables anyway, and perhaps loops are expected mostly end with `br_if.

lukewagner:
I'm in favor of this PR too.

rossberg-chromium:
I'm fine with this patch, too (although it's a bit sad to see the nice return=break unification go -- I'd be fine with leaving that alone, although my own consistency argument speaks against it :) ).

If we do not make this change and keep all implicit labels, then I would at least propose to switch the order of the two loop labels. Then you can explain the design consistently by saying that all block-like expression sequences have an implicit label directly enclosing them.

sunfishcode:
Updated this PR to just remove the bottom label of `loop`.

AstSemantics.md
```diff
@@ -367,7 +370,7 @@ The `nop`, `br`, `br_if`, `br_table`, and `return` constructs do not yield value
 Other control constructs may yield values if their subexpressions yield values:
 
 * `block`: yields either the value of the last expression in the block or the result of an inner branch that targeted the label of the block
-* `loop`: yields either the value of the last expression in the loop or the result of an inner branch that targeted the end label of the loop
+* `loop`: yields either the value of the last expression in the loop
```
lukewagner: nit: no "either"

sunfishcode:
Merged with lgtms above. This PR was against the master branch and github doesn't allow this to be changed, so I've now pushed the change directly to binary_0xc, and am closing this PR.

PR #716, opened 2016-06-30
Implement a simple NaN propagation scheme that still supports NaN boxing.

sunfishcode:
This is an alternative to https://github.com/WebAssembly/design/pull/713 which is simpler and should support the widest possible range of theoretical hardware while still following IEEE 754 rules and supporting the basic NaN boxing use case, at the cost of having more nondeterminism in the NaN bits.

At most one of these two PRs should be merged.


gahaas:
I like this PR better than #713.

flagxor:
looks good

PR #719, opened 2016-07-06
Function bodies deferred error reporting

Cellule:
## Proposal

This is a proposal to change the moment where we report an error during decoding of a function.
At this time, any malformed or not yet implemented feature in WebAssembly renders the module invalid.
This forces to parse and validate the full extent of the module.

This proposition would allow implementors to decide when is a good time to read the content of the function to generate code.
The implementor is still free to parse and validate the whole module ahead of time, but should only report any errors found when the function is called.
## Reasoning

First of all, at Microsoft, we believe that since most if not all WebAssembly module will be generated by tools, it should be fair to assume the module is well formed and valid until proved otherwise.
This means, we could ignore the functions that are not called during initialization of the module and save time and memory.
Thus, the main goal is to increase performance of WebAssembly and give flexibility on the implementation.

Using AngryBots as a reference, 
We measured that <25% of the functions in the module are called during the first minute of game play.
These functions represents ~26% of the size of all code in the module (3 169 049 / 12 164 730 bytes).
This means, about 75% of the functions in the module doesn't need to even be looked at for a long time which translate to precious cpu time saved and memory.

I implemented a prototype to parse and generate bytecode of the function on the first call.
With this, I went from ~500ms spent parsing and generating bytecode to ~150ms, about 70% cpu time gain.
I also looked at jit time and it goes from ~32sec to ~7sec to prejit the whole module and only the startup functions respectively. (This is with single thread jit).
The jit time is not as interesting to look at from our perspective, because we already delay the moment we jit the function with the current design.

However, the gain of ~350ms for parsing and bytecode is not negligible considering that AngryBots is still a fairly small game, this change could translate to wins in seconds of performance for us.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -61,6 +61,14 @@ Support for explicit tail calls is planned in
 which would add an explicit tail-call operator with well-defined effects
 on stack introspection.
 
+## Invalid AST Reporting
+
+An invalid AST can occur during compilation if there is an unknown or malformed
+construct in the source or if the implementor doesn't support a feature.
+
+This results in a [trap](#traps) that is reported once the function with the
+error is called.
+
 ## Types
 
 WebAssembly has the following *value types*:


titzer:
Would it be reasonable to implement a "verify only" pass? This is what we have done (mostly for testing) in V8. There is a simple check whether or not to generate any compiler IR when decoding. Our measured decode time is around 100mb/s, meaning verifying the entire module is about 100ms.

MikeHolman:
I prototyped that a while ago, but I think it is an insufficient mechanism. 75% of the verify only pass is purely unnecessary work (and I suspect that as applications get bigger, a smaller percent of code is used in startup). In AngryBots, it would be about 75ms wasted time.

But AngryBots is a tiny wasm module. I have seen asm.js apps with 15x more js than AngryBots (I'm looking at the game Prime World: Defenders right now as an example). 15x verify time puts that at 1500ms, 75% of which will be unnecessary. That is probably the upper bound on size for asm.js games, but I suspect with wasm we will be seeing even bigger apps than we ever saw with asm.js.

You can argue that this doesn't matter because with streaming it is IO bound, but I would rather be able to do something better with the CPU than validating unused functions, like running the optimizing JIT on startup functions. Also, streaming may not always be possible or desirable. In the past, we have discussed cases where people will be doing transformations/generating wasm on the client side. As an example, I could imagine it become common practice to do binary patching of modules to avoid requiring full download for updates. In these cases where you are not streaming, startup time will be bound by validation.

We have numbers that show 75% of the time spent validating is unnecessary. What is the point of requiring implementations to waste this time?

lukewagner:
I appreciate the argument of not wanting to waste work, but I'm not yet convinced that, after a few years of measurement on a variety of workloads/devices, all wasm engines won't be doing *some* tier of AOT compilation (which, in turns, requires validation) before execution begins.  If that becomes the norm, then lazy validation buys nothing and instead adds to the total complexity of the system (requiring special cases at various points in the pipeline and tooling to support degenerate throw-if-called functions) and then we'd regret requiring it.

The current asm.js workloads we're looking at are, by now, 3-5-year-old *mobile* workloads (which were certainly impressive from a JS pov, but won't be once people take wasm for granted) so the fact that they don't show significant stutter due to JIT compilation doesn't yet prove the point that some tier of AOT compilation isn't a prerequisite for reliable smooth startup performance.  And certainly *some* apps (esp. I can imagine for non-frame-based, e.g., productivity) will tolerate JIT compilation stutters quite well on any device.  But all it takes, I think, is to have a few important performance-sensitive apps that have that janky startup feel to motivate a general AOT tier, and I expect we'll see these pop up over the next few years.

MikeHolman:
Currently, we aren't seeing issues with stuttering on asm.js, and we aren't AOT JITing. But as you say, these workloads are not necessarily indicative of the wasm workloads to come.

And I agree with you that it is probably dependent on the scenario whether startup or stronger guarantee of smooth performance are more important.

One potential compromise is to put this decision into the hands of the developer. For example we can add a [form](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#type-entry) (say `0x41`) for functions with late error reporting. With such a mechanism, even the games that desire AOT performance characteristics can mark cold/dead functions as deferrable (e.g. via PGO).

Cellule:
Keep in mind that the late error reporting doesn't change if you do AOT or not and I don't believe it adds that much complexity since we already have mechanism in place to handle an invalid function, only the response changes.
If you believe AOT is the way to go, by all means go ahead.
We believe doing late error reporting could increase performance and give another option for feature detection/opt-in.
By doing so, it gives more flexibility to the implementors. You may not need that change, it doesn't mean it is not a good option.

lukewagner:
@MikeHolman With the finer level of granularity, and if engines reliably chose JIT/AOT based on lazy/eager, wasm modules annotated based on profiling could achieve the ideal AOT/JIT balance.  OTOH, without PGO, developers would likely go all-or-nothing specified by a single compiler flag so we probably couldn't in general expect the "lazy"/"eager" flag to be an accurate oracle for how the function was actually used.  FWIW, when we talked to gamedevs about the [asm.js hot/cold hints](https://discourse.wicg.io/t/allow-arbitrary-directives-in-asm-js-module-nested-function-directive-prologues/677) a while back, they weren't super excited since they didn't run PGO since it was such a pain.

@Cellule This flexibility to implementors comes at a cost to developers, though.  With AOT, the developer can treat compilation as just another load-time task that can be scheduled in parallel (via `WebAssembly.compile`) with other tasks like downloading initial assets and compiling WebGL shaders.  With JIT, the developer loses that ability.  (What you gain, of course, is not wasting cycles on cold/unused code, so I realize this is a tradeoff between two competing desirable qualities.)

This seems like the type of situation where developers would appreciate consistency between engines since it would give them a more predictable virtual machine and, if we added a lazy-validation flag, we need to get crisp on the implied-performance story or else it'll be a waste.  But it *also* feels too early to say anything definitive on load-time performance which makes me want to delay the addition until after the MVP.

LouisLaf:
The main request we have here is to keep the option open for delayed parsing.  Requiring ahead of time validation removes that possibility.  AOT and on demand JIT are on the opposite ends of aggressiveness for JIT'ing code.  There are many options in between to tune the aggressiveness to balance between responsiveness on one side, and startup perf, battery and memory on the other side.  @MikeHolman mentioned PGO as an option, but there are other approaches.  We could for example start in the interpreter and JIT on-demand in the background while JIT'ing of the whole module on other threads or using spare cycles.  We could also do static analysis of functions being JIT'ed to speculatively anticipate what to JIT next.

The modes could also be controlled by the user through API's requesting AOT or allowing lazy JIT'ing.  But it'd be nice to keep these modes semantically equivalent.  An API to force validation could also be nice to provide for the devs.

JSStats:
@cellule Would making any function body that failed to validate equivalent to `unreachable` address the issue too?

JSStats:
@MikeHolman Caching the validation result would avoid the need to re-validate so it might be a one-off cost and perhaps insignificant compared to the download time. How long does it take to download a 15x AngryBots app, 10+ minutes?

kripken:
Another option here might be to just allow flexible error reporting time. That is, an engine could either report an error immediately, or at the latest when the invalid function is actually called.

The downside is obviously compatibility risk, a program containing an invalid function that is never called might work in one browser but not in another. However, the risk might be worth it for the benefits. In particular, this would not require any extra work on engines (engines reporting errors during AOT can still do so, and other engines are free to report errors lazily). This would let engines explore this space while still being spec-compliant, and I think some of the options discussed in this thread (like initially interpreting) are very interesting.

To further reduce the risk, we could (1) make error reporting happen up front when devtools are open, so that no matter what browser a developer uses, in the devtools they see an error if there is one, and (2) we can tighten the spec later to one of the two options (report error immediately vs lazily) if things justify that.

rossberg-chromium:
Such a change may have lower value and higher cost than you make it sound.

As has been mentioned, the runtime cost of validation (as opposed to compilation) is pretty small and I/O bound. Once you implement streaming, the observable overhead should practically be zero.

On the other hand, this change effectively brings a form of dynamic typing to the language. You may not directly see it in the implementation, but conceptually it's right there, because a function will "raise a type error" only when it's actually called. That introduces a whole new dimension of behaviours that people will inevitably start relying on, besides adversely affecting user experience (and significantly complicating Wasm semantics on a more abstract level).

JS clearly shows that any abuse enabled by a language and existing implementations _will_ occur on the web -- locking a standard into supporting it forever, with often dire implications.

So from my POV, it would need strong, persistent evidence that validation is an unbearable cost even in production systems before we should consider such a change. Just avoiding some unnecessary work out of principle would be premature optimisation -- especially when it isn't actually all that unnecessary for the aforementioned reasons.

JSStats:
@rossberg-chromium Yes, so for example someone could embed arbitrary data in a function body so long as it is never called, e.g. 'fn1 body, fn1_meta invalid body with meta information, fn2 body, fn2_meta invalid body with meta information ...'  Could happen if it's allowed!

titzer:
Making errors lazy also makes feature testing by trying to load invalid modules more cumbersome, since the module must also attempt to run the invalid function(s). In the common (feature is supported) case, that means the engine does more work to compile the function, either JIT or AOT.

MikeHolman:
@JSStats You're right caching the validation result is something that should be done. Download time for a 15x AngryBots obviously depends on your internet connection (which seem to be increasing faster than CPU speeds FWIW). Personally, 150MB download would take about 12 seconds. IMO adding 1.5 seconds overhead to that is significant.

@rossberg-chromium How do you see this adversely affecting user experience? Malformed modules are unlikely to exist in production, anymore than malformed dlls in production applications. Maybe this complicates semantics at an abstract level, but I don't see this causing any practical complications in the present or future. One of our stated core tenets is to design for fast load time. Having a fully static type system is not on that list. We should not require an egregious problem to further one of our core tenets.

@titzer It's true that the proposal would alter how you do feature detection. But it gives way greater flexibility in how you do it by allowing for runtime feature detection. For example, you could do something like the following:
`
var m = Wasm.instantiate(blah);
try{ m.testSIMD(); } catch(){ m.disableSIMD();}
`
And in your module, you have code like:
`
if(supportsSIMD) { vectorizedAlgorithm() } else { scalarAlgorithm() }
`

I believe these runtime checks are probably lower cost than the current method (which I guess is test by loading a module that uses SIMD, and then choose to download the version compiled with/without it?) It's also not ideal if there is more than 1 feature you want to test for. However, my suggestion of having different function forms for eager/lazy error reporting would allow you to retain the current method of feature detection.

Cellule:
As I was going through my old issues, I found this one.
I thought in previous discussions we had settled on reporting compilation error of function bodies at call time except when using `WebAssembly.validate`.
However, as I went through the doc, I didn't find any mention to that effect.
Can anyone confirm the status of this please ?

MikeHolman:
>I thought in previous discussions we had settled on reporting compilation error of function bodies at call time except when using WebAssembly.validate.

It isn't that they will be reported at call time necessarily, but that it is not deterministic when the error will be thrown. Not sure if that is in documentation anywhere though.

PR #721, opened 2016-07-08
Update references to the "function" element type.

sunfishcode:
This is renamed to "anyfunc" in AstSemantics.md in #682.

Also, fix grammar in the Elements Section.


lukewagner:
Thanks!

PR #726, opened 2016-07-18
Add Table.prototype.length

lukewagner:
The `Table` object currently has `get`, `set` and `grow`, but no `length`.  This PR adds `length`.


rossberg-chromium:
lgtm

flagxor:
lgtm, seems no-brainer.

PR #727, opened 2016-07-18
BinaryEncoding.md changes implied by #682

lukewagner:
This PR makes the binary format changes to import, export, memory, table, global, code, data and element sections as implied by #682.  The PR is currently a branch off master but I'm expecting to rebase it against whatever `binary_0x*` branch we choose when we're ready to merge.

The existing s-expr text format can produce this new binary format, but new s-expr forms are necessary to activate the new functionality of #682:
 * `(import "m" "f" memory <min> <max?>)` / `(import "m" "f" table <min> <max?>)` to import (the default) module/table
 * `(export "x" memory)`/`(export "y" table)` to export (the default) module/table
 * `(elem <off> x y z)` to specify an element segment that writes functions `x`, `y` and `z` into the default table with offset specified by the <off> initializer s-expr
 * `(data <off> "data")` to specify a data segment that writes data into the default memory with offset specified by the <off> initializer s-expr
 * `(table x y z)` would now produce both a `table` section and an `elem` section
 * some new syntax is needed to define a table with `initial`/`maximum` (without elements); for SM prototype/testing purposes I used `(table (resizable initial maximum))`, but happy to have alternatives

This PR makes one change to Modules.md: the elements of the element section are changed to simply be indices (into the appropriate index space); the per-element type was redundant.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -66,8 +66,8 @@ A single-byte unsigned integer indicating the kind of definition being imported 
 * `2` indicating a `Memory` [import](Modules.md#imports) or [definition](Modules.md#linear-memory-section)
 * `3` indicating a `Global` [import](Modules.md#imports) or [definition](Modules.md#global-section)
 
-### resizable_definition
-A packed tuple that describes the import or definition of a resizable 
+### resizable_limits
+A packed tuple that describes the limits of a
 [table](AstSemantics.md#table) or [memory](AstSemantics.md#resizing):
 
 | Field | Type | Description |
@@ -200,13 +200,13 @@ or, if the `kind` is `Table`:
 | Field | Type | Description |
 | ----- | ---- | ----------- |
 | element_type | `varuint7` | `0x40`, indicating [`anyfunc`](AstSemantics.md#table) |
-| | `resizable_definition` | see [above](#resizable_definition) |
+| | `resizable_limits` | see [above](#resizable_limits) |
 
 or, if the `kind` is `Memory`:
 
 | Field | Type | Description |
 | ----- | ---- | ----------- |
-| | `resizable_definition` | see [above](#resizable_definition) |
+| | `resizable_limits` | see [above](#resizable_limits) |
 
 or, if the `kind` is `Global`:
 
@@ -236,18 +236,18 @@ The encoding of a [Table section](Modules.md#table-section):
 | Field | Type | Description |
 | ----- | ---- | ----------- |
 | element_type | `varuint7` | `0x40`, indicating [`anyfunc`](AstSemantics.md#table) |
-| | `resizable_definition` | see [above](#resizable_definition) |
+| | `resizable_limits` | see [above](#resizable_limits) |
 
 ### Memory section
 
 ID: `memory`
 
 The encoding of a [Memory section](Modules.md#linear-memory-section) is simply
-a `resizable_definition`:
+a `resizable_limits`:
 
 | Field | Type | Description |
 | ----- | ---- | ----------- |
-| | `resizable_definition` | see [above](#resizable_definition) |
+| | `resizable_limits` | see [above](#resizable_limits) |
 
 Note that the initial/maximum fields are specified in units of 
 [WebAssembly pages](AstSemantics.md#linear-memory).
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -83,8 +83,8 @@ The `varuint32` "flags" field assigns the following meaning to the bits:
 
 ### init_expr
 The encoding of an [initializer expression](Modules.md#initializer-expression)
-is simply the encoding of the operator (`*.const` or `get_global`) followed by
-its immediate (constant value or global index).
+is the normal encoding of the expression followed by the `end` opcode as a
+delimiter.
 
 # Definitions
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -200,7 +200,7 @@ or, if the `kind` is `Table`:
 
 | Field | Type | Description |
 | ----- | ---- | ----------- |
-| element_type | `varuint7` | `0x40`, indicating [`anyfunc`](AstSemantics.md#table) |
+| element_type | `varuint7` | `0x20`, indicating [`anyfunc`](AstSemantics.md#table) |
 | | `resizable_limits` | see [above](#resizable_limits) |
 
 or, if the `kind` is `Memory`:
@@ -236,7 +236,7 @@ The encoding of a [Table section](Modules.md#table-section):
 
 | Field | Type | Description |
 | ----- | ---- | ----------- |
-| element_type | `varuint7` | `0x40`, indicating [`anyfunc`](AstSemantics.md#table) |
+| element_type | `varuint7` | `0x20`, indicating [`anyfunc`](AstSemantics.md#table) |
 | | `resizable_limits` | see [above](#resizable_limits) |
 
 ### Memory section
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -140,6 +140,7 @@ The content of each section is encoded in its `payload_str`.
 * [Export](#export-section) section
 * [Start](#start-section) section
 * [Code](#code-section) section
+* [Element](#element-section) section
 * [Data](#data-section) section
 * [Name](#name-section) section
 
@@ -321,6 +322,26 @@ declaration corresponds to the `i`th function body.
 | count | `varuint32` | count of function bodies to follow |
 | bodies | `function_body*` | sequence of [Function Bodies](#function-bodies) |
 
+### Element section
+
+ID: `elem`
+
+The encoding of the [Elements section](Modules.md#elements-section):
+
+| Field | Type | Description |
+| ----- | ---- | ----------- |
+| count | `varuint32` | count of element segments to follow |
+| entries | `elem_segment*` | repeated element segments as described below |
+
+a `elem_segment` is:
+
+| Field | Type | Description |
+| ----- | ---- | ----------- |
+| index | `varuint32` | the [table index](Modules.md#table-index-space) (0 in the MVP) |
+| offset | `init_expr` | an `i32` initializer expression that computes the offset at which to place the elements |
+| num_elem | `varuint32` | number of elements to follow |
+| elems | `varuint32*` | sequence of [function indices](Modules.md#function-index-space) |
+
 ### Data section
 
 ID: `data`
@@ -341,26 +362,6 @@ a `data_segment` is:
 | offset | `init_expr` | an `i32` initializer expression that computes the offset at which to place the data |
 | size | `varuint32` | size of `data` (in bytes) |
 | data | `bytes` | sequence of `size` bytes |
-
-### Element section
-
-ID: `elem`
-
-The encoding of the [Elements section](Modules.md#elements-section):
-
-| Field | Type | Description |
-| ----- | ---- | ----------- |
-| count | `varuint32` | count of element segments to follow |
-| entries | `elem_segment*` | repeated element segments as described below |
-
-a `elem_segment` is:
-
-| Field | Type | Description |
-| ----- | ---- | ----------- |
-| index | `varuint32` | the [table index](Modules.md#table-index-space) (0 in the MVP) |
-| offset | `init_expr` | an `i32` initializer expression that computes the offset at which to place the elements |
-| num_elem | `varuint32` | number of elements to follow |
-| elems | `varuint32*` | sequence of [function indices](Modules.md#function-index-space) |
 
 ### Name section
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -199,14 +199,13 @@ or, if the `kind` is `Table`:
 
 | Field | Type | Description |
 | ----- | ---- | ----------- |
-| table_index | `varuint32` | [table index](Modules.md#table-index-space); must be 0 in the MVP |
+| element_type | `varuint7` | `0x40`, indicating [`anyfunc`](AstSemantics.md#table) |
 | | `resizable_definition` | see [above](#resizable_definition) |
 
 or, if the `kind` is `Memory`:
 
 | Field | Type | Description |
 | ----- | ---- | ----------- |
-| memory_index | `varuint32` | [linear memory index](Modules.md#linear-memory-index-space); must be 0 in the MVP |
 | | `resizable_definition` | see [above](#resizable_definition) |
 
 or, if the `kind` is `Global`:
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -59,7 +59,7 @@ A single-byte unsigned integer indicating a [value type](AstSemantics.md#types).
 * `3` indicating type `f32` 
 * `4` indicating type `f64`
 
-### definition_kind
+### external_kind
 A single-byte unsigned integer indicating the kind of definition being imported or defined:
 * `0` indicating a `Function` [import](Modules.md#imports) or [definition](Modules.md#function-and-code-sections)
 * `1` indicating a `Table` [import](Modules.md#imports) or [definition](Modules.md#table-section)
@@ -187,7 +187,7 @@ The import section declares all imports that will be used in the module.
 | module_str | `bytes` | module string of `module_len` bytes |
 | field_len | `varuint32` | field name length |
 | field_str | `bytes` | field name string of `field_len` bytes |
-| kind | `definition_kind` | the kind of definition being imported |
+| kind | `external_kind` | the kind of definition being imported |
 
 Followed by, if the `kind` is `Function`:
 
@@ -290,7 +290,7 @@ The encoding of the [Export section](Modules.md#exports):
 | ----- | ---- | ----------- |
 | field_len | `varuint32` | field name string length |
 | field_str | `bytes` | field name string of `field_len` bytes |
-| kind | `definition_kind` | the kind of definition being exported |
+| kind | `external_kind` | the kind of definition being exported |
 | index | `varuint32` | the index into the corresponding [index space](Modules.md) |
 
 For example, if the "kind" is `Function`, then "index" is a 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -214,7 +214,7 @@ or, if the `kind` is `Global`:
 | Field | Type | Description |
 | ----- | ---- | ----------- |
 | type | `value_type` | type of the imported global |
-| mutability | `uint8` | `0` if immutable, `1` if mutable; must be `0` in the MVP |
+| mutability | `varuint1` | `0` if immutable, `1` if mutable; must be `0` in the MVP |
 
 ### Function section
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -86,6 +86,10 @@ The encoding of an [initializer expression](Modules.md#initializer-expression)
 is the normal encoding of the expression followed by the `end` opcode as a
 delimiter.
 
+Note that `get_global` in an initializer expression can only refer to immutable
+imported globals and all uses of `init_expr` can only appear after the Imports
+section.
+
 # Definitions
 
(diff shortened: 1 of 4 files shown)


BinaryEncoding.md
```diff
@@ -170,52 +223,79 @@ The function section _declares_ the signatures of all functions in the
...
-| initial | `varuint32` | initial memory size in 64KiB pages |
-| maximum | `varuint32` | maximum memory size in 64KiB pages |
-| exported | `uint8` | `1` if the memory is visible outside the module |
+| ----- | ---- | ----------- |
+| | `resizable_definition` | see [above](#resizable_definition) |
+
+Note that the initial/maximum fields are specified in units of 
+[WebAssembly pages](AstSemantics.md#linear-memory).
+
+### Global section
```
bnjbvr: Can you indicate in the High Level structure section where this section is located? (and add a link to this anchor)

BinaryEncoding.md
```diff
@@ -249,18 +329,39 @@ The data section declares the initialized data that is loaded
...
 | Field | Type | Description |
-| ----- |  ----- | ----- |
-| offset | `varuint32` | the offset in linear memory at which to store the data |
+| ----- | ---- | ----------- |
+| index | `varuint32` | the [linear memory index](Modules.md#linear-memory-index-space) (0 in the MVP) |
+| offset | `init_expr` | an `i32` initializer expression that computes the offset at which to place the data |
 | size | `varuint32` | size of `data` (in bytes) |
 | data | `bytes` | sequence of `size` bytes |
 
+### Element section
```
bnjbvr: Ditto.

BinaryEncoding.md
bnjbvr: It appears that in this patch, the Global section is placed after the Import section. I assume there is a single Global index space; so this means that in the index space, all the imported global indexes will be strictly less than the first locally defined global, is that right? Independently of the final choice (imported global being before or after the locally defined globals), it might be worth telling something about it here or somewhere else (Modules.md), maybe?

Actually, if `call_import` is removed, function calls also share the functions index space and the same precision could be useful too.

BinaryEncoding.md
```diff
@@ -170,52 +223,79 @@ The function section _declares_ the signatures of all functions in the
...
+#### Global Entry
+
+Each `global_entry` declares a number of global variables of a given type and mutability.
+It is legal to have several entries with the same type.
+
+| Field | Type | Description |
+| ----- | ---- | ----------- |
+| count | `varuint32` | number of global variables of the following type/mutability |
+| type | `value_type` | type of the variables |
+| mutability | `uint8` | `0` if immutable, `1` if mutable |
```
bnjbvr: What about initial values? Are they embedded here as an initializer expression, or are they set with default values (0 or vectors of 0)? (in the latter case, the emitter can still define these with other values in the start function)

BinaryEncoding.md
```diff
@@ -59,6 +59,32 @@ A single-byte unsigned integer indicating a [value type](AstSemantics.md#types).
 * `3` indicating type `f32` 
 * `4` indicating type `f64`
 
+### definition_kind
```
rossberg-chromium: Nit: How about `export_kind` or `external_kind`? It's an annotation on imports/exports, not definitions, and only includes things that can actually be imported/exported.

BinaryEncoding.md
```diff
@@ -59,6 +59,32 @@ A single-byte unsigned integer indicating a [value type](AstSemantics.md#types).
...
+A single-byte unsigned integer indicating the kind of definition being imported or defined:
+* `0` indicating a `Function` [import](Modules.md#imports) or [definition](Modules.md#function-and-code-sections)
+* `1` indicating a `Table` [import](Modules.md#imports) or [definition](Modules.md#table-section)
+* `2` indicating a `Memory` [import](Modules.md#imports) or [definition](Modules.md#linear-memory-section)
+* `3` indicating a `Global` [import](Modules.md#imports) or [definition](Modules.md#global-section)
+
+### resizable_definition
```
rossberg-chromium: Nit: `resizable_limits`

BinaryEncoding.md
```diff
@@ -59,6 +59,32 @@ A single-byte unsigned integer indicating a [value type](AstSemantics.md#types).
...
+
+The `varuint32` "flags" field assigns the following meaning to the bits:
+* `0x1` : this is a *default* [table](AstSemantics.md#table)/[memory](AstSemantics.md#linear-memory)
+          (in the MVP, *every* memory/table import/definition must be flagged as default)
+* `0x2` : indicates that the memory/table has a specified maximum
+
+### init_expr
+The encoding of an [initializer expression](Modules.md#initializer-expression)
+is simply the encoding of the operator (`*.const` or `get_global`) followed by
+its immediate (constant value or global index).
```
rossberg-chromium: This may not be future proof. If we want to allow more operators in the future then we probably need to delimit the operator sequence somehow, e.g. with an `end` opcode.

BinaryEncoding.md
```diff
@@ -110,6 +136,7 @@ The content of each section is encoded in its `payload_str`.
 * [Function](#function-section) section
 * [Table](#table-section) section
 * [Memory](#memory-section) section
+* [Global](#global-section) section
```
rossberg-chromium: Add the element section here, too.

BinaryEncoding.md
```diff
@@ -149,18 +176,44 @@ ID: `import`
...
+
+| Field | Type | Description |
+| ----- | ---- | ----------- |
+| sig_index | `varuint32` | signature index of the import |
+
+or, if the `kind` is `Table`:
+
+| Field | Type | Description |
+| ----- | ---- | ----------- |
+| element_type | `varuint7` | `0x40`, indicating [`anyfunc`](AstSemantics.md#table) |
```
rossberg-chromium: This is a different type constructor from the 0x40 used for typed functions, so it should use a different opcode (imagine a future where we can have both heterogeneous and homogeneous tables, so both forms of function types).

BinaryEncoding.md
```diff
@@ -149,18 +176,44 @@ ID: `import`
...
+| Field | Type | Description |
+| ----- | ---- | ----------- |
+| | `resizable_definition` | see [above](#resizable_definition) |
+
+or, if the `kind` is `Global`:
+
+| Field | Type | Description |
+| ----- | ---- | ----------- |
+| type | `value_type` | type of the imported global |
+| mutability | `uint8` | `0` if immutable, `1` if mutable; must be `0` in the MVP |
```
rossberg-chromium: Is exporting a mutable global an error, too, in the MVP?

rossberg-chromium: Should this perhaps be `varuint1`?

BinaryEncoding.md
rossberg-chromium: `varuint1` perhaps?

BinaryEncoding.md
```diff
@@ -249,18 +329,39 @@ The data section declares the initialized data that is loaded
...
-| offset | `varuint32` | the offset in linear memory at which to store the data |
+| ----- | ---- | ----------- |
+| index | `varuint32` | the [linear memory index](Modules.md#linear-memory-index-space) (0 in the MVP) |
+| offset | `init_expr` | an `i32` initializer expression that computes the offset at which to place the data |
 | size | `varuint32` | size of `data` (in bytes) |
 | data | `bytes` | sequence of `size` bytes |
 
+### Element section
+
+ID: `elem`
```
rossberg-chromium: Nit: So far we have written out all section names. If we take "elem", then I'd suggest shortening "function" to "func", too. We probably should separate that Q, so I suggest `element` for this PR.

rossberg-chromium:
On 19 July 2016 at 00:33, Luke Wagner <notifications@github.com> wrote:

> The existing s-expr text format can produce this new binary format, but
> new s-expr forms are necessary to activate the new functionality of #682
> <https://github.com/WebAssembly/design/pull/682>:
>
>    - (import "m" "f" memory) / (import "m" "f" table) to import (the
>    default) module/table
>    - (export "x" memory)/(export "y" table) to export (the default)
>    module/table
>    - (elem off x y z) to initialize functions x, y and z int the default
>    table at offset
>    - (table x y z) would now produce both a table section and an elem
>    section
>    - some new syntax is needed to define a table with initial/maximum
>    (without elements); for SM prototype/testing purposes I used (table
>    (resizable initial maximum)), but happy to have alternatives
>
>
I was just starting to implement an S-expression format for this the other
day. My thoughts so far:

- Table syntax and memory syntax should be symmetric. So table should
become `(table <min> <max>? <table_segment>*)`.
- Correspondingly, an element segment has the form `(segment <offset>
<var>*)`, where <offset> is an S-expr now (same generalisation for memory
segments).
- We may keep the current table syntax as a shorthand.
- But in fact it may make sense to bring the S-expression format closer to
the current section format, i.e., pull out the segments from memory and
table into their own `data` and `elem` declarations.
- Import/export memory/table syntax like you propose, but imports also need
to include the limits. In accordance with the above, I would simply suggest
`(import "a" "b" table <min> <max>?)`, similarly for memory.
- We should make the syntax future proof wrt to multi-tables/memories.
Allowing identifier binders with their definitions and imports for that
purpose seems straightforward. For the default, maybe it's easiest to
introduce separate declarations `(default table <var>)` and `(default
memory <var>)`. These are only required if there is more than one of those
(but then you can't represent a program with just one table but no default,
hm...).

Come to think about it, would it make sense to declare defaults in their
own "default" section even in the binary format? It's a bit awkward to
mangle and distribute that flag in several possible places, because it's
kind of orthogonal. In particular, it's a bit weird that import specs have
this flag, since it has no relevance to the import semantics itself.

BinaryEncoding.md
lukewagner: That's right; that's how it works for all the importable definitions (functions, memories, tables, globals) as well as arguments+locals.  This is documented in Modules.md in the "X index space" sections which I've tried to link to, where appropriate, from this doc.  But I could try to add a few more links.

BinaryEncoding.md
lukewagner: I like `external_kind`; that's more precise.

BinaryEncoding.md
lukewagner: Also sounds good; more precise.

BinaryEncoding.md
lukewagner: Oh, hah, actually you're right; adding any non-nullary node would have this problem, so it is super un-future-proof.  I'll tack on an `end` for now.

BinaryEncoding.md
lukewagner: Done

BinaryEncoding.md
lukewagner: Right, both importing and exporting of mutable globals is disallowed (so we don't have to create some new `WebAssembly.Global` JS API; we can just import/export values).

Changed to `varuint1`

BinaryEncoding.md
lukewagner: So I was actually thinking forward to these cases and my intention was for `element_type` to use the same representation of type as we'd use for local variables when we had GC types.  In particular, IIRC, we'd have the positive "nullary constructors" for primitive types and then some way (prefix or the negative trick) to say "index into the type section" to describe function, struct, and array types.  So given all that, where do we stick "anyfuc"?  My though was that, since these non-nullary constructors can't show up directly, we could use them to mean "any X" (for X being "func", "struct", "array").  Maybe that's too cute; in which case I think "anyfunc" would just be another nullary constructor, but I think it would necessarily need to be disjoint from all the other constructors, right?

rossberg-chromium: Agreed. A while back I spent some thought on what the Wasm type structure would become with GC types (happy to talk about that next week). It quickly explodes into a zoo of categories like num_type, ref_type, value_type, field_type, mem_type, struct_type etc that combine in various ways. So I would rather avoid being too cute now, especially since we have plenty of room in the type opcode space. Perhaps just pick some free opcode with enough distance from others.

lukewagner:
> - So table should become `(table <min> <max>? <table_segment>*)`.

> - We may keep the current table syntax as a shorthand.

I think these two are in conflict since you won't be able to tell whether `(table 1 2)` is a table with min/max 1/2 or elements 1 and 2.  (That's why I put <min> <max> in its own nested expression.)

> But in fact it may make sense to bring the S-expression format closer to the current section format, i.e., pull out the segments from memory and table into their own `data` and `elem` declarations.

Agreed; we actually *need* to do this so that we can initialize imported table/memory.  (This is what we've done in SM's proto s-expr format which I forgot to mention above.)

> Import/export memory/table syntax like you propose, but imports also need to include the limits. In accordance with the above, I would simply suggest `(import "a" "b" table <min> <max>?)`, similarly for memory.

Oops, yes, that's what we've implemented in our s-expr proto, I forgot to mention it above; I'll update it.

> We should make the syntax future proof wrt to multi-tables/memories.

When we allow multiple tables/memories (which I think will necessarily block on GC, so it'll be a while, and by then I'm thinking we'll have a non-s-expr text format), I was thinking we could introduce it in a backwards-compatible way by saying: if you specify a name in the import/definition of a table/memory, then it's non-default.  Then to refer to a non-default memory/table you write `(memory $foo)` instead of `memory`.

> Come to think about it, would it make sense to declare defaults in their own "default" section even in the binary format? It's a bit awkward to mangle and distribute that flag in several possible places, because it's kind of orthogonal.

I kinda see the abstract point, but I don't see any practical downside to putting it on the import/def or practical upside to having a separate "defaults" section.  It might actually be a hair more complicated since it produces a window, in the decoding process, where you have a memory/table but don't yet know if it's default or not.  Until we allow non-default memories/tables it is definitely more work (b/c atm you trivially just check that the bit is set),

BinaryEncoding.md
rossberg-chromium: (Btw, I also discovered that it may be better to represent mutability with a type constructor `mut` instead of a flag on the holder, since that composes better. But we can talk that through after this PR.)

lukewagner: Righto, changed to 0x20.

rossberg-chromium:
On 19 July 2016 at 17:40, Luke Wagner <notifications@github.com> wrote:

>
>    -
>
>    So table should become (table <min> <max>? <table_segment>*).
>    -
>
>    We may keep the current table syntax as a shorthand.
>
> I think these two are in conflict since you won't be able to tell whether (table
> 1 2) is a table with min/max 1/2 or elements 1 and 2. (That's why I put
> in its own nested expression.)
>
Aw right. In that case, I propose removing the current syntax, given that
it no longer matches the binary structure. We don't need to optimise this
construct for terseness.

We should make the syntax future proof wrt to multi-tables/memories.
>
> When we allow multiple tables/memories (which I think will necessarily
> block on GC, so it'll be a while, and by then I'm thinking we'll have a
> non-s-expr text format), I was thinking we could introduce it in a
> backwards-compatible way by saying: if you specify a name in the
> import/definition of a table/memory, then it's non-default. Then to refer
> to a non-default memory/table you write (memory $foo) instead of memory.
>

Yeah, that has the disadvantage that you cannot represent a case that
actually exists, and for which you'd probably like to write tests.


> Come to think about it, would it make sense to declare defaults in their
> own "default" section even in the binary format? It's a bit awkward to
> mangle and distribute that flag in several possible places, because it's
> kind of orthogonal.
>
> I kinda see the abstract point, but I don't see any practical downside to
> putting it on the import/def or practical upside to having a separate
> "defaults" section. It might actually be a hair more complicated since it
> produces a window, in the decoding process, where you have a memory/table
> but don't yet know if it's default or not. Until we allow non-default
> memories/tables it is definitely more work (b/c atm you trivially just
> check that the bit is set),
>
Actually, to the contrary: it would completely avoid the need for
introducing any default flags/decls now. Even later you'd only require the
default section for "disambiguation" in modules that actually define/import
multiple tables/memories -- if a module has only one, that is automatically
taken as the default.  Wouldn't that factor things more nicely?

It's a bit weird right now that the default flag is part of the
resizable_limits, although it has nothing to do with them -- it's rather a
coincidence that only resizable things have defaults. With the above, the
flag field could be removed from the limits for good.

BinaryEncoding.md
```diff
@@ -59,6 +59,32 @@ A single-byte unsigned integer indicating a [value type](AstSemantics.md#types).
...
+| flags | `varuint32` | described below |
+| initial | `varuint32` | initial length (in units of table elements or wasm pages) |
+| maximum | `varuint32`? | only present if specified by `flags` |
+
+The `varuint32` "flags" field assigns the following meaning to the bits:
+* `0x1` : this is a *default* [table](AstSemantics.md#table)/[memory](AstSemantics.md#linear-memory)
+          (in the MVP, *every* memory/table import/definition must be flagged as default)
+* `0x2` : indicates that the memory/table has a specified maximum
+
+### init_expr
```
rossberg-chromium: Btw, are all uses of `init_expr` evaluated in order? What happens if a `get_global` in one of them refers to a not-yet-initialised global?

lukewagner: I should also state the requirement here but Modules.md#initializer-expression requires that the `get_global` only refer to an immutable import which will necessarily precede all the uses.

lukewagner:
> Aw right. In that case, I propose removing the current syntax

sgtm.  The current hack in SM is just to allow existing .wasts to work unchanged (so it's easy to test on existing demo builds), but I'm happy to unhack.

> Yeah, that has the disadvantage that you cannot represent a case that actually exists, and for which you'd probably like to write tests.

Do you mean an import/definition that is both default and used by index?  It seems like that could be support by allowing a `table`/`memory` to be explicitly marked as default with some `default` token.

> Actually, to the contrary: it would completely avoid the need for introducing any default flags/decls now.

Oh, then I guess I misunderstood what you were proposing; so you're just saying that, when there's only 1, it's the default by, well, default.  Then post-MVP, when we support >1, we can decide how to pick which is default.  That sounds good to me and keeps our options open.

rossberg-chromium:
Btw, shouldn't this be a PR against the binary_0xc (or binary_0xd) branch?

lukewagner:
Right, some binary_0x* branch; I'm not expecting to merge this into master.  As I noted in the OP, I'm waiting to see what people want to do for either folding this into binary_0xc or waiting until binary_0xd.

rossberg-chromium:
@lukewagner, actually, shouldn't the same have applied to #682? I just realised it got merged into master. ;)

lukewagner:
@rossberg-chromium Well, the nascent protocol we've been using is to only use the binary_0x* branch for BinaryEncoding.md changes (hence "binary_" :) for the purpose of coordinating landing the breaking binary format changes in engines.  Post-MVP, breaking changes will cease, so probably we should shift to a more regular "everything related to a proposed feature is on a branch until the whole thing merges".

BinaryEncoding.md
```diff
@@ -372,9 +478,10 @@ out of range, `br_table` branches to the default target.
 | `f32.const` | `0x13` | value : `uint32` | a constant value interpreted as `f32` |
 | `get_local` | `0x14` | local_index : `varuint32` | read a local variable or parameter |
 | `set_local` | `0x15` | local_index : `varuint32` | write a local variable or parameter |
-| `call` | `0x16` | argument_count : `varuint1`, function_index : `varuint32` | call a function by its index |
+| `get_global` | `0x18` | global_index : `varuint32` | read a global variable |
+| `set_global` | `0x19` | global_index : `varuint32` | write a global variable |
+| `call` | `0x16` | argument_count : `varuint1`, function_index : `varuint32` | call a function by its [index](Modules.md#function-index-space) |
```
rossberg-chromium: Thought I had commented on that already, but can't find the comment anymore: fyi, this conflicts with #711, will have to resolve somehow.

lukewagner: Since we're going to defragment this opcode space anyway, I'll change these codes.

flagxor:
Luke this seems good, can you resolve conflicts and merge?

lukewagner:
Righto.  Merged to `binary_0xc` (not `master`, as discussed above).  (GH doesn't let me simply retarget this PR, so I'll just close this now.)

PR #731, opened 2016-07-26
Rework f64.min and f64.max for 0.0 and -0.0.

gahaas:
This PR proposes that if the operands of f64.min or f64.max are `0.0` and `-0.0`, then the first operand is returned. Thereby it is not necessary for the implementations of f64.min and f64.max to distinguish between `0.0` and `-0.0`.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -579,7 +579,7 @@ follows:
   * `f64.min`: minimum (binary operator); if either operand is NaN, returns NaN
   * `f64.max`: maximum (binary operator); if either operand is NaN, returns NaN
 
-`min` and `max` operators treat `-0.0` as being effectively less than `0.0`.
+`min` and `max` operators return the first operand if the operands are `0.0` and `-0.0`.
 
 In floating point comparisons, the operands are *unordered* if either operand
 is NaN, and *ordered* otherwise.


sunfishcode:
This change would add overhead on ARM and Power, which can implement WebAssembly's current min/max with a single instruction.

Also, this change would make WebAssembly's min/max asymmetric, which would make it less convenient for some WebAssembly producers and tools.

And, it would make WebAssembly's min/max different from the min/max operations in ECMAScript and Java.

taisel:
No way to provide both kinds, and let the user/IDE sort it out?

jfbastien:
We'll eventually want IEEE754's `minNum` and `maxNum`, and maybe `totalOrder`. Until then I think having `min` and `max` as specified in most languages and ISAs is useful (x86's [`minss`](http://x86.renejeschke.de/html/file_module_x86_id_174.html) and [`maxss`](http://x86.renejeschke.de/html/file_module_x86_id_169.html) are the same).

The current specification does the equivalent of using the regular less-than (for `min`) and greater-than (for `max`) opcodes, which are as surprising as `min` / `max` are (i.e. `0.0 == -0.0`).

sunfishcode:
@taisel Rather than both of 2 kinds, there are 3 popular families of min/max semantics: JS/Java/etc, x<y?x:y, and IEEE754 minNum/maxNum, and the specific PR here is different yet (it's asymmetric for -0 and symmetric for NaN), so I'll at least advise against creating a 4th family.

(Note that x86's minss/maxss are asymmetric for both -0 and NaN, which is different from the current specification's min/max, which are fully symmetric. Also note that the current specification's min/max do distinguish between -0 and 0, unlike less-than and greater-than which don't.)

gahaas:
I was not aware that the current spec matches the JS/Java/etc spec when I created the pull request. I agree that we should not create a 4th family.

PR #734, opened 2016-08-01
Use section codes instead of section names

titzer:
This PR proposes uses section codes for known sections, which is more compact and easier to check in a decoder. It allows for user-defined sections that have string names to be encoded in the same manner as before. The scheme of using negative numbers proposed here also has the advantage of allowing a single decoder to accept the old (0xB) format and the new (0xC) format for the time being.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -89,40 +89,42 @@ The module starts with a preamble of two fields:
 | magic number | `uint32` |  Magic number `0x6d736100` (i.e., '\0asm') |
 | version | `uint32` | Version number, currently 10. The version for MVP will be reset to 1. |
 
-This preamble is followed by a sequence of sections. Each section is identified by an
-immediate string. Sections whose identity is unknown to the WebAssembly
-implementation are ignored and this is supported by including the size in bytes
-for all sections. The encoding of sections is structured as follows:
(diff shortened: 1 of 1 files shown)


BinaryEncoding.md
```diff
@@ -263,8 +249,6 @@ a `data_segment` is:
 
 ### Name section
 
-ID: `name`
-
 The names section does not change execution semantics and a validation error in
 this section does not cause validation for the whole module to fail and is
 instead treated as if the section was absent. The expectation is that, when a
```
lukewagner: One idea we discussed was to have the names section be defined as a user-defined section.  Then all the logic here (about how you can always ignore a names section -- that it doesn't have any semantic effect) could be hoisted and apply to all user-defined sections.

titzer: Would that basically mean that we move the names section to the JS embedding spec?

lukewagner: We could, although I think the names section makes sense more broadly.  If we change the meaning from "user-defined" to "metadata" or "ignorable", then it wouldn't seem too out of place to have the core wasm spec define a metadata/ignorable section.

titzer:
I screwed up the PR again by targeting master. Closing in favor of #739

titzer:
Make that #740

PR #735, opened 2016-08-01
Use LEB for br_table

titzer:
A very significant portion of the AngryBots and BananaBread demo binaries is taken by `br_table` opcodes. In fact, in both of these binaries, the total space consumed by `br_table` is more than `br` and `br_if` combined. This PR proposes using LEBs in the encoding of `br_table`. LEBs are more space-efficient than uint32 for the very common case of br_table target entries being smaller than 4 bytes, and they are more consistent with the rest of the format.

A possible followup would be require the length of the entire `br_table` as an immediate as well, to allow for bytecode iterators to skip ahead without decoding the entire table.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -354,9 +354,9 @@ The `br_table` operator has an immediate operand which is encoded as follows:
 | Field | Type | Description |
 | ---- | ---- | ---- |
 | arity | `varuint1` | number of arguments |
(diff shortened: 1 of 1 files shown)


lukewagner:
lgtm

sunfishcode:
lgtm

titzer:
Closing in favor of #738 (merge into 0xC instead of master)
https://github.com/WebAssembly/design/pull/738

PR #738, opened 2016-08-01
Use LEB for br_table

titzer:
(Rebasing onto binary_0xC branch--#735 was accidentally based on master)

A very significant portion of the AngryBots and BananaBread demo binaries is taken by br_table opcodes. In fact, in both of these binaries, the total space consumed by br_table is more than br and br_if combined. This PR proposes using LEBs in the encoding of br_table. LEBs are more space-efficient than uint32 for the very common case of br_table target entries being smaller than 4 bytes, and they are more consistent with the rest of the format.

A possible followup would be require the length of the entire br_table as an immediate as well, to allow for bytecode iterators to skip ahead without decoding the entire table.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -461,9 +461,9 @@ The `br_table` operator has an immediate operand which is encoded as follows:
 | Field | Type | Description |
 | ---- | ---- | ---- |
 | arity | `varuint1` | number of arguments |
-| target_count | `varuint32` | number of targets in the target_table |
-| target_table | `uint32*` | target entries that indicate an outer block or loop to which to break |
-| default_target | `uint32` | an outer block or loop to which to break in the default case |
+| target_count | `varuint32` | number of entries in the target_table |
+| target_table | `varuint32*` | target entries that indicate an outer block or loop to which to break |
+| default_target | `varuint32` | an outer block or loop to which to break in the default case |
 
 The `br_table` operator implements an indirect branch. It accepts an optional value argument
 (like other branches) and an additional `i32` expression as input, and 


mbebenita:
# Angry Bots
```
new.ab.wasm 11526505 
old.ab.wasm 11728278
a decrease of 1.72%
```
```
new.ab.wasm.gz 3761189 
old.ab.wasm.gz 3774923
a decrease of 0.36%
```

# BananaBread
```
new.bb.wasm 2192211
old.bb.wasm 2298984
a decrease of 4.64%
```
```
new.bb.wasm.gz 759592
old.bb.wasm.gz 764356
a decrease of 0.62%
```

I tried this out in Binaryen, @titzer is this roughly what you're seeing?

titzer:
Yes, those numbers are consistent with the measurements we made.

mbebenita:
Out of curiosity, I tried a delta encoding w/ signed LEBs and the results are not great at all.

```
delta.ab.wasm 11526127
delta.ab.wasm.gz 3761463
```

```
delta.bb.wasm 2192245
delta.bb.wasm.gz 760323
```

JSStats:
I assume this is conceding the use case of being efficient for a literal interpreter of the wasm binary code? I saw another PR closed recently with a similar suggestion? Is V8 going to abandon their wasm interpreter, or will there be a decoder from the wasm into an internal code before interpreting?

Might a sparse representation option also help here? I guess with compression the difference might be small but it might still save more space for the uncompressed wasm binary.

titzer:
The V8 WASM interpreter uses an internal lookup table for all branches, including `br_table`, so it is able to compute the target of any branch in O(logn). It's possible to do this in O(1) with a more sophisticated table, but so far the lookup time has not been a bottleneck.

rossberg-chromium:
lgtm

Issue #561, opened 2016-02-27
Have we considered an LEB128-like encoding for section string names? 


mbebenita:
Currently, section names are encoded as:

| Field | Type | Description |
| ----- |  ----- | ----- |
| size  | `varuint32` | size of this section in bytes, excluding this size |
| id_len | `varuint32` | section identifier string length |
| id_str | `bytes` | section identifier string of id_len bytes |

We could restrict `id_str` to 7-bit ASCII and then just use a LEB128-like encoding for the entire string. Might be too much of a cute trick, but would save us a byte in each section header. One could image that the section structure could be nested and used to encode a variety of data in the binary file, so a tiny header might be useful.

JSStats:
Here's another thought - choose names from the brotli dictionary!

JSStats:
Sorry if this is off topic, but in case it's of interest here's how brotli encodes the current section names. where `[]` are dictionary hits and `{}` are copy hits.

```
[memory]
sig[nature]s
[functions]
[global]s
[data]_[segment]{s
function}_[table]
end
[start]_[function]
[import]{_table
e}x{port_table}
```

Another option would be for the wasm CG to reserve all single character names for it's own use and use single character section names.

titzer:
That is a neat trick, however the current length+bytes design will match
the other places that strings can occur, e.g. as function names and
import/exports. Those we definitely want to treat as uninterpreted bytes
from the engine point of view, and leave the interpretation up to the
embedder.

jfbastien:
Agreed with @titzer, though we should write that rationale down somewhere! We simply don't want .`wasm` generation to require magical escaping, and wasm really doesn't care about the bytes in there so there's no reason to add extra constraints.

PR #740, opened 2016-08-01
Use section codes instead of section names

titzer:
(rebasing onto 0xC instead of master)

This PR proposes uses section codes for known sections, which is more compact and easier to check in a decoder.
It allows for user-defined sections that have string names to be encoded in the same manner as before.
The scheme of using negative numbers proposed here also has the advantage of allowing a single decoder to accept the old (0xB) format and the new (0xC) format for the time being.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -123,7 +123,7 @@ Known sections have negative ids, while user-defined sections have positive ids 
 the length of a string identifier immediately to follow.
 After the section identification, the section length and data follow.
 All sections unknown to the WebAssembly implementation are ignored.
-A validation error user-defined sections does not cause validation for the whole module to fail and is
+A validation error in a user-defined section does not cause validation for the whole module to fail and is
 instead treated as if the section was absent.
 
 | Field | Type | Description |
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -119,12 +119,15 @@ The module starts with a preamble of two fields:
 
 The module preamble is followed by a sequence of sections.
 Each section is identified by a 1-byte *section code* that encodes either a known section or a user-defined section.
-Known sections have non-zero ids, while unkown sections simply have a zero id and are ignored by the WebAssembly implementation.
+Known sections have non-zero ids, while unknown sections have a `0` id followed by an identifying string. 
+Unknown sections are ignored by the WebAssembly implementation.
 The section length and data immediately follow the section code.
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
 | id | `varint7` | section code |
+| name_len | `varuint32` ? | length of the section name in bytes, present if `id == 0` |
+| name | `bytes` ? | section name string, present if `id == 0` |
 | payload_len  | `varuint32` | size of this section in bytes |
 | payload_data  | `bytes` | content of this section, of length `payload_len` |
 
@@ -350,7 +353,7 @@ a `data_segment` is:
 User-defined section string: `"name"`
 
 The names section does not change execution semantics, and thus is not allocated a section code.
-It is encoded as an unknown section (id `0`) with the first few payload bytes identifying this section as a string.
+It is encoded as an unknown section (id `0`) followed by the identification string `"name"`.
 Like all unknown sections, a validation error in this section does not cause validation of the module to fail.
 The expectation is that, when a binary WebAssembly module is viewed in a browser or other development
 environment, the names in this section will be used as the names of functions
@@ -358,8 +361,6 @@ and locals in the [text format](TextFormat.md).
 
 | Field | Type | Description |
 | ----- | ---- | ----------- |
-| name_length | `varint7` | length of the "name" string = 4 |
-| name_string | `bytes` | the literal string "name" of length 4 |
 | count | `varuint32` | count of entries to follow |
 | entries | `function_names*` | sequence of names |
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -124,7 +124,7 @@ The section length and data immediately follow the section code.
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
-| id | `uint8` | section code |
+| id | `varint7` | section code |
 | payload_len  | `varuint32` | size of this section in bytes |
 | payload_data  | `bytes` | content of this section, of length `payload_len` |
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -119,17 +119,19 @@ The module starts with a preamble of two fields:
 
 The module preamble is followed by a sequence of sections.
 Each section is identified by a 1-byte *section code* that encodes either a known section or a user-defined section.
-Known sections have non-zero ids, while unknown sections have a `0` id followed by an identifying string. 
-Unknown sections are ignored by the WebAssembly implementation.
-The section length and data immediately follow the section code.
+The section length and payload data then follow.
+Known sections have non-zero ids, while unknown sections have a `0` id followed by an identifying string as
+part of the payload.
+Unknown sections are ignored by the WebAssembly implementation, and thus validation errors within them do not
+invalidate a module.
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
 | id | `varint7` | section code |
+| payload_len  | `varuint32` | size of this section in bytes |
 | name_len | `varuint32` ? | length of the section name in bytes, present if `id == 0` |
 | name | `bytes` ? | section name string, present if `id == 0` |
-| payload_len  | `varuint32` | size of this section in bytes |
-| payload_data  | `bytes` | content of this section, of length `payload_len` |
+| payload_data  | `bytes` | content of this section, of length `payload_len - sizeof(name) - sizeof(name_len)` |
 
 Each section is optional and may appear at most once.
 Known sections from this list may not appear out of order.
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -123,6 +123,8 @@ Known sections have negative ids, while user-defined sections have positive ids 
 the length of a string identifier immediately to follow.
 After the section identification, the section length and data follow.
 All sections unknown to the WebAssembly implementation are ignored.
+A validation error user-defined sections does not cause validation for the whole module to fail and is
+instead treated as if the section was absent.
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
@@ -351,13 +353,12 @@ a `data_segment` is:
 
 ### Name section
 
-Section string: `"name"`
+User-defined section string: `"name"`
 
 The names section does not change execution semantics, and thus is not allocated
 a section opcode.
-A validation error in this section does not cause validation for the whole module to fail and is
-instead treated as if the section was absent. The expectation is that, when a
-binary WebAssembly module is viewed in a browser or other development
+Like all user-defined sections, a validation error in this section does not cause validation of the module to fail.
+The expectation is that, when a binary WebAssembly module is viewed in a browser or other development
 environment, the names in this section will be used as the names of functions
 and locals in the [text format](TextFormat.md).
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -117,42 +117,45 @@ The module starts with a preamble of two fields:
 | magic number | `uint32` |  Magic number `0x6d736100` (i.e., '\0asm') |
 | version | `uint32` | Version number, currently 10. The version for MVP will be reset to 1. |
 
-This preamble is followed by a sequence of sections. Each section is identified by an
-immediate string. Sections whose identity is unknown to the WebAssembly
-implementation are ignored and this is supported by including the size in bytes
-for all sections. The encoding of sections is structured as follows:
+This preamble is followed by a sequence of sections. Each section is identified by `varint32` 
+that encodes either a known section or a user-defined section.
+Known sections have negative ids, while user-defined sections have positive ids that encode 
+the length of a string identifier immediately to follow.
+After the section identification, the section length and data follow.
+All sections unknown to the WebAssembly implementation are ignored.
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
-| id_len | `varuint32` | section identifier string length |
-| id_str | `bytes` | section identifier string of id_len bytes |
+| id | `varint32` | section identifier code |
+| id_str | `bytes` | section identifier string, of length `max(id, 0)` bytes |
 | payload_len  | `varuint32` | size of this section in bytes |
-| payload_str  | `bytes` | content of this section, of length payload_len |
+| payload_data  | `bytes` | content of this section, of length `payload_len` |
 
 Each section is optional and may appear at most once.
 Known sections (from this list) may not appear out of order.
-The content of each section is encoded in its `payload_str`.
+The content of each section is encoded in its `payload_data`.
 
-* [Type](#type-section) section
-* [Import](#import-section) section
-* [Function](#function-section) section
-* [Table](#table-section) section
-* [Memory](#memory-section) section
-* [Global](#global-section) section
-* [Export](#export-section) section
-* [Start](#start-section) section
-* [Code](#code-section) section
-* [Element](#element-section) section
-* [Data](#data-section) section
-* [Name](#name-section) section
+| Section Name | Code | Description |
+| ------------ | ---- | ----------- |
+| [Type](#type-section) | `-1` | Function signature declarations |
+| [Import](#import-section) | `-2` | Import declarations |
+| [Function](#function-section) | `-3` | Function declarations |
+| [Table](#table-section) | `-4` | Indirect function table and other tables |
+| [Memory](#memory-section) | `-5` | Memory attributes |
+| [Global](#global-section) | `-6` | Global declarations |
+| [Export](#export-section) | `-7` | Exports | 
+| [Start](#start-section) | `-8` | Start function declaration |
+| [Code](#code-section) | `-9` | Function bodies (code) |
+| [Element](#element-section) | `-10` | Elements section |
+| [Data](#data-section) | `-11` | Data segments |
+| [Name](#name-section) | `-12`| Names section|
+
 
 The end of the last present section must coincide with the last byte of the
 module. The shortest valid module is 8 bytes (`magic number`, `version`,
 followed by zero sections).
 
 ### Type section
-
-ID: `type`
 
 The type section declares all function signatures that will be used in the module.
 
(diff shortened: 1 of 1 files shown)


JSStats:
As I suggested long ago if short codes were reserved for use by the standard then it would be possible to test their length and content in a single 32 bit integer comparison which seems slightly simply than using the sign of the length to differentiate between codes and strings. This would address the use case of being 'easier to check in a decoder', and if the encoded size is a real show stopper then use single character strings for the high frequency section names.

Perhaps one issue with differentiating ignored sections using a binary 'sign' bit is that this offers no clean transition from a de-facto section name to become a standard section name, and if decoders need to still check for strings in future then the design is just more complex. Also decoders in general will be looking at the section name strings.

titzer:
Friendly ping for any more thoughts on this PR.

JSStats:
Is the suggestion to use short strings acceptable and would it address the performance and complexity issues raised?

rossberg-chromium:
lgtm

JSStats:
@rossberg-chromium Wouldn't short strings address this without adding more special cases.

lukewagner:
@titzer From the other PR: since Name is already being defined with the special "you can ignore it" / "validation error does not cause the module to fail to validate" semantics that presumably we want for all the other user-defined sections, can we make names be defined as a user-defined section and move the description of the special validation rules to be the case for all user-defined sections?

qwertie:
Seriously... what's wrong with the short names idea? It's simpler for consumers that read-modify-write a module, in that the new proposal will need an extra field to store the negative number. My thinking is that because a feature test _must_ be an entire module (ill-advised, imo), eliminating (even short) section names could significantly shorten the many "test" modules one might need.

JSStats:
I think the term 'user defined section' is an inappropriate distinction too. The web community will still want to experiment with extensions, might want them to be ignored by runtimes that do not recognize them, and to be a shared standard and calling these 'user defined' is not appropriate. Perhaps the example that @lukewagner mentions is helpful. Perhaps the real distinction is a flag noting if the section can be ignored for execution purposes.

If codes are added as an option to names then unrecognised codes need to be ignored too and then might as well just remove the names and use codes, which might then need to be managed, but it was deemed simpler to manage a name space so back to names.

titzer:
One drawback to short strings (e.g. less than 3 bytes so that a single word-sized check suffices) is that they don't really allow a decoder to just do a simple "load word and compare" operation. There's a couple subtleties here. For example, the module could end early, so the decoder has to first decode the (varint) length (also with bounds checks) and then check the length doesn't go past the end of the module. After that, even with short strings, the actual bytes may not be aligned in memory, so the decoder may still have to do a byte-by-byte comparison on some architectures (e.g. ARM and MIPS).

Sections are few, so the efficiency argument isn't that pressing. It's just that doing a single byte comparison without all the extra checks is a lot simpler and more robust.

titzer:
As for for graduating a user-defined section to a spec-blessed section, there would have to be a point where the string or the opcode is allocated by standardization and then becomes supported by all engines.

The only issue is for binaries created before this point which still use a user-defined section string as opposed to a short string or an opcode. Short strings don't help here, since presumably standardization only allocates short strings, while experiments should not be using short strings.

lukewagner:
@titzer Instead of thinking of user-defined sections as pre-standardization (such that they graduate into a spec-blessed integer code), I'd like us to consider saying instead that these are the just sections that can be completely ignored by an engine without having semantic effect (as is the case with the Names section).  I like this because it means that new (semantic) features that involve a new section should go through the normal spec process before shipping (so they'd *start* with a new section opcode, just behind a flag at first, etc).

JSStats:
The PR still has the ID as a varint, so there is no change in complexity compared with reading the length. So there is one less check that the length is within the module, hardly justifying the change. Loading a few unaligned bytes is a trivial matter, also does not justify a change, and this does not force a byte-by-byte comparison rather if the length of bytes can fit in a 32 bit integer or 64 bit integer then it can just pack them in and then dispatch on this integer code.

titzer:
If we really want to split hairs on this one, this proposal would basically amount to:

int8_t val = read_i8();
switch (val) {
  case -1: do_section_x();
  case -2: do_section_y();
  case -3: do_section_z();
  ...
  default: {
    if (val >= 0) { 
    rewind(1);
    unsigned str_length = read_leb128();
    skip(str_length);
    unsigned length = read_leb128();
    skip(length);
   }
  }
}

So there is only a single byte read and indirect branch in the normal case, and a compliant engine never has to implement a byte-by-byte comparison for string sections, since by definition they are not required to be parsed.

titzer:
@lukewagner PTAL. I made the names section into a user-string specified section. Not sure if I like that, since V8 will definitely want to (try to) recognize the names section.

BinaryEncoding.md
```diff
@@ -369,10 +351,11 @@ a `data_segment` is:
...
 
-ID: `name`
+Section string: `"name"`
 
-The names section does not change execution semantics and a validation error in
-this section does not cause validation for the whole module to fail and is
+The names section does not change execution semantics, and thus is not allocated
+a section opcode.
+A validation error in this section does not cause validation for the whole module to fail and is
 instead treated as if the section was absent. The expectation is that, when a
```
lukewagner: Could you move this sentence up above, below the para that introduces "user define sections", changing "this section" to "a user defined section"?  That way it's clear that this is true for all user-defined sections, not just the names section.

lukewagner:
lgtm with the one nit; much appreciated

BinaryEncoding.md
titzer: Done

JSStats:
From the v8 code in module-decoder.cc the short names could address the trivial performance use case as follows:

```
<<< existing code, that will still be needed >>>
      // Read the section name.
      int string_leb_length = 0;
      uint32_t string_length =
          consume_u32v(&string_leb_length, "section name length");
      const byte* section_name_start = pc_;
      consume_bytes(string_length);
      if (failed()) {
        TRACE("Section name of length %u couldn't be read\n", string_length);
        break;
      }

<<< fast path for short strings >>>
      if (string_length == 4) {
	switch read_u32(section_name_start){
	  case ...:
	  case ...:
	  default: // Skip and ignore section.
	  }
      } else {
<<< optional slow path if handling larger strings >>>
```

I see no rationale for this change.

There is no use case for 'user-defined' sections, app developers are not expected to be bundling in arbitrary payloads in this manner and there is no way to even access them. The use case is web standards, even if experimental or de-facto.

titzer:
In the V8 native prototype (before 0xA), sections were section codes, and after discussion with others we all decided on section strings to reduce the probability that user extensions would collide either with each other or with experiments. But now there really is a fixed set of sections that need to be supported, so it is easier to just give them codes. Having "blessed" sections have codes and "experimental" or "user" sections have names makes the distinction explicit, and has the benefit of making engines slightly more efficient, as shown above. Standardization will assign further fixed codes as part of its normal process.

titzer:
On Wed, Aug 24, 2016 at 2:31 AM, JSStats <notifications@github.com> wrote:

> From the v8 code in module-decoder.cc the short names could address the
> trivial performance use case as follows:
>
> <<< existing code, that will still be needed >>>
>       // Read the section name.
>       int string_leb_length = 0;
>       uint32_t string_length =
>           consume_u32v(&string_leb_length, "section name length");
>       const byte* section_name_start = pc_;
>       consume_bytes(string_length);
>       if (failed()) {
>         TRACE("Section name of length %u couldn't be read\n", string_length);
>         break;
>       }
>
>
>
> <<< fast path for short strings >>>
>       if (string_length == 4) {
>     switch read_u32(section_name_start){
>
>
This just hides the bounds-checking and unaligned-read inefficiencies
inside the read_u32 code.


>
>       case ...:
>       case ...:
>       default: // Skip and ignore section.
>       }
>
> This switch won't be dense, so it won't get turned into a jump table by
the C++ compiler.

BTW, there's a bunch of other complexity in the V8 implementation that is
the result of strings, if you look at the module-decoder.h definition of
all those strings and all the unittests. It's vastly simpler with
single-byte section codes. I'm not sure making the strings shorter fixes
any of that.

>
>       } else {
> <<< optional slow path if handling larger strings >>>
>
> I see no rationale for this change.
>
> There is no use case for 'user-defined' sections, app developers are not
> expected to be bundling in arbitrary payloads in this manner and there is
> no way to even access them. The use case is web standards, even if
> experimental or de-facto.
>
Actually, it's clear that they will, because debugging/source-mapping
sections are not going to be standardized for the engine, and will vary
across languages.

JSStats:
Would there be consensus to move to short strings as a trial, and then re-evaluate the situation? So use fixed four byte strings for all the core section names. V8 could clean out some of the string complexity, simplify unit tests etc. Would not be much different than having a 32 bit section id code, less room for collisions. Or how about 1 byte strings taking it to the limit?

titzer:
While implementing this, I thought better of using the negative number trick and instead introduced a sentinel 0 byte for user-defined (string) section names. That avoids a little wart where one reads a byte from the stream, checks its value, and if it's not negative, "backs up" the stream 1 byte to use the normal varint32-parsing routine.

Thoughts?

titzer:
Also, technically a negative number could be extended with more all-1's upper bits, which would be another case to handle :-(

rossberg-chromium:
In that case, can't we simply specify user-defined sections as a regular section type with index `0`, whose name string must be the first component in the payload? (While the remaining payload is custom?) Then you can remove all special language about these user sections.

KarlSchimpf:
I for one, do not like what is being proposed here. It adds a BIG wart for
decompression. The rules are very complex and make it much harder to
implement a "general" decompression tool.

Since there are so few sections (when considering the size of a WASM file),
I'm surprised how much time is spent saving a few dozen bytes.

I also don't like that some argue that we should presume all "sections" are
already known. This makes integrating "decompression" algorithms more
problematic, in that the decompression algorithms have no logical place to
be (optionally) added.

Karl Schimpf

Karl Schimpf

BinaryEncoding.md
```diff
@@ -117,19 +117,19 @@ The module starts with a preamble of two fields:
...
+The section length and data immediately follow the section identification.
 All sections unknown to the WebAssembly implementation are ignored.
 A validation error in a user-defined section does not cause validation for the whole module to fail and is
 instead treated as if the section was absent.
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
-| id | `varint32` | section identifier code |
-| id_str | `bytes` | section identifier string, of length `max(id, 0)` bytes |
+| id | `uint8` | section code |
```
lukewagner: The change lgtm; this is actually what I was assuming we'd do in the first place.  Nit: can you say `varuint7` instead of `uint8`: it means practically the same thing but points out that this could (in an unlikely future) become variable length in the future.

BinaryEncoding.md
```diff
@@ -356,7 +355,7 @@ a `data_segment` is:
 User-defined section string: `"name"`
 
 The names section does not change execution semantics, and thus is not allocated
-a section opcode.
+a section code.
 Like all user-defined sections, a validation error in this section does not cause validation of the module to fail.
```
JSStats: If making a distinct class of 'user-defined' sections then this wording is no long appropriate:

'like all user-defined sections' -> 'like all unknown sections'

BinaryEncoding.md
```diff
@@ -117,19 +117,19 @@ The module starts with a preamble of two fields:
...
-Known sections have negative ids, while user-defined sections have positive ids that encode 
-the length of a string identifier immediately to follow.
-After the section identification, the section length and data follow.
+The module preamble is followed by a sequence of sections.
+Each section is identified by a 1-byte *section code* that encodes either a known section or a user-defined section.
+Known sections have non-zero ids, while user-defined sections have a zero id followed by a string identifier.
+The section length and data immediately follow the section identification.
 All sections unknown to the WebAssembly implementation are ignored.
 A validation error in a user-defined section does not cause validation for the whole module to fail and is
 instead treated as if the section was absent.
```
JSStats: A validation error in an 'unknown' section does not cause validation for the whole module to fail either, the 'user-defined' distinction here is not necessary.

titzer:
I've now removed references to "user-defined" sections, replacing them simply with "unknown".

I've also followed a suggestion by @rossberg-chromium to drop all mention of strings. Unknown sections simply have section code `0` and the contents are user-defined. By convention, the names section begins with "\x04name".

titzer:
On Wed, Aug 24, 2016 at 5:43 PM, KarlSchimpf <notifications@github.com>
wrote:

> I for one, do not like what is being proposed here. It adds a BIG wart for
> decompression. The rules are very complex and make it much harder to
> implement a "general" decompression tool.
>
> Since there are so few sections (when considering the size of a WASM file),
> I'm surprised how much time is spent saving a few dozen bytes.
>
> I also don't like that some argue that we should presume all "sections" are
> already known. This makes integrating "decompression" algorithms more
> problematic, in that the decompression algorithms have no logical place to
> be (optionally) added.
>
>
I'm not sure I understand this comment, now that I've removed references to
strings for identifying unknown sections. The identification of sections is
just a single byte, as opposed to a string before. How does that make
decompression more complicated?


> Karl Schimpf
>
> Karl Schimpf
>

JSStats:
Guess it's workable. Naming of the 'unknown' sections is deferred to a de-facto convention.

lukewagner:
@titzer By removing the designated "here's where the eventual variety of unknown sections distinguish themselves" field, it seems likely we'll eventually end up with more ad hoc conventions that end up causing conflicts or ambiguities.  If it makes it any cleaner to describe, could we perhaps have a separate "unknown section header" that goes inside the payload of unknown sections and that includes the string?

JSStats:
Yes, why not at least specify where an optional name goes in the 'unknown section header'. The length might be zero if there is none.

qwertie:
I guess I should have kept my mouth shut before. I like having section names as an official part of the spec and separate from the payload.

BinaryEncoding.md
```diff
@@ -117,43 +117,41 @@ The module starts with a preamble of two fields:
 | magic number | `uint32` |  Magic number `0x6d736100` (i.e., '\0asm') |
 | version | `uint32` | Version number, currently 10. The version for MVP will be reset to 1. |
 
-This preamble is followed by a sequence of sections. Each section is identified by an
-immediate string. Sections whose identity is unknown to the WebAssembly
-implementation are ignored and this is supported by including the size in bytes
-for all sections. The encoding of sections is structured as follows:
+The module preamble is followed by a sequence of sections.
+Each section is identified by a 1-byte *section code* that encodes either a known section or a user-defined section.
+Known sections have non-zero ids, while unkown sections simply have a zero id and are ignored by the WebAssembly implementation.
```
naturaltransformation: Nit: typo unkown -> unknown

BinaryEncoding.md
```diff
@@ -369,17 +347,19 @@ a `data_segment` is:
...
+The names section does not change execution semantics, and thus is not allocated a section code.
+It is encoded as an unknown section (id `0`) with the first few payload bytes identifying this section as a string.
+Like all unknown sections, a validation error in this section does not cause validation of the module to fail.
+The expectation is that, when a binary WebAssembly module is viewed in a browser or other development
 environment, the names in this section will be used as the names of functions
 and locals in the [text format](TextFormat.md).
 
 | Field | Type | Description |
 | ----- | ---- | ----------- |
+| name_length | `varint7` | length of the "name" string = 4 |
```
commonlisp: Do we expect all "unknown" sections begin with name_length and name_string or can an unknown section conceivably have a payload begin with something else? Maybe unknown is its own section category and needs to be defined and Name section is a subcategory. Admittedly, this may be overkill.

hemobo:
Just floating an older idea here, feel free to shoot it down if it doesn't apply :-)

Did you consider adapting the proposal by @mbebenita for this purpose? (#561)

That is, just using LEB128 encoded ASCII for section codes while using short codes for official sections?

The argument against that proposal was uniform treatment of strings, but now with the whole point being that engines don't have to treat section codes as strings anymore, that doesn't seem to be too much of a concern here.

This could mean using the 128 "singleton" bytes as one byte codes for "official" sections ("ASCII strings of one character", but without paying any mind to the "value" of that character, i. e. just numbering sections from 0 to 128) and longer LEB values, that is "ASCII strings of more than one character", for non-standard sections.

This seems to share the advantage of engines only having to recognize one byte section codes while custom sections are just those that start with “continuation bytes”.
 There would be no reading of string sizes in any case.

This also means that all possible (ASCII-) strings are available as non standard section codes, as long as they are at least two characters long.

From an encoder perspective they're all just LEB values, then. There's also having one big uniform name-space with low risk of collision between non-standard sections while maintaining a distinction between official and non-standard by means of the length of the section code.

Of course implementations wouldn't need to know anything about ASCII in any case, the string interpretation of custom section codes would be more like mnemonics for longer LEB values.

Optionally, the spec may also want to reserve two character section codes (that is two byte LEB values) for a distant future where there might be more than 128 official sections :-)
Non-standard codes would then simply be defined to start at three bytes.

If that day ever comes, engines could still get away with reading and comparing just two bytes.

naturaltransformation:
Is there consensus on this one?

jfbastien:
Considering that https://github.com/WebAssembly/webassembly.github.io/pull/14 has already landed I think the train has left the station for 0xC. I think this should be moved to 0xD.

sunfishcode:
@jfbastien The binary in https://github.com/WebAssembly/webassembly.github.io/pull/14 does include the change here ("Section opcodes").

jfbastien:
> @jfbastien The binary in WebAssembly/webassembly.github.io#14 does include the change here ("Section opcodes").

Ah OK, it does seem like that merge was a tad optimistic then :-)

We should merge this PR then.

lukewagner:
Ok, it looks like the remaining difference between this PR and what we discussed earlier is saying all user-defined sections have the string field.

titzer:
Back by popular demand: string identifiers for unknown sections.

lukewagner:
Thanks, lgtm

rossberg-chromium:
I thought we wanted to make the string part of the payload, so that user sections still have normal structure?

lukewagner:
Oh right, I forgot.  So that effectively just switches the order of the id string and payload length in the current PR.

lukewagner:
Also: one consequence of the current wording which I think is desirable and perhaps should be more-explicitly called out is that unknown-to-the-engine non-zero section codes will now cause a validation error; only user-defined sections will be silently ignored if unknown.

titzer:
On Wed, Sep 14, 2016 at 8:33 PM, rossberg-chromium <notifications@github.com
> wrote:

> I thought we wanted to make the string part of the payload, so that user
> sections still have normal structure?
>
That's the way that string section names were originally implemented, but
that was actually more complex to check for (in the case where you actually
want to parse the name, you have to check if the string length and string
are within the payload size).

rossberg-chromium:
Sure, but it was a regular case then, now it's become irregular, and thus a minus for easy processing -- especially for consumers that just want to skip all but a few selected sections (like tools).

titzer:
Righty-o. PTAL.

rossberg-chromium:
lgtm

lukewagner:
lgtm

PR #742, opened 2016-08-02
Remove `loop`'s result value(s).

sunfishcode:
Following up on discussion in https://github.com/WebAssembly/design/pull/741, this PR proposes removing loop's exit value. If branch arities are moved to `block` and `if` as proposed, this change eliminates the need for `loop` to also have an arity immediate, which would simplify a common validation strategy.

I expect this change to have little negative effect on code size or expressiveness, because `(loop x y)` could usually be written as `(loop x) y`, and the change would also eliminate an immediate field on every `loop`.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -355,17 +355,17 @@ further see the parallel, note that a `br` to a `block`'s label is functionally
 equivalent to a labeled `break` in high-level languages in that a `br` simply
 breaks out of a `block`.
 
-Branches that exit a `block`, `loop`, or `br_table` may take a subexpression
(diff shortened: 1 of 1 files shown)


titzer:
This mostly looks good to me, but I think it does depend on the outcome of #741. It's probably the case the value-returning loops is quite rare, and in that case one can surround the loop with a block that does return a value and branch to it.

lukewagner:
+1

sunfishcode:
The current evolution of [block signatures](https://github.com/WebAssembly/design/pull/765) has them applying to `loop` as well, making this change less important.

PR #743, opened 2016-08-02
Update text about signed offsets for memory_immediate

titzer:
Update text to allow signed memory offsets in the encoding but disallow them for validation.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -523,13 +523,14 @@ The `memory_immediate` type is encoded as follows:
 | Name | Type | Description |
 | ---- | ---- | ---- |
 | flags | `varuint32` | a bitfield which currently contains the alignment in the least significant bits, encoded as `log2(alignment)` |
-| offset | `varuint32` | the value of the offset |
+| offset | `varint32` | the value of the offset |
 
 As implied by the `log2(alignment)` encoding, the alignment must be a power of 2.
 As an additional validation criteria, the alignment must be less or equal to 
 natural alignment. The bits after the
 `log(memory-access-size)` least-significant bits must be set to 0. These bits are reserved for future use
 (e.g., for shared memory ordering requirements).
+The offset is encoded a signed `varint32`, but negative offsets are currently disallowed.
 
 ## Simple operators ([described here](AstSemantics.md#32-bit-integer-operators))
 


sunfishcode:
lgtm

sunfishcode:
Actually, using `varint32` means the value is limited to the signed int32 range, which would affect the semantics since it would no longer cover the high half of the uint32_t range.

The intent of [the original proposal](https://github.com/WebAssembly/design/issues/584) was to support the full uint32_t range, because, for example, it allows linkers to relocate arbitrary absolute addresses using the offset field, should they wish to.

sunfishcode:
The most straightforward way to describe the original proposal is to make this field be a `varint33` instead, which would be a new type of field which is similar to `varint32` but which can hold up to a 33-bit signed integer value.

titzer:
It being only varint32 is a restriction we can lift in the future. Is that
a reasonable limitation for MVP?

sunfishcode:
Restricting it varint32 would mean tools would loose the ability to represent arbitrary absolute addresses in the offset field, as they can now with varuint32.

It seems we are going to need to forego one of the following:
 - reserving the ability to support signed offsets in the future
 - representing arbitrary absolute addresses in the offset field
 - avoiding adding a new LEB decoding path

It's not clear to me which to let go.

titzer:
Since a 5-byte LEB actually stores 35 bits, if we back out of this proposal, there is still room to change it in the future if we need it (since we could easily redefine it as varint33, which allows all current encodings to continue to work).

titzer:
@sunfishcode Thoughts on backing this out, with the idea that it's currently future-proof for a later redefinition to varint33?

sunfishcode:
@titzer If we back out this proposal, we'll be at varuint32, which is unsigned, and doesn't leave room for generalizing to being signed in the future (if we restrict ourselves to stanard LEB encodings). For example, the bit pattern 0x7f is 127 if unsigned and -1 if signed. If we use the text in the PR here, we loose the high half of the uint32 range now.

That said, I'm currently leaning towards just dropping this idea, since it's a minor issue and there don't seem to be any easy answers.

titzer:
I'm going to abandon this PR as per above comment.

PR #744, opened 2016-08-03
Add signature-match check to JS.md to match Modules.md

lukewagner:
Modules.md#imports says that if an imported wasm function's signature doesn't match the import signature, an instantiation-time error is raised.  This PR adds this check explicitly to the description of importing in JS.md.


lukewagner:
Hah, looks like I forgot and re-found this in #775.

PR #754, opened 2016-08-08
Define realm for ToWebAssemblyValue

gahaas:
At the moment it is not clearly defined which realm should be used for ToWebAssemblyValue. I think the only reasonable options are the realm of the WebAssembly.Instance constructor, and the current realm context at the time the ToWebAssemblyValue is called. 

In this PR I propose to use the realm of the WebAssembly.Instance constructor because this would match the semantics of asm.js and the JavaScript to asm.js interface. In asm.js, type conversion of parameters happens within the asm.js function and not in the JavaScript function which calls the asm.js function.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -255,7 +255,7 @@ Exported Functions do not have the normal function internal slots but instead ha
 as well as the internal slots required of all builtin functions:
  * [[Prototype]] : [%FunctionPrototype%](http://tc39.github.io/ecma262/#sec-well-known-intrinsic-objects)
  * [[Extensible]] : `true`
- * [[Realm]] : the [current Realm Record](http://tc39.github.io/ecma262/#current-realm)
+ * [[Realm]] : the [current Realm Record](http://tc39.github.io/ecma262/#current-realm) at the time the Exported Function was instantiated
  * [[ScriptOrModule]] : [`GetActiveScriptOrModule`](http://tc39.github.io/ecma262/#sec-getactivescriptormodule)
 
 Exported Functions also have the following data properties:
@@ -266,7 +266,7 @@ WebAssembly Exported Functions have a `[[Call]](this, argValues)` method defined
  * Let `argTypes` be the list of value types defined by the signature of [[FunctionIndex]].
  * Let `args` be an empty list of coerced values.
  * For each value type `t` in `argTypes` and value `v` in `argValues`:
-   * Append [`ToWebAssemblyValue`](#towebassemblyvalue)`(v)` to `args`.
+   * Execute [`ToWebAssemblyValue`](#towebassemblyvalue)`(v)` in the [[Realm]] of the Exported Function and append the result to `args`.
  * Let `ret` be the result of calling [`Eval.invoke`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/eval.ml#L327)
    passing [[Instance]], [[FunctionIndex]], and `args`.
  * Return [`ToJSValue`](#tojsvalue)`(ret)`.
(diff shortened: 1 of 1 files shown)


lukewagner:
You're right that the intended real is the current realm.   However, `ToWebAssemblyValue` is also called as part of the definition of an exported function's `[[Call]]`, so explicitly naming "the `Instance` constructor" inside `ToWebAssemblyValue` isn't quite right.  Just saying "the *current Realm Record*" would be sufficient.  However, there isn't any explicit use of realms in `ToWebAssemblyValue`; only the spec-procedures called by `ToWebAssemblyValue` like `ToInt32` care about the current realm, and those should already reference the current Realm Record, so I'm not sure there's any need to say anything explicit about realm in `ToWebAssemblyValue` at all.

gahaas:
You are right that the realm only matters for spec-procedures like `ToInt32` or `ToNumber`. However, it is not clear to me at the moment where the code which calls ToInt32 or ToNumber actually belongs to. Assume an exported WebAssembly function `fun_wasm` which is called from within a JavaScript function `fun_js` with a parameter `obj`. Does the call to `obj.ToNumber()` belong to `fun_wasm`, or does it belong to `fun_js`, or is the realm of `fun_wasm` the same as the realm of `fun_js` and therefore it does not matter where the call to `obj.ToNumber()` belongs to? At the moment I don't see where the spec says that the realms of `fun_wasm` and `fun_js` are the same, which would mean that a WebAssembly instance is context-independent.
I think we should state clearly whether WebAssembly instances are context-independent or not. If they are not context independent, then I propose that `obj.ToNumber()` should belong to `fun_wasm`, because if `fun_wasm` were an asm.js function, then it would also belong to `fun_wasm`.

lukewagner:
This situation also arises today in the Web with host-defined/builtin-functions that can create objects (the situation is compounded when said builtin can be called directly from the event loop via `setTimeout`/`then`).  In all these cases, the precedent is that it is always the *callee function object* which determines the current global/realm, so I think that's what we should do here too.  You're absolutely right that this should be explicitly documented; I think the right place, though, is in the description of an exported function's [[Call]], not in `ToWebAssemblyValue`.

gahaas:
PTAL. I define the realm of an exported function more explicitly now, and I mention in the definition of `[[Call]]` of an exported function that ToWebAssemblyValue is executed in the realm of the exported function.

JS.md
```diff
@@ -266,7 +266,7 @@ WebAssembly Exported Functions have a `[[Call]](this, argValues)` method defined
  * Let `argTypes` be the list of value types defined by the signature of [[FunctionIndex]].
  * Let `args` be an empty list of coerced values.
  * For each value type `t` in `argTypes` and value `v` in `argValues`:
-   * Append [`ToWebAssemblyValue`](#towebassemblyvalue)`(v)` to `args`.
+   * Execute [`ToWebAssemblyValue`](#towebassemblyvalue)`(v)` in the [[Realm]] of the Exported Function and append the result to `args`.
```
lukewagner: Since `ToJSValue` technically also depends on the current realm (for the prototype of the possibly-thrown `Error` object), I think we should say that the entire method body executes in [[Realm]].  Looking at the spec more (specifically 9.3 and 9.3.1), this is actually mandated, so we don't have a choice here.  Perhaps we could add, as a first bullet: "As \[required of all builtin function objects\](https://tc39.github.io/ecma262/#sec-built-in-function-objects), this procedure executes in the [[Realm]] of the callee Exported Function."

gahaas:
I do not like it that we call Exported Functions "builtin functions", or is this an official term in the WebAssembly design repo? I have never seen Exported Functions as builtin functions, and the document only relates Exported Functions to Exotic Objects. Could we say "Similar to the [requirement for builtin functions](https://tc39.github.io/ecma262/#sec-built-in-function-objects), this procedure executes in the [[Realm]] of the callee Exported Function."?

lukewagner:
I could be mis-reading, but 9.3 gave me the impression that "builtin function" is simply the term for all functions that aren't written by content JS.  So JS stdlib functions, DOM functions, Web APIs, and, by extension, the WebAssembly functions in JS.md.  Notice also the [current language](https://github.com/WebAssembly/design/blob/master/JS.md#exported-function-exotic-objects) in JS.md which describe exported function exotic objects as builtin functions and I think had some amount of vetting by some JS spec-y people.

rossberg-chromium:
\begin{languagelawyer} The relevant definitions in the ES spec are in [Section 4.3.9](https://tc39.github.io/ecma262/#sec-built-in-object), which defines a built-in object as an "object specified and supplied by an ECMAScript implementation", and [Section 4.3.29](https://tc39.github.io/ecma262/#sec-built-in-function), defining a built-in function as a "built-in object that is a function".

So I agree with Andreas, we shouldn't be using that term. Wasm exports and everything else JS.md defines are random exotic objects. Sorry if I have missed this in earlier reviews.

lukewagner:
Ah, I think see what you are both getting at: I keep thinking of exported functions as going through some "call wasm" built-in function (b/c that's literally what we have internally, in the non-JIT-optimized case), but you're saying that there is no such intermediate built-in; we're exporting the wasm function *itself* to JS as some exotic thing that isn't built-in (b/c the user wasm code isn't built-in, obv).  That makes sense.

gahaas:
PTAL

lukewagner:
Perfect, thanks!

PR #756, opened 2016-08-10
Change WebAssembly.compile to throw on bad args

lukewagner:
This is a tiny tweak to the JS API to have `WebAssembly.compile` throw an eager exception, instead of returning a rejected `Promise`, if it is given invalid arguments.  I don't actually know what the general JS API precedent is here, but this seems abstractly a bit cleaner so that you only get a `Promise` if an async task was actually started.


yurydelendik:
Per https://www.w3.org/2001/tag/doc/promises-guide#always-return-promises : "promise-returning functions should never synchronously throw errors, since that would force duplicate error-handling logic on the consumer"

lukewagner:
Ah, that makes sense, thanks for pointing that out.

PR #763, opened 2016-08-16
Binary 0xc varuint size

titzer:
This PR limits the size of varint1, varint7, varint32, varuint32, and varint64 values in the binary encoding to appropriate limits.


titzer:
Should be merged into binary_0xc instead.

PR #764, opened 2016-08-16
Limit varint sizes in Binary Encoding.

titzer:
This PR limits the size of varint1, varint7, varint32, varuint32, and varint64 values in the binary encoding to appropriate limits.


rossberg-chromium:
LGTM. `varuintN` with N <= 7 is now really just a `uintN`, but okay.

lukewagner:
lgtm

PR #765, opened 2016-08-16
Block signatures

sunfishcode:
https://github.com/WebAssembly/design/pull/741 proposed moving the arity immediates from branches to block/if, which has the advantage of giving decoders more information up front, however the arities alone aren't enough for some use cases. This PR implements @MikeHolman's [suggestion](https://github.com/WebAssembly/design/pull/741#issuecomment-237885553) to provide full type information up front.

This proposes a new "block signature" type to the Type section, so `block`/`if` only need a single index immediate to specify their full type, even if they are extended to returning multiple values post-MVP.

This gives us a greater benefit than #741 did at approximately the same code size cost -- one immediate per `block` and `if`. And if code size of the immediates becomes a concern, we can introduce an `if0` (and similar opcodes for future block-like opcodes).

In my experiments, the decode time cost of indirecting into the type table to obtain the signature data was not significant.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -464,7 +464,6 @@ The `br_table` operator has an immediate operand which is encoded as follows:
 
 | Field | Type | Description |
 | ---- | ---- | ---- |
-| arity | `varuint1` | number of arguments |
 | target_count | `varuint32` | number of entries in the target_table |
 | target_table | `varuint32*` | target entries that indicate an outer block or loop to which to break |
 | default_target | `varuint32` | an outer block or loop to which to break in the default case |
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -58,11 +58,11 @@ A single-byte unsigned integer indicating a [value type](AstSemantics.md#types).
 
 ### `inline_signature_type`
 A single-byte unsigned integer indicating a signature. These types are encoded as:
-* `0` indicating a signature with 0 arguments and 0 results.
-* `1` indicating a signature with 0 arguments and 1 result of type `i32`.
-* `2` indicating a signature with 0 arguments and 1 result of type `i64`.
-* `3` indicating a signature with 0 arguments and 1 result of type `f32`.
-* `4` indicating a signature with 0 arguments and 1 result of type `f64`.
+* `0` indicating a signature with 0 results.
+* `1` indicating a signature with 1 result of type `i32`.
+* `2` indicating a signature with 1 result of type `i64`.
+* `3` indicating a signature with 1 result of type `f32`.
+* `4` indicating a signature with 1 result of type `f64`.
 
 ### `external_kind`
 A single-byte unsigned integer indicating the kind of definition being imported or defined:
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -56,6 +56,14 @@ A single-byte unsigned integer indicating a [value type](AstSemantics.md#types).
 * `3` indicating type `f32` 
 * `4` indicating type `f64`
 
+### `inline_signature_type`
+A single-byte unsigned integer indicating a signature. These types are encoded as:
+* `0` indicating a signature with 0 arguments and 0 results.
+* `1` indicating a signature with 0 arguments and 1 result of type `i32`.
+* `2` indicating a signature with 0 arguments and 1 result of type `i64`.
+* `3` indicating a signature with 0 arguments and 1 result of type `f32`.
+* `4` indicating a signature with 0 arguments and 1 result of type `f64`.
+
 ### `external_kind`
 A single-byte unsigned integer indicating the kind of definition being imported or defined:
 * `0` indicating a `Function` [import](Modules.md#imports) or [definition](Modules.md#function-and-code-sections)
@@ -436,20 +444,21 @@ It is legal to have several entries with the same type.
 | Name | Opcode | Immediates | Description |
 | ---- | ---- | ---- | ---- |
 | `unreachable` | `0x00` | | trap immediately |
-| `block` | `0x01` |  | begin a sequence of expressions, the last of which yields a value |
-| `loop` | `0x02` |  | begin a block which can also form control flow loops |
-| `if` | `0x03` | | begin if expression |
+| `block` | `0x01` | sig : `inline_signature_type` | begin a sequence of expressions, yielding 0 or 1 values |
+| `loop` | `0x02` |  sig : `inline_signature_type` | begin a block which can also form control flow loops |
+| `if` | `0x03` | sig : `inline_signature_type` | begin if expression |
 | `else` | `0x04` | | begin else expression of if |
 | `select` | `0x05` | | select one of two values based on condition |
-| `br` | `0x06` | arity : `varuint1`, relative_depth : `varuint32` | break that targets an outer nested block |
-| `br_if` | `0x07` | arity : `varuint1`, relative_depth : `varuint32` | conditional break that targets an outer nested block |
+| `br` | `0x06` | relative_depth : `varuint32` | break that targets an outer nested block |
+| `br_if` | `0x07` | relative_depth : `varuint32` | conditional break that targets an outer nested block |
 | `br_table` | `0x08` | see below | branch table control flow construct |
 | `return` | `0x09` | return zero or one value from this function |
 | `drop` | `0x0b` | | ignore value |
 | `nop` | `0x0a` | | no operation |
 | `end` | `0x0f` | | end a block, loop, or if |
 
-The counts following the break operators specify how many operands are taken as transfer arguments; in the MVP, all these values must be either 0 or 1.
+The _sig_ fields of `block` and `if` operators specify function signatures
+which describe their use of the operand stack.
 
 The `br_table` operator has an immediate operand which is encoded as follows:
 


lukewagner:
What I like about this change is:
* Removes the only bit of "inference" from type checking, better matching the rest of wasm.
* Removes the need for an internal "unknown" type state for blocks during validation.
* For the same reason that a single-pass compiler would want to know the arity a priori, the type can be useful, e.g., to determine the register class of the block result.

MikeHolman:
lgtm, this is exactly what I wanted!

JSStats:
Could some of the common high frequency block signatures be given specified indexes. For example could block signature index 0 be zero result values, and could there be specified indexes for the single value core types, so index 1 be (i32), index 2 be (i64), etc. Further could it be invalid to have a duplicate signature in this table, to force a canonical encoding.

This might help the text format a lot because there would be canonical encodings for the common cases, and the text format might then not even need to encode the signature index in most cases and might still derive the signature from the text syntax.

JSStats:
If the signatures could also be required to be in a sorted order this might be even better. Then there is only one canonical table for a set of signatures and the validator can check this by simply checking that they are in order.

titzer:
Maybe there is an encoding trick that we could use to obviate the need for signature entries until the multi-value era. E.g. what if the immediate is an inline value type (i.e. void, i32, i64, f32, f64), with one encoding being an index into the type table?

JSStats:
If indexes are predefined and specified for the block types used in the MVP then the table will not be needed for the MVP, and there will be a canonical index for them. Perhaps consider sorting these in an order that will scale to multiple-values.

rossberg-chromium:
Ben and I have discussed the design space more extensively. There are 3 main proposals on the table now. Here are the pros and cons as we see them:

1. Status quo: arity annotations on branches.
  Pros:
  - No change to text format or tests required.
  - No change to implementations required.
  - Avoids forcing annotations on all future blockish constructs.
  - Avoids redundant information (e.g. on blockish constructs not using branches).
  - Probably the most size-efficient solution without additional measures.

  Cons:
  - Knowing the type of a block is deferred until the first exit point.
  - With multi-values, potential deferred allocation in decoder.

2. Proposed change #741: arity annotations on blocks
  Pros:
  - With multi-values, no deferred allocation in decoder.

  Cons:
  - Requires changes to implementations.
  - Requires changing text format and many tests.
  - Forces annotations on all future blockish constructs.

3. This proposal #765: type annotations on blocks
  Pro:
  - All types are known upfront.
  - For multi-values, no deferred allocation in decoder.

  Cons:
  - Requires changing implementations.
  - Requires extending type section?
  - Requires changing text format and many tests.
  - Forces annotations on all future blockish constructs.
  - Forces any code transformation tool to derive types if it wants to insert a blockish.
  - Least size-efficient without additional measures.

Overall, Ben and I still think option (1) remains the most attractive. We have implemented it, understand its pros and cons best, and don't need to worry about unforeseen implications late in the game. The trade-offs of the others are less obvious, and their advantages appear small (and biased towards consumers).

rossberg-chromium:
I'd also like to point out one further advantage of the status quo that I just ran into: for the linear text format (like will be visible in a debugger) having arities on branches is somewhat more user-friendly, because it is directly apparent how many values a branch operator consumes -- no need to search for the target (or rely on tooling). The information is where it belongs, that is, where it is used.

lukewagner:
> Requires changing implementations.

If the changes are quite simple, I don't think the bullet "Requires changing implementations" should be considered a con given that we are in the middle of changing this all anyway for 0xc.

> Requires extending type section?

@titzer's idea above seems fine as a way to avoid the indirection, but again the cost here is low, especially since we know we're adding more forms to the type section anyway (that's the point of the `form` field).

> Requires changing text format and many tests.

If there was a sixth, lowest [consituency](http://dev.w3.org/html5/html-design-principles/#priority-of-constituencies), I think this would be it.

> Forces annotations on all future blockish constructs.

I can't see any reason why we wouldn't want that for the same reason we want them for the current set of block constructs.  Also, other than, it sounds like, a try/finally block for which the type annotations seem fine, are we really anticipating a bunch more blocks?

> Forces any code transformation tool to derive types if it wants to insert a blockish.

Are there any concrete examples of such a transformation that is valuable and can't be bothered to understand types?  We put types on practically everything else, it seems hard to get away with being type-oblivious for anything non-trivial.

> Least size-efficient without additional measures.

How is this any bigger than adding arity?

lukewagner:
@rossberg-chromium Branch-with-value / block-with-signature is no different than call / function-signature.  For both, devtools could provide tooltips, highlighted, etc to make the information more immediate.

qwertie:
For the text format, there may be as much advantage as disadvantage: to the reader there may be some value in seeing up front what type a block will produce, even when there are no branches out. And type inference can be supported for anyone writing wasm, just as we're looking at supporting infix `+`.

JSStats:
@rossberg-chromium One key point, and perhaps even a show stopper, seems to be missing in the list of pros and cons:

We still need arity annotations on the fall-though for multi-value support. The strategy of just returning all the values remaining on the block at the end of the block will significantly frustrate code using multiple values, or with `pick` just making multiple use of values on the stack. With the status-quo it would require writing values to locals and reloading just those needed to clean up the stack for the fall-through. I think this is a show-stopper, so I strongly support annotating blocks with at least the number of values.

JSStats:
@qwertie I agree that some text formats might want to annotate the block with their return type, but it would only frustrate the text format if they were require to annotate them all with the index too which would be required to be lossless. Thus my suggesting that they be required to be in sorted order so there is always a canonical order and so the text format does not need to annotate them with the signature index too even if it annotates them with the type.

JSStats:
@rossberg-chromium I agree that a text format will be more familiar and readable if the number of result values are on the branches and even using `values` on the fall-through, but I still fully expect that a text format can do this with the change to putting the type on the block if it can be a lossless transform. Text format tool will need to be type-aware in order to deal with the stack machine encoding, and it seems consistent with the wasm approach to move some burden to tools and off the runtime.

rossberg-chromium:
On 17 August 2016 at 18:09, Luke Wagner <notifications@github.com> wrote:

> Requires changing implementations.
>
> If the changes are quite simple, I don't think the bullet "Requires
> changing implementations" should be considered a con given that we are in
> the middle of changing this all anyway for 0xc.
>
Hm, because there is other work to do it doesn't matter to produce more?

Requires extending type section?
>
> @titzer <https://github.com/titzer>'s idea above seems fine as a way to
> avoid the indirection, but again the cost here is low, especially since we
> know we're adding more forms to the type section anyway (that's the point
> of the form field).
>

It's not a proper type, though, so would represent an anomaly in the type
section.

Requires changing text format and many tests.
>
> If there was a sixth, lowest consituency
> <http://dev.w3.org/html5/html-design-principles/#priority-of-constituencies>,
> I think this would be it.
>

Since you cite that hierarchy it's worth noting that it also puts "users"
over "implementers". In our case, "users" = "producers", while we keep
arguing for design changes motivated the other way round, including with
the current proposal. ;)

Also, the text format actually is user-facing.

Forces annotations on all future blockish constructs.
>
> I can't see any reason why we wouldn't want that for the same reason we
> want them for the current set of block constructs. Also, other than, it
> sounds like, a try/finally block for which the type annotations seem fine,
> are we really anticipating a bunch more blocks?
>

It's extra redundancy, and it puts the cost on the wrong side: the
annotation is only needed for branches, but now you have to include it
everywhere even if there isn't a single branch around. That seems silly for
constructs like `if` and `try` in particular, for which not using their
label is most likely the 99% use case.


> Forces any code transformation tool to derive types if it wants to insert
> a blockish.
>
> Are there any concrete examples of such a transformation that is valuable
> and can't be bothered to understand types? We put types on practically
> everything else, it seems hard to get away with being type-oblivious for
> anything non-trivial.
>

No, that's not right. We don't have type annotations on any generic
operator right now. Even monomorphic operators only explicate their type as
a textual naming convention which has no actual representation in the
binary format. The only type annotations we currently have are on function
boundaries.

There are e.g. many kind of peephole optimisations that don't require
understanding types. Or various instances of partial evaluation. I can't
judge how relevant they are, but stuff like that is done for other bytecode
formats.

Least size-efficient without additional measures.
>
> How is this any bigger than adding arity?
>
Depends on the exact solution, but you either need extra entries in the
type section, or inline type vectors. In both cases, the size is generally
>1 without extra encoding hacks, and obviously linear with multi-values.

lukewagner:
> Hm, because there is other work to do it doesn't matter to produce more?

I'm saying it's in the area that is being changed anyway, so it's a change one way or another.  Anyhow, it's very little work, mostly involves simplifying existing code, so this shouldn't be a deterrent regardless.

> It's not a proper type, though, so would represent an anomaly in the type section.

Not every type in the types section will be able to show up in every context where an index into the types section is specified so I don't think it makes sense to say it's not a "proper type".  It's not like we're representing some cartesian closed category here ;)

> Forces annotations on all future blockish constructs.

Agreed and desired, for the same reason as adding block signatures in the first place.

> It's extra redundancy, and it puts the cost on the wrong side: the annotation is only needed for branches,
> but now you have to include it everywhere even if there isn't a single branch around. That seems silly for
> constructs like `if` and `try` in particular, for which not using their label is most likely the 99% use case.

If size is an actual concern, we can have the specialized opcodes.  But I expect layer 1 compression would obviate the need.

>>  We put types on practically everything else, it seems hard to get away with being type-oblivious for anything non-trivial.
>
> No, that's not right. We don't have type annotations on any generic operator right now.

What I mean is that the vast majority of ops are not generic.

> There are e.g. many kind of peephole optimisations that don't require understanding types. Or various
> instances of partial evaluation. I can't judge how relevant they are, but stuff like that is done for other
> bytecode formats.

Possibly, but I don't think we should constrain ourselves based on this level of speculation.

> Also, the text format actually is user-facing.

Since we're aiming at low-level clarity, not brevity, nor optimizing ease of writing by hand, with the current linear text format, I think it's actually useful to have the types returned by a block listed up front (just like having local types listed up front).  For the experimental sugared syntaxes, it seems natural to drop these annotations when they are implied by the types of branches/fallthrough.

> Depends on the exact solution, but you either need extra entries in the
> type section, or inline type vectors. In both cases, the size is generally
> without extra encoding hacks, and obviously linear with multi-values.

Let's not consider inlining the entire type vector into each block.  A few bytes of type section entries are negligible (and avoidable in the MVP).  That leaves the size effectively equivalent to arities-on-blocks.  We can measure the overall delta, though.

titzer:
On Thu, Aug 18, 2016 at 5:28 PM, Luke Wagner <notifications@github.com>
wrote:

> Hm, because there is other work to do it doesn't matter to produce more?
>
> I'm saying it's in the area that is being changed anyway, so it's a change
> one way or another. Anyhow, it's very little work, mostly involves
> simplifying existing code, so this shouldn't be a deterrent regardless.
>
After giving this more thought, I think there is no real advantage to
having arities only on blockish things versus having type annotations on
blockish things.

So one alternative, arities-only, seems to be out.

That leaves type-annotated blockish things (our set is now block, if,
try_finally, try_catch, and try_catch_finally--since we are prototyping
these), or backing off this change and keeping arities on branches.

I'm in the middle of implementing block type annotations throughout V8 on
top of the stack machine branch, and the changes aren't trivial. Overall I
haven't seen any code savings yet, either. Instead it makes decoding of
blocks more complex and there are several gotcha special cases. The
simplification of inference wasn't dramatic. I'm implementing the full
multi-value semantics (on top of the multi-value semantics I already
implemented). I'll report more when I am finished.

> It's not a proper type, though, so would represent an anomaly in the type
> section.
>
> Not every type in the types section will be able to show up in every
> context where an index into the types section is specified so I don't think
> it makes sense to say it's not a "proper type". It's not like we're
> representing some cartesian closed category here ;)
>
> Forces annotations on all future blockish constructs.
>
> Agreed and desired, for the same reason as adding block signatures in the
> first place.
>
> It's extra redundancy, and it puts the cost on the wrong side: the
> annotation is only needed for branches,
> but now you have to include it everywhere even if there isn't a single
> branch around. That seems silly for
> constructs like if and try in particular, for which not using their label
> is most likely the 99% use case.
>
> If size is an actual concern, we can have the specialized opcodes. But I
> expect layer 1 compression would obviate the need.
>
> We put types on practically everything else, it seems hard to get away
> with being type-oblivious for anything non-trivial.
>
> No, that's not right. We don't have type annotations on any generic
> operator right now.
>
> What I mean is that the vast majority of ops are not generic.
>
> There are e.g. many kind of peephole optimisations that don't require
> understanding types. Or various
> instances of partial evaluation. I can't judge how relevant they are, but
> stuff like that is done for other
> bytecode formats.
>
> Possibly, but I don't think we should constrain ourselves based on this
> level of speculation.
>
> Also, the text format actually is user-facing.
>
> Since we're aiming at low-level clarity, not brevity, nor optimizing ease
> of writing by hand, with the current linear text format, I think it's
> actually useful to have the types returned by a block listed up front (just
> like having local types listed up front). For the experimental sugared
> syntaxes, it seems natural to drop these annotations when they are implied
> by the types of branches/fallthrough.
>
Actually I am not sure what we are aiming at here. I admit that part of me
was enticed by the idea of all blocks being typed, but I am not
experiencing the simplification in the decoder yet. I'm in the middle of
changing over all of our internal tests for these experiments, and that's
an annoying amount of work. It will probably be the same for the spec
tests. I'd like to finish my implementation to say for sure, since it seems
we've bitten ourselves a couple times by making changes before they're
fully thought out.

> Depends on the exact solution, but you either need extra entries in the
> type section, or inline type vectors. In both cases, the size is generally
> without extra encoding hacks, and obviously linear with multi-values.
>
> Let's not consider inlining the entire type vector into each block. A few
> bytes of type section entries are negligible (and avoidable in the MVP).
> That leaves the size effectively equivalent to arities-on-blocks. We can
> measure the overall delta, though.
>
I think requiring a types section just to use typed blocks (even if it's
only for multi-values) is a no-go.

There are a couple of negatives that I already ran into, e.g. that one
doesn't know how many values a br* will pop without looking at the block;
so the control stack in the interpreter has to maintain that arity, too.

Other weird things are that with multi-values, br_table is actually
polymorphic; it would pop a different number of values depending on which
case was targeted. We'd have to disallow some (weird) legal cases and
require that the arity of all blocks targeted in a br_table match.

Before there was a nice bottleneck in the decoder where all merge-y
operations went through, and that was the natural place for both the
inference and checking step. Now that bottleneck must serve two masters;
fallthroughs have to be an exact match, while branches can have extra stuff
in the middle of the stack.

lukewagner:
> I'll report more when I am finished.

Great, and thanks for the experimentation.  We're also experimenting with how this works so we'll be able to compare notes in a bit.

> I think requiring a types section just to use typed blocks (even if it's only for multi-values) is a no-go.

It's not a types section, it'd just be a new `form` of entry in the existing types section.  Especially with GC types, there will be quite a few new `form`s so if the objection is the annoyance of going from 1->2 (which does turn the associated internal array from homogeneous to heterogeneous), then that will happen regardless with the current design and trajectory.

JSStats:
Re: 'br_table is actually polymorphic', it seems appropriate to 'require that the arity of all blocks targeted in a br_table match.'

Perhaps it is a the little extra work for the interpreter, to look at the target block's number of required values rather than for this to be an immediate argument, but this would have been necessary anyway for the fall-through, or the `end` opcode would have needed a count too.

I see problems if the index space of the block types is shared with all other types, it would make having a canonical order even more problematic, and it might have a negative impact on code compression, so a separate table seems needed.

Cellule:
For the type section, we could share the same type as Function types, but with no params. That way we could share an entry for blocks and functions if they happen to have the same signature.
The only I see to have a different form is to avoid the `param_count` field, but then again, if a function already use a signature with no params then we could simply reuse it.

> I'm in the middle of implementing block type annotations throughout V8 on
top of the stack machine branch, and the changes aren't trivial. Overall I
haven't seen any code savings yet, either. Instead it makes decoding of
blocks more complex and there are several gotcha special cases.

I don't really understand how having the signature ahead of time would make decoding more complicated.
Right now, in Chakra, we already have to keep track of the type of the block for br yielding values. Plus, when we enter the block, because we don't know the type nor how many (0 or 1 right now) values might be yielded, we have to pre-reserve a space for that possible value.
I don't even know, with our current model, how we'll handle multi-value yields.
Having the signature would definitely make that easier.

For the binary space, one could argue that if we keep arity on br then 1 block with multiple branches would have to spend 1 byte on every branches where as having that byte on the block would cost 1 byte for all the branches. Plus, if we really want to save more space, we could have the *0 opcodes to means there are no values.

titzer:
On Thu, Aug 18, 2016 at 7:12 PM, Luke Wagner <notifications@github.com>
wrote:

> I'll report more when I am finished.
>
> Great, and thanks for the experimentation. We're also experimenting with
> how this works so we'll be able to compare notes in a bit.
>
> I think requiring a types section just to use typed blocks (even if it's
> only for multi-values) is a no-go.
>
> It's not a types section, it'd just be a new form of entry in the
> existing types section. Especially with GC types, there will be quite a few
> new forms so if the objection is the annoyance of going from 1->2 (which
> does turn the associated internal array from homogeneous to heterogeneous),
> then that will happen regardless with the current design and trajectory.
>
No, I understand that part. I just don't think it makes sense to require a
types section entry for doing local control flow, even if it would just be
for multi-value blocks. It makes more sense to have the block type inline
in all cases, thus not requiring a non-local section for what is something
completely local.

lukewagner:
@titzer Ok; it'd just be a size optimization, similar to not including the signature of a `call_indirect` inline (esp. now that they are structurally-matched and thus the particular index has no meaning).  But if we did what you're suggesting for the MVP, we could make that decision post-MVP.

rossberg-chromium:
@Cellule, the size argument you are making is exactly the one I was making when I originally proposed this. However, it is invalidated by two things: (1) `br_table` which is a single branch to many blocks (IIRC, we have some with 100K targets), and more importantly, (2) `if`, which typically has no branches. When I proposed the change I still assumed we would remove labels from `if`, so that it wouldn't affect the equation (IIRC, AngryBots has like 140K uses of `if`).

However, if we want to go this way (I'd still prefer we don't) I like the suggestion to reuse the function signatures. That may in fact be more future-proof, because we might want to allow block operators to also _consume_ values from the stack eventually.

rossberg-chromium:
@lukewagner, getting a bit OT, but ever since we moved to structural types, and after looking at the design space for managed types, I have wondered about the distinction we are making between inline types and defined types. Once you add more type structure, it becomes increasingly ad-hoc.

Is there any particular reason left why we cannot make it uniform? That is, you can specify _all_ types inline, but you can also name _all_ types in the type section? Then the type section would just be a means of abbreviating common type expressions to save space. We could use the positive/negative index space for ids vs constructors that we discussed at some point to keep the encoding maximally compact (and fast).

lukewagner:
@rossberg-chromium Actually, function signatures may be the right thing here: in the MVP, input arity would be constrained to be 0, but post-MVP, we were realizing (as were you, in a recent email :) that there is no reason that blocks can't be given arguments in a consistent way (representing values on the stack that are popped by the block).  Arguments even make sense for `loop` (the values of a branch would be the loop's arguments), which has a nice sort of symmetry.

titzer:
Just some numbers for context from the AngryBots.wasm and BananaBread. Removing the arity from branches and introducing type signatures (typically one byte inline as I described in a previous comment), we end up with:

angrybots                  bananabread
--------------------------------------------
size         12072057 bytes         2444163 bytes

br             -40097           -10313
br_if          -16267            -6066
br_table        -2093             -852

block          +148224          +36292
if             +163429          +41173
loop            +17490           +5662
---------------------------------------
net            +270686          +65896
                  2.2%            2.7%

(apologies for the formatting)

AndrewScheidecker:
I'm in favor of declaring a signature for blocks up front for the simplification in validator state.

> Just some numbers for context from the AngryBots.wasm and BananaBread.

There may be a bunch of superfluous blocks in these test cases, since if this is just moving arity from `br` to `block` you would expect the size(excluding `if`) to only increase if there are blocks without corresponding branches to them. I can't quantify it, but I do see these superfluous blocks in binaryen output.

> Actually, function signatures may be the right thing here: in the MVP, input arity would be constrained to be 0, but post-MVP, we were realizing (as were you, in a recent email :) that there is no reason that blocks can't be given arguments in a consistent way (representing values on the stack that are popped by the block). Arguments even make sense for loop (the values of a branch would be the loop's arguments), which has a nice sort of symmetry.

I like this idea (I would use it for loops), but it makes me wonder: is there a good reason for function arguments to remain local variables, or should they just be implicitly pushed onto the operand stack on function entry?

titzer:
After having implemented this, I've come around to this idea. If we can move the block types inline for the common cases of void, i32, i64, f32, and f64, then this will LGTM.

AndrewScheidecker:
Thinking about https://github.com/WebAssembly/design/issues/778, I realized that with block signatures, `end` can have the same effect on the stack as `br 0`, which means you can save the drops to clean up values that aren't yielded by a block.

JSStats:
@AndrewScheidecker Right, so we are back to the old (current) block semantics, the fall through unwinds the stack, and some values can remain on the block value stack which might avoid some uses of `drop` and also enables `pick` style block stack constants.

sunfishcode:
@AndrewScheidecker That's true. There will be a temptation to use `br 0` as a way of avoiding the requirement to use `drop`. This highlights a question about explicit `drop`: Do we really want to require it at `end`? Or, if we really do mean for people to `drop` unused values, perhaps we should require it at `br` too. (If we do this, it dovetails well with #778.)

rossberg-chromium:
Keep in mind that the design of blocks and loops is intentionally such that they don't have any operational effect by themselves: their only purpose is to serve as a target for branches. For example, unused blocks can be inserted or removed freely with no effect (other than shifting the label space). That keeps their semantics simple and easy, and nicely separated from other concerns.

JSStats:
@rossberg-chromium Can't both allow block starts to be anywhere and have breaks unwind. Having blocks unwind matches the restriction inherent in the text format lexical blocks so at this point I would strongly object to removing the 'unwind' semantics of blocks and it seems a key feature of the wasm design, something we are doing 'better' differently to CIL and may other VMs. But it does mean that the following a very different:

```
block<1>
  i32.const 1
  i32.neg
end
```

versus

```
i32.const 1
block<1>
  i32.neg  ; Can't pop its argument.
end
```

Even with the suggested `drop` restrictions, the block start position is important.

@sunfishcode Yeh, the `drop` plan was a dead end, drop it.

rossberg-chromium:
@JSStats, I didn't say that you can insert blocks _anywhere_. That was never the case, regardless of `drop`.

That being said, there is the idea that block annotations should be function types, in which case they could later be allowed to consume stack, like e.g. in your example:
```
i32.const 1
block (param i32) (result i32)
  i32.neg
end
```
This may be useful for macrofication and other things.

sunfishcode:
Per feedback, this PR is now updated to use inline signature immediates for the common cases of void, i32, i64, f32, and f64.

BinaryEncoding.md
```diff
@@ -436,20 +444,21 @@ It is legal to have several entries with the same type.
 | Name | Opcode | Immediates | Description |
 | ---- | ---- | ---- | ---- |
 | `unreachable` | `0x00` | | trap immediately |
-| `block` | `0x01` |  | begin a sequence of expressions, the last of which yields a value |
-| `loop` | `0x02` |  | begin a block which can also form control flow loops |
-| `if` | `0x03` | | begin if expression |
+| `block` | `0x01` | sig : `inline_signature_type` | begin a sequence of expressions, yielding 0 or 1 values |
```
JSStats: There does not appear to be any compelling reason to restrict blocks to returning only '0 or 1 values` in the MVP. It is probably a small matter for the runtimes to be able to return multiple values here? But don't let that hold this up, can always revisit how it is going.

Also no mention if a function signature with arguments is usable in the MVP? Also seems a small matter, just not sure if it is really necessary yet.

BinaryEncoding.md
```diff
@@ -436,20 +444,21 @@ It is legal to have several entries with the same type.
...
 | `return` | `0x09` | return zero or one value from this function |
 | `drop` | `0x0b` | | ignore value |
 | `nop` | `0x0a` | | no operation |
 | `end` | `0x0f` | | end a block, loop, or if |
 
-The counts following the break operators specify how many operands are taken as transfer arguments; in the MVP, all these values must be either 0 or 1.
+The _sig_ fields of `block` and `if` operators specify function signatures
+which describe their use of the operand stack.
 
 The `br_table` operator has an immediate operand which is encoded as follows:
```
rossberg-chromium: The `br_table` immediate needs the arity removed, too.

rossberg-chromium:
Yeah, it's still not clear to me how we gonna future-proof this encoding for multiple values etc. But I guess we can still rectify that in 0xd. So to move forward for now: lgtm modulo br_table oversight.

sunfishcode:
Fixed the `br_table` oversight.

titzer:
The actual change here looks good to me, but can we simplify the text to leave out the discussion of arguments? We can always add that later when it becomes relevant.

BinaryEncoding.md
```diff
@@ -56,6 +56,14 @@ A single-byte unsigned integer indicating a [value type](AstSemantics.md#types).
 * `3` indicating type `f32` 
 * `4` indicating type `f64`
 
+### `inline_signature_type`
+A single-byte unsigned integer indicating a signature. These types are encoded as:
```
titzer: This will be simpler to explain in terms of a single result time, since the signatures won't be relevant until some future post-MVP version.

titzer:
lgtm

sunfishcode:
Merging with lgtms above.

PR #766, opened 2016-08-17
Binary 0xc

kmiller68:
What's the reason to have JS min/max? How would anyone ever generate these operations other than writing the binary by hand? Even if they did, what percentage of users would even use such an instruction? AFAIK, no C++ transpiler will produce this operation since it doesn't match what std::min  does.


sunfishcode:
The semantics are also those of Java's min/max. Also, these semantics are available in hardware on ARM and Power. Also, popular C++ compilers on ARM and Power, and now also [WebAssembly](https://github.com/llvm-mirror/llvm/blob/master/lib/Target/WebAssembly/WebAssemblyInstrFloat.td#L30), do utilize these instructions, when one of the operands of the min/max is constant, when one of the operands is known to be not NaN or zero, and/or when -ffast-math or similar is present.

Admittedly, the case for min/max isn't as strong as it is for many other instructions, but I think the language is stronger with them than without them.

(As an aside, "transpiler" means a compiler whose output is at close to the same abstraction level as its input, which is usually not the case for C++ compilers, or for compilers targeting WebAssembly.)

kmiller68:
I see. I wasn't aware that ARM supported this instruction. In that case I don't think it's unreasonable to support min/max.

PR #767, opened 2016-08-17
Clarify `TypeError` with a link to the JS spec.

gahaas:
At the moment it is not specified what a TypeError is, and I think the intention is right now that a JavaScript TypeError is thrown. This CL links every occurrence of TypeError to the JavaScript spec.


rossberg-chromium:
lgtm

lukewagner:
thanks, lgtm

PR #770, opened 2016-08-18
Evaluation order of parameters of exports.

gahaas:
This PR explicitly specifies the order in which ToWebAssemblyValue is called on the parameters of an Exported Function. Additionally it removes the definition of `argTypes`, which is not used at the moment.


lukewagner:
Oops, left some dead-prose there :)  lgtm!

PR #772, opened 2016-08-19
Make NaN handling more consistent.

gahaas:
This PR makes the handling of NaNs in the  promotion/demotion of floats more consistent to the handling of NaNs in arithmetic operations. Similar to #716, this PR requires that canonical NaNs have to be preserved by promotion/demotion. whereas the promotion/demotion of non-canonical NaNs can result in an arbitrary quiet NaN.


sunfishcode:
lgtm. This is indeed the intention of #716.

PR #779, opened 2016-08-29
Move the element section before the code section

lukewagner:
This PR changes the section order so that the elements section appears before the code section.  This is useful because it allows a streaming engine to know, before compiling function bodies:
 1. which functions are callable via `call_indirect` (which can affect prologue code generation)
 2. that `call_indirect` always stays within the same instance (in particular, that none of the elements that are used to initialize the table are imports)

(I think the data section still belongs after the code section b/c it can be quite large and having it later allows compilation to proceed in parallel with download of the data.)


titzer:
lgtm

PR #775, opened 2016-08-26
Tighten checks in JS.md for imported and table functions

lukewagner:
This PR fills in some missing checks in the JS API that we've discussed informally but never got recorded.  It's possible we'll want to relax the restriction on not being able to assign non-exported-wasm functions to Tables post-MVP, but for now it reduces the MVP impl burden of Tables while still supporting the core use case of dynamic linking.

PR #780, opened 2016-08-30
Move segment and start operations to the end of instantiation in JS.md

lukewagner:
The way JS.md is currently written, failures can occur after the Elements and/or Data segments have been written to globally-visible (imported) tables/memories and/or after `start` is called.  This sort of partial initialization seems undesirable so this PR moves all the globally-visible operations to the very end so failure makes more sense and you can't end up with partially-written segments.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -216,7 +216,8 @@ has been modified so that each export simply has a `name`, `type` and `index`:
     reusing an existing object if one exists for the given function definition,
     otherwise creating a new object given `instance` and `index`.
   * If the `type` is global:
-    * Assert: the global is immutable by MVP validation constraint.
+    * [Assert](https://tc39.github.io/ecma262/#assert): the global is immutable
+      by MVP validation constraint.
     * Let `v` be the global variable's initialized value.
     * Export [`ToJSValue`](#tojsvalue)`(v)`.
   * If the `type` is memory, then export a `WebAssembly.Memory` object, reusing
--- a/JS.md
+++ b/JS.md
@@ -247,7 +247,7 @@ Let `instanceObject` be a new `WebAssembly.Instance` object setting
 `[[Instance]]` to `instance` and `exports` to `moduleNamespace`.
 
 If any of the elements of an [Elements section](Modules.md#elements-section)
-refer to an imported function which is not an
+refers to an imported function which is not an
 [Exported Function Exotic Object](#exported-function-exotic-objects), throw a
 [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror).
 
@@ -260,8 +260,8 @@ Apply all Data and Element segments to their respective Memory or Table in the
 order in which they appear in the module. Segments may overlap and, if they do,
 the final value is the last value written in order. Note: there should be no
 errors possible that would cause this operation to fail partway through. After
-this operation completes, elements of `instance` may are now visible and callable
-through imported Tables, even if `start` fails.
+this operation completes, elements of `instance` are visible and callable
+through [imported Tables](Modules.md#imports), even if `start` fails.
 
 If a [`start`](Modules.md#module-start-function) is present, it is evaluated
 given `instance`. Any errors thrown by `start` are propagated to the caller.
--- a/JS.md
+++ b/JS.md
@@ -199,11 +199,10 @@ function, global, memory and table imports):
     throw a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror).
   * Otherwise, append `v.[[Table]]` to `imports`.
 
-Let `instance` be the result of evaluating 
-[`Eval.init`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/eval.ml#L319)
-with arguments `module` and `imports`.
-Note: this synchronously executes the [`start`](Modules.md#module-start-function)
-function, if present.
+Let `instance` be the result of creating a new
+[`Eval.instance`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/eval.ml#L15)
+given `module` and `imports`. Note: this step does not mutate imported memories
+or tables.
 
 Let `exports` be a list of (string, JS value) pairs that is mapped from 
 `module.exports` as follows (assuming the ML spec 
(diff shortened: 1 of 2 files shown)


JS.md
```diff
@@ -243,8 +242,30 @@ of `exports`. Let `moduleNamespace` be the result of calling
...
 
-Return a new `WebAssembly.Instance` object setting `[[Instance]]` to `instance`
-and `exports` to `moduleNamespace`.
+Let `instanceObject` be a new `WebAssembly.Instance` object setting
+`[[Instance]]` to `instance` and `exports` to `moduleNamespace`.
+
+If any of the elements of an [Elements section](Modules.md#elements-section)
+refer to an imported function which is not an
+[Exported Function Exotic Object](#exported-function-exotic-objects), throw a
+[`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror).
```
rossberg-chromium: Is this because they don't have a type that call_indirect can check? Why can't it use the respective import signature?

lukewagner: This is just mirroring the MVP constraints added in #775 that tables only contain WebAssembly functions.

lukewagner: (And so if we relaxed that restriction it'd be trivial to relax here; it's the same underlying implementation challenge.)

lukewagner:
Any other comments?  @gahaas?

JS.md
```diff
@@ -199,11 +199,10 @@ function, global, memory and table imports):
     throw a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror).
   * Otherwise, append `v.[[Table]]` to `imports`.
 
-Let `instance` be the result of evaluating 
-[`Eval.init`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/eval.ml#L319)
-with arguments `module` and `imports`.
-Note: this synchronously executes the [`start`](Modules.md#module-start-function)
-function, if present.
+Let `instance` be the result of creating a new
+[`Eval.instance`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/eval.ml#L15)
```
gahaas: I wonder if a link to (https://github.com/WebAssembly/design/blob/master/JS.md#webassemblyinstance-objects) would be more helpful here than a link to the spec. In (https://github.com/WebAssembly/design/blob/master/JS.md#webassemblyinstance-objects) the link to the spec is helpful because there is some text which tells you what to expect from that link. Here, however,  I did not know what this link actually means.

JS.md
```diff
@@ -215,11 +214,11 @@ has been modified so that each export simply has a `name`, `type` and `index`:
   * If the `type` is function, then export an 
     [Exported Function Exotic Object](#exported-function-exotic-objects),
     reusing an existing object if one exists for the given function definition,
-    otherwise creating a new object.
+    otherwise creating a new object given `instance` and `index`.
   * If the `type` is global:
-    * If the global is not immutable, then throw a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror).
+    * Assert: the global is immutable by MVP validation constraint.
```
gahaas: Is it defined somewhere what `Assert` means?

JS.md
```diff
@@ -243,8 +242,30 @@ of `exports`. Let `moduleNamespace` be the result of calling
 [`ModuleNamespaceCreate(moduleRecord, exportStrings)`](http://tc39.github.io/ecma262/#sec-modulenamespacecreate).
 Set `moduleRecord.[[Namespace]]` to `moduleNamespace`.
 
-Return a new `WebAssembly.Instance` object setting `[[Instance]]` to `instance`
-and `exports` to `moduleNamespace`.
+Let `instanceObject` be a new `WebAssembly.Instance` object setting
+`[[Instance]]` to `instance` and `exports` to `moduleNamespace`.
+
+If any of the elements of an [Elements section](Modules.md#elements-section)
+refer to an imported function which is not an
```
gahaas: I think it should be referS

JS.md
```diff
@@ -243,8 +242,30 @@ of `exports`. Let `moduleNamespace` be the result of calling
...
+If, after evaluating the `offset` [initializer expression](Modules.md#initializer-expression)
+of every [Data](Modules.md#data-section) and [Element](Modules.md#elements-section)
+Segment, any of the segments do not fit in their respective Memory or Table, throw a 
+[`RangeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-rangeerror).
+
+Apply all Data and Element segments to their respective Memory or Table in the
+order in which they appear in the module. Segments may overlap and, if they do,
+the final value is the last value written in order. Note: there should be no
+errors possible that would cause this operation to fail partway through. After
+this operation completes, elements of `instance` may are now visible and callable
```
gahaas: ... elements of `instance` are visible ...

JS.md
```diff
@@ -243,8 +242,30 @@ of `exports`. Let `moduleNamespace` be the result of calling
...
+through imported Tables, even if `start` fails.
```
gahaas: I am not sure what you mean by _imported Tables_. Could you add a link to the section which defines it?

JS.md
lukewagner: So this is a bit subtle, but there are two things here: the abstract wasm-spec-defined `Eval.instance` (which probably we'll rename and relink once we have a prose spec in the spec repo) and the JS `WebAssembly.Instance` object we're constructing which "wraps" it.  Here we're making the former; only at the end do we construct the latter.  We create the former here because it is needed to create the exported functions below.  Does that clear it up for you any?

JS.md
lukewagner: Oh here I just meant it in the "this is a condition that should always be true because it is checked elsewhere".  I think this is the same [meaning as in ecma262](https://tc39.github.io/ecma262/#assert).

lukewagner: I'll add a link to https://tc39.github.io/ecma262/#assert though.

JS.md
gahaas: I do understand that Eval.instance != WebAssembly.Instance. The reason why I mentioned (https://github.com/WebAssembly/design/blob/master/JS.md#webassemblyinstance-objects)  is because it is the section which contains the definition of Eval.instance in JS.md. Is there a way to create a link to the definition of Eval.instance directly?

lukewagner: Well, that's what I was attempting to do with https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/eval.ml#L15.  That's a stand-in for what will be the prose-spec's abstract definition of a wasm instnace.

lukewagner: Did you have something different in mind?

lukewagner:
@gahaas / @rossberg-chromium Any further comments?

JS.md
gahaas: I guess you are right that the prose spec will eventually fix this problem anyways.

rossberg-chromium:
LGTM

gahaas:
LGTM

PR #792, opened 2016-09-06
Small fixes

bnjbvr:
The (useful) commit in this merge request reorders two sentences in the Table section of AstSemantics: a sentence is referring to "this check", the previous sentence doesn't tell anything about a check although the one just before does.

### DIFF
--- a/AstSemantics.md
+++ b/AstSemantics.md
@@ -254,15 +254,15 @@ accessed by WebAssembly code indirectly through an integer index. This feature
 bridges the gap between low-level, untrusted linear memory and high-level
 opaque handles/references at the cost of a bounds-checked table indirection.
 
-The table's element type constrains the type of elements stored 
-in the table and allows engines to avoid some type checks on table use. 
-When a WebAssembly value is stored in a table, the value's type must precisely
-match the element type. Just like linear memory, updates to a table are
-observed immediately by all instances that reference the table. Depending on the
-operator/API used to store the value, this check may be static or dynamic. Host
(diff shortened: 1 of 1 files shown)


jfbastien:
There's still a bunch of whitespace change in the patch, could you drop these?

bnjbvr:
@jfbastien, yes that is intentional: I've split the whitespace changes into one patch (5d2bafc), so that if you prefer to not take them, you could just cherry-pick the reordering patch (81c0321).

Removing trailing spaces still sounds useful to me (it shouldn't affect authorship these days and it is just a good practice), but I do not have a very strong opinion about it.  ¯\\_(ツ)_/¯

bnjbvr:
Can we move this forward, please? Should I just remove the whitespaces changing commit?

lukewagner:
The reorder looks good.  Yeah, we generally have tons of trailing whitespace and don't try to avoid it.

jfbastien:
I'd rather review patches without all-whitespace changes in the middle. Makes it way easier to review. We can do a whitespace-only fix separately.

bnjbvr:
OK, will remove the whitespace patch.

@jfbastien, for what it's worth, Github allows you to do this (in this particular setup, since I've made two different patches):
- [first patch](https://github.com/WebAssembly/design/pull/792/commits/81c03214e33b16535c202ed0614270a69756a444)
- [second patch (whitespaces only)](https://github.com/WebAssembly/design/pull/792/commits/5d2bafc739bf0a8aff3d17620da1da8755f15e46)

PR #793, opened 2016-09-07
clarify return value of Memory.prototype.grow

lukewagner:
Right now JS.md doesn't mention anything about the return value of `grow`. For the same reason we made `grow_memory` return the size before resize (disambiguate races), I think `grow` should too.  Although not symmetric, I think, since it's a JS interface, `grow` should (continue to) throw an exception on failure instead of returning -1 like `grow_memory`, but I could be convinced otherwise.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -402,7 +402,7 @@ aliases `M.[[Memory]]` and whose
 [[[ArrayBufferByteLength]]](http://tc39.github.io/ecma262/#sec-properties-of-the-arraybuffer-prototype-object)
 is set to the new byte length of `M.[[Memory]]`.
 
-Return `ret`.
+Return [`ToJSValue`](#ToJSValue)(`ret`).
 
 ### `WebAssembly.Memory.prototype.buffer`
 
--- a/JS.md
+++ b/JS.md
@@ -377,13 +377,21 @@ Return a new `WebAssembly.Memory` instance with [[Memory]] set to `m` and
 
 ### `WebAssembly.Memory.prototype.grow`
 
+The `grow` method has the signature:
+```
+grow(delta)
+```
+
 Let `M` be the `this` value. If `M` is not a `WebAssembly.Memory`,
 a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
 is thrown.
 
-This method performs a [`grow_memory`](AstSemantics.md#resizing) on
-`M.[[Memory]]`, having first performed [`ToInteger`](http://tc39.github.io/ecma262/#sec-tointeger)
-on the first argument. On failure, a `WebAssembly.RuntimeError` is thrown.
+Let `d` be [`ToInteger`](http://tc39.github.io/ecma262/#sec-tointeger)(`delta`).
+
+Let `ret` be the result of performing a
+[`grow_memory`](AstSemantics.md#resizing) operation given delta `d`.
+
+If `ret` is `-1`, a `WebAssembly.RuntimeError` is thrown.
 
 If `M.[[Memory]].maximum` is `None`, perform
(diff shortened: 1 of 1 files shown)


JS.md
```diff
@@ -394,6 +402,8 @@ aliases `M.[[Memory]]` and whose
 [[[ArrayBufferByteLength]]](http://tc39.github.io/ecma262/#sec-properties-of-the-arraybuffer-prototype-object)
 is set to the new byte length of `M.[[Memory]]`.
 
+Return `ret`.
```
gahaas: Return [`ToWebAssemblyValue`](#towebassemblyvalue)`(ret)`.

lukewagner: I think you mean `ToJSValue` (because we're starting with a wasm `i32`), but point taken, I'll add this.

gahaas:
Is the reason why you think throwing an exception here is better because in JavaScript exceptions can be caught?

JS.md
gahaas: Seems like it has been a long day. Yes, I mean `ToJSValue`.

lukewagner:
Just because JS generally uses exceptions for errors.

lukewagner:
But yes, to your question, in wasm exceptions would be a no-go since we couldn't catch them.  Even if wasm MVP had exceptions, I still think perhaps we wouldn't want to throw since that's not a concept in some wasm source languages.  I also feel the desire for symmetry between the two; it's hard to weigh against having a more-idiomatic JS API, but I'm leaning toward the latter.

JS.md
```diff
@@ -377,13 +377,21 @@ Return a new `WebAssembly.Memory` instance with [[Memory]] set to `m` and
...
+```
+
 Let `M` be the `this` value. If `M` is not a `WebAssembly.Memory`,
 a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
 is thrown.
 
-This method performs a [`grow_memory`](AstSemantics.md#resizing) on
-`M.[[Memory]]`, having first performed [`ToInteger`](http://tc39.github.io/ecma262/#sec-tointeger)
-on the first argument. On failure, a `WebAssembly.RuntimeError` is thrown.
+Let `d` be [`ToInteger`](http://tc39.github.io/ecma262/#sec-tointeger)(`delta`).
```
jfbastien: https

jfbastien: This is necessarily positive, right?

gahaas:
I agree with the choice of throwing an exception, although I don't have a strong opinion either. LGTM

JS.md
lukewagner: So the goal here (and in the 3 other places that use `ToInteger`) is not to forcibly coerce everything via `ToUint32` since that prevents you from ever allowing values >2<sup>32</sup> and we may regret that when we do wasm64.  However, what I'm implicitly leaving out is the range error when the result of `ToInteger` is negative or crazy-big; I was sorta implicitly relying on the dependent operation to reject these values.  I could make another PR to factor out a new `ToUnsignedInteger` that is used by all four places, if that makes sense.

Also probably want a separate PR to switch all the http:// to https:// in this file (there are many).

jfbastien: sgtm

lukewagner:
Any more comments before merging?

rossberg-chromium:
Nit: There is a bit of a category error in this description because `grow_memory` is an instruction that is executed in a given module instance, whereas a JS memory object has no instance attached to it. In the spec we will probably want to introduce memories (and tables) as abstract entities that encapsulate such semantics and can be manipulated from both JS and the Wasm instructions. But for now it's fine, so LGTM.

lukewagner:
Heh, yeah, I realized that too and decided to gloss over it until we have things better factored in the real spec.

PR #794, opened 2016-09-07
Use ToNonWrappingUint32 to JS.md

lukewagner:
As suggested in #793, this PR clarifies how the JS API converts JS values to integers.  We can't just use `ToUint32` since it wraps which I think is a future hazard with wasm64.  Scanning the JS spec, most places that deal with lengths or indices funnel through `ToInteger`, so that's what this PR uses.


lukewagner:
Thumbs-up from @jfbastien; any other comments before merging?  @gahaas?

gahaas:
I like it.

rossberg-chromium:
lgtm

PR #800, opened 2016-09-12
Update BinaryEncoding.md memory/table sections to reflect the ml-proto encoding

AndrewScheidecker:
The difference being that ml-proto encodes an array of memories/tables instead of just one in each of their respective sections.


lukewagner:
I like the symmetry with every other section that has a count, so lgtm.

rossberg-chromium:
Yeah, LGTM too. I obviously had missed that this isn't the case already. :)

titzer:
lgtm

PR #802, opened 2016-09-14
Note that traps throw RuntimeErrors in JS.md

lukewagner:
This PR makes the previously-subtle connection between wasm traps and `RuntimeError` more explicit.  The PR also explains that traps/exceptions work like normal from a JS pov (which is I believe what we're all doing anyway).

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -13,6 +13,22 @@ basically equivalent to
 `new WebAssembly.Instance(new WebAssembly.Module(bytes), imports)`
 as defined below and will be removed at some point in the future.*
 
+## Traps
+
+Whenever WebAssembly semantics specify a [trap](AstSemantics.md#traps),
+a `WebAssembly.RuntimeError` object is thrown. WebAssembly code (currently)
+has no way to catch this exception and thus the exception will necessarily
+propagate to the innermost non-WebAssembly caller (either the browser or
+JavaScript) where it is handled like a normal JavaScript exception.
+
+If WebAssembly calls JavaScript via import and the JavaScript throws an
+exception, the exception is propagated through the WebAssembly activation to the
+enclosing caller.
+
+Because JavaScript exceptions can be handled, and JavaScript can continue to
+call WebAssembly exports after a trap has been handled, traps do not, in
+general, prevent future execution.
+
 ## The `WebAssembly` object
 
 The `WebAssembly` object is the initial value of the `WebAssembly` property of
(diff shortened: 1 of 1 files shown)


JS.md
```diff
@@ -13,6 +13,22 @@ basically equivalent to
 `new WebAssembly.Instance(new WebAssembly.Module(bytes), imports)`
 as defined below and will be removed at some point in the future.*
 
+## Traps
+
+Whenever WebAssembly semantics specify a [trap](AstSemantics.md#traps),
+a `WebAssembly.RuntimeError` object is thrown. WebAssembly code (currently)
+has no way to catch this exception and thus the exception will necessarily
+propagate to the innermost non-WebAssembly caller (either the browser or
+JavaScript) where it is handled like a normal JavaScript exception.
```
jfbastien: "innermost" seems a bit odd. "nearest on the call stack" or something else?

JS.md
```diff
@@ -13,6 +13,22 @@ basically equivalent to
...
+
+If WebAssembly calls JavaScript via import and the JavaScript throws an
+exception, the exception is propagated through the WebAssembly activation to the
+enclosing caller.
+
+Because JavaScript exceptions can be handled, and JavaScript can continue to
+call WebAssembly exports after a trap has been handled, traps do not, in
+general, prevent future execution.
```
jfbastien: I vaguely remember someone insisting that we neuter the memory if a trap occurred. Was that changed? I can't remember...

JS.md
lukewagner: I could go with "enclosing".

JS.md
lukewagner: That was originally brought up as an idea and then considered again a while ago and then dropped.  One issue is that it would prevent various forms of useful telemetry one might want to perform after a fault.  Also, that idea was for memories, not instances, and not all faults relate to memory, so it's technically orthogonal.

jfbastien: Right, that's vaguely what I recalled. Should you then mention something about observability of trap ordering? I think that's what we'd concluded (e.g. two accesses which would trap are observed to trap in program order). This seems like the right place to do so, but I'm happy with any other place you find more appropriate.

lukewagner: Well the abstract (not-JS-API-specific) wasm spec should take care of saying when a program traps and what effects have taken place at that point in time; in the JS API text here we're just saying what that trap turns into.  So I think if we wanted to add refinements to traps+memory-access, that would go in the AstSemantics.md#traps that I link to here.

titzer:
lgtm

PR #811, opened 2016-09-29
Binary 0xc

titzer:
Support for binary 0xC has landed in Firefox and Chrome, and the ml-proto spec is ready, so I think it's time we merged this as a signal that it's official.


lukewagner:
lgtm

titzer:
Both Andreas's over here are OOO, so if I can get another LGTM, I'll merge it.

titzer:
Looks like a git log dump. I can remove it.

jfbastien:
lgtm

PR #816, opened 2016-10-11
Tighten global import requirements

lukewagner:
This PR changes JS.md to require immutable global imports to already be number values, instead of trying to coerce them first, as discussed in [spec/#356](https://github.com/WebAssembly/spec/issues/356).

The PR also tidies up the `Get` logic to avoid reusing `v` and update ml-proto links.


rossberg-chromium:
lgtm

Issue #798, opened 2016-09-09
Reexporting JS imports has potentially surprising semantics


rossberg-chromium:
The first step of processing exports in JS.md says:

> - If index refers to an imported type definition, then simply re-export the imported JS value.

Is that really what we want? Globals and functions are converted/wrapped upon import, this requires to undo that, and store references to the original values.

It has some weird consequences, too:

- When importing-exporting-reimporting a global, the second module will repeat the conversion, and if the original value is a general object, may get a completely different result.
- When importing-exporting-reimporting a JS function, the second module may import it with a signature that is completely different from the original type under which it was exported.
- When exporting a JS function, the JS side may invoke it and see it returning values that are not Wasm values. That is weird, because it got it from a Wasm module after all.

An alternative semantics would be that reexporting a JS value produces the converted/wrapped item, which would presumably be less surprising and maybe simpler to implement, too. WDYT?

lukewagner:
I definitely agree that that seems like the natural thing for immutable globals.  I didn't realize this pre-/post-coercion nuance so that's actually what we've implemented for immutable global reexports.

Given that, I see the symmetry for functions.  So thinking through some of the details:
* There would be some new "`ToWebAssemblyFunction(f, sig)`" conversion that wraps a JS function `f` with a wasm `sig` and returns a new JS function object identity that we'd accumulate in some vector of coerced-imports that can be re-exported.
* The wrapped function object returned by `ToWebAssemblyFunction` would be the same kind of exported-wasm-function as all other wasm exports.
* `ToWebAssemblyFunction(f, sig)` on an exported-wasm-function `f` either returns `f` or throws if `f`'s signature doesn't match `sig`
* Because of the above, you'd never double-wrap a JS function: once it's been wrapped, it is now an exported wasm function and gets signature-checked on future import.
* That also means that, while I can't assign a JS function to a Table element directly via `set`, I *could* get it wrapped and then assign it: either by import+`elem` or import+reexport+`Table.set`.  I really like this b/c it removes an otherwise potentially-frustrating limitation on import+`elem` usage without allowing full-on untyped JS functions in tables (which I'd like to avoid).
* If a single imported JS function was reexported twice, you'd get the same identity (because there is one `ToWebAssemblyFunction` coercion on import).
Does that match your thinking?

[...]

PR #817, opened 2016-10-12
Allow host functions as table elements

rossberg-chromium:
As discussed here: https://github.com/WebAssembly/design/issues/798#issuecomment-245931553

Spec implementation: WebAssembly/spec#347

(Plus a few minor fixes, and updates to spec crossrefs.)

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -180,7 +180,7 @@ Let `imports` be an initially-empty [`import list`](https://github.com/WebAssemb
 * a [`Table.table`](#webassemblytable-objects)
 * a [`Values.value`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/values.ml#L9)
 
-Let `jafunctions` and `wrappers` be empty lists.
+Let `jsfunctions` and `wrappers` be empty lists.
 
 For each [`import`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/kernel.ml#L135)
 `i` in `module.imports` (assuming the ML spec `import` has been extended to have
@@ -202,7 +202,7 @@ function, global, memory and table imports):
       * Let `j` be the index of `v` in `jsfunctions`.
       * Let `wrapper` be `wrappers[j]`.
     * Otherwise:
-      * Let `wrapper` be an anonymous function
+      * Let `wrapper` be an anonymous [Exported Function Exotic Object](#exported-function-exotic-objects)
         which calls `v` by coercing WebAssembly arguments to JavaScript arguments
         via [`ToJSValue`](#tojsvalue) and returns the result by coercing
         via [`ToWebAssemblyValue`](#towebassemblyvalue).
@@ -434,12 +434,7 @@ A `WebAssembly.Table` object contains a single [table](Semantics.md#table)
 which can be simultaneously referenced by multiple `Instance` objects. Each
 `Table` object has two internal slots:
  * [[Table]] : a [`Table.table`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/table.mli)
- * [[Values]] : an array whose elements are either `null` or callable objects
-
-(Note: the ML spec currently represents tables as a single `int list` of
-function indices; we assume here it will be extended in the future with
-a more general `Table` similar to 
-[`Memory`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/memory.mli).)
+ * [[Values]] : an array whose elements are either `null` or [Exported Function Exotic Object](#exported-function-exotic-objects)
 
 ### `WebAssembly.Table` Constructor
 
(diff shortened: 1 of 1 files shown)


JS.md
```diff
@@ -262,11 +262,6 @@ Set `moduleRecord.[[Namespace]]` to `moduleNamespace`.
 Let `instanceObject` be a new `WebAssembly.Instance` object setting
 `[[Instance]]` to `instance` and `exports` to `moduleNamespace`.
 
-If any of the elements of an [Elements section](Modules.md#elements-section)
-refers to an imported function which is not an
-[Exported Function Exotic Object](#exported-function-exotic-objects), throw a
-[`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror).
```
lukewagner: There is a similar check above (GH won't give me a + to comment on it directly) preventing host functions in elems section that should be removed as well.

rossberg-chromium: Ah, thanks. Removed.

lukewagner:
Yes, sorry for the delay on this.  So two general comments:
* Because we're introducing new function identities here, I think we need to be explicit about where they are created and how they get propagated.  It seems like, in the initial pass through the import array, we need to coerce each element into some new array of coerced values that are referred to by exports and the elems section.
* By wholly removing the check on `set`, this allows arbitrary (non-webassembly-wrapped) JS functions which now don't have a destination signature and will require a totally new mechanism in the impl to handle in tables (esp. if the engine has an efficient indirect-call ABI).  I can see how it'd be useful, but could we keep `set` unchanged for now?

rossberg-chromium:
Ah, good points. Reinstantiated the restriction on `set` and introduced explicit caching for host imports.

JS.md
```diff
@@ -180,6 +180,8 @@ Let `imports` be an initially-empty [`import list`](https://github.com/WebAssemb
 * a [`Table.table`](#webassemblytable-objects)
 * a [`Values.value`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/values.ml#L9)
 
+Let `jafunctions` and `wrappers` be empty lists.
```
jfbastien: "JS"

JS.md
```diff
@@ -433,7 +434,7 @@ A `WebAssembly.Table` object contains a single [table](Semantics.md#table)
 which can be simultaneously referenced by multiple `Instance` objects. Each
 `Table` object has two internal slots:
  * [[Table]] : a [`Table.table`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/table.mli)
- * [[Values]] : an array whose elements are either `null` or Function Objects
+ * [[Values]] : an array whose elements are either `null` or callable objects
```
lukewagner: Instead of "callable", could this be "Exported Function Exotic Objects"?

JS.md
```diff
@@ -522,12 +520,8 @@ or `null`, throw a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-erro
 
 Let `i` be the result of [`ToNonWrappingUint32`](#tononwrappinguint32)(`index`).
 
-If `v` is an [Exported Function Exotic Object](#exported-function-exotic-objects):
-* Set the `i`th element of `T.[[Table]]` to the `v.[[FunctionIndex]]`th function
-  in the module's [function index space](Modules.md#function-index-space).
```
rossberg-chromium: Actually, looking at it now, I'm confused about the purpose of T.[[Table]], which doesn't seem to be used for anything. Also, the representation in the spec changed: a table no longer contains indices but actually a function values as well. Is this just essentially duplicating the same thing as T.[[Values]] now? Should I try to get rid of one?

rossberg-chromium: Yes, but when we allowed tables to be im/exported the spec's representation had to change, and table elements are now of type `closure`, as defined here:
https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/instance.ml#L11
This is pretty much the spec's rendering of the [[Values]] array. The index is gone, we need to find some language saying how JS functions map to the spec's "host functions" to make the connection. And at that point we are essentially just duplicating the information, I think.

rossberg-chromium:
This was *much* more work than I anticipated, but I think I managed to adapt the text to the current spec while defining the reuse of created objects more precisely.

JS.md
```diff
@@ -173,83 +173,90 @@ is thrown. If the list of
...
-  * Otherwise, append `v.[[Table]]` to `imports`.
+  * Append `v` to `tables`.
+  * Append `v.[[Table]]` to `imports`.
 
 Let `instance` be the result of creating a new
-[`Eval.instance`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/eval.ml#L15)
-given `module` and `imports`. Note: this step does not mutate imported memories
-or tables.
+[`Eval.init`](https://github.com/WebAssembly/spec/blob/master/ml-proto/spec/eval.ml#L416)
+given `module` and `imports`.
```
lukewagner: How about "creating a new [Eval.instance] by calling [Eval.init] given `module` and `imports`"?

rossberg-chromium: Good point, I actually misparsed the original sentence...

lukewagner:
Thanks for all the work this is much clearer! lgtm

PR #818, opened 2016-10-13
If yields value only with else

tsavola:


JSStats:
I don't think this is a useful change at this point. The block fall-through unwinding needs consideration, should probably unwind, so the `if` `end` should probably unwind too, the it would not be possible to trigger a validation due to values being left on the stack, they will just be discarded at the end of the `if`.

tsavola:
I wanted to mention the restriction on a conceptual level. Maybe I went for too formal wording.

tsavola:
I just created issue #819 like binji_ suggested.

PR #820, opened 2016-10-13
WebAssembly.compile: neuter received buffer

jfbastien:
It doesn't seem useful for the buffer to be available for any purpose after it's been handed off to the `compile` function. The current wording forces the implementation to copy the entire buffer, which is silly. Neutering the provided buffer allows for a more efficient implementation.

My suggested wording is similar to the wording for Worker.postMessage.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -82,9 +82,8 @@ with the resulting `WebAssembly.Module` instance. On failure, the `Promise` is
 [rejected](http://tc39.github.io/ecma262/#sec-rejectpromise) with a 
 `WebAssembly.CompileError`.
 
-The asynchronous compilation is logically performed on a copy of the state of
-the given `BufferSource` captured during the call to `compile`; subsequent mutations
-of the `BufferSource` after `compile` return do not affect ongoing compilations.
+The `bytes` argument becomes unusable (*neutered*) in the context it was sent from
+and becomes available only to the implementation of `compile`.
 
 In the [future](FutureFeatures.md#streaming-compilation), this function can be
 extended to accept a [stream](https://streams.spec.whatwg.org), thereby enabling


pizlonator:
I'm strongly opposed to this proposal.

This does not optimize time.  You gain nothing by saving a memcpy when after this memcpy, you will compile with an optimizing compiler.  Or any compiler.  A template compiler will be 10x more expensive than a memcpy of the input.

This does not optimize space.  You gain nothing by saving the input bits when you will be compiling to machine code.  Even optimized machine code will be bigger than the wasm.  I don't know how much, but I would get significantly so.  Unoptimized could be easily 20x bigger.  Also, if copied, the buffer would be alive while running an optimizing compiler.  The optimizing compiler will use more memory per function it compiles than many wasm inputs.

Neutering is a strange feature with surprising ramifications.  For smaller typed array sizes or other reasons, the VM will speculate that the array will never be neutered (at least in JSC).  This means that for many wasm inputs, this is a pessimization for us.  Personally, I feel that any use of neutering is a slap in the face of those who want to use typed arrays like you would use arrays in other languages - that is, it's not going to all just disappear all of a sudden.  We shouldn't make this worse.

lukewagner:
Agreed with @pizlonator.  Our `Module` actually holds onto the copy of the bytecode made by `compile` so that the bytecode can be included in possible future serializations to disk via IDB so that if, on deserialization, the machine code is not valid due to version/cpu/etc change, we can recompile the machine code.  (`Instance` doesn't keep `Module` alive so these bytes usually get freed with the `Module` on the next GC.)

JSStats:
It could be a burden to have to copy a large input array if only a few functions were used and by a baseline compiler. Perhaps the model is a poor fit. Could the function code and data segments be split out from the rest for this interfaces, and these be requested by a call back as needed.

rossberg-chromium:
Agreeing with @pizlonator and @lukewagner. Neutering is an odd and hostile semantics, to both programmers and engines. Depending on how a JS engine handles typed arrays, it can actually be costly, since it may cause deoptimisations and memory management oddities like mentioned. Hence it should be limited to an absolute bare minimum of cases where it is part of an explicit operation to give up ownership, not as a side effect of random other operations over buffers.

jfbastien:
I find the argument that it should act "like the rest of JS" compelling. Non-zero-copy is surprising IMO, I see no use to retaining the buffer once passed to `compile` (it's not "some random operation" IMO), but consistency matters and I agree that in the grand scheme of things it's a small cost.

Closing for now, happy to re-discuss if others have more to bring.

PR #821, opened 2016-10-14
Throw `RangeError` for invalid sizes or indices

rossberg-chromium:
The `WebAssembly.RuntimeError` exception should be reserved for actual traps in WebAssembly code. API functions should otherwise use appropriate regular JS exceptions.

In particular, this PR changes table/memory access and grow methods to throw a `RangeError` for out of bound sizes or indices. That is consistent with what their constructors do already.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -505,7 +505,7 @@ is thrown.
 
 Let `i` be the result of [`ToNonWrappingUint32`](#tononwrappinguint32)(`index`).
 
-If `i` is smaller or equal than the length of `T.[[Values]]`, a [`RangeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-rangeerror) is thrown.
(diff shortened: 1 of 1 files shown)


gahaas:
lgtm

lukewagner:
Reserving `RuntimeError` for actual traps in wasm sounds fine.  But do you think the `grow` operations should throw a `RangeError`?  That is, is there precedent for using `RangeError` to report resource exhaustion as opposed to programmer error?

rossberg-chromium:
@lukewagner, yes, in fact, some (most?, all?) engines throw `RangeError` on stack overflow (and there even was some proposal to make that standard).

lukewagner:
Ah ok; apparently we are the outlier, then, where we throw an `InternalError`...
lgtm then

jfbastien:
lgtm, but please next time leave issues open for more than 6 hours.

PR #823, opened 2016-10-19
Future-proof type opcodes

rossberg-chromium:
Post MVP we will likely extend Wasm with a few richer types, e.g., function types for blocks, struct types, or function pointers. That introduces a number of places where types will be given either directly or via reference to the type section.

To make Wasm future-proof for efficiently overlaying both these spaces, this PR changes the existing type constructors to have negative opcodes, such that they are distinguishable from indexed type definitions (which are naturally positive).

For example, this will in the future allow
- generalising block annotations to multiple or function types without requiring escape code hacks,
- defining local variables that have struct type (by reference into the type section),
- defining struct types whose fields are either value types, function (pointers), or other structs,
- if desirable, inlining of function types for uses where currently a type section entry is required.

The PR also cleans up the description of type encodings.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -53,6 +53,8 @@ Note: Currently, the only sizes used are `varint7`, `varint32` and `varint64`.
 
 ## Language Types
 
+Note: All types are represented by negative `varint7` values. This is so that they can coexist in a single space with (positive) indices into the type section, which may be relevant in future extensions of the type system.
+
 ### `value_type`
 A `varint7` indicating a [value type](Semantics.md#types). These types are encoded as:
 * `-0x01` (i.e., the byte `0x7f`) indicating type `i32`
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -53,40 +53,57 @@ Note: Currently, the only sizes used are `varint7`, `varint32` and `varint64`.
 
 ## Language Types
 
-Note: All types are represented by negative `varint7` values. This is so that they can coexist in a single space with (positive) indices into the type section, which may be relevant in future extensions of the type system.
+All types are distinguished by a negative `varint7` values that is the first byte of their encoding (representing a type constructor):
+
+| Opcode | Type constructor |
+|--------|------------------|
+| `-0x01` (i.e., the byte `0x7f`) | `i32` |
+| `-0x02` (i.e., the byte `0x7e`) | `i32` |
+| `-0x03` (i.e., the byte `0x7d`) | `f32` |
+| `-0x04` (i.e., the byte `0x7c`) | `f64` |
+| `-0x10` (i.e., the byte `0x70`) | `anyfunc` |
+| `-0x20` (i.e., the byte `0x60`) | `func` |
+| `-0x40` (i.e., the byte `0x40`) | pseudo type for representing an ampty `block_type` |
+
+Some of these will be followed by additional fields, see below.
+
+Note: Gaps are reserved for future extensions. The use of a signed scheme is so that types can coexist in a single space with (positive) indices into the type section, which may be relevant for future extensions of the type system.
 
 ### `value_type`
-A `varint7` indicating a [value type](Semantics.md#types). These types are encoded as:
-* `-0x01` (i.e., the byte `0x7f`) indicating type `i32`
-* `-0x02` (i.e., the byte `0x7e`) indicating type `i64`
-* `-0x03` (i.e., the byte `0x7d`) indicating type `f32`
-* `-0x04` (i.e., the byte `0x7c`) indicating type `f64`
+A `varint7` indicating a [value type](Semantics.md#types). One of:
+* `i32`
+* `i64`
+* `f32`
+* `f64`
+as encoded above.
 
 ### `block_type`
-A `varint7` indicating a signature. These types are encoded as:
-* `-0x40` (i.e., the byte `0x40`) indicating a signature with 0 results.
-* a [`value_type`](#value_type) indicating a signature with a single result
+A `varint7` indicating a block signature. These types are encoded as:
+* either a [`value_type`](#value_type) indicating a signature with a single result
+* or `-0x40` (i.e., the byte `0x40`) indicating a signature with 0 results.
 
 ### `elem_type`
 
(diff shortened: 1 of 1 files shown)


BinaryEncoding.md
```diff
@@ -49,19 +49,19 @@ where the former two are used for compatibility with potential future extensions
...
 
-Note: Currently, the only sizes used are `varint32` and `varint64`.
+Note: Currently, the only sizes used are `varint7`, `varint32` and `varint64`.
 
 ## Language Types
 
 ### `value_type`
-A single-byte unsigned integer indicating a [value type](Semantics.md#types). These types are encoded as:
+A `varint7` indicating a [value type](Semantics.md#types). These types are encoded as:
 * `-0x01` indicating type `i32`
```
titzer: Note that this is byte value 0x7f?

rossberg-chromium:
Online and offline comments addressed. Fixed out-of-bounds values.

BinaryEncoding.md
```diff
@@ -47,22 +49,69 @@ where the former two are used for compatibility with potential future extensions
...
+* `-0x10`  (i.e., the byte `0x70`) indicating [`anyfunc`](AstSemantics.md#table)
+
+Note: In the future, other element types may be allowed.
+
+### `func_type`
+The description of a function signature.
+
+| Field | Type | Description |
+| ----- | ---- | ----------- |
+| form | `varint32` | `-0x20` (i.e., the byte `0x60`) indicating a function type |
```
lukewagner: nit: so in this negative constructor opcode space, we have -1, -2, -3, -4, -0x10, -0x20, -0x40.  Do you think it makes sense to hoist these out into a new `type_constructor` subsection so that we can see them all in one list without having to know that these 7 values are all implicitly in the same index space?  Then each of the 3 uses could specify which subset are valid.

BinaryEncoding.md
```diff
@@ -47,22 +49,69 @@ where the former two are used for compatibility with potential future extensions
...
-* `4` indicating a signature with 1 result of type `f64`.
+A `varint7` indicating a [value type](Semantics.md#types). These types are encoded as:
+* `-0x01` (i.e., the byte `0x7f`) indicating type `i32`
+* `-0x02` (i.e., the byte `0x7e`) indicating type `i64`
+* `-0x03` (i.e., the byte `0x7d`) indicating type `f32`
+* `-0x04` (i.e., the byte `0x7c`) indicating type `f64`
+
+### `block_type`
+A `varint7` indicating a signature. These types are encoded as:
+* `-0x40` (i.e., the byte `0x40`) indicating a signature with 0 results.
```
titzer: Sorry to bike-shed here, but do we want to start at -0x40, and then -0x3F for `function` and -0x3E for `anyfunc`.

BinaryEncoding.md
rossberg-chromium: Good idea, done.

BinaryEncoding.md
rossberg-chromium: Hm, by starting with the lowest absolute value we keep an open-ended gap at the "end" for future extension. The inverse order would become rather weird if we were to grow out of the 6 bit negative range of a varint7, because then we'd have to wrap around and add additional types or categories "before" the basic types.

titzer:
On Thu, Oct 20, 2016 at 4:44 PM, rossberg-chromium <notifications@github.com
> wrote:

> *@rossberg-chromium* commented on this pull request.
> ------------------------------
>
> In BinaryEncoding.md <https://github.com/WebAssembly/design/pull/823>:
>
> > -### `inline_signature_type`
> -A single-byte unsigned integer indicating a signature. These types are encoded as:
> -* `0` indicating a signature with 0 results.
> -* `1` indicating a signature with 1 result of type `i32`.
> -* `2` indicating a signature with 1 result of type `i64`.
> -* `3` indicating a signature with 1 result of type `f32`.
> -* `4` indicating a signature with 1 result of type `f64`.
> +A `varint7` indicating a [value type](Semantics.md#types). These types are encoded as:
> +* `-0x01` (i.e., the byte `0x7f`) indicating type `i32`
> +* `-0x02` (i.e., the byte `0x7e`) indicating type `i64`
> +* `-0x03` (i.e., the byte `0x7d`) indicating type `f32`
> +* `-0x04` (i.e., the byte `0x7c`) indicating type `f64`
> +
> +### `block_type`
> +A `varint7` indicating a signature. These types are encoded as:
> +* `-0x40` (i.e., the byte `0x40`) indicating a signature with 0 results.
>
> Hm, by starting with the lowest absolute value we keep an open-ended gap
> at the "end" for future extension. The inverse order would become rather
> weird if we were to grow out of the 6 bit negative range of a varint7,
> because then we'd have to wrap around and add additional types or
> categories "before" the basic types.
>
Hmm. What about starting at -0x20 (-32) and counting towards negative
infinity also for constructors? I.e. -32 for void, -33 for function, -34
for anyfunc, etc.

BinaryEncoding.md
```diff
@@ -47,22 +49,84 @@ where the former two are used for compatibility with potential future extensions
...
+
+| Opcode | Type constructor |
+|--------|------------------|
+| `-0x01` (i.e., the byte `0x7f`) | `i32` |
+| `-0x02` (i.e., the byte `0x7e`) | `i32` |
+| `-0x03` (i.e., the byte `0x7d`) | `f32` |
+| `-0x04` (i.e., the byte `0x7c`) | `f64` |
+| `-0x10` (i.e., the byte `0x70`) | `anyfunc` |
+| `-0x20` (i.e., the byte `0x60`) | `func` |
+| `-0x40` (i.e., the byte `0x40`) | pseudo type for representing an ampty `block_type` |
```
lukewagner: *empty

lukewagner:
lgtm; no strong opinion on @titzer's suggestion to pack 'em closer

rossberg-chromium:
@titzer, the idea with the gaps is the same as for our instruction opcodes: there are different categories of types (currently, value types, element types, defined types), and it 's nice to leave space to extend each category.

titzer:
lgtm

titzer:
Alrighty

On Thu, Oct 20, 2016 at 6:27 PM, rossberg-chromium <notifications@github.com

PR #826, opened 2016-10-20
Reorganise opcode space

rossberg-chromium:
...as shown here: https://docs.google.com/spreadsheets/d/1RglIQM97cP83Sy-J3BT3D7IxdcH3QCKftiM1bjOL2vc

PR #828, opened 2016-10-21
Fix order of copysign

rossberg-chromium:
Apologies, I screwed up the order of `copysign` among the float operators in #826, putting it in the middle of unary operators. This PR fixes that by moving it last in the group.


titzer:
lgtm

PR #830, opened 2016-10-21
Global types can be mutable, just not when imported

binji:


jfbastien:
I don't remember: why was the restriction there before?

binji:
Not sure, but this PR is just fixing a copy-paste typo that was introduced.

rossberg-chromium:
Thanks, LGTM. I suppose I introduced that when I merged the description of global definitions and global imports into the `global_type` thing (the restriction only applies to the use).

PR #833, opened 2016-10-24
Add frontmatter for jekyll site

s3ththompson:
I'm planning to include the `design` repo as a git submodule of the `webassembly.github.io` repo in order to publish the markdown files in this repo as HTML on the WebAssembly site.

In order for the Jekyll static site generator to recognize these files as appropriate content, they must include YAML frontmatter, even if empty.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -1,3 +1,5 @@
+---
+---
 # Binary Encoding
 
 This document describes the [portable](Portability.md) binary encoding of the WebAssembly modules.
--- a/CAndC++.md
+++ b/CAndC++.md
@@ -1,3 +1,5 @@
+---
+---
 # Guide for C/C++ developers
 
 WebAssembly is being designed to support C and C++ code well, right from
--- a/CodeOfConduct.md
+++ b/CodeOfConduct.md
@@ -1,3 +1,5 @@
+---
+---
 # Code of Ethics and Professional Conduct
 
 WebAssembly operates under the W3C's
--- a/Contributing.md
+++ b/Contributing.md
@@ -1,3 +1,5 @@
+---
+---
 # Contributing to WebAssembly
 
 WebAssembly is initially designed and implemented by browser vendors who are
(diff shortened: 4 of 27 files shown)


s3ththompson:
cc @lukewagner @jfbastien

rossberg-chromium:
Note that some of the pages (especially JS.md) have various github cross-references to the spec repo. Will these carry over correctly?

lukewagner:
sgtm for the short term.  Longer-term, I was thinking that the prose spec would be able to serve all of the Getting Started docs' needs.

s3ththompson:
Ugh, I'm now running into [this issue](https://github.com/github/pages-gem/issues/69) so it looks like I won't be able to make things work without a manual step in the middle...

Agreed that this is not strictly necessary for Getting Started... there are just a number of files that would be nice to host on the site without duplicating: FAQ, High Level Goals, Rationale, etc.

s3ththompson:
Going to continue linking to GitHub for now

jfbastien:
I like the idea of including the design repo's content on the site. Would you just do it with some post-processing?

s3ththompson:
Two options:
- include a [Jekyll plugin](https://github.com/DanielRosenwasser/JekyllMarkdownLinkConverter) to rewrite links, which forces us to build site manually since GitHub pages doesn't allow arbitrary plugins for automatic building
- consider website the source of truth for viewing the design repo and chance design repo links from:
```md
[binary](BinaryEncoding.md)
```
to
```md
[binary](BinaryEncoding/)
```

Do people think either inconvenience is worth it? Which is preferable?

s3ththompson:
If we went with the first option, we could check in the Jekyll output, configure GH Pages to serve from the output directory, and even include a Travis CI test to confirm that the site was indeed built and checked in... might not be terrible

jfbastien:
Autogenerate sounds good, with Travis CI, bells, whistles, and all.

s3ththompson:
Honestly we might as well use this as an excuse to have Travis CI build the site and just push it to AWS with HTTPS set up...

jfbastien:
What's AWS? :wink:

PR #839, opened 2016-10-27
Tagging future features with icon.

flagxor:
Added brief item on multiple return.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -6,17 +6,19 @@ The binary encoding is a dense representation of module information that enables
 small files, fast decoding, and reduced memory usage.
 See the [rationale document](Rationale.md#why-a-binary-encoding) for more detail.
 
+[:unicorn:][future general] = Planned [future](PostMVP.md) feature
+
 The encoding is split into three layers:
 
 * **Layer 0** is a simple binary encoding of the bytecode instructions and related data structures.
   The encoding is dense and trivial to interact with, making it suitable for
   scenarios like JIT, instrumentation tools, and debugging.
-* **Layer 1** provides structural compression on top of layer 0, exploiting
+* **Layer 1** [:unicorn:][future compression] provides structural compression on top of layer 0, exploiting
   specific knowledge about the nature of the syntax tree and its nodes.
   The structural compression introduces more efficient encoding of values, 
   rearranges values within the module, and prunes structurally identical
   tree nodes.
-* **Layer 2** Layer 2 applies generic compression algorithms, like [gzip](http://www.gzip.org/) and [Brotli](https://datatracker.ietf.org/doc/draft-alakuijala-brotli/), that are already available in browsers and other tooling.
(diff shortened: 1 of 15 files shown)


s3ththompson:
Love this idea. It's simple and effective.

I [added support](https://github.com/WebAssembly/webassembly.github.io/commit/bf36461e4665d50acf7865868ac60a285a013d20) on the website for rendering these emojis as HTML images (just as GitHub does when it renders Unicode emoji characters) so that they're more accessible. I believe the unicorn, in particular, is relatively new to the Unicode spec.

One question: are you okay with some editors not being able to display these characters when editing? We could replace 🦄 with `:unicorn:` and still reap the above support for rendering as emoji...

lukewagner:
Heh, I like it too.

BinaryEncoding.md
```diff
@@ -6,17 +6,19 @@ The binary encoding is a dense representation of module information that enables
...
 
+[🦄](PostMVP.md) = Planned [future](PostMVP.md) feature
+
 The encoding is split into three layers:
 
 * **Layer 0** is a simple binary encoding of the bytecode instructions and related data structures.
   The encoding is dense and trivial to interact with, making it suitable for
   scenarios like JIT, instrumentation tools, and debugging.
-* **Layer 1** provides structural compression on top of layer 0, exploiting
+* **Layer 1** 🦄 provides structural compression on top of layer 0, exploiting
```
jfbastien: It would be better to make all of these links IMO. This style [should work](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links):

```
Hello, [🦄][🦄] World!

...

Way at the bottom of the page...

  [🦄][🦄]: PostMVP.md
```

Does this:

> Hello, [🦄][🦄] World!
> 
> ...
>
> Way at the bottom of the page...

  [🦄]: PostMVP.md

Modules.md
```diff
@@ -250,12 +250,12 @@ Each table definition declares an *element type*, *initial length*, and
 optional *maximum length*.
 
 In the MVP, the only valid element type is `"anyfunc"`, but in the
-[future](FutureFeatures.md#more-table-operators-and-types), more element
+[future 🦄e](FutureFeatures.md#more-table-operators-and-types), more element
```
flagxor: fixed

BinaryEncoding.md
flagxor: Agreed should always be a link. Moving to the footnote style.

flagxor:
Switched to ::unicorn:: style and footnotes.

PTAL

jfbastien:
I don't understand the argument for `:unicorn:` when Unicode has exactly what's needed, and it works. You can load 🦄 using fonts if thats necessary (just need to load one character too).

flagxor:
Not all browsers support it.
Also the OSX terminal treats it as a double wide character, so it covers the ] at the end of the tag when I edit. (And I don't have to keep it in the clipboard to type it).

s3ththompson:
LGTM, I think it's a tad more accessible across environments.

Issue #841, opened 2016-10-28
`new Memory` semantics could be better


wibblymat:
`new WebAssembly.Memory()` throws a TypeError if the first parameter isn't an Object, but doesn't require that any properties are set, i.e.

```
new WebAssembly.Memory(); // throws
new WebAssembly.Memory({}); // succeeds, memory with size 0
```

Two options
1. Loosen the restriction so that the parameter must be an Object _or_ undefined
2. Tighten the restriction so that the Object must contain the `initial` property

It doesn't really make sense to create a zero-sized `Memory`, so perhaps 2, with an additional requirement for `initial` to be > 0.

lukewagner:
Requiring `HasProperty(o, 'initial')` sounds fine.  0 size does actually make sense though, if there is no `maximum` or a `maximum` > 0 since you can grow memory later.

wibblymat:
Oh, good point on zero-sized memory. In which case I change my preference to not requiring the parameter at all.

lukewagner:
Hah, race condition.  Well, I can go either way.

wibblymat:
The restriction doesn't seem necessary. There is an obvious default value and it seems like making the parameter optional would be the more common pattern for JS APIs.

Still, I don't have much to put behind the argument beyond gut feeling.

jfbastien:
Sounds like `undefined` or `Object` is OK. Empty `Object` seems silly, but the semantics are that missing `initial` is equivalent to `0` so that's OK.

rossberg-chromium:
I'm sorry, but allowing `undefined` is another attempt of trying to "fix JavaScript". It's not how it works elsewhere. Look e.g. at `Object.defineProperty` for precedence. It takes a descriptor object as argument, but even though that may be empty, passing `undefined` instead will cause a type error, because the first thing that happens is a ToObject conversion. I think we should stick to that pattern.

Having a default value of 0 for the `initial` property is okay, though, and actually in line with JS elsewhere.

rossberg-chromium:
I think this should remain unchanged. The current behaviour, odd as it may be, is in line with what JS does elsewhere.

lukewagner:
So then could we leave this unchanged and close the issue?

jfbastien:
Sure thing.

PR #842, opened 2016-10-28
Require HasProperty('initial')

lukewagner:
As suggested in #841.


jfbastien:
I would instead allow `undefined` as well as `Object`, and not require `initial` (since `{ initial: undefined }` will have the result of being zero IIUC).

rossberg-chromium:
For the reasons I just argued in a [comment](https://github.com/WebAssembly/design/issues/841#issuecomment-257076325) on #841, I'd prefer not to make either change. Neither matches JavaScript practice.

lukewagner:
Sure, it sounds like the original request has gone away as well.

Issue #840, opened 2016-10-28
JS API details


MikeHolman:
I'm finishing up implementation of the new JS API, and there are a few specifics I'm unclear on.

Should we define length, name, toStringTag properties on WebAssembly.Memory, etc. objects? Are they enumerable, configurable, writable?

Should we spec 
WebAssembly.Memory.prototype.grow to throw TypeError if underlying ArrayBuffer is already detached. Since we expose the ArrayBuffer through getter, presumably we are allowing that to happen?

WebAssembly.Table.prototype.set has the restriction that elements must be 
Exported Function Exotic Objects. Does this apply for tables created internally as well, i.e. can you only place WebAssembly functions into the table or can you place JS functions as well?

lukewagner:
> Should we define length, name, toStringTag properties on WebAssembly.Memory, etc. objects? 
> Are they enumerable, configurable, writable?

We do add `length`, `name` and all standard constructor properties to the wasm constructors with their default descriptors.  For reference, here's our [test for all the properties and their descriptors](https://hg.mozilla.org/mozilla-central/file/tip/js/src/jit-test/tests/wasm/jsapi.js).

> Should we spec WebAssembly.Memory.prototype.grow to throw TypeError if underlying 
> ArrayBuffer is already detached. 
> Since we expose the ArrayBuffer through getter, presumably we are allowing that to happen?

We (and I believe V8) mark the `ArrayBuffer`s that are created for `Memory` as non-detachable; so they throw if any external operation like `postMessage` attempts to detach.  Agreed this should be added to JS.md.

> WebAssembly.Table.prototype.set has the restriction that elements must be
> Exported Function Exotic Objects. Does this apply for tables created internally as well,
> i.e. can you only place WebAssembly functions into the table or can you place JS functions as well?

After #817, when you import a JS function with a particular signature, it gets logically wrapped with a new JS function object that is considered an exported function and thus may be stored in tables (imported or internally-defined) via `elem` sections or `Table.set`.

lukewagner:
Any remaining open questions here or can we close?

jfbastien:
Assume good, closing.

PR #843, opened 2016-10-28
Prevent detachment of a Memory's ArrayBuffer

lukewagner:
As discussed in #840.


pizlonator:
+1

titzer:
This is what we do as well, although I think the attempt to detach is silently ignored.

rossberg-chromium:
LGTM

PR #845, opened 2016-10-28
Merge PostMVP into FutureFeatures for clarity.

flagxor:


s3ththompson:
I like this idea. Do we want to consider replacing "MVP" into "initial version" by the way? Although that word really helped us focus on the initial set of features, I feel as though it's somewhat confusing to explain in the context of the browser preview and/or launch. what do others think?

jfbastien:
I wouldn't rename "MVP" because that's what's already all over the internet.

This change looks good to me.

README.md
```diff
@@ -32,9 +32,8 @@ The actual WebAssembly specification is being developed in the [spec repository]
 
 We've mapped out features we expect to ship:
 
- 1. In [the Minimum Viable Product (MVP)](MVP.md);
- 2. Closely [after the MVP](PostMVP.md);
- 3. In [future versions](FutureFeatures.md).
+ 1. In initial [Minimum Viable Product (MVP)](MVP.md) release;
```
s3ththompson: nit: "In an initial..." or "In the initial..."

s3ththompson:
@jfbastien good point. changing things for clarity always runs the risk of the change itself being more confusing than the original :)

lukewagner:
lgtm

sunfishcode:
FutureFeatures.md (as it currently stands) is a collection of ideas, some strong, others vague, speculative, or contentious, meant to be things to think about. PostMVP.md is meant to address a specific audience of people who perceive the MVP feature set as incomplete even for its intended initial use cases. Merging the two, even with stars for priorities, may blur the PostMVP message, and may lend undue weight to some of the items in FutureFeatures today.

flagxor:
Dan, would marking some of these with :radioactive: address your concern about lending too much weight to some?

flagxor:
Or perhaps :confused: (confused) ?

flagxor:
Or :clown: (clown) ...

flagxor:
Or :chestnut: (chestnut) ... or :rocket: (rocket).... Anyhow.

sunfishcode:
My comment is that PostMVP addresses a difference audience than FutureFeatures (in its present state), and that keeping it as a separate page feels like it addresses each in a way that a single list, even with annotations, doesn't. However, if others disagree, I won't object further.

flagxor:
Adding the unicorn tags lead me to notice that the two are mixed together a good deal (and confusing).
I think if we retain the separation we'll probably want to move a few things from FutureFeatures into PostMVP and probably have two separate icons for links leading to the two.

What do others think?

jfbastien:
I'd rather merge the two.

s3ththompson:
@sunfishcode I agree that the bucketing between "near-term, we're definitely adding these" and "aspirational far-reaching future". however, I don't think that "post-mvp" as a term particularly captures the fact that they'll be implemented imminently.

How about changing the introduction to introduce two explicit buckets: "On deck for immediate design" and "Future proposals". Everything with stars would be in the first bucket. Everything afterwards would be in the second?

FutureFeatures.md
```diff
@@ -1,17 +1,99 @@
...
 These are features that make sense in the context of the
-[high-level goals](HighLevelGoals.md) of WebAssembly but are not considered part
-of the [Minimum Viable Product](MVP.md) or the essential
-[post-MVP :unicorn:](PostMVP.md)
-feature set which are expected to be standardized immediately after the
-MVP. These will be prioritized based on developer feedback, and will be
+[high-level goals](HighLevelGoals.md) of WebAssembly but won't be part of the
+initial [Minimum Viable Product](MVP.md) release.
+
+We except a several essential features to be standardized immediately after the
```
s3ththompson: expect

flagxor: done

flagxor:
Revised to separate into two sections.
Keeping the stars.
Dan would shuffling some of the more controversial stuff downward help?

rossberg-chromium:
LGTM

I for one think that replacing "MVP" with "initial version" would be much better. Outside the small circle of people that have followed this repo very closely I have yet to encounter a single person that knows what the heck I'm talking about when I use a term like "MVP". It's rather off-putting management/sales bla, TBH. I think we can happily risk the minor confusion among the few early people for the benefit of making it clearer to the broader future audience.

jfbastien:
"MVP" was a core thing in the initial messaging. I think we should keep it. It's late to change it now.

titzer:
lgtm

jfbastien:
Assuming this is good to merge, given the timing and all.

PR #850, opened 2016-10-30
Clarify status of text format, add example, and remove confusing language

s3ththompson:
While writing the intro section of the WebAssembly getting started page, I was frustrated by the lack of a clearer explanation of the text format. I attempted to clean up this file by explaining what exists and is currently specified, what we hope to design going forward, and what conventions the tools currently use.

I intended this change to be relatively non-controversial, so please let me know if this framing is inaccurate or misleading. Curious, too, to hear others' thoughts on the organization.

### DIFF
--- a/TextFormat.md
+++ b/TextFormat.md
@@ -1,64 +1,84 @@
 # Text Format
 
-The purpose of this text format is to support:
+WebAssembly does not yet have a standardized text format that encodes function
+bodies in addition to module structure, data segments and other program
+metadata in a way that is eqivalent to the [binary format](BinaryEncoding.md).
+WebAssembly does, however, have a specified [representation](#linear-bytecode)
+of function bodies which should be displayed in browsers and other tools when
+[debugging](#debug-symbol-integration) modules.
 
-* View Source on a WebAssembly module, thus fitting into the Web (where every
-  source can be viewed) in a natural way.
-* Presentation in browser development tools when source maps aren't present
-  (which is necessarily the case with [the Minimum Viable Product (MVP)](MVP.md)).
-* Writing WebAssembly code directly for reasons including pedagogical,
-  experimental, debugging, optimization, and testing of the spec itself.
+# Linear bytecode
 
-The text format is equivalent and isomorphic to the [binary format](BinaryEncoding.md).
+WebAssembly function bodies encode bytecode instructions which have a
+specified textual representation. A linear presentation of these sequences of
+instructions is the only textual encoding of function bodies which can
+accurately represent the computations of the
+[structured stack machine](Semantics.md). Therefore, these linear instructions
+are a minimum necessary part of any text format.
 
-The text format will be standardized, but only for tooling purposes:
+Here is an example function illustrated in C++, binary, and text (linear
+assembly bytecode):
 
-* Compilers will support this format for `.S` and inline assembly.
-* Debuggers and profilers will present binary code using this textual format.
-* Browsers will not parse the textual format on regular web content in order to
-  implement WebAssembly semantics.
+<table>
+  <tr>
+    <th>C++</th>
+    <th>Binary</th>
+    <th>Text</th>
+  </tr>
+  <tr>
+    <td><pre>
+int factorial(int n) {
+  if (n == 0)
+    return 1;
+  else
+    return n * fac(n-1);
+}
+    </pre></td>
+    <td><pre>
+20 00
+42 00
+51
+04 7e
+42 01
+05
+20 00
+20 00
+42 01
+7d
+10 00
+7e
+0b
+    </pre></td>
+    <td><pre>
+get_local 0
+i64.const 0
+i64.eq
+if i64
+    i64.const 1
+else
+    get_local 0
+    get_local 0
+    i64.const 1
+    i64.sub
+    call 0
+    i64.mul
+end
+    </pre></td>
+  </tr>
+</table>
 
-Given that the code representation is actually an
-[Abstract Syntax Tree](Semantics.md), the syntax would contain nested
-statements and expressions (instead of the linear list of instructions most
-assembly languages have).
+# Tool conventions
 
-There is no requirement to use JavaScript syntax; this format is not intended to
-be evaluated or translated directly into JavaScript. There may also be
-substantive reasons to use notation that is different than JavaScript (for
-example, WebAssembly has a 32-bit integer type, and it should be represented
-in the text format, since that is the natural thing to do for WebAssembly,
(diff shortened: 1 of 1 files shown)


TextFormat.md
```diff
@@ -1,64 +1,81 @@
...
+modules.
+
+# Linear bytecode
+
+WebAssembly function bodies encode bytecode instructions which have a
+specified textual representation. A linear presentation of these sequences of
+instructions is the only textual encoding of function bodies which can
+accurately represent the computations of the
+[structured stack machine](Semantics.md). Therefore, these linear instructions
+are a minimum necessary part of any text format.
```
kripken: I don't think that's true - it's not the only way, and therefore not a minimum necessary part of the text format.

It is reasonable, simple, and useful. So it makes some sense. But there are other options (arguably better ones, but less simple).

flagxor: How about something like:

WebAssembly function bodies encode bytecode instructions which have a specified canonical opcode names. A linear presentation of these sequences of instructions allows a direct human-readable order-preserving presentation of the binary format. This format is suitable for opcode by opcode inspection of a WebAssembly program and can readily be related to the [sematics](Semantics.md) of the format.

TextFormat.md
```diff
@@ -75,11 +92,40 @@ official format may look entirely different:
...
+## Debug symbol integration
 
 The binary format inherently strips names from functions, locals, globals, etc,
 reducing each of these to dense indices. Without help, the text format must
 therefore synthesize new names. However, as part of the [tooling](Tooling.md)
 story, a lightweight, optional "debug symbol" global section may be defined
 which associates names with each indexed entity and, when present, these names
 will be used in the text format projected from a binary WebAssembly module.
+
+# Future design :unicorn:
```
flagxor: Elsewhere it had been suggested we strive to make all unicorns links.
On the other hand this doesn't really have anywhere else to link to.

flagxor:
Aside from @kripken 's comment + one other observation, lgtm.

TextFormat.md
wllang: I don't think the statement is technical true either. A structured expression based text format can also encode the function code accurately, so the linear presentation is not the 'only textual encoding of function bodies which can accurately represent the computations'. If you have an example to demonstrate otherwise then please let us know? 

Expressions can be very useful to readers of the code as they present the data flow and live range for many common high frequency code patterns, so I am confident that there will be tools that use expression based patterns and other date flow patterns to assist developers.

Wasm is designed for quick decoding to validated SSA form, so this SSA form could also be considered a canonical form. There could be multiple ways to encode the same SSA form in the binary format, with the same basic sequence of instructions but with variation in use of local variables versus `pick` operators etc to connect operators. This SSA form can be presented in a range of text formats, including a functional style which is a long way from a linear format. I also think it is likely that some tools will decode to SSA form as part of the text format presentation canonicalizing away some of the encoding details, then when `pick` is added this can simply be an encoding efficiency option.

TextFormat.md
```diff
@@ -75,11 +92,40 @@ official format may look entirely different:
...
+cases:
+
+* View Source on a WebAssembly module, thus fitting into the Web (where every source can be viewed) in a natural way.
+* Presentation in browser development tools when source maps aren't present (which is necessarily the case with [the Minimum Viable Product (MVP)](MVP.md)).
+* Writing WebAssembly code directly for reasons including pedagogical, experimental, debugging, optimization, and testing of the spec itself.
+
+## Additional design considerations
+
+There is no requirement to use JavaScript syntax; this format is not intended to be evaluated or translated directly into JavaScript. There may also be substantive reasons to use notation that is different than JavaScript (for example, WebAssembly has a 32-bit integer type, and it should be represented in the text format, since that is the natural thing to do for WebAssembly, regardless of JavaScript not having such a type). On the other hand, when there are no substantive reasons and the options are basically bikeshedding, then it does make sense for the text format to match existing conventions on the Web (for example, curly braces, as in JavaScript and CSS).
```
jfbastien: 64-bit int is a better example.

TextFormat.md
```diff
@@ -75,11 +92,40 @@ official format may look entirely different:
...
+
+The text format isn't uniquely representable. Multiple textual files can assemble to the same binary file, for example whitespace isn't relevant and memory initialization can be broken out into smaller pieces in the text format.
```
jfbastien: Saying "the text format" is weird after saying there's isn't "one" text format.

TextFormat.md
```diff
@@ -75,11 +92,40 @@ official format may look entirely different:
...
+
+The text format is precise in that values that cannot be accurately represented in the binary format are considered invalid text. Floating-point numbers are therefore represented as hexadecimal floating-point as specified by the C99 standard, which IEEE-754-2008 section 5.12.3 also specifies. The textual format may be improved to also support more human-readable representations, but never at the cost of accurate representation.
```
jfbastien: Ditto.

jfbastien: Hex float is also in C++17 :smile:

TextFormat.md
s3ththompson: yeah, can't think of anywhere else to link it... will probably leave for now

TextFormat.md
s3ththompson: I just copied and pasted this from elsewhere, but agreed. will fix

s3ththompson:
Updated, PTAL

flagxor:
lgtm

TextFormat.md
```diff
@@ -1,64 +1,76 @@
...
-The purpose of this text format is to support:
-
-* View Source on a WebAssembly module, thus fitting into the Web (where every
-  source can be viewed) in a natural way.
-* Presentation in browser development tools when source maps aren't present
-  (which is necessarily the case with [the Minimum Viable Product (MVP)](MVP.md)).
-* Writing WebAssembly code directly for reasons including pedagogical,
-  experimental, debugging, optimization, and testing of the spec itself.
-
-The text format is equivalent and isomorphic to the [binary format](BinaryEncoding.md).
```
rossberg-chromium: Although this was inaccurate, we need to keep a statement like it in some variation. In particular, it is no longer clear from the new text that the text format is supposed to map _unambiguously_ to a _unique_ binary (modulo LEB encoding liberties), which is the most important design consideration of all.

s3ththompson: Given that this was already merged, can you create a new PR with a clarifying statement? Happy to merge immediately--I didn't intend to remove that consideration.

rossberg-chromium:
@s3ththompson, okay, please see #854.

PR #855, opened 2016-10-31
Unknown sections must be uniquely named

jfbastien:
This is left unspecified. Unique naming seems like the right answer.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -193,6 +193,7 @@ invalidate a module.
 | payload_data  | `bytes` | content of this section, of length `payload_len - sizeof(name) - sizeof(name_len)` |
 
 Each section is optional and may appear at most once.
+Unknown sections must be uniquely named (all bytes composing their names must be identical).
 Known sections from this list may not appear out of order.
 The content of each section is encoded in its `payload_data`.
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -418,6 +418,7 @@ User-defined section string: `"name"`
 The names section does not change execution semantics, and thus is not allocated a section code.
 It is encoded as an unknown section (id `0`) followed by the identification string `"name"`.
 Like all unknown sections, a validation error in this section does not cause validation of the module to fail.
+The name section may appear only once, and only after the [Data section](#Data-section).
 The expectation is that, when a binary WebAssembly module is viewed in a browser or other development
 environment, the names in this section will be used as the names of functions
 and locals in the [text format](TextFormat.md).
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -192,8 +192,8 @@ invalidate a module.
 | name | `bytes` ? | section name string, present if `id == 0` |
 | payload_data  | `bytes` | content of this section, of length `payload_len - sizeof(name) - sizeof(name_len)` |
 
-Each section is optional and may appear at most once.
-Unknown sections must be uniquely named (all bytes composing their names must be identical).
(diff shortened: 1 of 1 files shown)


titzer:
I see the argument for this, I don't think engines should be required to enforce this constraint. For example, it'd be unfortunate to have to keep a hashset of all unknown section names just to detect collisions to enforce this.

jfbastien:
I don't really care either way, as long as we specify the expectation.
A linear search in an array is probably way faster than a hash set in the case that matters.

lukewagner:
Given the preceding statement that "Unknown sections are ignored by the WebAssembly implementation, and thus validation errors within them do not invalidate a module.", I think the "must" is a bit ambiguous: it usually means validation error but here I don't think we want to create one special way that user-defined sections are validated.  How about "should"; that sets an expectation.

jfbastien:
"should" is kind of horrible standards weaselry IMO. One or the other is good, but the middle ground isn't.

sunfishcode:
I'm expecting we'll add this as a hard requirement in [convention documents](https://github.com/WebAssembly/tool-conventions/), as needed, as we start defining linking and other conventions. For example if we have a document that says that a section named "linking.static.relocations" (or whatever) contains relocations for static linking according to a certain convention, that same document could require that there be at most one section per module with that name in any object participating in the convention.

It's also possible that there will be situations where we don't need or want the uniqueness requirement. It would depend on the nature of the sections and the tools that will be processing them.

I guess this does raise a question: WebAssembly currently defines the "name" section as an official user-defined section. Should the spec specify how implementations should interpret the presence of mutiple "name" sections in a single module?

binji:
Along those lines, is there a defined location for the "name" section?

lukewagner:
Yeah, I was also thinking that perhaps one day we may want to allow multiple sections of a single type for specific section types (esp. for user-defined sections).  In that case, each section would specify whether it was valid to have more than one as part of that sections' validation rules.  Since all known sections may only appear once, it makes sense to say it once and for all.  But for user-defined sections, it's probably better to keep things open-ended and just say, in the names section, "may only appear once".

@binji You're right, that should also be stated in the Names section.  For us, the requirement is "after the Data section".

jfbastien:
> @binji You're right, that should also be stated in the Names section. For us, the requirement is "after the Data section".

"requirement"? You fail validation if it's before, or you just ignore it?

lukewagner:
Yes :)  Validation errors in user-defined sections don't cause the entire module to fail to validate.  So "only one" and "after data section" would just be two additional constraints in addition to all the existing validation criteria of a Names sections.

jfbastien:
Names section is starting to sounds like a Real Section which we're trying to specify 😕

binji:
wabt currently allows it anywhere after the function section. But if it occurs before then it is just ignored.

jfbastien:
Ping, does anyone really care about the polarity of this? What I propose seems like the most consistent approach.

lukewagner:
I'd propose removing the line added and instead adding one to the Names section that says "The Names section may only appear once and only after the Data section"

lukewagner:
(and to be clear, that line would go after the existing one that says "Like all unknown sections, a validation error in this section does not cause validation of the module to fail." and thus it doesn't force checking or a validation error)

jfbastien:
In that case, the line I add should also be reworded to say that unknown sections don't have to appear at most once.

lukewagner:
sgtm

jfbastien:
Updated as suggested, PTAL.

lukewagner:
Thanks, lgtm

PR #856, opened 2016-11-01
Fix broken links.

shurcooL:
The Markdown parser used on the http://webassembly.org/ website doesn't support having a newline character between the [] and () parts of the Markdown link syntax. E.g.:

[image](http://webassembly.org/docs/c-and-c++/)

Remove that newline character in order for links to render properly on the website.

As I understand it, the http://webassembly.org/ website is generated in https://github.com/WebAssembly/website repo, which has a git submodule of this repo, that's why I'm doing the change in this repo. First time contributing here, so please let me know if I'm doing something wrong.

### DIFF
--- a/CAndC++.md
+++ b/CAndC++.md
@@ -82,10 +82,9 @@ expected to be defined in accompaniment.
 
 WebAssembly doesn't change the C or C++ languages. Things which cause
 undefined behavior in C or C++ are still bugs when compiling for WebAssembly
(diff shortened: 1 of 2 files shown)


jfbastien:
Even if it's the wrong place, this can't hurt. Merging, will let @s3ththompson figure out the rest :wink:

Thanks!

s3ththompson:
Thanks, will pull in the new changes now

PR #858, opened 2016-11-01
Add Module.(import|exports)

lukewagner:
@yurydelendik pointed out that if people want to integrate WebAssembly modules into existing module loaders like SystemJS, they'll need to get the string names of imports/exports of a `Module`.  Failing this, they'll need to resort to partially decoding the binary themselves (as Yury's [prototype](https://github.com/yurydelendik/system-wasm/blob/38363ebef655ac298a5bc80619e02a2c520ff835/index.js#L27) did) which could lead to brittleness.  It seems like a good idea to expose these as own data properties on the `Module` object since the engine needs to have these sitting around *anyway*.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -168,10 +168,18 @@ Otherwise, this function performs synchronous compilation of the `BufferSource`:
 
 ### `WebAssembly.Module.exports`
 
-The initial value of the `exports` property of a `Module` `m` is an `Array` of
-Object values produced by mapping each 
+The `exports` function has the signature:
+
+```
+Array exports(moduleObject)
+```
+
+If `moduleObject` is not a `WebAssembly.Module` instance, a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
+is thrown.
+
+This function returns an `Array` produced by mapping each
 [`Ast.export`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/ast.ml#L152)
-`e` of [m.[[Module]].exports](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/ast.ml#L187)
+`e` of [moduleObject.[[Module]].exports](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/ast.ml#L187)
 to the Object `{ name: String(e.name), kind: e.ekind }` where `e.name` is [decoded as UTF8](Web.md#names)
 and `e.ekind` is mapped to one of the String values `"function"`, `"table"`, `"memory"`, `"global"`.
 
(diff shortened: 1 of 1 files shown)


JS.md
```diff
@@ -166,6 +166,29 @@ Otherwise, this function performs synchronous compilation of the `BufferSource`:
   the validated `Ast.module`.
 * On failure, a new `WebAssembly.CompileError` is thrown.
 
+### `WebAssembly.Module.exports`
```
jfbastien: The section names are `import` and `export`, so I wouldn't pluralize for consistency.

lukewagner: Just evaluating it as a JS API, though, I think the plural makes more sense and matches other APIs that return an aggregate like `Object.keys`, `Response.headers`, `Node.childNodes`.  How we name the section headers is just a spec-internal detail anyway and could change at any time.  IIRC, some of the section names even started as plural and the s's were stripped when the string names were embedded in the section header to save bytes.

jfbastien: OK sounds fine.

jfbastien:
FWIW that's a bug I ran into a while ago as well:
https://github.com/jfbastien/musl/blob/wasm-prototype-1/arch/wasm32/wasm.js#L935

rossberg-chromium:
LGTM

JS.md
yurydelendik: Will it make sense to add "prototype" to denote object instance property (vs "static" property of Module): `WebAssembly.Module.prototype.exports`. Alternative is to start section "Properties of the WebAssembly.Module prototype" with subsection for every property.

rossberg-chromium: Ah, because of the heading I hadn't noticed that these are actually meant to be methods on module _objects_ (I suppose?). We probably should not do that -- it is a kind of reflection mechanism, and the general trend for ES has been to introduce new reflection mechanisms as static methods only, for a number of reasons (see all the `Object.*` methods). To be in line with that, the API here should rather be `WebAssembly.Module.exports(moduleObj)` and similarly for imports.

lukewagner: As written, these are actually describing own data properties of the `Module` instance objects.  One can use internal hooks for data properties to do lazy allocation on the first read, so that isn't an issue, but maybe it is a bit simpler and more explicit to express this lazy allocation as a getter...

lukewagner: (race posting)  Hah, three different interpretations of the text.  So would static `Module.imports` methods be preferred even over getters on `Module.prototype` (which is almost equivalent, really...)?

rossberg-chromium: Yes, that's the pattern since ES5. The difference is that it enables a more centralised way for policing reflection. Not that I'd have an immediate use case for that with modules, but we can just as well follow the pattern.

lukewagner: Agreed it's good to follow the pattern here, I'll update.

lukewagner:
Updated, PTAL.

rossberg-chromium:
lgtm

PR #859, opened 2016-11-02
Clarify names section indexing

lukewagner:
This PR clarifies something that was technically already implied but might have been overlooked: now that function imports are in the function index space, they are named by the Names section.

### DIFF
--- a/Modules.md
+++ b/Modules.md
@@ -300,7 +300,7 @@ The *function index space* indexes all imported and internally-defined
 function definitions, assigning monotonically-increasing indices based on the
 order of definition in the module (as defined by the [binary encoding](BinaryEncoding.md)).
 Thus, the index space starts at zero with the function imports (if any) followed
-by the functions defined within the module
+by the functions defined within the module.
 
 The function index space is used by:
 
@@ -312,7 +312,7 @@ The *global index space* indexes all imported and internally-defined
 global definitions, assigning monotonically-increasing indices based on the
 order of definition in the module (as defined by the [binary encoding](BinaryEncoding.md)).
 Thus, the index space starts at zero with the global imports (if any) followed
-by the globals defined within the module
+by the globals defined within the module.
 
 The global index space is used by:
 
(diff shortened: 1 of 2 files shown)


binji:
lgtm

wllang:
The order of the function indexes also does not appear to be well defined. Could a clarification be added that the imports are assigned an index starting at zero in the order they appear and then the functions assigned an index in the order their appear in the functions section.

Could the start function index also be clarified to be in the function index space, so could be an import.

jfbastien:
> The order of the function indexes also does not appear to be well defined. Could a clarification be added that the imports are assigned an index starting at zero in the order they appear and then the functions assigned an index in the order they appear in the functions section.
>
> Could the start function index also be clarified to be in the function index space, so could be an import.

PR?

wllang:
@jfbastien Well this PR is trying to clarify the relationship between the names section names and the function index space, but if the function index space is not well defined then it has not really done so.

lukewagner:
Oops, I left off the "Modules.md" in "Modules.md#function-index-space"; this seems to define the order precisely.

wllang:
Thank you, I see the definition of the function index space now. Could suggest also noting that it is a zero based index space, and that the import and function indexes are defined in the imports section and functions function section and it might help the reader to repeat that the imports section must be before the functions section so the imports are first in the function index space.

lukewagner:
Righto

Modules.md
```diff
@@ -299,6 +299,8 @@ available before compilation begins.
 The *function index space* indexes all imported and internally-defined
 function definitions, assigning monotonically-increasing indices based on the
 order of definition in the module (as defined by the [binary encoding](BinaryEncoding.md)).
+Thus, the index space starts at zero with the function imports (if any) followed
+by the functions defined within the module
```
jfbastien: Period at the end of sentence (here and below)

jfbastien:
lgtm

PR #871, opened 2016-11-14
JS API algorithms: use numbering

jfbastien:
It's easier to refer to JS APIs using numbered lists rather than bullet indentations. This isn't perfect because Markdown tops out after some amount of numbering, but it's better than before!

I think we'll separately want to rejigger some of the ordering of the algorithms. This change should be a no-op.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -157,14 +157,14 @@ exception is thrown.
 
 Otherwise, this function performs synchronous compilation of the `BufferSource`:
 
-* The byte range delimited by the `BufferSource` is first logically decoded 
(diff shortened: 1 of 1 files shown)


lukewagner:
So is the way this works that if you give everything "1." then markdown auto-numbers it for you?

jfbastien:
@lukewagner [yup](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#lists)! Any number works.

lukewagner:
Oh hah, I see.  Yeah, well lgtm/

PR #872, opened 2016-11-14
Function index space is used by exports

jfbastien:


Modules.md
```diff
@@ -304,7 +304,8 @@ by the functions defined within the module.
 
 The function index space is used by:
 
-* [calls](Semantics.md#calls), to identify the callee of a direct call
+* [Calls](Semantics.md#calls), to identify the callee of a direct call.
+* [Exports](BinaryEncoding.md#export-section), to determine which functions are exposed to the embedder.
```
lukewagner: Pre-existing, but since you're here, could you add the start section?

lukewagner:
lgtm

Modules.md
binji: The table section as well.

Issue #867, opened 2016-11-10
Exception type for stack overflows


gahaas:
At the moment stack overflows in WebAssembly should throw a WebAssembly.RuntimeError object (see https://github.com/WebAssembly/design/blob/master/Semantics.md#traps and https://github.com/WebAssembly/design/blob/master/JS.md#traps). JavaScript, however, deals with stack overflows by throwing a RangeError in Chrome, or an InternalError in Firefox. I think stack overflows should cause the same exception in JavaScript and in WebAssembly. The reason is that if an application which uses both JavaScript code and WebAssembly code has a stack overflow, then the thrown exception depends on whether the stack overflow happened in a JavaScript function or in a WebAssembly function. 

In V8 we have the additional problem that we use the same stack overflow detection for WebAssembly and JavaScript. This makes it difficult for us to throw a different exception for stack overflows in WebAssembly.

For these reasons I think that in the case of stack overflows, being consistent with JavaScript is more important than being consistent within WebAssembly. Since the exception for stack overflows is browser dependent (at the moment), I suggest to leave the exception type of stack overflows undefined.

jfbastien:
Sounds fine.

lukewagner:
I feel like this might've been discussed recently with @rossberg-chromium in an issue.  The resolution though agreed with you: while all *traps* may throw a `RuntimeError`, stack overflow isn't a trap.

rossberg-chromium:
@lukewagner, I believe stack overflow only came up in the context of what errors to use for signalling resource exhaustion in the JS API grow methods.

Anyways, agreed with the conclusion. Since browsers disagree on the exception used for SO, the design doc should probably just state that Wasm uses the same exn.

rossberg-chromium:
Pondering a bit, I guess the actual question here is whether we want future Wasm to be able to handle traps and whether that should include stack overflow. If so, then stack overflow should remain a regular trap/RuntimeError, and _moreover_ any stack overflow caused while calling back from Wasm into JS should also be mapped to a trap.

lukewagner:
[...]

PR #875, opened 2016-11-16
Define the exception type for stack overflows

gahaas:
This pull request explicitly defines the exception thrown in JavaScript for WebAssembly stack overflows. The changes were discussed in #867.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -9,14 +9,11 @@ any other Web API that loads ES6 modules via URL—as part of
 
 ## Traps
 
-Whenever WebAssembly semantics specify a deterministic [trap](Semantics.md#traps),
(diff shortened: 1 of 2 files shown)


rossberg-chromium:
I think this should become a separate section, because it would no longer be a trap. Also, you'll need to adjust Semantics.md as well.

rossberg-chromium:
Alternatively, you could leave Semantics.md alone and say here that stack overflow is an exception and the only case of a trap that does not map to RuntimeError.

gahaas:
I changed Semantics.md now so that stack overflows are not called traps anymore, and I introduced a separate section in JS.md to define WebAssembly stack overflows.

rossberg-chromium:
lgtm

PR #877, opened 2016-11-22
Add Module.sections

lukewagner:
This PR adds a reflective static `Module` function for returning the contents of user-defined sections.  The intended use case is dynamic linking: after compiling a `Module` which is to be dynamically linked to other instances via shared `Memory`/`Table`, the linker needs metadata to know, e.g., how much space to allocate for global data and `Table` entries (including zero/empty entries, so simply exposing the length of `data` and `elem` sections isn't sufficient).  Providing access to metadata through user-defined sections avoids the need for some out-of-band metadata source or the need to do ad hoc binary decoding at runtime.

Two interesting details:
* this proposal only returns *user-defined* sections (known sections should have separate reflective methods like `exports`/`imports`).  However, by rejecting `sectionName` arguments that aren't already a string, we leave open the possibility of passing Number values in the future to extract known sections
* because sections are not necessarily unique, the function returns a (possibly-empty) array of results when querying a given user-defined section name

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -205,12 +205,12 @@ to the Object `{ module: String(i.module_name), name: String(i.item_name), kind:
 
 Note: other fields like `signature` may be added in the future.
 
-### `WebAssembly.Module.sections`
+### `WebAssembly.Module.userSections`
 
-The `sections` function has the signature:
+The `userSections` function has the signature:
 
 ```
-Array sections(modueObject, sectionName)
+Array userSections(modueObject, sectionName)
 ```
 
 If `moduleObject` is not a `WebAssembly.Module`, a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
--- a/JS.md
+++ b/JS.md
@@ -221,8 +221,8 @@ Let `sectionNameString` be the result of [`ToString`](https://tc39.github.io/ecm
 This function returns an `Array` produced by mapping each
 [user-defined section](BinaryEncoding.md#high-level-structure) (i.e., section with
 `id` 0) whose `name` field ([decoded as UTF-8](Web.md#names)) is equal to
-`sectionNameString` to an `ArrayBuffer` containing the section's `payload_data`.
-(Note: `payload_data` does not include `name` or `name_len`.)
+`sectionNameString` to an `ArrayBuffer` containing a copy of the section's
+`payload_data`. (Note: `payload_data` does not include `name` or `name_len`.)
 
 ### Structured Clone of a `WebAssembly.Module`
 
--- a/JS.md
+++ b/JS.md
@@ -216,13 +216,12 @@ Array userSections(modueObject, sectionName)
 If `moduleObject` is not a `WebAssembly.Module`, a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
 is thrown.
 
-If `Type(sectionName)` is not String, a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
-is thrown.
+Let `sectionNameString` be the result of [`ToString`](https://tc39.github.io/ecma262/#sec-tostring)(`sectionName`).
 
 This function returns an `Array` produced by mapping each
 [user-defined section](BinaryEncoding.md#high-level-structure) (i.e., section with
 `id` 0) whose `name` field ([decoded as UTF-8](Web.md#names)) is equal to
-`sectionName` to an `ArrayBuffer` containing the section's `payload_data`.
+`sectionNameString` to an `ArrayBuffer` containing the section's `payload_data`.
 (Note: `payload_data` does not include `name` or `name_len`.)
 
 ### Structured Clone of a `WebAssembly.Module`
--- a/JS.md
+++ b/JS.md
@@ -205,6 +205,26 @@ to the Object `{ module: String(i.module_name), name: String(i.item_name), kind:
 
 Note: other fields like `signature` may be added in the future.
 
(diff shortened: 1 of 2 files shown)


JS.md
```diff
@@ -205,6 +205,26 @@ to the Object `{ module: String(i.module_name), name: String(i.item_name), kind:
...
 
+### `WebAssembly.Module.sections`
+
+The `sections` function has the signature:
+
+```
+Array sections(modueObject, sectionName)
+```
+
+If `moduleObject` is not a `WebAssembly.Module` instance, a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
```
rossberg-chromium: Perhaps s/instance/object/, to avoid confusion

rossberg-chromium:
Looks good, the only thing I wonder about is the method name, which suggests that it can return any kind of section. Would `userSections` be more adequate? As you say, known sections should rather get their own methods, so I wouldn't worry about extending this one. Even if we ever wanted to return raw known sections, that would probably still be addressed better with a separate method (that e.g. takes a symbolic section name instead of a meaningless number).

JS.md
lukewagner: Hey, yeah, that is confusing.  In other places we just say "if `x` is not a `Y`" (no "instance" or "object"), so I'll do that here too.

lukewagner:
Hah, I was going back and forth on exactly this.  That's a good point re: known sections and symbolic names, though.  I'll switch.

lukewagner:
(I also put back the `ToString` on the `sectionName` argument, not because I want all the crazy coercions it allows, but just for regularity since it seems everything string-taking does a `ToString`.)

rossberg-chromium:
LGTM

JS.md
```diff
@@ -205,6 +205,25 @@ to the Object `{ module: String(i.module_name), name: String(i.item_name), kind:
 
 Note: other fields like `signature` may be added in the future.
 
+### `WebAssembly.Module.userSections`
```
jfbastien: We call them "unknown" sections in the binary format. I'd rather be consistent (either name is fine).

JS.md
```diff
@@ -205,6 +205,25 @@ to the Object `{ module: String(i.module_name), name: String(i.item_name), kind:
...
+If `moduleObject` is not a `WebAssembly.Module`, a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
+is thrown.
+
+Let `sectionNameString` be the result of [`ToString`](https://tc39.github.io/ecma262/#sec-tostring)(`sectionName`).
+
+This function returns an `Array` produced by mapping each
+[user-defined section](BinaryEncoding.md#high-level-structure) (i.e., section with
+`id` 0) whose `name` field ([decoded as UTF-8](Web.md#names)) is equal to
+`sectionNameString` to an `ArrayBuffer` containing the section's `payload_data`.
+(Note: `payload_data` does not include `name` or `name_len`.)
```
jfbastien: This `Array` is a copy of the original content, correct?

JS.md
```diff
@@ -205,6 +205,25 @@ to the Object `{ module: String(i.module_name), name: String(i.item_name), kind:
 
 Note: other fields like `signature` may be added in the future.
 
+### `WebAssembly.Module.userSections`
+
+The `userSections` function has the signature:
+
+```
+Array userSections(modueObject, sectionName)
```
jfbastien: Typo "modueObject".

JS.md
lukewagner: Ugh, I hadn't noticed; it seems we call them *both* user-defined and unknown sections.  Seems like we should just go with one.  I actually like "unknown" better than "user defined"; it contrasts better with "known".  @rossberg-chromium ?

JS.md
lukewagner: Well, technically `Array`*`Buffer`*, but yes.  I'll add "copy of" to clarify.

lukewagner:
@jfbastien Agreed, I was going to wait as well.

JS.md
jfbastien: TBH I don't like the term "user". Seems condescending.
But "unknown" is weird.
So yeah, I don't care for either name ;-)

How about "named sections" everywhere?

lukewagner: Yeah, I have the same vague feeling re "user".  I like "named".

rossberg-chromium:
"Unknown" is weird, "named" a bit too generic. How about "custom"?

JS.md
sbc100: We moved to calling these "user" sections in wabt, and in my proposed llvm binutils patch.

I think "user" is fine, and has more meaning than either "named" or "unkown" which don't seem to convey much to me.   We're all used to "user_data" pointers in callbacks.

JS.md
binji: I was wondering about this. Why make a copy? Don't we have the original buffer copied already (to prevent user from modifying it while compiling)? Maybe not a big deal, but it seems wasteful when we could just provide a view over the existing copy.

lukewagner: If we had the ability to return a read-only view, I'd do that, but otherwise we're in trouble if the user tries to write to the view.

lukewagner:
"Custom sections" sounds fine to me.  Anyone else?

JS.md
binji: Hm, too bad. Though I suppose users who are concerned about this cost for very large modules can parse the module themselves.

lukewagner:
Updated, PTAL.

jfbastien:
lgtm

binji:
Custom is fine. But wouldn't you prefer "Bespoke"? :-)

rossberg-chromium:
lgtm

PR #880, opened 2016-11-28
simplify exports object

lukewagner:
This PR removes the big hand-wavy hole in the middle of instantiation which is the creation of the Module Namespace Objects/Module Records as the `exports` property of `Instance` objects.  The reasoning is that Module Namespace objects are actually pretty magical proxies that deal some ES6-isms that we can avoid in wasm (in particular, when we allow importing/exporting mutable globals, it would be through a new `WebAssembly.Global` object (which contained the mutable value) and thus the property value of an exported mutable global would be a *constant* reference to a `WebAssembly.Global`, not the mutable value itself, which is the source of the ES6 complexity, iiuc).

Currently in SM (and I believe other engines) we just create a plain object for the exports object.  This PR changes that to be a frozen object with a null prototype.  This is almost mostly behaviorally equivalent to a Module Namespace Object but much simpler while allowing us to do fancier things with proxies in the future, if need be, without having to preserve extensibility.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -231,11 +231,6 @@ internal slot:
 * [[Instance]] : an [`Instance.instance`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/instance.ml#L17)
   which is the WebAssembly spec definition of an instance
 
-as well as one plain data property (configurable, writable, enumerable)
-added by the constructor:
-
-* exports : a [Module Namespace Object](http://tc39.github.io/ecma262/#sec-module-namespace-objects)
-
 ### `WebAssembly.Instance` Constructor
 
 The `WebAssembly.Instance` constructor has the signature:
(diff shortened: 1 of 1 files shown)


jfbastien:
I'm not sure I understand: you *don't* want the `Module Record` anymore? Why?
IIUC this made is easier to interact with ES6 modules, no?

Or are you suggesting this conservative spec (frozen, etc) which we can integrate with ES6 modules later, using `Module Record` and friends?

lukewagner:
Right, this is a conservative starting point.  One path forward would be to fully define and produce real Module Namespace Objects.  But another path would be to just add wasm modules into the module loading algorithm directly (not relying on wasm modules producing Namespace Objects) which may end up being necessary anyhow for other reasons.  User code shouldn't really notice one way or another if all they can see is this frozen object with the obvious properties/values.

rossberg-chromium:
LGTM. I'm in favour of this simplification, as it is conservative. At least right now, using a Module Namespace Object has no observable benefit but unclear implications and less than fitting semantics.

jfbastien:
If this proposal means we can implement this as either a Module Namespace Object or a different object with the properties you describe, then I'm OK with this.

lukewagner:
Yes, iiuc, that would be a valid implementation given that Module Namespace Objects present a non-extensible, non-writable, non-configurable set of properties.

jfbastien:
@lukewagner except TC39 just decided to remove its `@@iterator`. (see https://github.com/tc39/ecma262/issues/693 and https://github.com/tc39/ecma262/issues/710).

lukewagner:
@jfbastien Oh good, objects don't have one either.

jfbastien:
@lukewagner indeed, but IIUC we still need `.keys()` on it, no? I think your change removes that.

lukewagner:
With the change you linked to, both objects and Module Namespace Objects would both need `Object.keys()` to iterate keys, so this is not a source of incompatibility iiuc.

lukewagner:
Ok, so we're good then?

jfbastien:
Yup.

PR #887, opened 2016-12-02
Change exported function name to string

### LINKED ISSUE #883: Export JSFunctions should have string names
I just noticed that in JS.md, we specify that the exported Function exotic objects, we specify the name of a function to be the index as a number. That seems to be at odds with ECMAScript, which seems to require that all function names be either a string or a symbol.

Should we change JS.md to specific the function index is stored as a string?

lukewagner:
This PR fixes #883 by making the `.name` of a function be the stringified function index.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -397,14 +397,14 @@ Return `instanceObject`.
 
 ## Exported Function Exotic Objects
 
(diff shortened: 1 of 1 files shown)


rossberg-chromium:
lgtm, thanks

titzer:
lgtm

PR #890, opened 2016-12-08
Specify that inconsistent limits cause a RangeError

rossberg-chromium:
Both JS `Memory` and `Table` constructors require that the upper limit, if any is no smaller than the lower, otherwise a `RangeError` is produced.


titzer:
lgtm

jfbastien:
lgtm, addresses some of the issues from #889.

PR #892, opened 2016-12-08
Non-canonical LEB128

jfbastien:
We've discussed this before, but the spec doesn't mention it.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -53,6 +53,15 @@ represented by _at most_ ceil(_N_/7) bytes that may contain padding `0x80` or `0
 
 Note: Currently, the only sizes used are `varint7`, `varint32` and `varint64`.
 
+### Non-canonical LEB128
+
+Both `varuintN` and `varintN` can be encoded with non-canonical LEB128 values where:
+
+* non-significant zero `0x80` bytes are present in an over-large encoding; and / or
+* non-significant LEB128 bits are ignored.
+
+In both cases, the _N_ bit limitation applies.
+
 ## Language Types
 
 All types are distinguished by a negative `varint7` values that is the first byte of their encoding (representing a type constructor):
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -57,8 +57,8 @@ Note: Currently, the only sizes used are `varint7`, `varint32` and `varint64`.
 
 Both `varuintN` and `varintN` can be encoded with non-canonical LEB128 values where:
 
-* non-significant zero `0x80` bytes are present in an over-large encoding; and / or
-* non-significant LEB128 bits are ignored.
+* non-relevant zero `0x80` bytes are present in an over-large encoding; and / or
+* non-relevant LEB128 bits (bits past the size) are ignored.
 
 In both cases, the _N_ bit limitation applies.
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -55,12 +55,12 @@ Note: Currently, the only sizes used are `varint7`, `varint32` and `varint64`.
 
 ### Non-canonical LEB128
 
-Both `varuintN` and `varintN` can be encoded with non-canonical LEB128 values where:
+Both `varuintN` and `varintN` can be encoded with *non-canonical* LEB128 values where:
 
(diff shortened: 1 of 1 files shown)


BinaryEncoding.md
```diff
@@ -53,6 +53,15 @@ represented by _at most_ ceil(_N_/7) bytes that may contain padding `0x80` or `0
...
 Note: Currently, the only sizes used are `varint7`, `varint32` and `varint64`.
 
+### Non-canonical LEB128
+
+Both `varuintN` and `varintN` can be encoded with non-canonical LEB128 values where:
+
+* non-significant zero `0x80` bytes are present in an over-large encoding; and / or
+* non-significant LEB128 bits are ignored.
+
+In both cases, the _N_ bit limitation applies.
```
binji: An additional constraint (AIUI) is that the maximum length of a valid non-canonical LEB128 is equal to the maximum length of the canonical LEB128. So, for example, varint32 has a max length of 5 bytes so a non-canonical varint32 cannot be 6 bytes long.

binji: Well, I interpret this as saying the maximum value must be a 32-bit value (say). So a 6-byte LEB128 that decodes to a value < 2**32 would be valid.

binji:
Oh, I just noticed this actually is mentioned, including the maximum length:

> A LEB128 variable-length integer, limited to N bits (i.e., the values [0, 2^N-1]), represented by at most ceil(N/7) bytes that **may contain padding 0x80 bytes**.

> A Signed LEB128 variable-length integer, limited to N bits (i.e., the values [-2^(N-1), +2^(N-1)-1]), represented by at most ceil(N/7) bytes that **may contain padding 0x80 or 0xFF bytes**.

rossberg-chromium:
Yes, as @binji points out, this is already stated. No need to add anything.

jfbastien:
That's half of the non-canonical stuff. It's missing the other half. And the maximum length. I can put everything in one section if that's clearer.

binji:
What other half? And it has the maximum length part too (`ceil(N/7)` bytes). The only thing I see that's missing is that the padding will actually be 0 or 0x7f for the last byte.

jfbastien:
"non-relevant LEB128 bits (bits past the size) are ignored"

binji:
Ah, but I think that's incorrect. If we want to extend a varint (from varint32 to a varint64, say) in the future, then it's important that the bits past the size are a zero-extension (for LEB) or sign-extension (for SLEB) of the most significant bit.

jfbastien:
Is it incorrect? That's the main reason I opened this :)
The wikipedia entry were refer to so authoritatively leads me to believe past-the-end bits can be anything!

binji:
Hm, I don't read it that way:
> To encode an unsigned number using unsigned LEB128 first represent the number in binary. Then zero extend the number up to a multiple of 7 bits...

> A signed number is represented similarly, except that the two's complement number is sign extended up to a multiple of 7 bits...

Since it's extended up to a multiple of 7 bits, it should only have zeroes or ones past the size.

Anyway, I think it's pretty important that the bits are not ignored. For example, in https://github.com/WebAssembly/design/pull/895 you changed the resizable_limits flags from a varuint32 to a varuint1. If we ignore the bits past the size, then we can't safely extend the value.

lukewagner:
@jfbastien If I'm reading the `varuintN` and `varintN` sections correctly, they are already defining their respective types as a restriction of LEB128 and no bytes are ignored.

jfbastien:
@binji @lukewagner are you saying that non-canonical LEB128 can *only* be extra long (either zero or sign extended, up to max size), and that insignificant bits are and should be disallowed?

rossberg-chromium:
@jfbastien, that was the intention of the text, and it still reads that way to me.

lukewagner:
Agreed, assuming "insignificant" means "bits that otherwise wouldn't fit in the target `uintN`".

lukewagner:
@jfbastien So I think we can close this out now?  Or perhaps you'd like to create PR that clarifies the wording?

jfbastien:
@lukewagner yeah I'll update the PR to clarify wording. Soon :)

sunfishcode:
As discussed above, [this is already documented](https://github.com/WebAssembly/design/pull/892#issuecomment-265958714), so there's no need to add anything.

PR #893, opened 2016-12-09
Instance should assert globals are immutable

kmiller68:
As far as I can tell, validation of the Imports section guarantees that all imported globals are immutable. It's unnecessary to have an explicit check in the Instance constructor.


rossberg-chromium:
LGTM

lukewagner:
lgtm

Issue #778, opened 2016-08-29
Terminator operators that imply "end"


AndrewScheidecker:
WebAssembly has several operators that imply that any code following them until the next `end` or `else` operator will never be executed: `ret`, `br`, `br_table`, `unreachable`.

I propose that these operators act as implicit `end`/`else` operators. This would imply that they may only occur at the end of a control flow block. The `end` operator would still be useful as a special-case `br` operator that targets a fixed depth of 0.

This also makes it clear that `else` can be eliminated, and that an `if` operator just needs to be followed by two terminator instructions.

I believe this simplifies code generation, since you don't need to worry about the stack having the right type going into an unreachable `end` operator. I believe it also simplifies the spec and validation, since it completely avoids the question of how to handle code between these operators and the next `end`.

JSStats:
There seems to be some merit exploring this. The type validation of dead code does appear to be a problem when relating the wasm code back to a presentation format, and this might help.

Would this mean that `end == br 0` except that the branch currently unwinds the values stack?

This might help reduce code size when a block fall through needs to unwind the values stack and end the block with a branch, so it can omit the `end` in this case.

Then would `if` with an `eise` be equivalent to `block block br_if`, and `if` without an else be equivalent to `block br_if`?

JSStats:
Correction:

*  `<cond>, if, <A>, end` might then be equivalent to `block<0>, <cond>, br_unless 0, <A>, br 0` or `block<0>, <cond>, br_unless 0, <A>, end` where `end == br 0`.

* `<cond>, if, <A>, else, <B>, end` might then be equivalent to `block<n>, block<0>, <cond>, br_unless 0, <A>, br 1, <B>, br 0` or `block<n>, block<0>, <cond>, br_unless 0, <A>, else, <B>, end` where `else == br 1`.

Perhaps `if` could be defined as effectively `block<0>, <cond>, br_unless`, and if-else could require an explicit `block<n>` around this. This might go some way to addressing the code bloat when adding types to blocks as there would be a short opcode for this common case.

 E.g. `<cond>, if<n>, <A>, else, <B>, end` => `block<n>, <cond>, if, <A>, br 1, <B>, br 0` or `block<n>, <cond>, if, <A>, else, <B>, end`

AndrewScheidecker:
> Would this mean that end == br 0 except that the branch currently unwinds the values stack?

[...]

PR #894, opened 2016-12-09
Explain validation and checking after branches

rossberg-chromium:
Validation is currently only defined in the spec, which is not approachable by everybody. This PR describes the basic idea.

Based on various online and offline discussions it also makes the handling of unreachable code more precise, trying to address a couple of concerns that keep coming up with the status quo, namely:

1. it allows implementations to diverge to some degree, by not requiring them to fully check,
2. it is not easy to understand for everybody what that lenience implies,
3. non-lenient validators need a certain amount of inference, especially with potential future feature like `dup`.

The description in this PR makes the treatment of code after branches precise (removing any lenience) and tries to adopt the simplest possible rules. For most implementations, it should not imply any real change. The spec interpreter and the test suite would need some adjustment (removing the "soft failure" mode).

EDIT: Changed PR to reflect new consensus to require full checking.

### DIFF
--- a/Semantics.md
+++ b/Semantics.md
@@ -688,14 +688,17 @@ outside the range which rounds to an integer in range) traps.
 ## Validation
 
 A module binary has to be _validated_ before it is compiled.
-Validation ensures that the module is well-defined and that its code cannot exhibit any undefined behaviour.
+Validation ensures that the module is well-defined and that its code cannot exhibit any undefined behavior.
 In particular, along with some runtime checks, it ensures that no program can access or corrupt memory it does not own.
 
 Validation of code is mostly defined in terms of type-checking the use of the operand stack.
 It sequentially checks for each that the expected operands can be popped from the stack, and tracks which new operands are pushed onto it. At the start of a function the stack is empty; at its end it has to match the return type of the function. In addition, instructions inside a `block` (or `loop` or `if`) cannot consume operands pushed outside; at the end of the block the remaining inner operands must match the block signature.
 
-A special case are instructions performing an unconditional control flow transfer (`br`, `br_table`, `return`, `unreachable`), because execution never proceeds after them.
-The stack after such an instruction is defined as _unreachable_.
-An unreachable stack is valid input for any instruction, and an instruction with unreachable input will produce unreachable output.
-An unreachable stack also matches any possible signature at the end of a block or function.
+A special case are instructions performing an unconditional control transfer (`br`, `br_table`, `return`, `unreachable`), because execution never proceeds after them.
+The stack after such an instruction is unconstrained, und thus said to be _polymorphic_.
+Instructions after an unconditional control transfer still need to type-check,
+but conceptually, values of any type can be popped off a polymorphic stack for the sake of checking consecutive instructions.
+A polymophic stack also matches any possible signature at the end of a block or function.
 After the end of a block, the stack is determined by the block signature and the stack before the block.
+
+The details of validation are currently defined by the [spec interpreter](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/valid.ml#L162).
--- a/Semantics.md
+++ b/Semantics.md
@@ -683,3 +683,19 @@ outside the range which rounds to an integer in range) traps.
 [future types]: FutureFeatures.md#more-table-operators-and-types
 [future large pages]: FutureFeatures.md#large-page-support
 [future ieee 754]: FutureFeatures.md#full-ieee-754-2008-conformance
+
+
+## Validation
+
+A module binary has to be _validated_ before it is compiled.
+Validation ensures that the module is well-defined and that its code cannot exhibit any undefined behaviour.
+In particular, along with some runtime checks, it ensures that no program can access or corrupt memory it does not own.
+
+Validation of code is mostly defined in terms of type-checking the use of the operand stack.
+It sequentially checks for each that the expected operands can be popped from the stack, and tracks which new operands are pushed onto it. At the start of a function the stack is empty; at its end it has to match the return type of the function. In addition, instructions inside a `block` (or `loop` or `if`) cannot consume operands pushed outside; at the end of the block the remaining inner operands must match the block signature.
+
+A special case are instructions performing an unconditional control flow transfer (`br`, `br_table`, `return`, `unreachable`), because execution never proceeds after them.
+The stack after such an instruction is defined as _unreachable_.
+An unreachable stack is valid input for any instruction, and an instruction with unreachable input will produce unreachable output.
+An unreachable stack also matches any possible signature at the end of a block or function.
+After the end of a block, the stack is determined by the block signature and the stack before the block.
--- a/Semantics.md
+++ b/Semantics.md
@@ -692,11 +692,11 @@ Validation ensures that the module is well-defined and that its code cannot exhi
 In particular, along with some runtime checks, it ensures that no program can access or corrupt memory it does not own.
 
 Validation of code is mostly defined in terms of type-checking the use of the operand stack.
-It sequentially checks for each that the expected operands can be popped from the stack, and tracks which new operands are pushed onto it. At the start of a function the stack is empty; at its end it has to match the return type of the function. In addition, instructions inside a `block` (or `loop` or `if`) cannot consume operands pushed outside; at the end of the block the remaining inner operands must match the block signature.
+It sequentially checks for each instruction that the expected operands can be popped from the stack, and tracks which new operands are pushed onto it. At the start of a function the stack is empty; at its end it has to match the return type of the function. In addition, instructions inside a `block` (or `loop` or `if`) cannot consume operands pushed outside; at the end of the block the remaining inner operands must match the block signature.
 
 A special case are instructions performing an unconditional control transfer (`br`, `br_table`, `return`, `unreachable`), because execution never proceeds after them.
-The stack after such an instruction is unconstrained, und thus said to be _polymorphic_.
-Instructions after an unconditional control transfer still need to type-check,
+The stack after such an instruction is unconstrained, and thus said to be _polymorphic_.
+The following instructions still need to type-check,
 but conceptually, values of any type can be popped off a polymorphic stack for the sake of checking consecutive instructions.
 A polymophic stack also matches any possible signature at the end of a block or function.
 After the end of a block, the stack is determined by the block signature and the stack before the block.
--- a/Semantics.md
+++ b/Semantics.md
@@ -687,16 +687,16 @@ outside the range which rounds to an integer in range) traps.
 
 ## Validation
 
-A module binary has to be _validated_ before it is compiled.
+A module binary must be _validated_ before it is compiled.
 Validation ensures that the module is well-defined and that its code cannot exhibit any undefined behavior.
-In particular, along with some runtime checks, it ensures that no program can access or corrupt memory it does not own.
+In particular, along with some runtime checks, this ensures that no program can access or corrupt memory it does not own.
 
 Validation of code is mostly defined in terms of type-checking the use of the operand stack.
-It sequentially checks for each instruction that the expected operands can be popped from the stack, and tracks which new operands are pushed onto it. At the start of a function the stack is empty; at its end it has to match the return type of the function. In addition, instructions inside a `block` (or `loop` or `if`) cannot consume operands pushed outside; at the end of the block the remaining inner operands must match the block signature.
+It sequentially checks for each instruction that the expected operands can be popped from the stack and tracks which new operands are pushed onto it. At the start of a function the stack is empty; at its end it must match the return type of the function. In addition, instructions inside a `block` (or `loop` or `if`) cannot consume operands pushed outside. At the end of the block the remaining inner operands must match the block signature.
 
-A special case are instructions performing an unconditional control transfer (`br`, `br_table`, `return`, `unreachable`), because execution never proceeds after them.
+A special case is unconditional control transfers (`br`, `br_table`, `return`, `unreachable`), because execution never proceeds after them.
 The stack after such an instruction is unconstrained, and thus said to be _polymorphic_.
-The following instructions still need to type-check,
+The following instructions still must type-check,
 but conceptually, values of any type can be popped off a polymorphic stack for the sake of checking consecutive instructions.
 A polymophic stack also matches any possible signature at the end of a block or function.
 After the end of a block, the stack is determined by the block signature and the stack before the block.


rossberg-chromium:
To add to the OP, the goals here are:

- addressing the existing concerns,
- while minimising changes,
- and not requiring producers to care about reachability.

Semantics.md
```diff
@@ -683,3 +683,19 @@ outside the range which rounds to an integer in range) traps.
...
+
+Validation of code is mostly defined in terms of type-checking the use of the operand stack.
+It sequentially checks for each that the expected operands can be popped from the stack, and tracks which new operands are pushed onto it. At the start of a function the stack is empty; at its end it has to match the return type of the function. In addition, instructions inside a `block` (or `loop` or `if`) cannot consume operands pushed outside; at the end of the block the remaining inner operands must match the block signature.
+
+A special case are instructions performing an unconditional control flow transfer (`br`, `br_table`, `return`, `unreachable`), because execution never proceeds after them.
+The stack after such an instruction is defined as _unreachable_.
+An unreachable stack is valid input for any instruction, and an instruction with unreachable input will produce unreachable output.
+An unreachable stack also matches any possible signature at the end of a block or function.
+After the end of a block, the stack is determined by the block signature and the stack before the block.
```
binji: Maybe this could be clarified a bit, as it currently reads backward to me. How about "After the end of a block, the stack contains the contents of the stack before the block with the block's signature appended."

lukewagner:
Definitely agreed on the goals of precisely defining validation so that spec/ref-interp/engines can all precisely match, and not making significant changes (e.g., that would require a new binary version).

While this PR does avoid the major problems of the currently-spec'd polymorphic validation, it seems unfortunate to add a whole cross-cutting mode (in both the spec and all impls) just to handle the special case of instructions between an unconditional block and the following `end`/`else`, particularly because practically all wasm used in production will have no such instructions.

The alternative I prefer would be to introduce no such mode and simply reject any instructions between an unconditional branch and the following `end`/`else`.  This option:
* is the most conservative, and could be loosened later
* should accept practically all wasm output today
* does not prohibit unreachable code
* requires simple syntax-directed compilers to either use a flag that inhibits codegen (set when emitting an unconditional branch and unset when the block is closed) or to wrap unconditional branches in `block`/`end`

The only downside I'm aware of is that syntax-directed compilers would have to do the above instead of ignoring the issue entirely.  Given that the syntax-directed compiler's choices are simple and that optimal syntax-directed translation is not thus far a high-level goal, I think that this is the right tradeoff.  I'd be interested to hear what others think about this tradeoff.

Either way, though, I agree we need to nail this down so we're validating the exact same thing.  [I'm afraid I'll be on PTO for the next week so sorry in advance for the week delay in response.]

lukewagner:
One other thing I forgot to mention: having implemented the more general version of this PR in SM (unreachable-mode continues out of blocks), we can confirm it is a source of impl bugs in practice, particularly when some code that has been given validated-wasm forgets about the special cases introduced by unreachable-mode.

Semantics.md
```diff
@@ -683,3 +683,19 @@ outside the range which rounds to an integer in range) traps.
 [future types]: FutureFeatures.md#more-table-operators-and-types
 [future large pages]: FutureFeatures.md#large-page-support
 [future ieee 754]: FutureFeatures.md#full-ieee-754-2008-conformance
+
+
+## Validation
+
+A module binary has to be _validated_ before it is compiled.
+Validation ensures that the module is well-defined and that its code cannot exhibit any undefined behaviour.
```
jfbastien: "behavior" is the spelling throughout this repo.

titzer:
This has been a discussed a bit offline as well, so I'll throw in my 2 cents here for the record.

Although we haven't made supporting syntax-directed translation a first class goal for WASM, I think we should avoid introducing undue cost on simple producers. If WASM is successful there will likely be hundreds of producers want to generate WASM from various representations, many of whom are targeting JavaScript now. The clarification in this PR preserves that ability for producers to not worry about unreachable code, a property that the previous compromise had. The suggestions from @lukewagner that we should disallow unreachable code do exact some cost and don't preserve those properties.

So lgtm from my side.

jfbastien:
@titzer I'm not sure I understand your answer. IIUC you're saying " let's allow 'dumb' producers, and push the validation cost onto wasm engines". Correct?

kmiller68:
I think this needs to clarify what happens with unreachable control flow. The current OCaml implementation has bizarre behavior in this area. There are, at least, two points that I dislike. First, an unreachable `if` still reads an `i32` condition from the stack prior to the unreachable code. For example, the following does not compile:
<pre>(module
 (func (export "foo")
  (unreachable)
  (i32.const 1)
  (if i32 (then i32.const 1))
 ) 
)</pre>Produces the following error message: "type mismatch: operator requires [i32] but stack has []"
 Then there is the issue that the OCaml code validates (as normal code) any interior block in the unreachable code. For example:
<pre>(module
 (func (export "foo")
  (i32.const 1)
  (unreachable)
  (if i32 (then f32.const 1))
 )
)</pre>
Produces the following error message: "invalid module: type mismatch: operator requires [i32] but stack has [f32]".

binji:
@kmiller68 interesting, that behavior seems to be at odds with what is described in the text here. FWIW, wabt validates both of these modules, and AFAIK implements the behavior described.

kmiller68:
@binji I mostly agree that it differs from what is specified here. Although, I would prefer if the text was explicit in saying that the structure of unreachable control flow is still preserved. i.e. each `if`, `block`, and `loop` has an associated `end`. 

Also, it would be good if this explicitly explained whether or not `else` in unreachable code must be associated with an `if` when nested inside other control flow. i.e. is the validation of unreachable control flow just a count of the number of `end`s we need to look for. This has the advantage of not requiring modification on an actual control flow stack while decoding unreachable code.

Also, Do we decode immediates from the opcodes we parse? I assume the answer is yes since it's going to be messy if an immediate is the same as the `end` opcode. 

Also, Do we allow garbage instructions?

sunfishcode:
[Luke's proposal](https://github.com/WebAssembly/design/pull/894#issuecomment-266178431) could evolve in a backwards-compatible way to Andreas' proposal if desired. Andreas' proposal could not evolve in a backwards-compatible way to Luke's proposal.

Given where we are, and how we still have confusion about both the current spec and the proposal here, I support Luke's proposal.

Luke's proposal is inconvenient for some kinds of producers (and it's inconvenient for LLVM too, fwiw), however it's still usable, and it gives us some time and flexibility to think about the problem and design for it.

Cellule:
I can attest that from a consumer point of view, this unreachable mode is a big source of bugs and it took a lot of iterations to get it right.
Then again, the way I think we got it right is from the spec test, but as @kmiller68 showed, even the spec interpreter has some bugs there.

I don't really know the impact Luke's proposal could have on producers, but I can say it would make it much easier on consumer and also allow to loosen this in the future.
However, I am still a little confused by these 2 statement
> and simply reject any instructions between an unconditional branch and the following end/else
> - does not prohibit unreachable code

If I understand this correctly, we would reject unreachable code within a block, but there could be unreachable blocks like the following
```
(module
  (func
    (block
      (block (br 1))
      (block
        ;; This is unreachable, but not in unreachable mode
        (drop (i32.const 0))
      )
    )
  )
)
```

pipcet:
@sunfishcode I agree with you entirely. I think it's interesting that my asm.js backend already required a variation of Luke's "inhibit emission" flag: SpiderMonkey would complain about unreachable code following a return statement, so it was better not to generate any. So this isn't totally new.

wllang:
Some context:

* There was discussion of not allowing code between an unconditional branch and a block end in https://github.com/WebAssembly/design/issues/778 and then optimizing the encoding around that. It was pointed out by @sunfishcode code that if a producer still wanted to insert dead code they could do so by wrapping it in another block.

* There was also some discussion in https://github.com/WebAssembly/spec/pull/345 where the spec changes landed. There was an objection to landing those changes that was not resolved, it landed without consensus, and an alternative proposal was put forward.

Lets look at some of the technical realities.

* There are good reasons why the language should support a block ending with an unconditional branch etc, this is a requirement. This can create a conflicting type at the end of the block for the fall-through and it would be a burden to have to push values after the branch just to have a valid fall-through state, so an exception i need to be added for this case.

* It is technically trivial for a consumer to flag the state or look ahead one instruction to detect this case and to permit an unconditional branch as the last instruction in a block and to ignore the fall-through type check in this case.

* Surely it is technically trivial for a producer to flush code between an unconditional branch and a block end, so that the block end would immediately follow an unconditional branch, so there is no need for a runtime consumer to have to accept unreachable code in this region. Wasm could just prohibit such dead code, and better yet optimize the encoding for this semantics.

The solution was supplied in https://github.com/WebAssembly/spec/pull/345#issuecomment-254034480 both adding the above special case handling, and another option to allow dead code without introduction new type checking by simply having the unconditional branch push no operations onto the values stack (consuming the values passed out) and continuing with the regular validation rules.

This needs to be settled because it affects the optimization of the encoding.

I strongly object to the dead code validation changes introduced in https://github.com/WebAssembly/spec/pull/345 They add a considerable burden on consumers and the technical need has not been articulated.

If the use case is simply to allow a producer to avoid flushing dead code between an unconditional branch and a block end then I do not find that a sufficient use case for adding these complex dead code validation rules because flushing this dead code is trivial for the producer and a design principle of wasm is to offload complexity to the producer.

Simply ignoring dead code does not seem a great outcome, but in my opinion better than the introduced validation rules, because the producer still needs to scan the operators to determine their size and this dead code may not follow the wasm language constraints making it effectively impossible to present.

A design goal of wasm is that it be presentable in a familiar structured format. Neither allowing non-validated dead code or the dead code validation rules have been demonstrated to meet this design goal without adding a considerable burden and I have good reason to expect it would be a very significant burden to do so.

MikeHolman:
I still prefer the design @sunfishcode came up with where br ends blocks, and we avoid the unreachable state by construction. That seemed to be a very elegant solution.

With that proposal, there were concerns where it might be untenable for producers. As a consumer, this argument is not something I can debate. But it seems like Luke's proposal imposes identical constraints, while being somewhat more conservative change to semantics.

As I see it, this PR is standardizing the status quo. I like Luke's change as a second best to Dan's, but it is definitely a change.

wllang:
@MikeHolman Yes I liked that too and credit to @AndrewScheidecker too. From what I understand the producers simply have to not emit the code between the branch (or some other operator or call that has an unreachable result) and the block end which seems a small matter. For some producers it might require a second pass, but that may well just be the encoder pass.

Once that is settled we can look at optimizing the encoding for that case, to avoid emitting the `end` and how that works with `if-else-end` etc, and also eliminating the distinction between `end` and a branch out so the `end` also unwinds the stack. There seems to be a clear path for cleanup. If this is all too big a step and no agreement is in short sight then I really think we might need a more exploratory fork to build a counter proposal so that progress can be made.

titzer:
I see a potentially large can of worms in the br-is-the-end-of-a-block (br-as-end) proposal. For one, it feels uncomfortable to have a bytecode serve as the both the end and the beginning of a block, depending on context. For example,

`block ... A ... br` seems pretty straightforward
`if ... A ... br ... B ... end` is ambiguous, it would require disambiguation between `if` and `if_else`
`block block ... A ... br end` is also weird, since we can use `br` to end one block, but not the nested outer block.
`if_else ... A.. br ... B ... end` is surprising since `br` in this case not only ends the true block, but it starts the false block. 

In general it gets more and more confusing the more constructs we have. For example, adding:

`try ... catch ... end` (as currently implemented in V8)
Could legally become `try ... br ... end`, which seems pretty weird to have a try without ever finding a catch bytecode (!) One has to model the nesting stack to determine which br is actually introducing the catch....very confusing, IMO.

The alternative as Luke mentioned, is just to require an `end` (or `else`, or `catch`) immediately after a `br` or other unconditional control flow transfer. That avoids all the weird syntactic situations above at the small cost of being explicit with the `end` byte.

I don't think we can realistically redo all the toolchain and engines to make the br-as-end proposal fly before MVP launch. It'd be a lot of churn, and to be honest, I think the bytecode becomes very confusing.

titzer:
That above is not to say that I would prefer the end-must-follow-br proposal, either, it just makes the br-as-end proposal look like a lot of unnecessary confusion to establish a simple property.

rossberg-chromium:
On 12 December 2016 at 20:03, Keith Miller <notifications@github.com> wrote:

> I think this needs to clarify what happens with unreachable control flow.
> The current OCaml implementation has bizarre behavior in this area.
>

@kmiller68, good point. The proposed text says that `unreachable` doesn't
propagate _out_ of blocks, but it doesn't say whether it propagates _into_
blocks. I should clarify.

The current interpreter simply type-checks _all_ the code precisely (though
we don't require impls to do that). That is, it checks that for all control
flow edges, both ends have consistent assumptions about the stack --
regardless of whether they are reachable. That's pretty much the canonical
definition of well-typedness, not sure what's bizarre about it? (Both your
examples are essentially the same: in both the `if` has one arm whose type
doesn't match the signature, in the first case it's the else-arm, in the
other the then-arm. The `unreachable` before doesn't affect the
ill-typedness of that.)

This proposal tries to adjust the trade-off between simplicity and
completeness of type checking so that all engines can agree. Allowing some
"morally" ill-typed code is not the _goal_ but the _price_, so keeping that
fall-out minimal seems preferable.

In that light, the idea is to minimise the scope of `unreachable`. There is
no particular need to propagate it into blocks (although we could), because
all necessary type information to type-check a block is in its signature --
which was kind of the point of introducing block signatures.

Does that make sense?

wllang:
@titzer Locking in the 'end-must-follow-br proposal' would be the incremental step. If and how to follow up could be a separate discussion. It seems obvious that at least a wasm compressor would optimize its codes for the. You could look at `br 0` as equivalent to `end` and this would even work for `if-else` which would used `else` and not `br 0` to end the true branch.

I don't see the 'confusion' in some of the case you note.

* The `else` operator would need to remain and be required even if preceded by a `br`.

* `block block ... A ... br end` == `block block ... A ... br br`. The first `br` ends the inner block and the second `br` ends the outer block, it is just like two `end` operators.

rossberg-chromium:
One important thing to note is that there is no precedent, in any widely
used external code format that I'm aware of, to restrict the placement of
unreachable code in any way. That alone should make us very cautious,
especially this late in the process. Even if the restriction is just
partial, and there are ways to work around it, it is likely to trip up
producers, and surely causes extra work for many of them.

It is generally easy for consumers to ignore unreachable code, since they
gonna do the analysis anyway. It is potentially more work for producers and
tools to suppress it, because they'd all have to implement control flow
analysis of some form then. Many compilers don't, especially when peak
performance is not their primary concern.

Multiply this with numbers: we should think in terms of a world with a
producers/consumers ratio of 100:1. Where producers are very heterogeneous.
That's what we see with JavaScript right now, and Wasm regressing over the
convenience of JS as a compilation target would be really sad.

AndrewScheidecker:
I concur that producer complexity should carry more weight than consumer complexity, but I would like to note that in this case the risks of getting it wrong are also asymmetric:
* A producer that emits certain kinds of unreachable code will be rejected as invalid by a correctly implemented consumer.
* vs a consumer that incorrectly handles unreachable code is potentially a security vulnerability

I can back up what Luke said that implementing this area of the spec is error prone: my implementation had at least one potentially exploitable bug that was only uncovered by fuzz testing.

I still prefer #778, but I understand the arguments against it, and I think all of the suggestions in this thread would be an improvement on what's there now.

> You could look at br 0 as equivalent to end and this would even work for if-else which would use else and not br 0 to end the true branch.

Tangentially, with or without #778, `br 0; end` has different behavior from `end` because the `br` will implicitly drop extra stack entries while `end` requires that the stack only contains entries for the control structure's expected result.

wllang:
@AndrewScheidecker The `end` operator can (and needs to) change to unwind the stack, which would make it the same as `br 0`. There is technical no use case for restricting the stack at the `end` to only containing the valid values, and a very compelling use case for unwinding excess values otherwise `pick` and other code styles require a `br 0; end` to clean up the block.

If the design principle of wasm were changed to put more weight on producer complexity then it would be a very different language, probably not something most people came here to work on. The existing burden of writing an encoder, of optimizing the encoding, is far in excess of the burden of dropping the unreachable code between a branch and the block end. If wasm followed a design principle of minimizing the burden on producers it would have a text format with lots of aids to help the producer - it's not that.

@AndrewScheidecker What compelling arguments do you understand against https://github.com/WebAssembly/design/issues/778 ?

@rossberg-chromium Can you even explain the technical relevance of the claimed precedence?

Design work needs to be done. At least one of the Chairs has disowned responsibility for doing it. Other people are claiming the MVP is close as an excuse for not advancing development. Perhaps we should approach the organizations currently funding development and suggest a different model, such as a competition for designs meeting certain encoding efficiency and decoding performance tests, and meeting other design principles, etc.

lukewagner:
(Ok back :)

So one thing we should've been more clear about above is that both this PR and what I proposed above are different from what's in spec/interpreter (and will imply a spec/interpreter change).  What's in spec/interpreter matches what @rossberg-chromium described above, but when looking into implementing this precisely in SM, I realized that, as soon as we add `dup`/`pick`, implementing the polymorphic type checking will require some form unification that surely no will want to implement solely for dead code and so we'll be back in a situation where *noone* precisely implements the spec, and that's bad.  That much it seems we all agree on.

So this PR accepts a *superset* of what spec/interpreter accepts and what I proposed accepts a *subset*.  *Both* I think will accept all wasm out there today (because binaryen strips trivially-dead code) and thus won't require a whole binary version iteration.

Looking at the above comments: it seems we have multiple confirmations that a cross-cutting mode is indeed a non-trivial implementation hazard.  So again, given that the trivial-compiler use cases *can* express unreachable code—so we're just talking about whether they have to do a tiny bit more work—I think the conservative starting point I proposed is the right tradeoff.

@MikeHolman I agree that branches-ending-blocks is abstractly more elegant and also essentially equivalent to what I'm proposing; you can think of branches-ending-blocks as just an optimization which saves a few bytes on `end` opcodes (which wins about 0.5%, by my measurements, on AngryBots, so in "meh" territory, esp. with compression).  My main reason to prefer what I proposed is that it's a relatively trivial and conservative change whereas branches-ending-blocks is, at this late stage, a pretty big change.

wllang:
@lukewagner Thank you for the analysis and results. Yes, it seems fine to just lock in a change that makes it a validation failure if an `end` does not follow an operator that ends the control flow in a block, so the `br` and `unreachable` operators. But is progress blocked and how to move forward.

I have not checked, but it seems it would not break existing code if the `end` and `else` operators were changed to unwind the stack too. It seems this would just be making more code valid, so not break anything so that might be another easy change to get locked in, and then the final change to allow a branch to end a block is a smaller step and almost trivial.

A number of people keep noting 'at this late stage' etc as an excuse for not continuing the design work, but some of these changes have been in the system some time now, and I appeal to people that the best outcome is to just sort them out and get a great design out the door. The only other choice is to go with wasm as-is with the expectation that future runtimes will deprecate old versions, and probably at the next roll, but I don't think the community will view that well. Please lets get the design work done, it seems within sight, but it has appeared in signed for much of the year now.

wllang:
@rossberg-chromium @titzer There appear to be no substantive and reasonable grounds for the 'partially-validated-unreachable-code' design and you have not responded recently on the technical matters so could I ask if you could withdraw your demands for this design and back the 'unreachable-op-ends-block' design?  Can we have a cleaner start for the new year without have the stress of resolving this blocker.

jfbastien:
@wllang I believe many folks are on vacation and you're unlikely to get an answer soon.

wllang:
@fbastien Ok, lets adjourn on the design in this area until late Jan at least.

rossberg-chromium:
(Back too. :)

On 18 December 2016 at 02:56, Luke Wagner <notifications@github.com> wrote:

> So this PR accepts a *superset* of what spec/interpreter accepts and what
> I proposed accepts a *subset*. *Both* I think will accept all wasm out
> there today (because binaryen strips trivially-dead code) and thus won't
> require a whole binary version iteration.
>
Yes, that's a good characterisation. Another way to describe it is that the
current interpreter accepts (1) well-typed code anywhere and (2) ill-typed
code nowhere. In that sense, it represents the canonical semantics (in a
way that could be made very precise). Ben and my overarching criterion is
to maintain (1), because otherwise every producer in the world will have to
worry about reachability and dead code. I'm willing to give up (2) if
that's what it takes, because that property is less practically important.
That's what this proposal does, while yours does the opposite.

wllang:
@rossberg-chromium  Re: "Another way to describe it is that the current interpreter accepts (1) well-typed code anywhere"

The description in this PR is as follows and this is not "well-typed code":
"An unreachable stack is valid input for any instruction"
"An unreachable stack also matches any possible signature at the end of a block or function."

It certainly fails the type test of matching the expected number of values, so it does not seem very "well-typed code". I think a number of groups have expressed how this creates a separate type system just for this case, a complexity that is just not proportional to the benefit.

I did propose an alternative that would have ensured that the expected number of values were validated, but both yourself and @titzer ignored it and landed the changes anyway. That proposal would have allowed the validated unreachable code to be presented with the expected number of arguments to operators and retaining the connectivity between operators.

If you really want to validate the unreachable code then do it properly and make sure that the dataflow structure is still valid, otherwise lets just exclude unreachable code from wasm. Your solution is neither and complex and adds a burden on consumers.

wllang:
Would also add that the use case of allowing producers to include arbitrary blobs of code in unreachable areas could be handled by having them drop a position and the blob in an unnamed section and it would be efficiently ignored by the decoder.

rossberg-chromium:
@wllang, by "current interpreter" I am referring to the status quo, not this PR. As I said in my previous comment, this PR suggests dropping property (2), and as I mention in the OP, that would necessitate changing the interpreter.

wllang:
@rossberg-chromium Sorry, I am a little lost on what the state of the interpreter is and loath to dig into it, but I was referring to the text in this PR, again:
"An unreachable stack is valid input for any instruction"
"An unreachable stack also matches any possible signature at the end of a block or function."

It's not just allowing 'unreachable' type single values, rather introducing a concept of an 'unreachable stack' which is effectively a stream of 'unreachable' single values as demanded by the code and this is not a familiar concept in the context of this language. If you had just allow single value 'unreachable' types then the dataflow structure could have been validated and retained too. Obviously there would need to be an exception for the case of a block end immediately following an unreachable operator to all matching values in this casem and that was basically my counter proposal. On the balance it seems easier to just avoid the unreachable code in wasm - producers can put it in another section if they must.

rossberg-chromium:
@wllang, the point isn't that producers can generate unreachable code somewhere, but that they _don't need to care_ whether and when they are generating unreachable code. Reachability simply shouldn't matter for producers, like it doesn't for other code formats either. Unless they choose to care.

lukewagner:
> Reachability simply shouldn't matter for producers, like it doesn't for other code formats either.

Reachability may not matter for unsafe machine code, but it *does* matter for the code formats of JVM (modern, with stack maps) and CLR, both of which require their own particular dance after unconditional branches to establish the stack type to use going forward.  So you can't simply make this assertion.

Rather, we need evidence that we'll actually be deriving significant-enough value from adding a *whole new cross-cutting mode* to validation which has very real, demonstrable hazards and costs.  In the absence of evidence, I think we should start conservatively.

rossberg-chromium:
@lukewagner, I don't agree with that characterisation of the JVM. The JVM
does not care about reachability: in classic verification mode it requires
nothing, in new mode it unconditionally requires a stack map after every
branch, regardless of reachability; that may be more type-explicit, but
reachability is immaterial in both modes. The amount of annotation required
has its own drawbacks, of course.

Fair enough that the CLR is a bit weirder: it makes a rather curious
distinction between code targeted by forward branches (or sequential flow)
and code that isn't. Unreachable code is not distinguished from code
targeted by backward branches, but both may need producers' extra care in
some cases.

However, neither code format just disallows unreachable code after
branches, nor does any other, do you agree?

lukewagner:
@rossberg-chromium I think the key perspective difference is that I view what I proposed above as still totally allowing unreachable code:
```
(block ... (block ... br end) ... this is unreachable and still allowed ...)
```
Where the `end` is basically just the stack map that new-mode JVM requires after unconditional branches (the block signature being a "delta encoding").  All that's prevented is writing unreachable code at a point where the type state is not otherwise clearly defined.

wllang:
@rossberg-chromium Your response misses a key point I was try to communicate: producers already have to generate valid code but what your proposal for validation allows is code that would not validate if the unreachable operator was removed, that was the test in my proposal. Give it a try, try removing the unreachable operator (a branch or the `unreachable` operator) and see if the remaining code will still validate under your proposal - you will find that it does not because the stack might not have enough values. So if you are going to propose that code between an unreachable op and a block end is validated then please ensure it follows the existing type validation rules. Your proposal creates new encoding semantics, new validation rules, which are an extra burden for consumers (both the runtime and tools).

Also as @lukewagner notes, this is not just 'unreachable' code in general, rather a very specific encoding matter of the code between an unreachable operator and the end of a block, and it does seem fair to judge it as a trivial matter for the producer to handle, and there is no requirement that producers in need to infer and remove unreachable code in general.

rossberg-chromium:
@lukewagner, I would call that a _workaround_. From my perspective, it would be a rather silly design choice to ask producers to wrap all their branches into spurious extra blocks as the only alternative to reachability analysis. I understand that you view this as just a "type annotation", but I beg to differ. Blocks are control constructs, not type annotations. We could of course require proper type annotations for all branches, but that seems like overkill as well.

rossberg-chromium:
@wllang, yes, again, that's what I described as dropping property (2) above. I would happily keep it, but this PR is trying to match implementations more closely while addressing the concern that validation could become too involved under the "ideal" semantics. I don't actually share this concern myself, but we have to reach some compromise.

lukewagner:
@rossberg-chromium Well, we have to weigh the silliness of having toy compilers do a tiny bit of extra work (wrapping or tracking reachability) with the silliness of adding an entire cross-cutting validation mode to the spec and engines, which demonstrably leads to serious bugs, to handle a case that should never happen in practice.

rossberg-chromium:
@lukewagner, we just have to agree to disagree. For consumers. the "cross-cutting validation mode"  you allude to amounts to <20 lines of code that can be well localised. For some producers, your proposal OTOH can affect every single algorithm that generates or transforms code, and they cannot necessarily easily be funneled through a single abstraction taking care of reachability. It has _far_ more cross-cutting impact.

Also, your portrayal of all such compilers as "toys" is very dismissive. Such a property affects every non-CFG compiler, of which there are many, including production ones. After all, both asm2wasm and V8's internal asm.js-to-Wasm pipeline fall into that class. E.g. for V8, your proposal would very likely result in a net increase in complexity, because the complication in its internal producer would already outweigh any win in its validator.

lukewagner:
@rossberg-chromium "Toy" doesn't mean "non-CFG", it means "isn't stripping trivially-dead code", something that any production compiler would do for size reasons alone.  Also, it's not about how many lines of code a mode-check can be factored into; it's about how easy it is to miss it and have a security bug.  And just *adding a whole new mode*.

wllang:
@rossberg-chromium How many lines of code do you expect it would add to a consumer that is a wasm text formatter, to have to implement a text format that fits the semantics of the validation checking, it could be a good deal of code and also a burden on the human reader to comprehend for a corner case that they will not practically care about anyway.

Lets keep some perspective, this does not affect 'type annotations' or CFG etc etc, it is a simple encoding matter - at the encoding stage the producer just flags it's state as being *post-unreachable-op* and does not emit code until this flag is cleared at the end of a block. Can we agree on this point, that implementing this flag in a v8 internal producer would likely be a small matter and that in turn v8 can probably clean out other bloat in the wasm decoder. Also v8 is free to implement a non-standard wasm code variation internally, without having to agree on this here, but I guess your point was to raise an example.

I would be happy to patch v8 to try to objectively settle this point if that is necessary, but it would be a burden to do so and distract from other critical work so if this should be obvious to people engaged here then that might not be a tenable way to work if that has to be repeated over and over.

@kripken couldn't such a flag being implement in your tool to demonstrate an implementation, without the full DCE?

rossberg-chromium:
@lukewagner, I find that portrayal no more agreeable. In most cases, dead code elimination is practically irrelevant as an optimisation, and I'm pretty sure you'll find plenty of compilers that don't bother. Again, V8's asm-to-Wasm is an existing example in the Wasm space.

FWIW, the alternative you prefer implies a "whole new mode" as well, in which all instructions but `end` are an error. At least conceptually. It would not avoid the Unreachable type in the type system. And most likely a simple "mode" flag also is the easiest way to implement it.

@wllang, the point isn't what V8 could do, but that it is a concrete example of how that solution creates artificial extra work for producers. And it would presumably be worse for less trivial compilers.

wllang:
@rossberg-chromium The responses read back as not having seen the point trying to be communicated. So the key point is that it is not dead-code-elimination in general, rather could be seen as just an encoding matter. That the producer could emit the code to its encoder without having done DCE and only the encoder would need to flag and remove that code.

Why would it be 'worse for less trivial compilers', it's just a last stage encoding matter, and both simple and very complex producers need to encode to wasm binary, or perhaps the v8 producer does not but then it's not a standard path. Again such responses read back as there being a communication issue on this point.

> the alternative you prefer implies a "whole new mode" as well, in which all instructions but end are an error. At least conceptually. It would not avoid the Unreachable type in the type system. And most likely a simple "mode" flag also is the easiest way to implement it.

Look ahead to the next step, having the unreachable operator end the block, and then the problem of 'all instructions but end are an error' does not even exist.

There is no unreachable type in the type system, the `unreachable` operator does not push a value so there is no definition and thus no type.  Wasm does not have an operator that pushes an `unreachable` type definition. Perhaps you could give an example of the problem you see here to help other people understand?

Perhaps coding it up for v8 and looking at the change might be one effective way to help communicate this. If you feel people are not understanding your side then can you see from the responses what is being missed?

jfbastien:
> @lukewagner, I find that portrayal no more agreeable. In most cases, dead code elimination is practically irrelevant as an optimisation, and I'm pretty sure you'll find plenty of compilers that don't bother. Again, V8's asm-to-Wasm is an existing example in the Wasm space.

I don't think you're talking about the same thing. Code that's already been through a compiler (such as asm-to-wasm) won't usually benefit from DCE. That doesn't mean DCE is useless.

At the same time, code that is dead doesn't need to be structurally unsound. For example, LLVM has passes which create the equivalent of `if (true)` because changing the CFG in a pass isn't always desirable, it then expects a subsequent DCE pass. That doesn't mean that the content of `if (true)` is typed incorrectly: it is valid IR.

My point is: we're not talking about **all** DCE, and @lukewagner's proposal isn't about forcing all dead code from being removed.

I think his characterization of "toy" compiler is quite apt: do something quick-and-dirty, but don't create unsound results.

rossberg-chromium:
@jfbastien, sure no compiler should create unsound code, but the question under discussion is whether and why we should _arbitrarily define_ some very specific dead code as "unsound", while dead code remains fine in all other places.

jfbastien:
@rossberg-chromium how is it arbitrary? That's a pretty dismissive qualification of @lukewagner's proposal 🙃

rossberg-chromium:
@jfbastien, because we'd just define something as "unsound" although there is nothing inherently unsound about it. And because we don't do so for other cases.

titzer:
I don't think the characterization of producers as "toys" is helpful. Disallowing this type of dead code is just a gotcha for producers and there really isn't much justification for it if clients can trivially ignore it. There are a lot of situations where a producer might just want to do something quick and dirty (e.g. a JIT) and an arbitrary disallowing of perfectly valid code is just friction. The code in a consumer to check for reachability is basically the same to disallow it (i.e. check for appropriate end/else) as to suppress a type error after unconditional control transfer in a block. They both amount to a bit on entries in the control stack.

wllang:
@titzer @rossberg-chromium 

* It might be a small matter for a consumer to simply ignore the code, but ignoring the code leaves it's validation status as "sound" unknown. It might attempt to pop values from the stack that do not exist, and it might make use of definitions that do not have the correct type, and thus may not be "sound" code.

* The wasm encoding already fails to encode in the code section much code that has no effect on the execution semantics, and this dead code has no effect on the semantics too. For example a debug name for a symbol.

* If a consumed were required to validate this dead code as valid, to check that the code were valid with the unreachable operator removed then that might cross the line and be a much more significant burden, and certainly a much more significant burden than having the producer simply drop this code.

* The validation rules being proposed in the PR are also a much more significant burden for the consumer than the burden on the produce simply dropping the dead code. It is not just "to suppress a type error after unconditional control transfer in a block" rather it needs to allow invalid operations on the stack, and code with a type error is not "sound" anyway.

* People keep failing to note the burden on a consumer that formats the code in a familiar language, it can't just drop the code, and how can it present code that pops values that are not there (references non-existent definitions), and even type invalid code could be a burden. A consumer that translates the code into another related format would have a similar burden, for example a tool to convert to SSA form, it would have to materialize definitions if popping definitions not on the stack and it would have to introduce an `unreachable` type for these definitions (a type not otherwise in wasm), and converting that mess back would be a huge burden.

Thus wasm needs to either not included this dead code or ensure that it is fully validated. The middle group "compromise" proposed in this PR is a poor outcome. It looks simpler to just not include this dead code and that also appears to support some follow up optimizations of the encoding.

lukewagner:
@titzer Client's *can't* just "trivially ignore it", though.  This has been discussed above, by multiple implementors, but an additional data point is that our baseline compiler has no inherent notion of reachability, so dead-mode is actually much more of a burden than in our optimizing SSA compiler.

It does seem we're at a tough point, though, where there's simply a fundamental disagreement on the relative value and cost of both options.

wllang:
@lukewagner It appears that the Baseline compiler already discards the unreachable code, 67 lines of code access the dead-code flag. The burden there appears to be having to parse all these operators enough to be able to step over them. I don't think this burden is so great, not for a consumer that is just discarding them (even if skipping them), but for a consumer that actually has to pass them through and perhaps transform them then the burden is much greater.

Perhaps this process is not productive. I am sure the web community would appreciate consensus but we also want a good outcome, and if compromises are made and if there is no experimentation by web browser vendors and thus no diversity then perhaps the community is not going to get a great result anyway. So could I suggest building in the ability for web browser vendors to disagree and to handle some diversity in the wasm standard. So make a user defined decoder a priority and core feature of wasm, so that this decoder can probe the target and emit code specific to the target. This would allow one browser vendor to offer the dead-code-part-validated feature and others a more compact encoding that does can not encode this dead-code and producers can still get by and community feedback can settle the matter eventually.

This strategy would also set us up for other areas of incompatible experimentation, beyond just feature additions. For example, perhaps one vendor would offer a really fast compiler accepting SSA form code, and producers could have the option to encode in this form but could also translate out to other encodings for other web browsers (it's pretty quick translating out of wasm structured SSA form).

qwertie:
Fwiw, so far I am persuaded that @lukewagner's proposal to reject instructions between an unconditional branch and the following end/else is best. It's easy for consumers, not an excessive burden on producers, and could be changed in the future should we change our minds.

Has the possibility been considered of treating unconditional branches as if they were `(i32.const 1) (br_if label)` for validation purposes?

rossberg-chromium:
@qwertie, unfortunately, that equivalence doesn't answer how to deal with `return` or `unreachable`, where you often want the stack to be "cleared" or otherwise adjusted (otherwise you would need to add e.g. dead code `drop` and `const` instructions just to make the stack type match up).

wllang:
@qwertie @rossberg-chromium That was basically my proposal if allowing this dead code, to fully validate it as if the unreachable operator was not there. For the `return` operator it would pop it's arguments, and for the `unreachable` operator the values stack would not change, for `br_table` it would pop its condition argument and target values, etc.

It does seem reasonable that if a producer includes this dead code that the use of the stack should be valid. The only case we really need an exemption for (for encoding efficiency reasons) is when the block ends directly after an unreachable operator, so that a block can end with a branch or `return` without requiring following operators to fill the stack - that was the only exemption in my proposal. The block end needs to be changed to clear the stack anyway, the problem without the exemption is if it needed to fill some fall-through values of the correct type.

wllang:
Just to put another corner case on the table. Noticed the wasm llvm backend generates loops with a return type expecting a single value but for which the last operation in the loop is a branch to the repeat label (the loops exits via a `return`), see zlib `crc32_combine_`. A `loop` has no exit label, so if the fall-through is unreachable then there is no definition for the result of the loop requiring the consumer to fill some value here. This case seems to be distinctly different to the case in which the block has other exits to its label as in those cases the consumer can just ignore the fall-through value.

Thus for the counter proposal I would add that it be invalid for a block expecting values to have no branches to its label or a fall-through value. So in the case of the current `loop` it would be invalid for it to have no fall-through value on the values stack if it declared a result value.

For example the following would be invalid: `(block $l1 i32 (i32.eqz (block $l2 i32 (br $l1 (i32.const 1)))))`

This would also make the following pattern invalid `(block $l1 i32 (unreachable))` which would need to be encoded as `(block $l1 i32 (unreachable) (i32.const 0))`.

The only practical alternative would appear to be for the consumer to fill in a value in these cases if the block result is consumed, rather than requiring it to be explicit, or to introduce some 'unreachable' type which does not seem warranted for many reasons explored above.

For the llvm generated code, the `loop` would just change to return no values, but it was also the last operation in the function and the function returned an `i32`, so that creates other problems to consider - if the loop had an exit label then that could have been used to exit. The encoding I am exploring currently has the `loop` result being unreachable, requiring a loop end with a branch, and that solves that problem, but having the `loop` have an implicit branch to the repeat label would also work and will probably be a better option.

wllang:
Sorry, looking over this more suggests that `(block $l1 i32 (unreachable) (i32.const 0))` would be unworkable, e.g. would not allow flushing code between the unreachable operator and the end of the block.

So perhaps materializing definitions is a good solution, but that appears to only be necessary for declared results of block like operators.

Defining the problem: code might want to have a block with a declared return type that includes values but for which there are *no* paths that reach the target and thus no definitions to consume. For example: `(func f1 (result i32) (loop $r i32 (br $r)))`  or `(func f1 (result i32) (unreachable))`.

Proposed solution: An operator with a declared result type for which there are *no* reachable paths supplying definitions will materialize definitions for each declared result type for the purposes of validation, and for the purpose of type derivation (should that need to be well defined in this dead code region) these have the type of the value zero for the numeric types.

Implementation: A operator with an unreachable result ends a block. The decoder knows the expected result types for a block at the start of the block, and it decodes the body of the block checking that all branches to this target match the expected types, and at the end of the block if checks the last operator and if that operator has an unreachable result (`br` `return` `unreachable`) then it does not pop values for the fall-through path. After decoding the block if there are no definitions targeting the results then the decoder materializes definitions for each expected result with the type of the numeric value zero for the numeric types.

The above proposal seems to agree with the following in this PR "After the end of a block, the stack is determined by the block signature and the stack before the block." However it does not require introducing an "unreachable stack" because instructions after the unreachable operators are not allowed. Might this be the proposal of the other three web browser vendors, and might it help for them to submit a counter proposal PR to define it well?

wllang:
Sorry again for the noise, but let me add one more case to ponder. When using `pick` and an SSA style of code, the end of the block would better unwind the stack, and that raises the following case: `(block $l1 i32 (f64.const 1.0) (unreachable))` Here the `(f64.const 1.0)` remains on the stack at the end of the block but if the `unreachable` operator can not be followed by a branch to exit the block and unwind the stack then that would frustrate an SSA encoding working with operators that have unreachable results. Further, it might be tempting to just pop the `(f64.const 1.0)` definition at the end of the block (which would have the wrong type in this example on purpose) but that would not appear to be a general solution (they may be not value to pop in general), so the proposal is to materialize an `i32` result in this example and the stack would unwind so this example would be valid.

Perhaps there is some small area of common ground namely "After the end of a block, the stack is determined by the block signature and the stack before the block.", and I think everyone agrees on that, so perhaps that could be landed alone without the "unreachable stack" clauses and we have made some small progress.

rossberg-chromium:
After more offline discussion, we have converged on requiring full type-checking of unreachable code instead. So I changed this PR to reflect that.

Also added a reference to the spec interpreter to address #974.

Semantics.md
```diff
@@ -683,3 +683,22 @@ outside the range which rounds to an integer in range) traps.
...
+
+Validation of code is mostly defined in terms of type-checking the use of the operand stack.
+It sequentially checks for each that the expected operands can be popped from the stack, and tracks which new operands are pushed onto it. At the start of a function the stack is empty; at its end it has to match the return type of the function. In addition, instructions inside a `block` (or `loop` or `if`) cannot consume operands pushed outside; at the end of the block the remaining inner operands must match the block signature.
+
+A special case are instructions performing an unconditional control transfer (`br`, `br_table`, `return`, `unreachable`), because execution never proceeds after them.
+The stack after such an instruction is unconstrained, und thus said to be _polymorphic_.
+Instructions after an unconditional control transfer still need to type-check,
+but conceptually, values of any type can be popped off a polymorphic stack for the sake of checking consecutive instructions.
+A polymophic stack also matches any possible signature at the end of a block or function.
```
wllang: It does not read well. What can 'Instructions after an unconditional control transfer still need to type-check' mean if 'values of any type can be popped off a polymorphic stack'. How can they 'type-check' if the values materialized are always of the appropriate type?

Semantics.md
```diff
@@ -683,3 +683,22 @@ outside the range which rounds to an integer in range) traps.
...
+
+A module binary has to be _validated_ before it is compiled.
+Validation ensures that the module is well-defined and that its code cannot exhibit any undefined behavior.
+In particular, along with some runtime checks, it ensures that no program can access or corrupt memory it does not own.
+
+Validation of code is mostly defined in terms of type-checking the use of the operand stack.
+It sequentially checks for each that the expected operands can be popped from the stack, and tracks which new operands are pushed onto it. At the start of a function the stack is empty; at its end it has to match the return type of the function. In addition, instructions inside a `block` (or `loop` or `if`) cannot consume operands pushed outside; at the end of the block the remaining inner operands must match the block signature.
+
+A special case are instructions performing an unconditional control transfer (`br`, `br_table`, `return`, `unreachable`), because execution never proceeds after them.
+The stack after such an instruction is unconstrained, und thus said to be _polymorphic_.
```
Cellule: nit: `and thus`

Semantics.md
Cellule: From what I understand, the unconditional branch creates the polymorphic type, and thus is valid for any instructions.
If the instruction uses other values, not polymorphic, they must type-check.
Also, any following instructions still needs to correctly type check.
```
(module (func (i32.add (i64.const 1) (unreachable)))) ;; Invalid because [i64 poly] does not type-check for i32.add
(module (func (i32.add (i32.const 1) (unreachable)))) '' valid because [i32 poly] is valid for i32.add
(module (func (param i32)
  (unreachable) ;; code becomes unreachable after this
  (i32.add (i32.const 1) (get_local 0)) ;; still needs to do type-checking on this
))
```

lukewagner:
lgtm

Semantics.md
wllang: @Cellule That still does not read well, how can "The stack after such an instruction is unconstrained" yet values of specific types can be pushed onto the stack and must type check? 

Also what does "A polymophic stack also matches any possible signature at the end of a block or function." if the code after the unreachable operator is to be typed checked, are the values discarded and ignored, or must the stack have just the expected number of values?

There is the following in the spec, and this PR just references this, but honestly I find this cryptic and it is impossible to evaluate the merits of the PR in a reasonable time and it need not be take cryptic. Looking at the `push` does it fail if pushing a value to a stack with `Ellipses`, and if so that would break the checking after the unreachable operator, or am I just reading this wrong (highly likely but a good description would help).  The `Ellipses` stack property also appears sticky, so it's not clear it is doing what you expect in your examples. Sorry I don't even follow what the `pop` operation is doing, a description would help.

```
let pop (ell1, ts1) (ell2, ts2) at =
  let n1 = List.length ts1 in
  let n2 = List.length ts2 in
  let n = min n1 n2 in
  let n3 = if ell2 = Ellipses then (n1 - n) else 0 in
  check_stack ts1 (Lib.List.make n3 None @ Lib.List.drop (n2 - n) ts2) at;
  (ell2, if ell1 = Ellipses then [] else Lib.List.take (n2 - n) ts2)

let push (ell1, ts1) (ell2, ts2) =
  assert (ell1 = NoEllipses || ts2 = []);
  (if ell1 = Ellipses || ell2 = Ellipses then Ellipses else NoEllipses),
  ts2 @ ts1
```

Semantics.md
rossberg-chromium: Fixed

Semantics.md
rossberg-chromium: @Cellule, right. @wllang, this is not actually a change, but an informal explanation of what the spec interpreter and the test suite have defined as valid code for a long time. The only intended change is that we now want to require engines to check this strictly.

Agreed that the interpreter snippet is not particularly readable, because it is factored differently. A more direct pseudo code rendering of the algorithmic explanation given in this PR can be seen [here](https://docs.google.com/document/d/1JHK1wznji1LphTxNjCseeUHRq6BsGvouWChqtMZ_3Q4/) and should materialise in the prose spec soon.

Semantics.md
```diff
@@ -683,3 +683,22 @@ outside the range which rounds to an integer in range) traps.
 [future types]: FutureFeatures.md#more-table-operators-and-types
 [future large pages]: FutureFeatures.md#large-page-support
 [future ieee 754]: FutureFeatures.md#full-ieee-754-2008-conformance
+
+
+## Validation
+
+A module binary has to be _validated_ before it is compiled.
```
titzer: s/has to be/must be/

Semantics.md
```diff
@@ -683,3 +683,22 @@ outside the range which rounds to an integer in range) traps.
...
+Validation ensures that the module is well-defined and that its code cannot exhibit any undefined behavior.
+In particular, along with some runtime checks, it ensures that no program can access or corrupt memory it does not own.
```
titzer: s/it ensures/this ensures/

Otherwise the two its in the last clause could be confusing.

Semantics.md
```diff
@@ -683,3 +683,22 @@ outside the range which rounds to an integer in range) traps.
...
+
+Validation of code is mostly defined in terms of type-checking the use of the operand stack.
+It sequentially checks for each instruction that the expected operands can be popped from the stack, and tracks which new operands are pushed onto it. At the start of a function the stack is empty; at its end it has to match the return type of the function. In addition, instructions inside a `block` (or `loop` or `if`) cannot consume operands pushed outside; at the end of the block the remaining inner operands must match the block signature.
```
titzer: s/, and tracks/and tracks/

titzer: s/has to match/must match/

titzer: s/; at the end/. At the end/

Semantics.md
```diff
@@ -683,3 +683,22 @@ outside the range which rounds to an integer in range) traps.
...
+
+A special case are instructions performing an unconditional control transfer (`br`, `br_table`, `return`, `unreachable`), because execution never proceeds after them.
```
titzer: s/are/is/

titzer: s/instructions performing an unconditional control transfer/unconditional control transfers/

Semantics.md
```diff
@@ -683,3 +683,22 @@ outside the range which rounds to an integer in range) traps.
...
+The stack after such an instruction is unconstrained, and thus said to be _polymorphic_.
+The following instructions still need to type-check,
```
titzer: s/still need to/must still/

rossberg-chromium:
Comments addressed, merging with above LGTMs and off-line consent.

PR #901, opened 2016-12-12
Introduce WebAssembly.LinkError

rossberg-chromium:
In order to distinguish WebAssembly link-time failure from JS-level errors, introduce a new error class `WebAssembly.LinkError`, which arises from instantiation in analogy to `CompileError` and `RuntimeError`. It indicates link-time type errors (function signatures, memory/table bounds mismatches), data/elem initialisation errors (out-of-bounds), and other potential future error causes. Traps from the start function will still raise a `RuntimeError`.

Addresses #889.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -113,7 +113,7 @@ On success, the `Promise` is [fulfilled](http://tc39.github.io/ecma262/#sec-fulf
 with a plain JavaScript object pair `{module, instance}` containing the resulting
 `WebAssembly.Module` and `WebAssembly.Instance`. On failure, the `Promise` is
 [rejected](http://tc39.github.io/ecma262/#sec-rejectpromise) with a 
-`WebAssembly.CompileError`, `WebAssemlby.LinkError`, or `WebAssembly.RuntimeError`, depending on the cause of failure.
+`WebAssembly.CompileError`, `WebAssembly.LinkError`, or `WebAssembly.RuntimeError`, depending on the cause of failure.
 
 The asynchronous compilation is logically performed on a copy of the state of
 the given `BufferSource` captured during the call to `instantiate`; subsequent mutations
@@ -131,7 +131,7 @@ from `moduleObject` and `importObject` as described in the
 On success, the `Promise` is [fulfilled](http://tc39.github.io/ecma262/#sec-fulfillpromise)
 with the resulting `WebAssembly.Instance` object. On failure, the `Promise` is
 [rejected](http://tc39.github.io/ecma262/#sec-rejectpromise) with a 
-`WebAssembly.CompileError`, `WebAssemlby.LinkError`, or `WebAssembly.RuntimeError`, depending on the cause of failure.
+`WebAssembly.CompileError`, `WebAssembly.LinkError`, or `WebAssembly.RuntimeError`, depending on the cause of failure.
 
 ## `WebAssembly.Module` Objects
 
--- a/JS.md
+++ b/JS.md
@@ -41,7 +41,7 @@ The following intrinsic objects are added:
 * `WebAssembly.CompileError` : a [NativeError](http://tc39.github.io/ecma262/#sec-nativeerror-object-structure)
    which indicates an error during WebAssembly decoding or validation
 * `WebAssembly.LinkError` : a [NativeError](http://tc39.github.io/ecma262/#sec-nativeerror-object-structure)
-   which indicates an error during WebAssembly instantiating a mdoule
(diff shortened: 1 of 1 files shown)


jfbastien:
Should we call out out-of-memory errors separately, or do they cause Compile / Link errors? This seems similar to @gahaas' clarification w.r.t. stack error in #875.

If we split out memory error then let's not do it in this PR, but if we don't then let's address it in this PR (as a Link error).

JS.md
```diff
@@ -40,6 +40,8 @@ The following intrinsic objects are added:
 * `WebAssembly.Table` : the [`WebAssembly.Table` constructor](#webassemblytable-constructor)
 * `WebAssembly.CompileError` : a [NativeError](http://tc39.github.io/ecma262/#sec-nativeerror-object-structure)
    which indicates an error during WebAssembly decoding or validation
+* `WebAssembly.LinkError` : a [NativeError](http://tc39.github.io/ecma262/#sec-nativeerror-object-structure)
+   which indicates an error during WebAssembly instantiating a mdoule
```
JosephPecoraro: Typo: "mdoule"

rossberg-chromium: Done.

rossberg-chromium:
@jfbastien, excellent question, but I think it's a somewhat tricky issue that's better discussed separately. Nothing specified in this PR depends on how that's resolved.

jfbastien:
The proposed wording lgtm

JS.md
```diff
@@ -111,7 +113,7 @@ On success, the `Promise` is [fulfilled](http://tc39.github.io/ecma262/#sec-fulf
 with a plain JavaScript object pair `{module, instance}` containing the resulting
 `WebAssembly.Module` and `WebAssembly.Instance`. On failure, the `Promise` is
 [rejected](http://tc39.github.io/ecma262/#sec-rejectpromise) with a 
-`WebAssembly.CompileError`.
+`WebAssembly.CompileError`, `WebAssemlby.LinkError`, or `WebAssembly.RuntimeError`, depending on the cause of failure.
```
titzer: s/WebAssemlby/WebAssembly/

JS.md
```diff
@@ -129,7 +131,7 @@ from `moduleObject` and `importObject` as described in the
 On success, the `Promise` is [fulfilled](http://tc39.github.io/ecma262/#sec-fulfillpromise)
 with the resulting `WebAssembly.Instance` object. On failure, the `Promise` is
 [rejected](http://tc39.github.io/ecma262/#sec-rejectpromise) with a 
-`WebAssembly.CompileError`.
+`WebAssembly.CompileError`, `WebAssemlby.LinkError`, or `WebAssembly.RuntimeError`, depending on the cause of failure.
```
titzer: typo here too

titzer:
lgtm

JS.md
rossberg-chromium: Both typos fixed.

JS.md
```diff
@@ -129,7 +131,7 @@ from `moduleObject` and `importObject` as described in the
 On success, the `Promise` is [fulfilled](http://tc39.github.io/ecma262/#sec-fulfillpromise)
 with the resulting `WebAssembly.Instance` object. On failure, the `Promise` is
 [rejected](http://tc39.github.io/ecma262/#sec-rejectpromise) with a 
-`WebAssembly.CompileError`.
+`WebAssembly.CompileError`, `WebAssembly.LinkError`, or `WebAssembly.RuntimeError`, depending on the cause of failure.
```
lukewagner: driveby nit: IIUC, no `CompileError` is possible here.

JS.md
```diff
@@ -288,10 +290,10 @@ For each [`import`](https://github.com/WebAssembly/spec/blob/master/interpreter/
 1. Let `v` be the value of performing [`Get`](http://tc39.github.io/ecma262/#sec-get-o-p)(`o`, [`i.item_name`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/ast.ml#L171))
```
jfbastien: The omission of the above `TypeError` ("If `Type(o)` is not Object") is on purpose, right? It kinda happens before the constructor is called I guess?

Same question for "If the `importObject` parameter is not `undefined` and `Type(importObject)` is not Object".

rossberg-chromium: The required type of `v` differs depending on what type of import they match, so is tested in the individual cases below. In particular, for globals it _must not_ be an object.

Issue #897, opened 2016-12-09
Data segment overlaps


jfbastien:
[Data segments](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#data-section) aren't well defined.

* If they overlap, does it fail?
* If so, what's the exception?
* When is it thrown? Some of these issues can be known at Module time (if it was an `i32.const`), and others at Instance time (if it was a global).
* If they're out of bounds, does it fail?

binji:
This is specified in [JS.md](https://github.com/WebAssembly/design/blob/master/JS.md#webassemblyinstance-constructor), but probably should be mentioned elsewhere too.

> Segments may overlap and, if they do, the final value is the last value written in order.

> If, after evaluating the offset initializer expression of every Data and Element Segment, any of the segments do not fit in their respective Memory or Table, throw a RangeError.

jfbastien:
What if the initializer is a constant, and it's known at Module time that it's out of bounds? The failure still occurs at Instance time?

jfbastien:
Oh also, what if there's a Data section but no memory (neither imported, nor in a section)?

jfbastien:
I'm also not sure we specify that data segments are initialized before start is called.

jfbastien:
Oh, and if linking fails, is an imported memory changed by data segments? Or do data segments get populated after link succeeds, but before start is called?

jfbastien:
Oh and if initializing segments fails halfway, I assume it's visible up to when it failed? i.e. if the first segment is valid but the second isn't, then the first is set, second isn't?

And what if a segment is only partly valid? i.e. it starts in range but ends out of range. Do we see a partial write, or none of that segment's write? I'd rather not do partial, but that's what wasm does when executing.

jfbastien:
Another one: what if there's an empty data section, and no memory (neither imported, nor memory section)? Is that an error at Module creation time?

jfbastien:
This is fun. What if the memory exists but is zero-sized, and we initialize with an empty data segment?

rossberg-chromium:
@jfbastien, have you checked the spec interpreter and test suite?

> * If they overlap, does it fail?
> * If so, what's the exception?
> * When is it thrown? Some of these issues can be known at Module time (if it was an i32.const), and others at Instance time (if it was a global).

What @binji says, no failure (a new module could be linked in at any later time, so that would require engines to track what parts of tables have been initialised).

We are missing more tests for the element stuff. Contributions welcome.

[...]

PR #902, opened 2016-12-12
Clarify Data and Element segments

jfbastien:
Addresses #897. Related to https://github.com/WebAssembly/spec/pull/399.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -166,6 +166,7 @@ Otherwise, this function performs synchronous compilation of the `BufferSource`:
    the validated `Ast.module`.
 1. On failure, a new `WebAssembly.CompileError` is thrown.
 
+
 ### `WebAssembly.Module.exports`
 
 The `exports` function has the signature:
@@ -379,16 +380,19 @@ the internal `[[Instance]]` slot to `instance`.
 
 Perform [`CreateDataProperty`](https://tc39.github.io/ecma262/#sec-createdataproperty)(`instance`, `"exports"`, `exportsObject`).
 
-If, after evaluating the `offset` [initializer expression](Modules.md#initializer-expression)
-of every [Data](Modules.md#data-section) and [Element](Modules.md#elements-section)
-Segment, any of the segments do not fit in their respective Memory or Table, throw a 
-[`RangeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-rangeerror).
+Apply all Data segments, followed by all Element segments, to their respective Memory or Table in the order in which they appear in the Module.
 
-Apply all Data and Element segments to their respective Memory or Table in the
-order in which they appear in the module. Segments may overlap and, if they do,
-the final value is the last value written in order. Note: there should be no
-errors possible that would cause this operation to fail partway through. After
-this operation completes, elements of `instance` are visible and callable
+Note: validation rules prevent a Module from having a Data section without having a Memory section or import, as well as prevent a Module from having an Element section without having a Table.
+
+* The `offset` [initializer expression](Modules.md#initializer-expression) of every [Data](Modules.md#data-section) and [Element](Modules.md#elements-section) segment is evaluated, any of the segments do not fit in their respective Memory or Table, throw a [`RangeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-rangeerror).
+
+  - Segments which are partially out-of-bounds cause the entire evaluation of that segment to fail.
+  - Zero-sized segments do not cause failure, even if their offset is out-of-bounds.
+  - Any segment evaluation performed prior to such an error is visible.
+
+* Segments may overlap and, if they do, the final value is the last value written in order.
+
+After this operation completes, elements of `instance` are visible and callable
 through [imported Tables](Modules.md#imports), even if `start` fails.
 
 If a [`start`](Modules.md#module-start-function) is present, it is evaluated
--- a/JS.md
+++ b/JS.md
@@ -384,7 +384,7 @@ Apply all Data segments, followed by all Element segments, to their respective M
 
 Note: validation rules prevent a Module from having a Data section without having a Memory section or import, as well as prevent a Module from having an Element section without having a Table.
 
-* The `offset` [initializer expression](Modules.md#initializer-expression) of every [Data](Modules.md#data-section) and [Element](Modules.md#elements-section) segment is evaluated, any of the segments do not fit in their respective Memory or Table, throw a [`RangeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-rangeerror).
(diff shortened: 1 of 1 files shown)


JS.md
```diff
@@ -379,16 +380,19 @@ the internal `[[Instance]]` slot to `instance`.
...
 
-If, after evaluating the `offset` [initializer expression](Modules.md#initializer-expression)
-of every [Data](Modules.md#data-section) and [Element](Modules.md#elements-section)
-Segment, any of the segments do not fit in their respective Memory or Table, throw a 
-[`RangeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-rangeerror).
+Apply all Data segments, followed by all Element segments, to their respective Memory or Table in the order in which they appear in the Module.
+
+Note: validation rules prevent a Module from having a Data section without having a Memory section or import, as well as prevent a Module from having an Element section without having a Table.
+
+* The `offset` [initializer expression](Modules.md#initializer-expression) of every [Data](Modules.md#data-section) and [Element](Modules.md#elements-section) segment is evaluated, any of the segments do not fit in their respective Memory or Table, throw a [`RangeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-rangeerror).
```
kmiller68: "segment is evaluated, if any of the segments"

binji:
I agree that the common case is that linking succeeds, but I can imagine a scenario where your application has a plugin architecture which allows importing arbitrary modules. It would be nice to know that you can do this without worrying about link failure putting your application in an inconsistent state.

pipcet:
@binji I agree, but isn't what's in the PR good enough for that? I'd imagine dlopen() would malloc() the right amount of memory, try linking, and free() it on failure, without worrying about the contents of the free()d block; similarly for table entries, though those should probably be cleared to null explicitly to make debugging easier.

Are you saying it would be better to snapshot memory before attempting any link and reverting to the pre-link state upon failure?

jfbastien:
Agreed with @pipcet : if you dynamically load random stuff and it fails halfway then you *have* to know that the memory you gave it to initialize is now invalid, regardless of whether it wrote to that memory or not. Otherwise you leak. You're already trusting that rando-load to stick to the memory you gave it, so "corrupting" is strong wording!

/meme the writes are coming from *inside the module*

binji:
Sorry, saying "dynamic-linking" is confusing, I guess. I agree that as soon as user code is running, it doesn't make much sense for any kind of rollback of tables or memories. But if we are still instantiating a module, I think it's reasonable to make sure that the memory and tables are not modified if the new module cannot be instantiated.

jfbastien:
Why? I can't see a useful thing that could be done with the memory / table in these cases since they were handed out to a sub-module which failed. They need to be leaked, or reclaimed. In both cases their content is irrelevant.

Maybe I'm thinking about this wrong though!

pipcet:
Well, I'm not sure this is a useful case, but a simple application might save its state by dumping the (entire, in the worst case) memory contents into a wasm module, unexec-style. If we want to support that, and if we want to further support recovery from failed reloads, we probably would have to guarantee the all-or-nothing semantics @binji proposes.

jfbastien:
That seems highly wasteful, and I don't think we should encourage it. We're talking about a minor performance thing (do two passes or one) and I don't see why we'd do the slower thing. In the grand scheme of things the cost is minor, but why pay it if there's no worthwhile usecase for the two-pass approach?

binji:
Yeah, I'm not concerned about the memory being trashed if you instantiate the new module, which runs its start function, which traps in some way so the memory state is not saved. At that point, I'd guess you'd have to just say that the application failed to load.

I was just thinking that if we could try to preserve some shared state, and it wasn't too expensive (two passes over the segments probably wouldn't be), it may be useful. The use case of an optional plugin, popped into my mind, and it seemed reasonable. But it sounds like you think that this isn't that useful because there are a number of other ways in which a plugin module could fail, and we can't prevent indeterminate state in those cases?

I'm not sure what you mean by "leaked or reclaimed". If you hand them out to a sub-module, but they haven't been modified yet, then why can't you just pretend as if that failed instantiation never happened?

jfbastien:
> I was just thinking that if we could try to preserve some shared state, and it wasn't too expensive (two passes over the segments probably wouldn't be), it may be useful. The use case of an optional plugin, popped into my mind, and it seemed reasonable. But it sounds like you think that this isn't that useful because there are a number of other ways in which a plugin module could fail, and we can't prevent indeterminate state in those cases?

Correct.

> I'm not sure what you mean by "leaked or reclaimed". If you hand them out to a sub-module, but they haven't been modified yet, then why can't you just pretend as if that failed instantiation never happened?

I meant: you hand out memory, whoever you give it to fails and maybe trashes that memory. You either shrug and leave that memory trashed forever (leaked), or you zero it and give it back to your allocator (reclaimed).

binji:
OK, makes sense. Now that I think of it, if a user really wants to catch this error ahead of time, they can parse the module they're about to load to determine whether all the segments would fit.

jfbastien:
(I'm waiting for the @horse_wasm quote: "wasm: why bother?").

rossberg-chromium:
I'm torn. What's specified in this PR is slightly simpler (and both V8 and the spec interpreter actually implement it that way right now -- which WebAssembly/spec#399 intended to fix). Yet, what the existing text requires seems cleaner to me: it consistently checks all types and bounds of imports and internal defs before doing anything observable.

All-or-nothing is a fallacy. Declarative(-ish) definitions leaving data structures in an inconsistent state is bad IMO, even when you cannot protect against other possible failures. Asking apps to defend themselves by complex reflection is both asking a lot and encouraging bad practice. And it cannot address start function failures either, so is no more a complete solution.

jfbastien:
I'm not asking apps to defend themselves by complex reflection. If they have to defend themselves then they've already lost. Don't load untrusted / untested code. In my experience the failure path isn't tested, so any "defense" is likely buggy.

titzer:
It makes more sense to me to check everything up front, before performing any operation with observable effects, even though that is strictly more work in the engine. That way, applications don't need to handle inconsistent states at all.

lukewagner:
I also think it's better to do all checking up front before performing any side effects.

rossberg-chromium:
I also prefer the hygiene of up-front checking.

pipcet:
Is it actually true that there's a non-negligible performance penalty for up-front checking? I don't think it is, since you have to validate the module anyway, and I don't think there's a reason other than performance not to do the checks.

lukewagner:
@pipcet Yes, there's ordinarily very few segments (often just 1) so the difference between checking segments in a loop up-front vs. in the loop that copies segments to memory is negligible.

jfbastien:
Looks like this won't happen. Closing.

PR #905, opened 2016-12-13
Start function clarifications

### LINKED ISSUE #896: Can start functions be imports?
The [current wording](https://github.com/WebAssembly/design/blob/master/Modules.md#module-start-function) isn't so clear about this:

> If a module contains a start node, the function must be defined in the module

I want to clarify some of the other wording in this entire section, but this question merits a bit more thought I think.

jfbastien:
Resolves #896.

### DIFF
--- a/Modules.md
+++ b/Modules.md
@@ -175,8 +175,8 @@ to allow *explicitly* sharing state between module instances.
 ## Module start function
 
 If the module has a start node defined, the function it refers should be called
-by the loader after the instance is initialized, including its Data and Element
-sections, and before the exported functions are called.
+by the loader after the instance is initialized, including its Memory and Table
+though Data and Element sections, and before the exported functions are callable.
 
 * The start function must not take any arguments or return anything
 * The function is identified by [function index](#function-index-space), can be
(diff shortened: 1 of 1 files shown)


Modules.md
```diff
@@ -175,12 +175,12 @@ to allow *explicitly* sharing state between module instances.
 ## Module start function
 
 If the module has a start node defined, the function it refers should be called
-by the loader after the instance is initialized and before the exported functions
-are called.
+by the loader after the instance is initialized, including its Data and Element
```
rossberg-chromium: Wording nit: sections don't get initialised, memories and tables do (from them).

rossberg-chromium:
LGTM modulo nit

Modules.md
jfbastien: Right. Clarified, ty.

jfbastien:
We don't specify that this happens before any instance-creating promise returns because that's JS.md stuff. I changed the phrasing a bit to say "before exported functions are callable", that's probably good enough to clarify the promise thing?

rossberg-chromium:
SGTM. the original wording was weird, this is better.

Issue #907, opened 2016-12-13
It's not clear in JS.md what values processing the Elements list creates during instantiation


saambarati:
For example, lets consider the following JS program:
```
module = new WebAssembly.Module({
    ;; single function (at index 0)
    ;; export function at index 0 as "foo"
    ;; import a table
    ;; elements section with a single element:
    ;;      element = { offset 0, function indices: [0] }
    (all of this is in the binary of course)
});

let table = new WebAssembly.Table({element: "anyfunc", initial: 1});
let instance = new WebAssembly.Instance(module,  {import: {table}});

instance.exports.foo === table.get(0) // What is this? true or false?
```

- Are the Web Assembly Function objects that get stored into table supposed to be exactly the same object that gets stored into the exports dictionary?

rossberg-chromium:
Yes, they have to be the same. The wording of the algorithm in JS.md is pretty explicit on that, note the construction and querying of the `funcs` list.

saambarati:
@rossberg-chromium I'm specifically talking about the elements section. There is hardly any text inside Instance construction talking about the Element section, and it doesn't reference the issue I stated above:
https://github.com/WebAssembly/design/blob/master/JS.md#webassemblyinstance-constructor

The querying you're talking about is done for function imports/exports and table exports. It's not talked about with respect to the Elements section.

If we want the behavior you describe, JS.md should have some text about it. FWIW, I agree that this should be the specced behavior.

rossberg-chromium:
@saambarati, I don't quite understand, your example _does_ test function exports against table exports, so AFAICT, its behaviour is fully specified. And other than that, elements don't have any observable identity, so what is missing?

saambarati:
@rossberg-chromium My example doesn't export a table. It imports a table and has the element section write an imported function into index zero in the table

rossberg-chromium:
@saambarati, ah, yes, you're right, thanks! Probably the fix is to include all elements of imported tables in `funcs`and later run over the table to update it, independent of whether it is exported. I'll try to prepare a PR.

But to clarify the question in the abstract, the overarching idea is that every Wasm function should be wrapped into a JS function at most once.

rossberg-chromium:
See #911.

rossberg-chromium:
This is fixed.

PR #911, opened 2016-12-14
Fix value creation for imported tables

rossberg-chromium:
To address #907.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -358,13 +358,12 @@ The following steps are performed _before_ the `start` function executes:
   1. For each index `i` of `t`:
     1. Let `c` be the `i`th element of `t`
     1. If `c` is a [`closure`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/instance.ml#L7) `c`:
-      1. If there is an [Exported Function Exotic Object](#exported-function-exotic-objects) `in `funcs` whose `[[Closure]]` equals `c`:
+      1. If there is an [Exported Function Exotic Object](#exported-function-exotic-objects) in `funcs` whose `[[Closure]]` equals `c`:
         1. Let `func` be that function object.
       1. (Note: At most one wrapper is created for any closure, so `func` is uniquely determined. Moreover, if the item was an import that is already an [Exported Function Exotic Object](#exported-function-exotic-objects), then the original function object will be found. For imports that are regular JS functions, a new wrapper will be created.)
       1. Otherwise:
         1. Let `func` be an [Exported Function Exotic Object](#exported-function-exotic-objects) created from `c`.
         1. Append `func` to `funcs`.
-        1. Return `func`.
       1. Set the `i`th element of `table.[[Values]]` to `func`.
 
 (Note: The table and element function objects created by the above steps are only observable for tables that are either imported or exported.)
--- a/JS.md
+++ b/JS.md
@@ -318,6 +318,10 @@ For each [`import`](https://github.com/WebAssembly/spec/blob/master/interpreter/
      throw a `WebAssembly.LinkError`.
   1. Append `v` to `tables`.
   1. Append `v.[[Table]]` to `imports`.
+  1. For each index `i` of `v.[[Table]]`:
+    1. Let `e` be the `i`the element of `v.[[Table]]`.
+    1. If `e` is a [`closure`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/instance.ml#L7) `c`:
+      1. Append the `i`th element of `v.[[Values]]` to `funcs`.
 
 Let `instance` be the result of creating a new
 [`instance`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/instance.ml#L17)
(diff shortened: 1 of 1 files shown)


JS.md
```diff
@@ -337,11 +341,34 @@ Among other things, this function performs the following observable steps:
...
+  1. If there is no element in `tables` whose `table.[[Table]]` is `t`:
+    1. Let `table` be a new `WebAssembly.Table` object with [[Table]] set to `t` and [[Values]] set to a new list of the same length all whose entries are `null`.
+    1. Append `table` to `tables`.
+  1. Otherwise:
+    1. Let `table` be the element in `tables` whose `table.[[Table]]` is `t`
+  1. (Note: At most one `WebAssembly.Table` object is created for any table, so the above `table` is unique, even if there are multiple occurrances in the list. Moreover, if the item was an import, the original object will be found.)
+  1. For each index `i` of `t`:
+    1. Let `c` be the `i`th element of `t`
+    1. If `c` is a [`closure`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/instance.ml#L7) `c`:
+      1. If there is an [Exported Function Exotic Object](#exported-function-exotic-objects) `in `funcs` whose `[[Closure]]` equals `c`:
```
lukewagner: I think you have a stray backtick here that throws off the rest.

JS.md
```diff
@@ -337,11 +341,34 @@ Among other things, this function performs the following observable steps:
...
+        1. Let `func` be that function object.
+      1. (Note: At most one wrapper is created for any closure, so `func` is uniquely determined. Moreover, if the item was an import that is already an [Exported Function Exotic Object](#exported-function-exotic-objects), then the original function object will be found. For imports that are regular JS functions, a new wrapper will be created.)
+      1. Otherwise:
+        1. Let `func` be an [Exported Function Exotic Object](#exported-function-exotic-objects) created from `c`.
+        1. Append `func` to `funcs`.
+        1. Return `func`.
```
lukewagner: Since this isn't a reduction (and since the other branch doesn't have one), I'd remove the "Return" step.

lukewagner:
Thinking about this patch, I realized a rather subtle consequence of the current spec text (that is made worse by this patch) that I'm guessing wasn't intentional (and isn't what SM has implemented, maybe V8 too).

So for the current spec text, the question is whether this assert holds:
```
var js = () => {};
var m = new Module(wasmTextToBinary`(module
    (import $a "" "a")
    (export "a" $a)
    (import $b "" "b")
    (export "b" $b))
`));
var i = new Instance(m, {'':{a: js, b: js}});
assert(i.exports.a !== i.exports.b);
```
In SM the assert passes, because we're basically saying that each (instance, function-index) pair is given a unique exported-wasm-function-object, regardless of whether the function-index is an import.  I find this easy to reason about and it also simplifies the implementation a good deal by allowing imported/internal functions to be treated much more uniformly.

I think the current spec text says the assert() should fail, though, because the associated closures would be equal (because the underlying JS object identities would be equal).  To implement the spec, you'd need to maintain a separate hash-set of imported JS objects and query this set separately.  Then, taking this PR into account, you'd have to eagerly walk an imported Table to put any wrapped-JS-functions into the hash-set.  For example, this assert passes in SM (and I bet V8) but I think fails according to this PR:
```
var js = () => {};
var m1 = new Module(wasmTextToBinary(`(module
    (import $i "" "")
    (table (export "tbl") anyfunc (elem $i)))
`));
var i1 = new Instance(m1, {'':{'':js}});
var m2 = new Module(wasmTextToBinary(`(module
    (import $fun "" "fun")
    (export "fun" $fun)
    (import "" "tbl" (table 1 anyfunc)))
`));
var i2 = new Instance(m2, {'':{fun:js, tbl:i1.exports.tbl}})
assert(i2.exports.fun !== i1.exports.tbl.get(0));
```

I think the fix for both issues is to give *all* closures an (instance, func-index), on which equality is defined (and `HostFunc`s just have additional fields).

JS.md
rossberg-chromium: Done

JS.md
rossberg-chromium: Done

rossberg-chromium:
@lukewagner, good point. I actually don't know what V8 does (can't check right now). However, I think this should perhaps be addressed/discussed in a separate PR.

titzer:
In V8 we always wrap imported functions with a stub that does the argument conversion, unless the imported function is an export from another WASM module. So if we import a JS function J into a module M, then calls to J in M go through a wrapper Stub->J. When we reexport from M, Stub->J gets wrapped again, the code becoming Stub'->Stub->J and getting a new JSFunction with a new identity.

lukewagner:
@titzer So does the first assert I wrote above succeed or fail in V8?  It passes in SM which, based on what @rossberg-chromium said in #920, is according to spec.  But it sounds like, from your description, the two exports might get the same identity.

jfbastien:
I'd like to make sure that the spec allows an implementation to remove as many stubs as possible.

titzer:
The assert passes in V8; i.e. the two exported functions are different.

titzer:
On Mon, Dec 19, 2016 at 5:55 PM, JF Bastien <notifications@github.com>
wrote:

> I'd like to make sure that the spec allows an implementation to remove as
> many stubs as possible.
>
That's tempting, and IIRC we did this also initially, but with JS type
conversions being observable, it seems more consistent that a re-exported
imported JS function has a wrapper around it that does type conversions. It
also seems consistent (as in this PR) that each exported function has a
separate identity; otherwise the engine is essentially required to
canonicalize imports based on their (JS) identity.

lukewagner:
@titzer Ah, nice.  And just to confirm, this assert also passes?
```
var js = () => {};
var m = new Module(wasmTextToBinary(`(module
    (import $js "" "js")
    (export "a" $js)
    (export "b" $js))
`));
var i = new Instance(m, {'':{js}});
assert(i.exports.a === i.exports.b);
```
i.e., multiple exports of the same func-index produce the same identity?  If so then it sounds like we're on the same page of exported-function-object-per-(instance,func-index)-pair, which is great.

lukewagner:
@jfbastien I think that re-exports will be quite rare, so it's probably not worth optimizing this case, but if it ever became so, one could introduce a new kind of JS function object that recorded the wasm signature to coerce "through" and called the imported JS function directly and this could be optimized by the JITs just like they currently do for bound function objects.

titzer:
On Mon, Dec 19, 2016 at 6:19 PM, Luke Wagner <notifications@github.com>
wrote:

> @titzer <https://github.com/titzer> Ah, nice. And just to confirm, this
> assert also passes?
>
> var js = () => {};
> var m = new Module(wasmTextToBinary(`(module
>     (import $js "" "js")
>     (export "a" $js)
>     (export "b" $js))
> `));
> var i = new Instance(m, {'':{js}});
> assert(i.exports.a === i.exports.b);
>
> i.e., multiple exports of the same func-index produce the same identity?
> If so then it sounds like we're on the same page of
> exported-function-object-per-(instance,func-index)-pair, which is great.
>
Yes. We keep a table of JS-exported-exotic-function indexed by function
index, so we only ever create one just JS-exported-exotic-function per
index.

lukewagner:
Excellent.  So pulling it all back together, I think this PR lgtm and matches existing SM/V8 behavior and there is no actual problem in #920.  But as a followup we could stand to clear up the spec wording so it's  more obvious that there is exactly one exported function object per (instance, func-index) (where func-index is either an imported or internal definition).

rossberg-chromium:
Okay, landing this one then, and thinking about a way to make wrapper generativity more explicit in a follow-up (as I just commented in https://github.com/WebAssembly/design/issues/920#issuecomment-271274680, (instance, index) pairs may not be the ideal solution, however).

Issue #910, opened 2016-12-13
Should it be valid to have empty Data/Element sections


saambarati:
Right now, I think it's valid to have zero elem_type in an Element section. Should we allow this?
Ditto for Data.

https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#element-section

jfbastien:
Pretty much all sections. Code without Code? etc.

pipcet:
I was surprised that right now, it's illegal to have a zero-table table section if there is a table import, or a zero-memories memory section if there is a memory import. If there is no import, though, zero tables are okay.

I think it would be cleaner to allow those, and zero-element sized sections generally. I don't think that would cost consumers anything, and producers could cut a corner.

(By "sized sections", I mean sections beginning with a count of entries and then containing the entries. Right now, that's all sections but the start section; maybe the start section should have a count (1 in the MVP) as well, for consistency?)

saambarati:
@pipcet My gut feeling is to go in the opposite direction. It's likely a bug to have zero sized sections of anything that are meaningful only if they have a size. So my vote would be to make zero sized sections an error.

pipcet:
@saambarati I'm not sure I can think of a common bug that would result in a valid, but zero-sized, section when a valid non-zero-sized section is the intended output. Instead, it seems to me more likely that the case in which there is code but no data or data but no code isn't usually well-tested, and might well result in a zero-sized section.

I think a simple producer should only be expected to fill in a fixed wasm skeleton with its data, rather than conditionalizing part of the skeleton based on whether there is any data. That's what I do for intermediate files, which are produced as binary output by GNU ld (so, in particular, they have a linker map allowing me to trace back file offsets) but should also be valid wasm. (I realize I'm a bit of a special case :-) ).

rossberg-chromium:
I don't see any good reason for introducing all these special cases for zero-sized things. It generally is unnecessary hassle to disallow them, for both producers and consumers. Catching internal problems in producers is not the purpose of validation.

@pipcet, where do you read that zero-sized table sections are illegal in the presence of an import? The spec interpreter makes no such restriction. If that's stated somewhere then it's simply an oversight and should be corrected (probably a left-over from the pre-multi-table design?).

saambarati:
[...]

PR #912, opened 2016-12-14
Fix outdated formulations regarding singular tables/memories

rossberg-chromium:
To address @pipcet's observation in #910.


jfbastien:
lgtm

Issue #918, opened 2016-12-16
What's the deal with the TODO inside JS.md


saambarati:
https://github.com/WebAssembly/design/blob/master/JS.md#todo

I don't know what this is addressing, but we should either delete it or address it.

cc @lukewagner

lukewagner:
Yeah, that TODO list was longer when JS.md was first started.  Probably it'll be quite a while before adding cyclic imports, so I'm fine removing it.

Issue #917, opened 2016-12-16
Table.prototype.grow needs to specify a return value


saambarati:
It doesn't right now:
https://github.com/WebAssembly/design/blob/master/JS.md#webassemblytableprototypegrow

saambarati:
My vote would be for returning `undefined`. Like Memory.prototype.grow, I'm not sure what else would  make sense.

binji:
Well, the `grow_memory` operator returns the old page size, so I asssume that `Table.prototype.grow` would as well.

See https://github.com/WebAssembly/design/blob/master/Semantics.md#resizing

lukewagner:
Agreed, this is missing from JS.md and that, for the same reasons that `grow_memory`/`Memory.prototype.grow` return the previous size, so should `Table.prototype.grow`.  (That's actually what we're currently doing in SM.)

saambarati:
Ok sounds good to me. We will do that in JSC. Let's just make sure we specify it in the API

rossberg-chromium:
Whatever we do, it clearly should match Memory.p.grow.

lukewagner:
(btw, I'm accumulating these JS.md TODOs on a list I'll get to when I'm back up to speed, but I'm quite happy for anyone else to make PRs first)

lukewagner:
I believe this is resolved by #922.

Issue #916, opened 2016-12-16
What's the return value of Memory.prototype.grow


saambarati:
https://github.com/WebAssembly/design/blob/master/JS.md#webassemblymemoryprototypegrow
Says to return the value of:
https://github.com/WebAssembly/design/blob/master/Semantics.md#resizing
I'm not quite sure what that means. Is the link stale? Should it be to the ML interpreter here:
https://github.com/WebAssembly/spec/blob/e7632dc00c7ea0743d64a42d1ed3c39ed3cd5dfa/interpreter/spec/eval.ml#L213
?

As a matter of style, if Memory.prototype.grow succeeds, why would it return anything other than `undefined`? Since an error is thrown if it fails, I don't think anything else is needed to indicate success.

lukewagner:
When wasm gets threads, the return value is useful for knowing which ranges of bytes were just claimed by this `grow` operation and not some interleaved `grow_memory` from a different thread.

wllang:
It's to make it an atomic operation, see the discussion in https://github.com/WebAssembly/design/issues/525

saambarati:
Ok. Sounds good to me. I'll implement returning the previous size inside JSC. Is it returning the previous size in bytes or pages?

rossberg-chromium:
Returning the previous value SGTM.

lukewagner:
Pages (to match [`grow_memory`](https://github.com/WebAssembly/design/blob/master/Semantics.md#resizing)).

lukewagner:
I believe this is resolved by #922.

PR #922, opened 2016-12-20
Fix wording of (Memory|Table).prototype.grow in JS.md

lukewagner:
This PR is intended to address the omissions identified by #916 and #917.  The dangling TODO is removed as well (#918).


jfbastien:
lgtm

saambarati:
LGTM too

Issue #913, opened 2016-12-15
When should i64 exceptions happen in calls


kmiller68:
When calling a Wasm function that has an i64 as and argument or return value there are four places an exception can occur:

1. When calling an exported Wasm function that takes i64 argument from JS.
1. When calling an exported Wasm function that returns an i64 result from JS.
1. When importing a JS function to a Wasm module that takes an i64 argument.
1. When importing a JS function to a Wasm module that returns an i64.

The current behavior for all of these (or at least from what got out of JS.md) is to trigger the exception exactly when the i64 would be "touched". For example, in case 1, if you have a Wasm function with the signature: `(i32 * i64) -> i32`, and you called it with `instance.exports.func({ toNumber() { throw "gotcha"; } }, foo);` `"gotcha"` would be thrown. Is this the desired behavior? I would prefer if, rather, the host function immediately threw a TypeError.

Similarly, In the second case, the exported Wasm function would be called and any of its effects executed. Then, as the function returns, the TypeError would be generated. Here I would also prefer the host function be a TypeError stub.

In the last two cases, when a JS function is imported that claims to interface with i64, the current spec, as above throws an error at runtime. It seems to me that should be an instantiation time error, since we know if the import is JS or Wasm.

rossberg-chromium:
Isn't this pretty unambiguous from JS.md's definition of ToWebAssemblyValue? For i64, you immediately throw.

rossberg-chromium:
Ah, scratch that. I take it you are not asking what the specified semantics is but whether we should change it, specifically, the spec of the [[Call]] method of Exported Function Objects?

kmiller68:
@rossberg-chromium Yeah, I would like to change the semantics. I think we should also change the spec for Imported JS Function Objects as well. Since, at least for imports we know that the Imported JS Function Object cannot handle an i64, thus we should fail to instantiate. e.g. 

<pre>(module (import "foo" "bar" (func (param i64)) )</pre>

passed to

<pre>new WebAssembly.instance(module, { foo: { bar() { } } }); // should throw a WebAssembly.LinkError</pre>

lukewagner:
For exported functions, it sounds good to throw if there is `i64` in any argument/return slot at the beginning of the [[Call]].

[...]

Issue #906, opened 2016-12-13
|this| value in wasm -> JS function?


jfbastien:
Should we just make |this| undefined?

Similar to #900.

saambarati:
+1 for `undefined`

jfbastien:
`undefined+1` would be `NaN`. 🙃

rossberg-chromium:
I don't think there is any other option if we want to have any consistency with function calls in JavaScript.

Nit: You mean to refer to the _receiver_, not the `this` binding, which is an internal detail. Not all JS-side functions even have a notion of the latter, e.g. primitives. Moreover, sloppy functions convert an undefined receiver to the global object (themselves).

lukewagner:
Agreed on `undefined` and that we should call this out explicitly in the spec.

lukewagner:
Oops, forgot to close this when resolved by #923.

PR #923, opened 2016-12-20
Throw immediately when calling into or out of JS with i64 arg/ret

lukewagner:
As discussed in #913.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -303,16 +303,20 @@ For each [`import`](https://github.com/WebAssembly/spec/blob/master/interpreter/
     1. Let `closure` be `v.[[Closure]]`.
   1. Otherwise:
     1. Let `closure` be a new [host function](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/instance.ml#L9)
-       of the given signature
-       which calls `v` by coercing WebAssembly arguments to JavaScript arguments
-       via [`ToJSValue`](#tojsvalue) and returns the result, if any, by coercing
-       via [`ToWebAssemblyValue`](#towebassemblyvalue).
+       of the given signature:
+      1. If the signature contains an `i64` (as argument or result), the host
+         function immediately throws a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
+         when called.
+      1. Otherwise, the host function calls `v` by coercing WebAssembly
+         arguments to JavaScript arguments
+         via [`ToJSValue`](#tojsvalue) and returns the result, if any, by coercing
+         via [`ToWebAssemblyValue`](#towebassemblyvalue).
   1. Append `v` to `funcs`.
   1. Append `closure` to `imports`.
 1. If `i` is a global import:
   1. [Assert](https://tc39.github.io/ecma262/#assert): the global is immutable
      by MVP validation constraint.
-  1. If `Type(v)` is not Number, throw a `WebAssembly.LinkError`.
+  1. If `v` is an `i64` or `Type(v)` is not Number, throw a `WebAssembly.LinkError`.
   1. Append [`ToWebAssemblyValue`](#towebassemblyvalue)`(v)` to `imports`.
 1. If `i` is a memory import:
   1. If `v` is not a [`WebAssembly.Memory` object](#webassemblymemory-objects),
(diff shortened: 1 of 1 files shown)


lukewagner:
Since I'm changing the same line, I also added the clarification requested in #906 that the receiver for a wasm->JS call is `undefined`.

kmiller68:
sgtm!

JS.md
```diff
@@ -303,16 +303,20 @@ For each [`import`](https://github.com/WebAssembly/spec/blob/master/interpreter/
     1. Let `closure` be `v.[[Closure]]`.
   1. Otherwise:
     1. Let `closure` be a new [host function](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/instance.ml#L9)
-       of the given signature
-       which calls `v` by coercing WebAssembly arguments to JavaScript arguments
-       via [`ToJSValue`](#tojsvalue) and returns the result, if any, by coercing
-       via [`ToWebAssemblyValue`](#towebassemblyvalue).
+       of the given signature:
+      1. If the signature contains an `i64` (as argument or result), the host
+         function immediately throws a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
```
saambarati: Should this be WebAssembly.RuntimeError?

lukewagner: Thus far, we've only thrown those for core-wasm-defined traps (enumerating my list: bad call_indirect, unreachable, idiv, bad convertion to int, oob), not for JS API-level errors.

saambarati: Makes sense to me.

JS.md
```diff
@@ -303,16 +303,20 @@ For each [`import`](https://github.com/WebAssembly/spec/blob/master/interpreter/
...
+         and WebAssembly arguments coerced to JavaScript arguments
+         via [`ToJSValue`](#tojsvalue). The result is returned by coercing
+         via [`ToWebAssemblyValue`](#towebassemblyvalue).
   1. Append `v` to `funcs`.
   1. Append `closure` to `imports`.
 1. If `i` is a global import:
   1. [Assert](https://tc39.github.io/ecma262/#assert): the global is immutable
      by MVP validation constraint.
-  1. If `Type(v)` is not Number, throw a `WebAssembly.LinkError`.
+  1. If `v` is an `i64` or `Type(v)` is not Number, throw a `WebAssembly.LinkError`.
```
saambarati: I know this is how things used to be, but is there a reason we don't just call toNumber/toInt here?

lukewagner: That's actually how we started out and the requirement was tightened by #816 after a bit of [discussion](https://github.com/WebAssembly/spec/issues/356#issuecomment-252851799) in [spec/#356](https://github.com/WebAssembly/spec/issues/356).

rossberg-chromium: Don't understand this line. How can `v` (which is a JS value) be an `i64`?

JS.md
```diff
@@ -361,6 +365,7 @@ each [external](https://github.com/WebAssembly/spec/blob/master/interpreter/spec
 1. If `e` is a [global](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/instance.ml#L15) `v`:
   1. [Assert](https://tc39.github.io/ecma262/#assert): the global is immutable
      by MVP validation constraint.
+  1. If `v` is an `i64`, throw a `WebAssembly.LinkError`.
```
rossberg-chromium: Oh, this actually means that you cannot export an `i64` on the Web, not even to another Wasm module. Rather unfortunate. Should we do something about that?

jfbastien: That is indeed undesirable.

JS.md
lukewagner: Oops, I meant the type (as in declared global type) of \*`i`\*.

JS.md
lukewagner: Since we can't instantiate one wasm module knowing that its exports will only be used by another wasm module, unfortunately I don't think there's anything we can do in the MVP.  If we either add "`WebAssembly.instantiateGraph`" (which takes and simultaneously instantiates multiple modules, allowing them to reference each other) or we add integration with ES6 modules, we could allow this to work b/c we could know that "both sides" were wasm.

Practically speaking, I think this is fine b/c the only need for global imports in the MVP is to provide (32-bit) offsets/pointers for dynamic linking purposes; actual C++ `int64_t` global variables will go into linear memory.

rossberg-chromium: Okay, makes sense.

rossberg-chromium:
lgtm

PR #927, opened 2016-12-21
Remove TODO in JS.md

distransient:
Doesn't seem like it's worth it to have a TODO like this floating around, instead tracking it as an issue and putting it on [the roadmap thing](https://github.com/WebAssembly/design/blob/master/FutureFeatures.md) if we want people to know we're working on it

(From #918)


distransient:
Closing this, didn't see #922 lol oops!

Issue #921, opened 2016-12-20
Make file header into a valid section


qwertie:
In #860 I suggested changing the 8 header bytes of Wasm files to

    00 06 04 77 61 73 6D 01

This makes the header appear to be a valid user-defined section with the name "wasm".

The idea is to view the Wasm format as an instance of a more general "list of sections" or "LOS" format. Because files in LOS format have a header that is also a valid section, LOS files can be concatenated. Concatenation seems not to be a useful operation for Wasm modules themselves since the result would not be a valid module, but it might be useful for related formats, such as object files or debug info files defined outside the Wasm spec.

The "magic number" that identifies an LOS file would be the initial three bytes 00 06 04. Any LOS-based format would need a name that is four bytes long, and a one-byte version number.

wllang:
@qwertie If you prepare a PR, a change to BinaryEncoding.md and to Rationale.md if you wish to explain it, then I will review it and approve the change subject to any substantive and reasonable technical objections.

sunfishcode:
Now that we have a PR tracking this, https://github.com/WebAssembly/design/pull/928, we should continue discussion there.

PR #928, opened 2016-12-22
Change WebAssembly module file signature

qwertie:
Changes the file signature in BinaryEncoding.md as suggested in #921. This PR also increments the version number to 0xE so that we can talk about 0xE as a shorthand way of to referring to this change, but to minimize disruption, module readers can continue to accept the old signature in addition to the new one.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -169,17 +169,23 @@ The following documents the current prototype format. This format is based on an
 
 ## High-level structure
 
-The module starts with a preamble of two fields:
+The module starts with a preamble of the following eight bytes:
 
-| Field | Type | Description |
-| ----- |  ----- | ----- |
-| magic number | `uint32` |  Magic number `0x6d736100` (i.e., '\0asm') |
-| version | `uint32` | Version number, currently 0xd. The version for MVP will be reset to 1. |
+    00 06 04 77 61 73 6D 0E
 
-The module preamble is followed by a sequence of sections.
+WebAssembly modules consist entirely of a list of *sections*, and these initial
+eight bytes are the first section, which is a *custom* section with a 6-byte 
+payload, consisting of the name "wasm" in ASCII and a version number (0x0E) 
+which, in the MVP, will be reset to 1, i.e.
+
+    00 06 04 77 61 73 6D 01
+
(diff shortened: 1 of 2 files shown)


wllang:
It looks highly unlikely that there is going to be one compatible wasm encoding due to the difficulty in getting agreement across web browser vendors and perhaps it was not a health expectation anyway.

So perhaps a single version is not the best approach moving forward, perhaps files need to be tagged somewhat like the 'User-Agent' header in HTTP requests with the features that are used by the wasm file.

This might allow different opcode tables for different vendors. Even if they had some cross compatibility support they might have a preferred ordering (for example there could the `google-block` and the `moz-block` operators with different unreachable code semantics etc).

The user layer translator would be responsible for generating code suitable for the target web browser and it could fill in the features used in the wasm file header which might also be input into cache keys. For example, if local variables are not used then this feature could just not be listed in the header and it would be a validation error if it were used and an opcode table without these opcodes might be chosen.

So perhaps changing the header could be deferred until a little more is sorted out.

Kardax:
@wllang I passionately hope that never happens.

wllang:
@Kardax Without room for variation and experimentation I expect there will be very slow and difficult improvement. Sometimes you have to give a little to get something good back, and the web community can give the vendors some freedom here by implementing a translation stage that targets the vendors features and in return we get to take advantage of those improvements and fill otherwise. What we need from the vendors is a description of features so we can target them clearly and good support for a translation stage. I don't object to this PR if that is what the vendors can agree on, but I could live without it and even different headers for each vendor.

sunfishcode:
@wllang This PR proposes changing the wasm binary format header to support the List-Of-Sections idea. This idea is independent of whether the binary format should also have other features (even if those features would want additional changes to the header). For new topics of discussion, please file new issues.

wllang:
@sunfishcode It does still include a single version number and that was what I was drawing attention to which is not going to cut it if web browser vendors have different well versions, and it bakes in a fixed size of `6`, but sure it could be further modified as necessary, enough said.

jfbastien:
This is now obsolete and can't be changed since Firefox and Chrome are shipping WebAssembly.

Issue #929, opened 2016-12-22
Should provably out of bounds loads/stores be a validation error?


saambarati:
Right now, they're not. There is a spec test:
https://github.com/WebAssembly/spec/blob/master/interpreter/test/address.wast
That assumes it's not a validation error, but instead, a runtime error.

It's possible to prove some loads/stores are out of bounds when
offset + sizeof(load/store)  > UINT_MAX

lukewagner:
While the bounds checks may be provably out-of-bounds if executed, it's possible they are intentionally never executed, so this may frustrate more than it helps.  Also, to have deterministic validation, we'd have to precisely define what it means to be "provably out of bounds" which would get hairy and inevitably leave out some cases where you could still end up with a constant value in the compiler backend.

saambarati:
@lukewagner By provably out of bounds, I'd mean the simplest possible proof, which is what I wrote:
offset + sizeof(load/store)
These are known to the world's most naive wasm compiler.

saambarati:
@lukewagner To elaborate, I never want to propose something that would require a compiler churning on code as part of the validation step. But that's obviously not the case here.

For example, something I'd *not* be in favor of is if the following would be a validation error:
- wasm code emits a load from a constant UINT_MAX pointer.

This is trivial a compiler to figure out, but it should not be part of the validation. However, my point is that nothing like that is happening here. The `offset` and `sizeof(load/store)` are part of the actual load/store instruction. It requires no work extra work by the compiler to figure out their values.

lukewagner:
Ah, I see; I thought you were just giving an example, not proposing the exact rule.  Yes, if we're only talking about constraining the `offset` immediate w.r.t `UINT32_MAX`, that sounds reasonable (and potentially avoids weird corner cases in the compiler backend :).

saambarati:
@lukewagner Yes, sorry for not making that more clear. I am proposing the exact rule, which is based on looking at the constant operand to a load/store and the size of number of bytes of memory it loads/stores from.

lukewagner:
sgtm.  Perhaps you could write up the PR for Semantics.md#addressing and BinaryEncoding.md#memory-related-operators?

saambarati:
@lukewagner Sure, I'm happy to do that. Lets first make sure other implementors are also with this change. If they are, I will write the PR.

lukewagner:
Oh sure, I was expecting that to block merging, but waiting before is cool too.

sunfishcode:
[...]

PR #930, opened 2016-12-28
Make it a validation error to have memory operations that would trivially be out of bounds because their (offset + load/store size) exceeds the maximum possible memory size

saambarati:
Relevant discussion here:
https://github.com/WebAssembly/design/issues/929

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -584,6 +584,8 @@ natural alignment. The bits after the
 are reserved for [future :unicorn:][future threads] use
 (e.g., for shared memory ordering requirements).
 
+It is a validation error if `offset + numBytesTouched(load/store) - 1 > 2^32 - 1`. If this condition is true, the load or store is guaranteed to be out of bounds.
+
 The `reserved` immediate to the `current_memory` and `grow_memory` operators is
 for [future :unicorn:][future multiple tables] use and must be 0 in the MVP.
 
--- a/Semantics.md
+++ b/Semantics.md
@@ -174,6 +174,12 @@ the addition of the offset to the address never causes wrapping, so if the
 address for an access is out-of-bounds, the effective address will always also
 be out-of-bounds.
 
+It is a validation error if a load or store has an offset immediate where:
+```
+offset + numBytesTouched(load/store) - 1 > 2^32 - 1
+```
+If this condition is true, the load or store is guaranteed to be out of bounds.
+
 In wasm32, address operands and offset attributes have type `i32`, and linear
 memory sizes are limited to 4 GiB (of course, actual sizes are further limited
 by [available resources](Nondeterminism.md)). In wasm64, address operands and


BinaryEncoding.md
```diff
@@ -584,6 +584,8 @@ natural alignment. The bits after the
 are reserved for [future :unicorn:][future threads] use
 (e.g., for shared memory ordering requirements).
 
+It is a validation error if `offset + numBytesTouched(load/store) - 1 > 2^32 - 1`. If this condition is true, the load or store is guaranteed to be out of bounds.
```
jfbastien: How about `sizeof(memoryAccess)` instead? We use `sizeof` in this page. I'd also only talk about "memory access" instead of "load or store" so that we don't have to change the text when we add atomics.

sunfishcode: It might clarify the rule to express it as `... >= 2^32` instead of `... > 2^32 - 1`.

sunfishcode: Emphasizing how this catches accesses "guaranteed to be out of bounds" could seem to suggest this feature is intended to help catch bugs in application code. However, for reasons you noted earlier, the specific check here wouldn't be very effective for that purpose. To me, the more important motivation for this check is eliminating certain special cases that would arise as a result of integer overflows.

So perhaps change "If this condition is true, the load or store is guaranteed to be out of bounds" to "This frees consumers from some special cases that could arise from integer overflows in addressing arithmetic" or so.

Semantics.md
```diff
@@ -174,6 +174,12 @@ the addition of the offset to the address never causes wrapping, so if the
 address for an access is out-of-bounds, the effective address will always also
 be out-of-bounds.
 
+It is a validation error if a load or store has an offset immediate where:
+```
+offset + numBytesTouched(load/store) - 1 > 2^32 - 1
+```
+If this condition is true, the load or store is guaranteed to be out of bounds.
```
sunfishcode: Both of my comments apply here as well.

Alternatively, instead of having two copies of this rule, it might be sufficient to just state it here in Semantics.md and omit it in BinaryFormat.md.

rossberg-chromium:
Is there any specific benefit in adding ad-hoc validation errors like this one? Note that the purpose of validation isn't debugging of faulty producers. Unless it somehow simplifies the lives of either (correct) producers or consumers, I would prefer not to introduce special errors for random corner cases.

sunfishcode:
Yes. I initially had the same concern. However, this isn't about catching producer or application errors. It's about simplifying consumers by eliminating a case where integer overflow could occur inside the consumer. It doesn't eliminate all integer overflow concerns from wasm, but integer overflow is pernicious enough that rejecting it when we can seems worthwhile.

AndrewScheidecker:
I agree with @rossberg-chromium; the existing runtime error is enough to test this corner case in consumers, and to provide feedback to producers.

Should a C program like this fail to validate?
`int* x; ... x[2000000000] = 0;`

sunfishcode:
@AndrewScheidecker (That specific example doesn't trigger the overflow condition proposed here, but it's similar to one that would.) In code like that, it would be the responsibility of the C compiler to refrain from folding the constant into the offset field of the store.

rossberg-chromium:
@AndrewScheidecker, good point. So this change would actually burden producers with more special cases they have to look out for (or just as likely, cause them to have subtle bugs).

jfbastien:
Can you explain what the burden on producers is? The intent is to forbid `offset + mem_size` exceeding 32-bit. This is never useful, but forces consumers to handle values as 64-bit and / or tread lightly around all accesses.

titzer:
I know we had to be careful for these overflows in our implementation (and
that was annoying), but that said, I don't think they should be validation
errors, since IMO they fall into the same class as statically-known
division by zero.

lukewagner:
I think the difference is that allowing this corner case can make more work (and chance for security bugs) for otherwise-regular bounds-checking schemes.  So it's more than just "watch out when you're doing constant optimizations"; because in those cases you can just conservatively fall back to not doing the optimization.  That being said, it doesn't seem like a huge burden.

rossberg-chromium:
@jfbastien, the extra case is a pitfall that producers have to be aware of and actively work around. It breaks "obvious" equivalences they might rightly expect, such as
```
(load (i32.const N))  ==  (load offset=N (i32.const 0))
```
for all i32 values N. Compilers may want to exercise transformations along these lines, and would now be bothered with one more (fairly non-obvious, because artificial) special case that they have a good chance of missing.

saambarati:
I agree with @lukewagner here. That said, I appreciate both arguments and am not strongly in favor of adding the rule. Just mildly in favor of it. I understand that wanting to add validation error rules for all corner cases is not a tenable strategy.

(I'll leave this open until other folks are ok with closing it too.)

sunfishcode:
The arguments in favor of it are indeed modest, but they still seem to me stronger than those against. 

The transformation `(i32.load (i32.const N))  == (i32.load offset=N (i32.const 0))` would indeed become invalid. However on the other hand, the following equivalence, that one also might rightly expect, is currently just as invalid, and would become valid under the proposed rule here:

```
(i32.load offset=N (i32.const 0) == (i32.or (i32.load16_u offset=N (i32.const 0))
                                            (i32.shl (i32.load16_u offset=N+2 (i32.const 0)) (i32.const 2)))
```
Split loads/etc like this might occur in SIMD polyfills, for example. So the rule cuts two ways.

titzer:
@sunfishcode I'm still not sure how this rule helps, even in your example.

let A = `(i32.load offset=N (i32.const 0)`
let B = `(i32.or (i32.load16_u offset=N (i32.const 0))
                                            (i32.shl (i32.load16_u offset=N+2 (i32.const 0)) (i32.const 2))`

It is already the case that the offset is required to be a varuint32, so the transformation `A => B` isn't legal if `N + 2 > MAXUINT32`, since there is no way to legally encode that value. In effect, the wraparound would occur in encoding. So `B` is already not legal to encode, as you pointed out. But reacting to this by also making `A` not legal to encode is not the right answer IMO.

eholk:
I would be in favor of keeping this case as a runtime error rather than requiring it to be rejected during validation. It strikes me as more consistent with the overall behavior of WebAssembly to have all out of bounds accesses behave the same way.

jfbastien:
@saambarati would you like to keep things as-is, or move this forward?

saambarati:
@jfbastien I've come around to simply leaving it as a runtime error.

PR #932, opened 2016-12-29
Export name uniqueness

jfbastien:
spec test exports.wast makes it a validation failure for exported names to be duplicate. I find this somewhat surprising (I see exports as a dictionary, so duplicates simply overwrite previous entries), I'd therefore be happy if others want to reverse the polarity of this (make it OK to duplicate export names).

### DIFF
--- a/Modules.md
+++ b/Modules.md
@@ -48,7 +48,9 @@ In the future, other kinds of imports may be added. Imports are designed to
 allow modules to share code and data while still allowing separate compilation
 and caching.
 
-All imports include two opaque names: a *module name* and an *export name*. The
+All imports include two opaque names: a *module name* and an *
+
+name*. The
 interpretation of these names is up to the host environment but designed to
 allow a host environments, like the [Web](Web.md), to support a two-level
 namespace.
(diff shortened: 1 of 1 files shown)


Modules.md
```diff
@@ -48,7 +48,9 @@ In the future, other kinds of imports may be added. Imports are designed to
 allow modules to share code and data while still allowing separate compilation
 and caching.
 
-All imports include two opaque names: a *module name* and an *export name*. The
+All imports include two opaque names: a *module name* and an *
+
+name*. The
```
lukewagner: I expect this change wasn't intended.

lukewagner:
Heh, I said "request changes", but it looks so negative with all the red and Xs; I'd prefer an "approve with nits" that is happy and green.

jfbastien:
Though the check isn't during compilation :-)

lukewagner:
Actually, maybe you want to make that weak in this PR?

lukewagner:
Oops, I see #933 now.

jfbastien:
> Actually, maybe you want to make that weak in this PR?

Yeah editing github through their UI doesn't seem to let me add more files. I could pull the branch but that's so much work!

I'll merge this PR for now since it makes the status quo official, but I'm happy to reverse the polarity if anyone feels we got it wrong.

PR #935, opened 2017-01-04
customSections order

jfbastien:
Duplicates currently aren't ordered.


lukewagner:
lgtm

jfbastien:
I'll assume this won't be contended since it seems like a simple oversight. Merging.

PR #944, opened 2017-01-11
Add @@toStringTag to JS API prototypes

lukewagner:
This PR adds `@@toStringTag` properties to the new builtin WebAssembly objects.  As with the other JS builtin types, this allows `Object.prototype.toString` to produce, e.g., `[object WebAssembly.Instance]` instead of the default `[object Object]`.  The `WebAssembly.` prefix was added to avoid ambiguity with ES6 Module objects which are already have the `@@toStringTag` `Module`.  The `WebAssembly` object is given the tag `WebAssembly` for symmetry with `Math` and `JSON` which are similarly tagged.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -36,6 +36,13 @@ the global object. Like the `Math` and `JSON` objects, the `WebAssembly` object
 is a plain JS object (not a constructor or function) that acts like a namespace
 and has the following properties:
 
+### `WebAssembly [ @@toStringTag ]` Property
+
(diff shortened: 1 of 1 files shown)


rossberg-chromium:
LGTM, thanks!

rossberg-chromium:
FWIW, I noticed that this implies that `String(WebAssembly.Module.prototype) == "[object WebAssembly.Module]"`, although the prototype isn't a valid instance. That's stupid, but once more it's how JavaScript works, the same holds for `Map.prototype` and others.

Issue #943, opened 2017-01-10
Imported memory validation step


Cellule:
According to https://github.com/WebAssembly/design/blob/master/Modules.md#imports, there needs to be a validation for an imported memory.
We should also add this to [JS.md](https://github.com/WebAssembly/design/blob/master/JS.md#webassemblyinstance-constructor) in the steps when importing a memory or at least make a reference to the validation.

lukewagner:
Technically this validation is done within `Eval.init` (in [`add_import`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/eval.ml#L390) which calls [`check_limits`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/eval.ml#L380) for memory and table imports) since this is part of core wasm semantics, not the JS binding, but you're probably right that it'd be more clear to make a mention in JS.md with link to Modules.md in the associated memory/table import cases.

lukewagner:
Should be resolved by #945.

PR #945, opened 2017-01-11
Add memory/table import validation note

lukewagner:
This PR adds a clarification as suggested by #943.


rossberg-chromium:
LGTM, but I just noted that for symmetry, we should also change step 4.ii.a about checking functions signatures into such a note, since conceptually that is an analogous import type check performed in Eval.init.

lukewagner:
Ah, good observation.

rossberg-chromium:
LGTM

lukewagner:
@Cellule Is this what you were looking for?

Cellule:
LGTM thanks !

Issue #948, opened 2017-01-12
WebAssembly.imports/exports: why Array and not Iterable?


mtrofin:
The size of imports/exports may be large. Instead of pre-allocating an Array, why not return an object conforming to the Iterable protocol, thus avoiding memory pressure?

Kardax:
Arrays have many useful features and characteristics, and would represent a very small portion of the memory used by any non-trivial WASM file.  I don't see any issue with this aspect of the design.

jfbastien:
@Kardax right, but if you want an array then you can use spread syntax with something using the iteration protocol. You can't do the reverse.

Which one is more desirable is based on what's more idiomatic "modern" JS. If we change here we should also change other APIs, such as `Module.customSections`.

rossberg-chromium:
The size of imports/exports may be large but it is clearly dominated by the size of the binary, so shouldn't matter in practice.

It is normal in JS to use an array for this kind of thing, see e.g.  the analogue `Object.keys` and friends.

mtrofin:
It matters when you are memory constrained. By returning an array, we increase memory pressure.

jfbastien:
@rossberg-chromium normal in ye-olden JS. Is it normal in "modern" JS? If so I think we should follow "modern".

Agreed import/export aren't big, but `customSections` could be.

lukewagner:
While I expect the payload of some custom sections to be big, I expect the number of custom sections (esp., with a particular name) to be pretty small.

My impression is that, even for huge modules, we're talking 100s of imports/exports, and then only for huge modules where the bytecode is 1000x bigger than the aggregate size of the returned arrays.  So I think an Iterable would be overkill.  Also, these functions are really just a convenience to allow clients to avoid manually decoding the binary, but manually decoding is actually pretty easy (that's what @yurydelendik and @kripken did in their respective module loaders before we added these functions) so that's always a workaround for users in pathological situations (who might want to avoid the GC allocation of an iterable altogether).

mtrofin:
@lukewagner What about customSections, as @jfbastien mentioned?

If these APIs are just convenience, why even include them - did we have a strong ask/scenario for them? We can always add later (apologies if this was discussed elsewhere - if so, pointer, pls)

jfbastien:
It was added because it's very simple, and otherwise tools such as emscripten's generated code end up decoding the sections themselves in the browser, which is kinda silly. Adding it later basically means we're stuck with silly code forever.

[...]

PR #952, opened 2017-01-13
Make Module.customSections more closely match Module.imports and Module.exports

titzer:
The previous text was not precise about the return value of the WebAssembly.Module.customSections function. This PR updates the text and precisely defines the return value of this function to more closely match the WebAssembly.Module.imports and WebAssembly.Module.exports functions.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -224,12 +224,10 @@ Array customSections(moduleObject, sectionName)
 If `moduleObject` is not a `WebAssembly.Module`, a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
 is thrown.
 
-Let `sectionNameString` be the result of [`ToString`](https://tc39.github.io/ecma262/#sec-tostring)(`sectionName`).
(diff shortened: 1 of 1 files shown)


lukewagner:
This PR completely changes `Module.customSections` to return a copy of *all* custom sections.  I know `Module.imports` and `Module.exports` return a copy of all imports/exports, but in that case we're expecting a relatively small number of imports/exports.  A single module may have multiple huge custom sections though, so returning a copy of all of them seems problematic.

titzer:
Sure, I can see that behavior being better, but the current text is not very precise. E.g. it mentions returning an 'Array', but there doesn't seem to any real reason why this needs to be a requirement. Or is it an Array of names, which has additional getters for each custom section's name?

jfbastien:
Sections can have duplicate names, the Array contains each section with the queried name. This seems like the desirable behavior (or, as @mtrofin suggests in #948, an iterable of all sections with that name). Agree with @lukewagner: I don't think we want to return all sections.

PR #953, opened 2017-01-13
Clarify when TypeError is thrown for exotic fcts

mtrofin:
Making it crystal clear when TypeError is thrown for invalidly typed exotic function arguments/return vals.


jfbastien:
lgtm

PR #962, opened 2017-01-20
WebAssembly.instance's pair return property attributes

mtrofin:
Clarified the attributes of the {module, instance} return


lukewagner:
lgtm

PR #963, opened 2017-01-21
Clarify WebAssembly.Table "maximum" property

mtrofin:
Text was ambiguous when RangeError is thrown and when the property is None.


lukewagner:
lgtm

Issue #838, opened 2016-10-27
Can WebAssembly.Instance recompile an already-compiled Module?


jfbastien:
Say I have this code:

``` js
let descriptor = /* ... */;
let memories = Array.apply(null, {length: 1024}).map(() => new WebAssembly.Memory(descriptor));
let instance = fetch('foo.wasm')
  .then(response => response.arrayBuffer())
  .then(buffer => WebAssembly.compile(buffer))
  .then(module => new WebAssembly.Instance(module, { memory: memories[1023] }));
```

Is `WebAssembly.Instance` allowed to block for a substantial amount of time. Could it for example recompile the `WebAssembly.Module`?

In most cases I'd say no, but what if the already-compiled code doesn't particularly like the memory it receives? Say, because that memory is a slow-mode memory and the code was compiled assuming fast-mode? maybe `memories[0]` was a fast-mode memory, but `memories[1023]` sure won't be.

What about this code instead:

``` js
let instances = [0,1,2,3,4,5,6,7].map(v => fetch(`foo${v}.wasm`)
  .then(response => response.arrayBuffer())
  .then(buffer => WebAssembly.compile(buffer))
  .then(module => new WebAssembly.Instance(module)));
```

Are those calls to `WebAssembly.Instance` allowed to cause recompilation?

Assuming the above makes sense, here are a few related questions:
- Do we want a promise-returning async function which can compile _and_ instantiate? I'm not saying that we should drop any of the synchronous and asynchronous APIs that we already have, I'm proposing a new asynchronous API.
- How does a browser expose that compiled code in a `WebAssembly.Module` is fast, and that a `WebAssembly.Memory` instance is suitable for such fast code? Right now the answer seems to be "try it and see if you can notice".
- How does a user know how many `WebAssembly.Memory` instances they're allowed before they get slow code (counting the implicit ones, e.g. as created by the second example)?

mbebenita:
Would be nice for `WebAssembly.Instance` to sometimes cause recompilation, this way immutable globals vars could be constant folded in the generated code. For instance, Emscripten generates relocatable code by offsetting all pointers to static data. The offset is passed in as an immutable global var when the module is instantiated. If `WebAssembly.Instance` can recompile, it could specialize the generated code.

rossberg-chromium:
The specification doesn't define what "compilation" is, nor would it make
sense for it to do so, because implementation approaches may wildly differ
(including interpreters). So it cannot have any normative say about this
[...]

PR #991, opened 2017-02-17
Adding Response based versions of validate, compile, and instantiate.

flagxor:
This allows us to support streaming compilation and origin bound
sources for WebAssembly modules.

Proposes a mime type of application/wasm for WebAssembly modules.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -88,7 +88,7 @@ If the given `bytes` argument is not a
 the returned `Promise` is [rejected](http://tc39.github.io/ecma262/#sec-rejectpromise)
 with a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror).
 
-Otherwise, this function starts an asychronous task to compile a `WebAssembly.Module`
+Otherwise, this function starts an asynchronous task to compile a `WebAssembly.Module`
 as described in the [`WebAssembly.Module` constructor](#webassemblymodule-constructor).
 On success, the `Promise` is [fulfilled](http://tc39.github.io/ecma262/#sec-fulfillpromise)
 with the resulting `WebAssembly.Module` object. On failure, the `Promise` is 
--- a/Web.md
+++ b/Web.md
@@ -22,6 +22,107 @@ and retrieve compiled modules from offline storage, instantiate compiled modules
 with JavaScript imports, call the exported functions of instantiated modules,
 alias the exported memory of instantiated modules, etc.
 
+The Web embedding includes additional methods useful in that context.
+Non-web embeddings are not required to support these additional methods.
+
+### Additional Web Embedding API
+
+#### `WebAssembly.compile`
+
+:cyclone: Added for milestone 2, developers must feature detect.
+
+In Web embeddings, the following overloads are added (in addition to the core
+JS API method of the same name).
+
+```
+Promise<WebAssembly.Module> compile(Response source)
+
+Promise<WebAssembly.Module> compile(Promise<Response> source)
+```
+
+Developers can set the argument `source` with either a promise that resolves
+with a
+[`Response`](https://fetch.spec.whatwg.org/#response-class)
+object or a
+[`Response`](https://fetch.spec.whatwg.org/#response-class)
+object (which is automatically cast to a
+promise).
+If when unwrapped that `Promise` is not a `Response` object, then the returned `Promise` is
+[rejected](http://tc39.github.io/ecma262/#sec-rejectpromise)
+with a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror).
+Renderer-side
+security checks about tainting for cross-origin content are tied to the types
+of filtered responses defined in
+[`Fetch`](https://fetch.spec.whatwg.org/#concept-fetch).
+
+This function starts an asynchronous task to compile a `WebAssembly.Module`
+as described in the [`WebAssembly.Module` constructor](#webassemblymodule-constructor).
+On success, the `Promise` is [fulfilled](http://tc39.github.io/ecma262/#sec-fulfillpromise)
+with the resulting `WebAssembly.Module` object. On failure, the `Promise` is
+[rejected](http://tc39.github.io/ecma262/#sec-rejectpromise) with a
+`WebAssembly.CompileError`.
+
+The `Promise<Response>` is used as the source of the bytes to compile.
+MIME type information is
+[`extracted`](https://fetch.spec.whatwg.org/#concept-header-extract-mime-type)
+from the `Response`, WebAssembly `source` data must have a MIME type of `application/wasm`,
+extra parameters are not allowed (including empty `application/wasm;`).
+MIME type mismatch or `opaque` response types
+[reject](http://tc39.github.io/ecma262/#sec-rejectpromise) the Promise with a
+`WebAssembly.CompileError`.
+
+#### `WebAssembly.instantiate`
+
+:cyclone: Added for milestone 2, developers must feature detect.
+
+In Web embeddings, the following overloads are added (in addition to the core
+JS API method of the same name).
+
+```
+Promise<{module:WebAssembly.Module, instance:WebAssembly.Instance}>
+  instantiate(Response source [, importObject])
+
+Promise<{module:WebAssembly.Module, instance:WebAssembly.Instance}>
+  instantiate(Promise<Response> source [, importObject])
+```
+
+Developers can set the argument `source` with either a promise that resolves
+with a
+[`Response`](https://fetch.spec.whatwg.org/#response-class)
+object or a
+[`Response`](https://fetch.spec.whatwg.org/#response-class)
+object (which is automatically cast to a
+promise).
+If when unwrapped that `Promise` is not a `Response` object, then the returned `Promise` is
+[rejected](http://tc39.github.io/ecma262/#sec-rejectpromise)
+with a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror).
+Renderer-side
+security checks about tainting for cross-origin content are tied to the types
+of filtered responses defined in
+[`Fetch`](https://fetch.spec.whatwg.org/#concept-fetch).
+
+This function starts an asynchronous task that first compiles a `WebAssembly.Module`
+based on bytes from `source` as described in
+the [`WebAssembly.Module` constructor](#webassemblymodule-constructor)
+and then instantiate the resulting `Module` with `importObject` as described in the
+[`WebAssembly.Instance` constructor](#webassemblyinstance-constructor).
+On success, the `Promise` is [fulfilled](http://tc39.github.io/ecma262/#sec-fulfillpromise)
+with a plain JavaScript object pair `{module, instance}` containing the resulting
+`WebAssembly.Module` and `WebAssembly.Instance`. The 2 properties `module` and `instance` of the returned pair are  configurable, enumerable and writable.
+
+On failure, the `Promise` is
+[rejected](http://tc39.github.io/ecma262/#sec-rejectpromise) with a
+`WebAssembly.CompileError`, `WebAssembly.LinkError`, or `WebAssembly.RuntimeError`, depending on the cause of failure.
+
+The `Promise<Response>` is used as the source of the bytes to compile.
+MIME type information is
+[`extracted`](https://fetch.spec.whatwg.org/#concept-header-extract-mime-type)
+from the `Response`, WebAssembly `source` data must have a MIME type of `application/wasm`,
+extra parameters are not allowed (including empty `application/wasm;`).
+MIME type mismatch or `opaque` response types
+[reject](http://tc39.github.io/ecma262/#sec-rejectpromise) the Promise with a
+`WebAssembly.CompileError`.
+
 ## Modules
 
 WebAssembly's [modules](Modules.md) allow for natural [integration with


rossberg-chromium:
Can we make these separate functions instead of overloading further? Also, I think we need to separate Web-specific extensions from the "general" JS API spec.

flagxor:
Agree on the overloads.
What do you and others think of:
downloadAndCompile
downloadAndValidate
downloadAndInstantiate

Should we separate into separate file (or clean up Web.md) or just a section in this one?

rossberg-chromium:
Names sound okay to me. Even though less convenient, I'd have a slight preference for a separate file, since technically, it is a separate spec (a supplement) -- I'm thinking of it in analogy to the DOM supplementing the ES lib on the web.

JS.md
```diff
@@ -62,51 +62,87 @@ The following intrinsic objects are added:
...
-The `validate` function has the signature:
+The `validate` function is overloaded based on types of its arguments.
+If none of the following overloads match, then the returned `Promise` is
+[rejected](http://tc39.github.io/ecma262/#sec-rejectpromise)
+with a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror).
 
 ```
 Boolean validate(BufferSource bytes)
+
+Boolean validate(Response source)
```
titzer: Hmm, since this is a synchronous method (it returns a Boolean), how can it take a Response? Should it instead create a Promise<Response>? In that case, I think it should not be an overload, but a separate function.

titzer:
Agree with @rossberg-chromium that these things are specific to a Web embedding and should probably move to their own .md file.

JS.md
lukewagner: `validate` is only useful for feature testing, so I'm not sure it's worth adding a whole Response overload and Promise type for this.  I don't think anyone should be calling validate on a whole codebase.  Furthermore, validation, unlike compilation, is crazy fast.

lukewagner:
Separate functions and files sound fine.  Not to bikeshed, though, but "download" sounds a bit off.  What about the more literal names `compileResponse`/`instantiateResponse`?

lukewagner:
Actually, could these go in Web.md?

JS.md
flagxor: Agreed, hadn't thought of it until making the PR, will remove.

flagxor:
Moved and renamed.
PTAL

Web.md
```diff
@@ -22,6 +22,82 @@ and retrieve compiled modules from offline storage, instantiate compiled modules
 with JavaScript imports, call the exported functions of instantiated modules,
 alias the exported memory of instantiated modules, etc.
 
+In addition to the core JavaScript API, the Web embedding includes additional
```
jfbastien: You have an "additional" "addition" in this sentence.

Web.md
```diff
@@ -22,6 +22,82 @@ and retrieve compiled modules from offline storage, instantiate compiled modules
...
+[rejected](http://tc39.github.io/ecma262/#sec-rejectpromise)
+with a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror).
+
+```
+Promise<{module:WebAssembly.Module, instance:WebAssembly.Instance}>
+  instantiateResponse(Response source [, importObject])
+
+Promise<{module:WebAssembly.Module, instance:WebAssembly.Instance}>
+  instantiateResponse(Promise<Response> source [, importObject])
+```
```
jfbastien: How does one use `instantiateResponse` with an `importObject` that contains a `WebAssembly.Memory` from another `instantiateResponse`? i.e. create multiple instances which share a memory, using the response APIs.

Web.md
```diff
@@ -22,6 +22,82 @@ and retrieve compiled modules from offline storage, instantiate compiled modules
...
+This function starts an asynchronous task to compile a `WebAssembly.Module`
+as described in the [`WebAssembly.Module` constructor](#webassemblymodule-constructor).
+On success, the `Promise` is [fulfilled](http://tc39.github.io/ecma262/#sec-fulfillpromise)
+with the resulting `WebAssembly.Module` object. On failure, the `Promise` is
+[rejected](http://tc39.github.io/ecma262/#sec-rejectpromise) with a
+`WebAssembly.CompileError`.
+
+The `Response` or `Promise<Response>` is used as the source of the
+bytes to compile.
+WebAssembly `source` data is expected to have a mime type of `application/wasm`.
```
jfbastien: `s/mime/MIME/` here and elsewhere.

jfbastien:
I think [this comment](https://github.com/WebAssembly/design/pull/991#discussion_r101837953) contains an important design point:

> How does one use `instantiateResponse` with an `importObject` that contains a `WebAssembly.Memory` from another `instantiateResponse`? i.e. create multiple instances which share a memory, using the response APIs.

I'm worried that the proposed API only allows us to use `response` for a single `.wasm`, but not to combine multiple `.wasm` that'll share a memory (or, later, more than just a memory).

Do you think that's an important usecase, and that this API should get it right from the beginning?

Web.md
Constellation: `application/wasm` sounds reasonable. In the current `<script type="module">` pipeline, we have a restriction that the MIME type of the response should be one of JS MIME type. This is great since we can choose the different way to produce different modules (JS v.s. WASM modules) based on the MIME type of the given Response.

Web.md
```diff
@@ -22,6 +22,82 @@ and retrieve compiled modules from offline storage, instantiate compiled modules
...
+
+The `compileResponse` function is overloaded based on types of its arguments.
+If none of the following overloads match, then the returned `Promise` is
+[rejected](http://tc39.github.io/ecma262/#sec-rejectpromise)
+with a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror).
+
+```
+Promise<WebAssembly.Module> compileResponse(Response source)
+
+Promise<WebAssembly.Module> compileResponse(Promise<Response> source)
```
domenic: How this works on the web is that the first step of the algorithm wraps whatever it gets in a promise, then unwraps the promise before continuing. I'm not sure how that translates to the notation you're using here (TypeScript??) but it should be noted in the algorithm. In Web IDL it's just `Promise<Response> source` as the argument, and then any incoming argument gets converted by the Web IDL bindings.

Web.md
```diff
@@ -22,6 +22,82 @@ and retrieve compiled modules from offline storage, instantiate compiled modules
...
+[rejected](http://tc39.github.io/ecma262/#sec-rejectpromise) with a
+`WebAssembly.CompileError`.
+
+The `Response` or `Promise<Response>` is used as the source of the
+bytes to compile.
+WebAssembly `source` data is expected to have a mime type of `application/wasm`.
+Other mime types
+[reject](http://tc39.github.io/ecma262/#sec-rejectpromise) the Promise with a
+`WebAssembly.CompileError`.
+
```
domenic: Should mention that the returned promise is rejected with a TypeError if the incoming value, after unwrapping, is not a `Response`.

Web.md
lukewagner: You'd need to Promise-chain them; but I think this is already the case with `instantiate`.  To avoid the sequential dependency, one would want to instead create the memory/table up front (via their JS constructor) and pass them simultaneously into N `instantiateResponse` calls.

Web.md
flagxor: oops, fixed.

Web.md
And good point.

Web.md
jfbastien: It'd like to see what the code looks like, and have folks such as @domenic and @slightlyoff chime in.

Web.md
flagxor: Sounds better. Changed.

Web.md
flagxor: Done

Web.md
flagxor: I would think you'd either in this case want the memory to be external, or if you wanted it to originate from another module, you'd have to first instantiate that module (i.e. like a libc).
That does have the unfortunate property of serializing things.
The current instantiate method has this same issue, though or am I missing something?
I suppose we could also accept multiple modules bound to one internal memory, but at that point I wonder if we wouldn't be better off sorting out ES6 modules (as they should have a similar behavior).

lukewagner: Sequential:
```
WebAssembly.instantiate(fetch(moduleA)).then(resultsA =>
  WebAssembly.instantiate(fetch(moduleB), resultsA.instance.exports).then(resultsB =>
    ...
  )
);
```
Parallel:
```
var imports = { libc: {
  mem: new WebAssembly.Memory(...),
  tbl: new WebAssembly.Table(...)
} };
Promise.all([
  WebAssembly.instantiate(fetch(moduleA), imports),
  WebAssembly.instantiate(fetch(moduleB), imports)
]).then(([resultsA, resultsB]) =>
  ...
);

**Comment by flagxor** [comment]
PTAL

File: Web.md (hunk 3 above)
Comment (flagxor) [inline Web.md]: I think that only works if you don't need to import exports between A and B.
In that case you can build a web of promises that serialize things as little as possible, but you wouldn't be able to get maximum concurrency on the compiles.
If there's dependency chains like that, then you can do better if you use the two-step compile instantiate (as you can get all the compiles going in parallel, download already are).

That's no worse that with the non-response versions of these APIs.

We could try to fix that also, but as I said before its probably better to do that with ES6 modules.
If we did it for the one-step instantiate methods, it would probably end up requiring a declarative way to describe all the hookups (which seems to re-do a lot of the ideas from the ES6 modules space).

Comment (lukewagner) [inline Web.md]: Agreed.

> If we did it for the one-step instantiate methods, it would probably end up requiring a declarative way to describe all the hookups (which seems to re-do a lot of the ideas from the ES6 modules space).

IIUC, there is a JS function to dynamically load an ES6 module proposed now (I forget the name).  With wasm modules integrated with ES6 modules, that would basically give us this function.

File: Web.md [hunk 7]
```diff
@@ -22,6 +22,67 @@ and retrieve compiled modules from offline storage, instantiate compiled modules
...
+This function starts an asynchronous task to compile a `WebAssembly.Module`
+as described in the [`WebAssembly.Module` constructor](#webassemblymodule-constructor).
+On success, the `Promise` is [fulfilled](http://tc39.github.io/ecma262/#sec-fulfillpromise)
+with the resulting `WebAssembly.Module` object. On failure, the `Promise` is
+[rejected](http://tc39.github.io/ecma262/#sec-rejectpromise) with a
+`WebAssembly.CompileError`.
+
+The `Promise<Response>` is used as the source of the bytes to compile.
+WebAssembly `source` data is expected to have a MIME type of `application/wasm`.
```
Kardax: I'd rather see this as application/webassembly for symmetry with JavaScript, which is application/javascript, not application/js.

domenic: For the record JS is all over the place: https://html.spec.whatwg.org/multipage/scripting.html#javascript-mime-type

Kardax: Indeed, but never abbreviated.

wllang:
@flagxor Thank you for exploring the space.

> This allows us to support streaming compilation and origin bound sources for WebAssembly modules.

Great to see some effort to address this space of the design.

It is not clear (to me anyway) that this meets the needs of resource planning and caching. I understand that the web browser could cache the source data, but it is not clear if that caching will work well if there is a streaming translation stage?

The compiler might cache all the input stream, and use that as a key, but that seems a huge burden.

What would you think about having a build file, or Makefile of sorts, and that being the input:

```
Promise<{module:WebAssembly.Module, instance:WebAssembly.Instance}> build(makefile [, importDescriptor])
```

The makefile could be text or json, whatever suits people, and would describe the steps required to get the app running, and in a way that allows the web browser to schedule the fetches and to schedule linear memory allocation and to cache parts of the compilation work. Perhaps it could even launch multiple threads all using the same linear memory, and thus be able to optimize for that if that was a significant matter for the web browser. This could also include provision for a custom translation stage.

I believe that your proposal could be used to implement parts of such a build process, but that a web browser could do much better with the description than your proposal could ever do. Also, should the build file prove to be a dead end then it might be a small matter to create a polyfill for it.

Perhaps part of the problem with Responses is that they have no version or provenance. Can the compiler even trust that the stream is from a URL with a name and version without any modification? 

For caching to be effective the web browser needs keys, and if it can use named sources with versions and data flow based on these as keys then it is a lot better off than having to use the data at each stage as a key.

jfbastien:
Continuing on [this comment](https://github.com/WebAssembly/design/pull/991#discussion_r101837953), IIUC:

* Sharing a memory seems OK as long as the memory is instantiated explicitly, no through the module's memory section.
* This API doesn't allow us to instantiate two modules where one imports the other's functions, unless we serialize manually.
* We'll get something based on ES6 modules that'll allow non-blocking instantiation where imports are shared (even cyclic?) and can even be wasm <-> JS seamlessly.

Is the intent then to have:

1. Blocking APIs for wasm (already in).
2. Non-blocking promise APIs for wasm. (already in)
3. Non-blocking promise+response APIs for wasm (this proposal).
4. ES6 APIs for wasm, shared with JS (later).

?

wllang:
While looking at the potential of ES6 modules, I see references to a Loader, and find pages such as the following https://whatwg.github.io/loader/#request-translate and https://gist.github.com/dherman/7568080 which includes discussion of setting up a pipeline and includes a source-to-source translation stage. This all sounds much closer to what might be needed.

I also note that the ES6 Modules feature **static** definitions, but they appear to be just top level forms in a JS file, and ES6 Modules alone do not appear to have what we need, rather we already have imports and exports, and simply adding a `require` value might give us all that ES6 Modules offer and that does not seem enough to address the issues?

I also see a lot of uses of Promises, but Promises do not appear to be **static** descriptions (is that right?), so they do not seem to be the appropriate way to express the loader.

Perhaps we need a static description of the loader and need to extend the ideas to meet the specific needs of wasm, so to be able to define a binary-to-wasm translation stage (rather than source-to-source), and to work with the constraints imposed by the linear memory.

Would @dherman be willing and able to help on the state of the Loader?

annevk:
I realize this is not the formal language, but I wanted to point out some things with respect to MIME types that would be good to have clarified for implementations and tests:

1. It would be good to define extraction in terms of https://fetch.spec.whatwg.org/#concept-header-extract-mime-type.
2. It would be good to state whether MIME type parameters need to be ignored or if their presence leads to failure.
3. If you decide on failure, make sure to test `application/wasm;`, which would then technically have to fail if I remember correctly.

It would also be good to point out an opaque filtered response leads to failure as you don't explicitly define yet how you extract the data out of a response. (I.e., cross-origin without CORS won't work.)

lukewagner:
@jfbastien That matches my understanding.  I found the [proposal](https://github.com/tc39/proposal-dynamic-import), currently Stage 3, for the dynamic import I mentioned above.  So if we made `application/wasm`-MIME-typed modules loaded by the ES6 module loader Just Work, I think the cases you're talking about would ride along for free.

jfbastien:
@lukewagner 
> @jfbastien That matches my understanding.  I found the [proposal](https://github.com/tc39/proposal-dynamic-import), currently Stage 3, for the dynamic import I mentioned above.  So if we made `application/wasm`-MIME-typed modules loaded by the ES6 module loader Just Work, I think the cases you're talking about would ride along for free.

OK, so the ES6 module loader API will Just Work for WebAssembly, and be strictly more powerful than this proposal? What would its timing look like, compared to this proposal?

@domenic maybe?

domenic:
I'm not sure what I'm supposed to be commenting on, exactly? It seems definitely false that a Response based version is less powerful than an import()-based one; for example, you can create synthetic streaming Responses easily, whereas without a service worker there is no such ability for import(). import() also has no ability to pass in a custom WebAssembly.Memory, or separate out instantiation and evaluation.

jfbastien:
@domenic 
> I'm not sure what I'm supposed to be commenting on, exactly? It seems definitely false that a Response based version is less powerful than an import()-based one; for example, you can create synthetic streaming Responses easily, whereas without a service worker there is no such ability for import(). import() also has no ability to pass in a custom WebAssembly.Memory, or separate out instantiation and evaluation.

Basically: I want to understand the **four** JavaScript APIs we're proposing for WebAssembly, and listed [here](https://github.com/WebAssembly/design/pull/991#issuecomment-280821206). I want a list of features each offers.

From my summary it sounds like Response is missing features compared to `import()`, you're saying `import()` is also not offering all the features of Response. That's fine! My point is: we're designing APIs without a clear picture of what each offers. Let's make a little table, one row per feature and one column per JS API, and agree that's the table we want users to see on MDN.

flagxor:
@jfbastien What you describe above with the 4 items also matches my understanding.

The ES6 modules based version does seem to get cyclic imports (to whatever extent they currently work generally). This could be emulated by mutating Tables in the current API. So I think they're of equal power in that regard. My suspicion is that we'll end up having to restrict ES6 module style wasm modules in ways we haven't considered (for instance not sure if the init function works ok with cyclic imports).

As @domenic points out the Response object will allow for more interception etc, without service workers. (I would imagine any performance difference with this would be dependent on the implementation, rather than innate).

rossberg-chromium:
It's probably best to view ES module integration as a convenience mechanism
on top of the explicit API. That is, it should be expressible in terms of
the Wasm API and the ES loader spec, as a custom loader that can be
polyfilled. I would be very reluctant to introducing it as a separate
primitive mechanism with whole new capabilities -- that would likely result
in a semantic mess.

jfbastien:
@rossberg-chromium 
> It's probably best to view ES module integration as a convenience mechanism
> on top of the explicit API. That is, it should be expressible in terms of
> the Wasm API and the ES loader spec, as a custom loader that can be
> polyfilled. I would be very reluctant to introducing it as a separate
> primitive mechanism with whole new capabilities -- that would likely result
> in a semantic mess.

IIUC, you're saying the `import()` API should have zero capabilities that the Response API doesn't already have. Is that correct?

flagxor:
Revised to overload again (but from the Web.md file).
PTAL

flagxor:
Any objections to this land?
Open on mime type (application/wasm vs application/webassembly), but got preferences both ways.

annevk:
I think you missed my comment with respect to parsing and processing the MIME type.

rossberg-chromium:
@jfbastien

> It's probably best to view ES module integration as a convenience mechanism
> on top of the explicit API. That is, it should be expressible in terms of
> the Wasm API and the ES loader spec, as a custom loader that can be
> polyfilled. I would be very reluctant to introducing it as a separate
> primitive mechanism with whole new capabilities -- that would likely result
> in a semantic mess.
>
> IIUC, you're saying the import() API should have zero capabilities that
> the Response API doesn't already have. Is that correct?
>

Well, response + base API. The two already have complementary capabilities.

jfbastien:
@flagxor see [this comment](https://github.com/WebAssembly/design/pull/991#issuecomment-281202429). I'd like to see a list of features for all 4 APIs, because this one seems designed without consideration of `import()`.

I *like* the Response API, but it comes out of concerns that the first two APIs were mis-designed, in isolation from other Web APIs. Let's avoid this happening again.

Web.md
```diff
@@ -22,6 +22,73 @@ and retrieve compiled modules from offline storage, instantiate compiled modules
...
+
+### Additional Web Embedding API
+
+#### `WebAssembly.compile`
+
+In Web embeddings, the following overload is added (in addition to the core
+JS API method of the same name).
+
+```
+Promise<WebAssembly.Module> compile(Promise<Response> source)
```
slightlyoff: I'd like to also see this overload to allow:

`Promise<WebAssembly.Module> compile(Response source)`

for symmetry with [`respondWith`](https://w3c.github.io/ServiceWorker/#fetch-event-respondwith)

Web.md
slightlyoff: I don't have a strong opinion here except to note that many other mimetypes abbreviate: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Complete_list_of_MIME_types

Web.md
```diff
@@ -22,6 +22,73 @@ and retrieve compiled modules from offline storage, instantiate compiled modules
...
+`WebAssembly.CompileError`.
+
+#### `WebAssembly.instantiate`
+
+In Web embeddings, the following overload is added (in addition to the core
+JS API method of the same name).
+
+```
+Promise<{module:WebAssembly.Module, instance:WebAssembly.Instance}>
+  instantiate(Promise<Response> source [, importObject])
```
slightlyoff: Per line 36 comment please allow `instantiate(Response source [, importObject])`

Web.md
```diff
@@ -22,6 +22,73 @@ and retrieve compiled modules from offline storage, instantiate compiled modules
...
+```
+
+The source of the material to compile and instantiate is a Promise of a
```
slightlyoff: To support direct use of `Response` you'll need to add wrapping, e.g.: https://w3c.github.io/ServiceWorker/#fetch-event-respondwith

Web.md

Web.md

Web.md
flagxor: done (I think I grok your meaning, have both signatures + verbiage, please check).

flagxor:
@slightlyoff PTAL

@jfbastien 
After chatting with folks about ES6 Modules, I'm unsure we can immediately conclude how wasm should interoperate with them. As @rossberg-chromium pointed out, I suspect we'll end up with polyfills based on the Response API, and need to experiment with what's more ergonomic, as the point of modules is partially JS use ergonomics. The table below lists my best guesses for ES6, but note there's a ton of maybes in the ES6 column, because how it works depends on how we choose to have it work (does es6 module == wasm module or wasm instance etc).

@slightlyoff @domenic Would you say the Response API as done here has any concerns in terms of things we should be looking out for with modules?

| Capability | Synchronous API | Asynchronous API | Response API | ES6 Modules / import() API |
| --- | --- | --- | --- | --- |
| Bind to an external Memory | Yes | Yes | Yes | Maybe? |
| Bind to an external Table | Yes | Yes | Yes | Maybe? |
| Allow cyclic imports | No | No | No | Maybe? |
| Allow multiple instances per page | Yes | Yes | Yes | Maybe? |
| Program visible module object | Yes | Yes | Yes | No |
| Easy dependency setup | No | No | No | No | Yes |
| Implicit caching | No | No | Yes | Yes |
| Explicit caching | Yes | Yes | Yes | Partial? |
| Share code cross-Worker | Yes | Yes | Yes | Yes (implicit) |
| Share memory cross-Worker | Yes | Yes | Yes | Maybe? |
| Validation without compile | Yes | No | No | No |
| Size caps on main thread in some implementations | Yes | No | No | No |

slightlyoff:
I've got a lot of concerns about `import()` in this context. A few of them in no particular order:

 - The `compile()` method operates at the level of bytes. `Response`s represent a bag of bytes, making them more-or-less symmetrical. Imports have another indirection (resolution) and conceptually represent `Module`s, not bytes
 - Having a way to import a WASM module into ES makes some sense and will probably need to be sorted out at the loader level there
 - Having WASM import ES or WASM modules is much less straightforward until/unless WASM defines blocking semantics for those operations

All of that makes me thing that `import()` is either at a different level of abstraction or represents a much larger of (likely compatible) future work...but @domenic is the expert here. Thoughts?

jfbastien:
Thanks I'll take a closer look soon.

I'd like to make sure @kripken also signs off from the POV of usability for Emscripten. We expect it to be our main code generator for the foreseeable future, so it had better like the APIs we offer.

domenic:
A much largely set of (likely compatible) future work sounds about right to me. Figuring out how import() works and what it gives you depends on figuring out how wasm and ES modules interoperate, which I understand is not settled at this time. In the meantime, the Response variation makes a lot of sense.

jfbastien:
@kripken specifically: I'd like to understand whether Emscripten can use this Response API for dynamic linking of wasm modules, where one module imports another's exports.

kripken:
@jfbastien I don't see anything specific to Emscripten or these Response versions here, so probably I'm missing something. But the constraint has to be that if module A imports module B's exports, then B must be fully loaded&compiled before A does so. Given these APIs, it seems we could either

1. Fire off an instantiate for B and a compile for A. When both complete, we can use the exports from B and module from A to make an instance of A using B's exports.
2. Fire off an instantiate for B. When that completes, fire off an instantiate for A using the exports from B.

Are you asking if 1 and/or 2 are possible with Emscripten? Both should be.

jfbastien:
@kripken yup that sounds right. I'd rather check because I was surprised last time with the promise API being more complex to use for Emscripten than I'd envisioned.

I'm not sure an importObect promise would make sense to break this sequencing (or importObject containing some promises which are checked after compilation). We can always add that as an overload later if it makes sense.

annevk:
Final ping for https://github.com/WebAssembly/design/pull/991#issuecomment-280851429.

dschuff:
As @kripken said, I think we could do dynamic linking with async using the same mechanisms we have now. Currently Emscripten also supports emulation of `dlopen()` which of course is synchronous, so there's probably no good way around that.

jfbastien:
> As @kripken said, I think we could do dynamic linking with async using the same mechanisms we have now. Currently Emscripten also supports emulation of `dlopen()` which of course is synchronous, so there's probably no good way around that.

I agree it can be done. I'm asking if this is a good API for it, if Emscripten would use it, and if the API could be better suited to that purpose.

It sounds like a resounding "meh" so far. Given that the Response API is probably useful for non-Emscripten uses as well, I expect that Emscripten just isn't a good client. Maybe it would be one if the `importObject` could somehow be obtained asynchronously (promise or whatever stream combination mechanism). It's not clear that's necessary now, and IIUC it can be added as a simple overload later.

annevk:
@jfbastien `Response` objects are a stream mechanism...

dschuff:
@annevk I think he's referring to a the `importObject` argument to `instantiate`, not the `source` argument. I do think a promise for an `importObject` or some other async way to get that might be useful, but I agree it doesn't have to be now.

flagxor:
@annevk Sorry missed your comment earlier. I've just added verbiage I believe addresses those points, please confirm. I figure strict seems better in this context, as there's a single format.

flagxor:
@jfbastien If I grok your concern correctly, you're pointing out that properly compiling and instantiating a collection of modules with complex dependencies while ensuring minimum latency and full asynchrony will require a very verbose pile of promise laden JS?

I think this is true. But such collections of modules will at this point be the product of something like Emscripten, which will no doubt have the plumbing to do it well. Such modules are also likely to have a complex internal contract between them anyhow. We might be able to make the set up less verbose by accepting promise in a few more places as Derek suggests, but I suspect it will only help some (as those promises won't have simple sources).

I think ultimately an asynchronous dynamic loader is a complex beast and making the API more ergonomic for this won't help as much as ensuring it has all the raw low-level rope.
There aren't likely to be a huge number of these. We'll probably end up with the one for C++ baked into Emscripten (I suppose it already is), maybe others for Rust etc, and then later a polyfill for wasm ES6 modules might.
The languages that need a more ELF like loader might all be able to share, but maybe not, as that also comes with other ABI issues.

I suppose it's a little sad that the recipe for C++ style dynamic loading won't be small, but then again even nacl's "simple" nmf manifests for dynamic loading required a bunch of tooling in practice to get right.

I'm hopeful that if we make wise choices for ES6 modules we'll get something super-pithy there, which seems like what you'd want if you're mixing in tiny wasm modules with JS code.

annevk:
@flagxor thanks, that seems good enough for now.

jfbastien:
@annevk as @dschuff said, I'm talking about APIs such as these:

 ```
 Promise<{module:WebAssembly.Module, instance:WebAssembly.Instance}>
   instantiate(Response source [, importObject])
 
 Promise<{module:WebAssembly.Module, instance:WebAssembly.Instance}>
   instantiate(Promise<Response> source [, importObject])
 ```

The optional `importObect` is very important to dynamic linking, and isn't a stream. Here's the most basic dynamic linking usecase:

* Create an object of JavaScript "syscalls" which WebAssembly code calls to (draw to canvas, touch the DOM, etc) with `syscalls = { log: console.log, draw: (x, y) => { /* something... */ }`
* Create `mem = WebAssembly.Memory(...)`
* Instantiate a module with `i0 = WebAssembly.instantiate(myResponse, { myImports: { memory: mem}, syscalls: syscalls })`
* Instantiate another module which will share the *same* memory, and be able to call functions exported by `i0` with `i1 = WebAssembly.instantiate(myOtherResponse, { myImports: { memory: mem }, importsFromI0: i0, syscalls: syscalls })`.

This serializes the response handling. **Compilation of `i1` cannot start before that of `i0` completes**.

It seems weird to me to add a Response API, yet not handle this case up front, because the entire point is to handle streams and my example shows how it doesn't for what I think is a common usecase. We can add an overload later, it's not an MVP API so the overload may ship at the same time as this Response API.

flagxor:
@jfbastien 
So it's true that a bare instantiate that takes a Response won't be able to handle these hook-ups (without serializing). You'd have to use the explicit compilation method to get maximum parallelism.

```
m0 = WebAssembly.compile(myResponse);
m1 = WebAssembly.compile(myOtherResponse);

(Then later promise chained in fun ways):
i0 = WebAssembly.instantiate(m0, { myImports: { memory: mem}, syscalls: syscalls });
i1 = WebAssembly.instantiate(m1, { myImports: { memory: mem }, importsFromI0: i0, syscalls: syscalls });
```

I suppose if you used the MIME type to aggressively hint the need to compile, you'd do nearly as well.

I don't think any simple change to importObjects will let you just use plain instantiate for complex dynamic linking cases, since you have to express all the ways you might extract an export from an instance.
My suspicion is that you either end up effectively duplicating our compile/instantiate split, or end up doing the same thing, but under a declarative veneer (and maybe missing cases you'd want to support).

I suppose this is somewhat at odds with instantiate(fetch(url)) being the "recommended" pattern.

Maybe I'm still not groking your core concern?

jfbastien:
@flagxor that's correct. Because of this, I'm worried that we're back to advocating for the API that we had prior to #838. We'd really like `WebAssembly.instantiate` to be what people use because then our engines know *at compilation time* the amount of memory that'll be used, and *which* memory will be used.

If developers use something else than `WebAssembly.instantiate` then we have to use heuristics, or do gymnastics to try to peephole common usecases in the wild. It's not impossible, but has pitfalls and is kinda silly to force upon ourselves when we could avoid it.

What I'd like is for folks familiar with Response / streams to help design a solution which addresses this as well (basically, an API which makes stream combination possible for WebAsembly). If they think this can be done on top of the currently proposed Response API then I'm good with it. I like the high-level bits of the Response API.

lukewagner:
This is already a limitation of the existing `WebAssembly.instantiate` APIs.  I think to allow these complex instantiation graphs (Response overload or not), we'd need to add a more powerful API that let you pass a map of modules to instantiate that could reference each other:
```
WebAssembly.instantiateGroup({a:fetch(urlA), b:fetch(urlB), otherImports:importObject})
```
With this, the module at `urlA` could import the `foo` export of the module at `urlB` by importing from `"b" "foo"`.  (Whether we allow cycles is orthogonal; we could limit to DAGs initially.)

(I actually had a TODO in JS.md (that I later removed) to add an API just like this to allow instantiating whole cyclic batches of modules at the same time.)

But, as I said, this is a more complicated API and thus it makes sense to have the simple version here for the simple case.

jfbastien:
What @lukewagner says sounds good to me. As long as we go back to a multi-stream API for `importObject` akin to what he sketched, then I'm OK with this going in. Just file an issue and assign the right person to it.

flagxor:
Agreed, API Luke suggests sounds close.
Opened this, which we should sort out soon:
https://github.com/WebAssembly/design/issues/997

jfbastien:
Great, thanks @flagxor . Can you mark this Response API as post-MVP (so developers don't get mislead, and know that they have to feature-test for it)? After that I'm good to go, commit this, iterate on #997.

flagxor:
Added, PTAL

jfbastien:
Update LGTM, let's merge this and continue in #997.

Issue #993, opened 2017-02-21
What should we throw for implementation specific limits?


flagxor:
We have a number of implementation specific limits, such as number of imported functions, that are relatively synced between browsers.
https://github.com/v8/v8/blob/master/src/wasm/wasm-limits.h
We currently in practice throw WebAssembly.CompileError for these (or return false for validation).
Would it make sense to throw something else such as RangeError, or add WebAssembly.ImplementationLimitError ?

kripken:
If those limits are not temporary, it would be nice to document them somewhere, including rationale? For example,
````
const size_t kV8MaxWasmMemoryPages = 16384; // = 1 GiB
````
seems like a surprising limit, important asm.js apps exist today that need more memory than that.

jfbastien:
I'd rather have these as per-embedder limits, and would document them as such.
This issue is just about acknowledging that such limits can exist, and documenting what behaviors to expect when faced with these limits.

rossberg-chromium:
The underlying question is whether these should be considered validation failures in respective engines. If they are, then `CompileError` seems adequate, but that also affects what the `validate` method returns (false vs exception or other value).

Pro validation failure:
- Devs only need to deal with one kind of compile failure.
- `validate` will never fail.

Cons validation failure:
- De facto, "validation" becomes an implementation-dependent notion.
- Cannot programmatically distinguish between compiler bugs and resource limits (might be a bummer for e.g. error reporting in a JIT compiler).

While the simplicity of the first option is nice, I don't feel entirely comfortable with making the meaning of "valid module" appear implementation-dependent.

MikeHolman:
I agree it would be more elegant for validate to succeed in such situations, but storage types for implementations are probably the same for both compile and validate, and changing this would in practice result in unnecessary overhead and add extra risk to the parser.

For example, we use uint16 for holding argument counts (this is an implementation detail, but we share this with js and on 32 bit platforms it allows to use 16 bits for argument count and 16 bits for flags in a single register). Making argument count infinite would be required to prevent non-normative output, but is obviously impractical.

Also, validate necessarily requires stack/memory allocation, which can already lead to non-normative results.

rossberg-chromium:
@MikeHolman, ah, I didn't assume we would require `validate` to succeed, only that it results in something else than false. An exception would be adequate (but perhaps annoying for client code?).

MikeHolman:
[...]

PR #995, opened 2017-02-22
Change `validate` to return error messages

rossberg-chromium:
Following the discussion in #993, this PR changes `WebAssembly.validate` to return an object
```
{ok: Boolean, error?: String}
```
so that users are given a useful cause of failure.

This is a fairly late change, but the current version arguably is broken -- it's difficult to come up with a use case for explicitly calling `validate` but not caring about a diagnostics. For example, the test harness currently invokes full compilation after a failed `validate` just to get a decent error message, which is silly -- and moreover, may run afoul of a possible cap on synchronous compiles. Other users are likely to run into the same issues.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -65,7 +65,7 @@ The following intrinsic objects are added:
 The `validate` function has the signature:
 
 ```
-Boolean validate(BufferSource bytes)
+validate(BufferSource bytes): {ok: Boolean, error: String}
 ```
 
 If the given `bytes` argument is not a
@@ -73,7 +73,11 @@ If the given `bytes` argument is not a
 then a `TypeError` is thrown.
 
 Otherwise, this function performs *validation* as defined by the [WebAssembly
-specification](https://github.com/WebAssembly/spec/blob/master/interpreter/) and returns `true` if validation succeeded, `false` if validation failed.
+specification](https://github.com/WebAssembly/spec/blob/master/interpreter/).
+It returns a plain JavaScript object with a property `ok`
+whose falue is `true` if validation succeeded, and `false` otherwise.
+Furthermore, iff `ok` is `false`, a property `error` is present indicating the cause of failure with an implementation-dependent error message.
+Both properties are configurable, enumerable and writable.
 
 #### `WebAssembly.compile`
 


lukewagner:
The only *production* use case for `WebAssembly.validate` I know of is to do cheap feature testing:
```
let hasFoo = WebAssembly.validate(Uint8Array.of(0, 0x61, 0x73, 0x6d, 1, 0, 0, 0, ...));
```
where the implementation-defined limits would be a non-issue.

Rather, I'm having a hard-time coming up with a use case where I want to eagerly validate, but not compile, a large module in production.  Development-time/debugging is a separate issue, but for this, one can just use compilation to get the full error message.

lukewagner:
And, given the above, having `validate` fail when implementation limits are hit could be exactly the desired behavior if the limit is the feature being tested.

rossberg-chromium:
@lukewagner, good point, forgot about feature testing. Of course, that use case would cost a mere 3 character `.ok` extra.

Another use case (besides tests) I can see is downloading modules and storing/caching them for later use. For example, in an installation-like procedure where you don't necessarily want to compile already. In such a case you might want to ensure that they are actually valid, and provide some diagnostics if they aren't.

lukewagner:
@rossberg-chromium It also means wasting the time to generate the error string.  If one can assume errors are never on the hot path, an engine might even want to take some effort to make a really helpful string, but validate and failures will be on the hot path, so this could create a tension.

For the AOT download case, though, you'd just `compile` because you actually want the `Module`.

rossberg-chromium:
@lukewagner, somewhat hard to imagine a feature test on such a hot path. I am also not convinced that all download use cases coincide with download-time AOT. The compiled module will be larger. Or you might want to distribute the module further over a local network.

lukewagner:
@rossberg-chromium The feature test would happen on every load and so it's on the load-time critical path.  I still don't understand this use case for downloading maybe-invalid modules.

jfbastien:
IIUC we would now have:

```js
r = WebAssembly.validate(buffer);
console.log(r ? "yes" : "no");
```

This will always print "yes", where it used to be a bool. This is a misleading API since I think folks might expect a bool, and misuse it as above, which wasn't possible before.

RyanLamansky:
A new name would be needed; like `getValidationErrors`, which returns an array of objects that indicate compilation errors and exceeded implementation limits.  An empty array means compilation is guaranteed to succeed.  An engine could speculatively compile during validation if it heuristically determines that it's likely to be requested.

jfbastien:
I don't think we want to collect all possible errors (and specify all associated things). First one seems fine.

rossberg-chromium:
@lukewagner, I don't understand. Loading or linking a module is a very
expensive operation. How is generating a string a relevant cost on such a
path? (FWIW, V8 generates detailed messages anyway, but throws them away
for validate.)

Download use case: consider e.g. a node package manager installing Wasm
modules. Would not necessarily be desirable to compile them all.

I guess I just fail to see what the downside is of providing more
information.

@jfbastien, yes, that's why I'm proposing the change now. The window is
about to close, we can no longer change the method after releasing the API
and such code starts existing in the wild.

lukewagner:
@rossberg-chromium That compilation will happen in parallel and feature tests may block the initial fetch.  Also, I think it's too late to make this change.

rossberg-chromium:
@lukewagner, a fetch is even slower, so computing the string even less significant? Or are you imagining a scenario where an app needs to do so many feature tests that creating a string for each blocks the main thread for too long?

But okay, unless @bnjbvr or @MikeHolman want to argue in favour I gonna close the PR.

bnjbvr:
I agree it is actually pretty late for such a change, and just checking the result as a boolean is the intuitive and obvious usage here. If compilation time is low enough in general, people could use this in the meanwhile.

Another idea: what if there was an optional `onError` function callback passed to the `validate` function that gets called with a descriptive string parameter if there is an error?

I was wondering, the same way there are Promise-based and non-Promise-based functions for compiling and instantiating, could there be a Promise-based function for validation, which rejection value would be an error string if the module is not valid? It doesn't quite make sense to have an async API though, since validation should be pretty fast in general, but there would be a nice symmetry.

So maybe the obvious choice is introducing another API that throws, or returns an object as defined in the first comment. That is cumbersome to have to introduce a new API just for this use case. It just seems wasteful to have to recompile the full module if there's an error somewhere, especially since the Validate function seems to generate it, at least in V8 and SpiderMonkey. Probably this can be deferred to later, according to feedback from the community on the matter.

lukewagner:
@rossberg-chromium Right, I'm just talking about not wasting main-thread time for the many-small-modules-each-doing-many-feature-tests case when all all these cases need is the boolean result.

@bnjbvr The optional `onError` callback sounds like a great option to consider: backwards-compatible, pay-as-you-go, and who knows, maybe we want to consider more structured output than just a string if the user is explicitly asking for a diagnosis...

jfbastien:
@rossberg-chromium 
> @jfbastien, yes, that's why I'm proposing the change now. The window is
> about to close, we can no longer change the method after releasing the API
> and such code starts existing in the wild.

Let me clarify [my comment](https://github.com/WebAssembly/design/pull/995#issuecomment-281725790): I agree validation returning the first error is useful. I don't think it should be called `validate` because the example I gave looks like the "right" usage of the API, but is in fact totally broken. I therefore think we need a different API.

I quite like @bnjbvr's optional `onError` proposal.

rossberg-chromium:
Adding an optional `onError` sounds okay, though a bit clumsy. Closing.

@jfbastien, to nitpick: it'd be natural to assume that a method named `valid` returns a Boolean, but `validate` not so much.

jfbastien:
@bnjbvr can you follow up with `onError`?

bnjbvr:
> @bnjbvr can you follow up with onError?

Yes, will open a new issue to discuss it.

Issue #997, opened 2017-02-24
Support multi-module 1-step Instantiate with dependencies


flagxor:
We should support a version of instantiate that will continue to allow us to recommend it as the preferred instantiation method one dynamic linking / multiple modules become more popular.

`WebAssembly.instantiateGroup({a:Promise<Response>, b:Promise<Response>, otherImports:importObject})`
With this, the module at urlA could import the foo export of the module at urlB by importing from "b" "foo". 

Questions:
* Do we want to support cycles or just DAGs?
* Do we want to support the BufferSource overload?
* How does this interact with ES6 modules? (or does it?)

rossberg-chromium:
Just repeating my usual observation that allowing cycles is much more than a mere API addition. It would introduce full-fletched recursive modules into Wasm, which is a tricky design space with numerous potential implications on present and future versions of the language. (FWIW, recursive modules for typed languages are still an active and rather involved research topic.) We should be very careful not to do something hasty there.

So for a Milestone 2 feature it would be much safer to separate the problem of bulk instantiation from recursive linking and only allow DAGs.

One other nit: `otherImports` should not be part of the object but a separate parameter.

lukewagner:
Agreed on starting with DAGs, just as a practical first step.

On the `otherImports` bit: good point.  I had been thinking that it was simpler to put them all in the same "namespace", but thinking more about this: in the future it is likely we'd want to allow importing `Module`, `ArrayBuffer`, etc and then there would be an ambiguity: are we asking to *compile* these or just use them as import values.  So yeah, definitely better to keep `otherImports` a separate parameter.

So the original reason I thought it was simple to keep them in the same namespace: let's say I'm instantiating some module in the DAG of some call `instantiateGroup(modules, importObj)` and that module imports "a" "b"; where do we look up "a"?  We could merge `modules` and `importObj` into a single namespace, but then what about duplicates: do we require there are none or do we define precedence (e.g., `modules` over `importObj`)?  The latter seems fine, actually, and no more limiting than providing a single namespace object.

mtrofin:
@flagxor - this would be a helper API, correct? 

I worry we would enter composition engine territory with this API. Questions of policy arise. Besides cycles, as @rossberg-chromium illustrates, there's also: how do you match imports and exports; what if there are 2 imports you want satisfied with the same export.

I would prefer we let third parties come up with such policies, and us just provide the basic building blocks, API-wise.

jfbastien:
[...]

PR #999, opened 2017-02-27
Split out compileAndInstantiate from instantiate

lukewagner:
I think there would be a number of benefits from having `instantiate` *always* return a `Promise<Instance>` and having a separate `compileAndInstantiate` that returns the `Promise<{module,instance}>`.  In this PR, `instantiate` is still able to take all the same things it does now, and thus, this PR is strictly more expressive than the current state.  (It's also a pretty simple change to implement.)

The benefits I see of this PR are:
* `instantiate` is less surprising; `compileAndInstantiate` returning a pair is implied by the name.
* It's easier to teach wasm if `instantiate` is just a sink for all manner of byte-y things that always produces an `Instance`.  I was feeling this acutely while writing MDN docs/tutorials recently.
* With `instantiate`, the engine would know up front that the `Module` won't be needed and that the generated code can't be instantiated multiple times, serialized or shared. This can allow for more efficient compilation, more aggressive specialization and less garbage generated by compilation.
* Resolves awkward design question for `instantiateGroup` (#997): given that this function takes a map of inputs and produces a map of outputs: are the output values `Instance`s or `{module,instance}` pairs? Do we have it depend on the dynamic types of values in the input map?  What if some of the input values are `Module`s and some are not?  Do we instead break symmetry and have it always return a `{module,instance}` pair?  After this PR, the obvious, symmetric thing would be to have both `instantiateGroup` and `compileAndInstantiateGroup` which makes the output type clear.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -111,8 +111,7 @@ If neither of the following overloads match, then the returned `Promise` is
 with a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror).
 
 ```
-Promise<{module:WebAssembly.Module, instance:WebAssembly.Instance}>
-  instantiate(BufferSource bytes [, importObject])
+Promise<WebAssembly.Instance> instantiate(BufferSource bytes [, importObject])
 ```
 
 This description applies if the first argument is a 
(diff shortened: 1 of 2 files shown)


lukewagner:
Sadly, it's too late to make such a breaking change, sorry for the noise.  This would've been nice to have, but it's not necessary.

PR #1006, opened 2017-03-04
Version is now 0x1

### LINKED ISSUE #1005: Is the official version number 0x1 now?
In https://bugzilla.mozilla.org/show_bug.cgi?id=1342060, IIUC, SpiderMonkey has stopped accepting 0xd as a version number and now requires version 0x1.

BinaryEncoding.md still says the current version number is 0xd, and that it will be reset to 0x1 in the MVP.

As far as I can tell, that means I can no longer build a module that conforms with BinaryEncoding.md (which allows only (0xd) and test it on SpiderMonkey (which allows only 0x1). I think that's an unfortunate state of affairs.

At the very least, BinaryEncoding.md should be updated to reflect the new reality.

Changing the emitted version number is a minor issue, but if there are now browsers out there which think they understand wasm version 1, which has not yet been released, we've effectively closed the door on significant changes to WebAssembly that would otherwise have gone into version 1. If that's the case, the MVP might as well be released now.

jfbastien:
Resolves #1005.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -183,7 +183,7 @@ The module starts with a preamble of two fields:
 | Field | Type | Description |
 | ----- |  ----- | ----- |
 | magic number | `uint32` |  Magic number `0x6d736100` (i.e., '\0asm') |
-| version | `uint32` | Version number, currently 0xd. The version for MVP will be reset to 1. |
+| version | `uint32` | Version number, `0x1` as of the MVP |
 
 The module preamble is followed by a sequence of sections.
 Each section is identified by a 1-byte *section code* that encodes either a known section or a custom section.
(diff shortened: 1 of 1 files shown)


RyanLamansky:
Is there any published information on when the browser previews will be updated to accept WASM files with the new version?

jfbastien:
Information for each browser would indeed be useful documentation for this transition period. It should be a separate PR, though.

BinaryEncoding.md
```diff
@@ -183,7 +183,7 @@ The module starts with a preamble of two fields:
 | Field | Type | Description |
 | ----- |  ----- | ----- |
 | magic number | `uint32` |  Magic number `0x6d736100` (i.e., '\0asm') |
-| version | `uint32` | Version number, currently 0xd. The version for MVP will be reset to 1. |
+| version | `uint32` | Version number, `0x1` as of the MVP |
```
lukewagner: It seems like "as of the MVP" will become decreasingly relevant and may end up being ambiguous as to whether the version number is expected to change, so I'd remove it.

jfbastien: Done.

pipcet:
I'd really like to see this merged. Are we waiting for resolution of the `"name"` section rename?

PR #1014, opened 2017-03-10
Remove Mutable Gobals from future features

wanderer:


jfbastien:
This removes some features which aren't in MVP. I suggest editing out the MVP global feature set instead.

wanderer:
ok @jfbastien  re-added the global array types

lukewagner:
lgtm

PR #1016, opened 2017-03-14
Require import/export names to be UTF-8.

sunfishcode:
This implements the UTF-8 proposal described in
https://github.com/WebAssembly/design/issues/989#issuecomment-284757788.

This does not currently rename "name" to "utf8-name", because if UTF-8 is
required for import/export names, there's a greater appeal to just saying
that all strings are UTF-8, though this is debatable.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -253,9 +253,9 @@ The import section declares all imports that will be used in the module.
 | Field | Type | Description |
 | ----- | ---- | ----------- |
 | module_len | `varuint32` | module string length |
-| module_str | `bytes` | module string of `module_len` bytes |
+| module_str | `bytes` | module name: `module_len` bytes holding valid utf8 string |
 | field_len | `varuint32` | field name length |
-| field_str | `bytes` | field name string of `field_len` bytes |
+| field_str | `bytes` | field name: `field_len` bytes holding valid utf8 string |
 | kind | `external_kind` | the kind of definition being imported |
 
 Followed by, if the `kind` is `Function`:
@@ -356,7 +356,7 @@ The encoding of the [Export section](Modules.md#exports):
 | Field | Type | Description |
 | ----- | ---- | ----------- |
 | field_len | `varuint32` | field name string length |
-| field_str | `bytes` | field name string of `field_len` bytes |
+| field_str | `bytes` | field name: `field_len` bytes holding valid utf8 string |
 | kind | `external_kind` | the kind of definition being exported |
 | index | `varuint32` | the index into the corresponding [index space](Modules.md) |
 
@@ -471,7 +471,7 @@ where a `naming` is encoded as:
 | ----- | ---- | ----------- |
 | index | `varuint32` | the index which is being named |
 | name_len | `varuint32` | number of bytes in name_str |
-| name_str | `bytes` | binary encoding of the name |
+| name_str | `bytes` | utf8 encoding of the name |
 
 #### Function names
 
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -253,7 +253,7 @@ The import section declares all imports that will be used in the module.
 | Field | Type | Description |
 | ----- | ---- | ----------- |
 | module_len | `varuint32` | length of `module_str` in bytes |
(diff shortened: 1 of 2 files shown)


BinaryEncoding.md
```diff
@@ -253,9 +253,9 @@ The import section declares all imports that will be used in the module.
 | Field | Type | Description |
 | ----- | ---- | ----------- |
 | module_len | `varuint32` | module string length |
-| module_str | `bytes` | module string of `module_len` bytes |
+| module_str | `bytes` | module name: `module_len` bytes holding valid utf8 string |
```
jfbastien: "UTF-8" here and elsewhere.

Modules.md
```diff
@@ -48,7 +48,8 @@ In the future, other kinds of imports may be added. Imports are designed to
 allow modules to share code and data while still allowing separate compilation
 and caching.
 
-All imports include two opaque names: a *module name* and an *export name*. The
+All imports include two opaque names: a *module name* and an *export name*,
```
jfbastien: In JS.md, which type of exception should occur if import or export are invalid UTF-8 strings?

BinaryEncoding.md
sunfishcode: done

Modules.md
RyanLamansky: JS.md already seems to specify `WebAssembly.CompileError` in this case.

sunfishcode: It'd a validation requirement, so WebAssembly.validate would return false, and APIs that throw would throw WebAssembly.CompileError. The design docs don't specify the details of validation, so there doesn't seem to be a clear place to specify this; perhaps we could just handle this in eventual spec PR?

jfbastien: sgtm

BinaryEncoding.md
```diff
@@ -253,9 +253,9 @@ The import section declares all imports that will be used in the module.
 | Field | Type | Description |
 | ----- | ---- | ----------- |
 | module_len | `varuint32` | module string length |
-| module_str | `bytes` | module string of `module_len` bytes |
+| module_str | `bytes` | module name: `module_len` bytes holding valid UTF-8 string |
```
annevk: This would be a valid UTF-8 byte sequence. A string is what you get after you decode.

sunfishcode: Thanks for the correction! This is now fixed.

BinaryEncoding.md
```diff
@@ -252,10 +252,10 @@ The import section declares all imports that will be used in the module.
 
 | Field | Type | Description |
 | ----- | ---- | ----------- |
-| module_len | `varuint32` | module string length |
-| module_str | `bytes` | module name: `module_len` bytes holding valid UTF-8 string |
-| field_len | `varuint32` | field name length |
-| field_str | `bytes` | field name: `field_len` bytes holding valid UTF-8 string |
+| module_len | `varuint32` | length of `module_str` in bytes |
+| module_str | `bytes` | module name: valid UTF-8 byte sequnce |
```
jfbastien: typo

sunfishcode:
@rossberg-chromium Yes, allowing embedders to impose additional constraints in this space would be fine with me.

Are there any other comments on this PR?

wanderer:
less validation rules on the base spec === smaller code base. I think it should be left to the upper layers to decided on the string format

sunfishcode:
@wanderer The amount of code needed is quite small. Some implementations will already have a Unicode library linked in for other purposes, and for those that don't, here's a simple standalone implementation in C, for example:

  https://gist.github.com/sunfishcode/c050d4f60633c49ae6e54a3d45385031

In my [experiment](https://github.com/WebAssembly/design/issues/989#issuecomment-282082105) adding this to a production wasm decoder, the performance impact was negligible.

An implementation which only accepted ASCII strings, as [mentioned above](https://github.com/WebAssembly/design/pull/1016#issuecomment-286663852), could be even simpler -- just check that no byte has the MSB set.

RyanLamansky:
Anything left to discuss before this can be merged?

sunfishcode:
I believe all the concerns raised have been answered.

PR #1018, opened 2017-03-21
WebAssembly.instantiate could be specified more explicitly

kmiller68:
The current wording for WebAssembly.instantiate is a little confusing. It's not clear exactly what it means to run a task that does JS asynchronously, as JS is not specified to be asynchronous. This change clarifies that instantiate should queue a task to do the JS work.

Additionally, an implementor may wish to do more asynchronous work after they have looked at a module's imports. This change also explicitly allows implementors to do more asynchronous work post instantiation.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -120,8 +120,9 @@ This description applies if the first argument is a
 
 This function starts an asynchronous task that first compiles a `WebAssembly.Module`
 from `bytes` as described in the [`WebAssembly.Module` constructor](#webassemblymodule-constructor)
-and then instantiate the resulting `Module` with `importObject` as described in the
-[`WebAssembly.Instance` constructor](#webassemblyinstance-constructor).
+and then queue a task to instantiate the resulting `Module` with `importObject` as described in the
+[`WebAssembly.Instance` constructor](#webassemblyinstance-constructor). After the instantiation task runs
+and before any subsequent steps are taken, other unspecified asynchronuous tasks may be run.
 On success, the `Promise` is [fulfilled](http://tc39.github.io/ecma262/#sec-fulfillpromise)
 with a plain JavaScript object pair `{module, instance}` containing the resulting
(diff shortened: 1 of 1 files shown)


JS.md
```diff
@@ -120,8 +120,9 @@ This description applies if the first argument is a
 
 This function starts an asynchronous task that first compiles a `WebAssembly.Module`
 from `bytes` as described in the [`WebAssembly.Module` constructor](#webassemblymodule-constructor)
-and then instantiate the resulting `Module` with `importObject` as described in the
-[`WebAssembly.Instance` constructor](#webassemblyinstance-constructor).
+and then queue a task to instantiate the resulting `Module` with `importObject` as described in the
+[`WebAssembly.Instance` constructor](#webassemblyinstance-constructor). After the instantiation task runs
+and before any subsequent steps are taken, other unspecified asynchronuous tasks may be run.
```
jfbastien: "asynchronous"

JS.md
```diff
@@ -140,9 +141,10 @@ Promise<WebAssembly.Instance> instantiate(moduleObject [, importObject])
 
 This description applies if the first argument is a `WebAssembly.Module` instance.
 
-This function starts an asynchronous task that instantiates a `WebAssembly.Instance`
+This function asynchronously queues a task that instantiates a `WebAssembly.Instance`
 from `moduleObject` and `importObject` as described in the
-[`WebAssembly.Instance` constructor](#webassemblyinstance-constructor).
+[`WebAssembly.Instance` constructor](#webassemblyinstance-constructor). After the instantiation task runs
+and before any subsequent steps are taken, other unspecified asynchronuous tasks may be run.
```
jfbastien: "asynchronous"

lukewagner:
lgtm

littledan:
The wording used here doesn't really make it explicit that you can do the instantiation work in parallel. It's just done in a task, where the exact timing can't be observed. Is this preferred way to specify it, or should there be some specific "in parallel" wording? cc @domenic

Issue #950, opened 2017-01-13
allow Memory.buffer to be a SharedArrayBuffer


bpowers:
Some background: I'm working on [Browsix](https://browsix.org), a tool for running multiple C, C++, Go and node.js programs as processes in the browser.  We do this for C programs by compiling with a (fork) of Emscripten and running the compiled "binary" in a Web Worker, where our fork mainly replaces `src/library_syscalls.js` with an implementation that calls into a kernel (library) that runs on the main browser thread.

Our preferred method for system calls works like this:  When the process/Web Worker starts up, the backing buffer for the heap is initialized as a SharedArrayBuffer, and a reference is sent to the kernel (over a message sent with `postMessage`), along with an offset into that SharedArrayBuffer reserved as an address for `Atomics.wait`.  When the program issues future system calls, the runtime sends a message (with `postMessage`) to the kernel with the system call number and arguments (integers or "pointer" offsets into the heap), and `Atomics.wait()` on the pre-arranged request.  The kernel processes the system call, places results in the SharedArrayBuffer, and `Atomics.wake()`s the process.

This works great with `asm.js` code emitted from Emscripten (and is much faster than our first approach, which relied on the Emterpreter to save and restore the stack), but I don't think is possible with the current WebAssembly spec (I would love to be wrong!).

My proposal is to add an additional optional argument to the options object that `WebAssembly.Memory` takes, a boolean "`isShared`".  When this argument is present and `true`, the buffer backing the Memory is created as a SharedArrayBuffer.

I've implemented this in v8 here: https://github.com/bpowers/v8/commit/bda41b7e9467fc8a0df5d38db83bb413bb6fb9ac and so far in my limited testing seems to do the trick -- I am able to run small C programs compiled with a modified emscripten/binaryen as Browsix processes in Web Workers.

Thoughts/suggestions?

jfbastien:
This would fall under [future features: threads](https://github.com/WebAssembly/design/blob/master/FutureFeatures.md#threads). We definitely want to do this, and it's great if you want to explore what it looks like.

A few thoughts:

* Can you create `isShared` from an exported `Memory` (i.e. not created with `WebAssembly.Memory`, but created and then exported through a memory section).
* What are the semantics when importing such a shared memory into a WebAssembly instance?
* What semantics do you propose for `Memory.grow` / `grow_memory`?
* How does the integration with workers work? We need to clarify requirements on running wasm from different workers.
* How does it interact with `SharedArrayBuffer` from JS?

[...]

PR #1019, opened 2017-03-24
Atomic and thread operators

binji:
This is an initial pass at adding some atomic and thread operators, based on some work that @flagxor did earlier.

There's still a lot that needs to be fleshed out, but I wanted to get something going.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -161,6 +161,8 @@ A packed tuple that describes the limits of a
 | initial | `varuint32` | initial length (in units of table elements or wasm pages) |
 | maximum | `varuint32`? | only present if specified by `flags` |
 
+Note: In the [future :unicorn:][future threads], the "flags" field may be changed to `varuint32`, e.g., to include a flag for sharing between threads.
+
 ### `init_expr`
 The encoding of an [initializer expression](Modules.md#initializer-expression)
 is the normal encoding of the expression followed by the `end` opcode as a
@@ -181,7 +183,7 @@ The module starts with a preamble of two fields:
 | Field | Type | Description |
 | ----- |  ----- | ----- |
 | magic number | `uint32` |  Magic number `0x6d736100` (i.e., '\0asm') |
-| version | `uint32` | Version number, currently 0xd. The version for MVP will be reset to 1. |
+| version | `uint32` | Version number, `0x1` |
 
 The module preamble is followed by a sequence of sections.
 Each section is identified by a 1-byte *section code* that encodes either a known section or a custom section.
@@ -193,8 +195,8 @@ part of the payload.
 | ----- |  ----- | ----- |
 | id | `varuint7` | section code |
 | payload_len  | `varuint32` | size of this section in bytes |
-| name_len | `varuint32` ? | length of the section name in bytes, present if `id == 0` |
-| name | `bytes` ? | section name string, present if `id == 0` |
+| name_len | `varuint32` ? | length of `name` in bytes, present if `id == 0` |
+| name | `bytes` ? | section name: valid UTF-8 byte sequence, present if `id == 0` |
 | payload_data  | `bytes` | content of this section, of length `payload_len - sizeof(name) - sizeof(name_len)` |
 
 Each known section is optional and may appear at most once. Custom sections all have the same `id` (0), and can be named non-uniquely (all bytes composing their names may be identical).
@@ -250,10 +252,10 @@ The import section declares all imports that will be used in the module.
 
 | Field | Type | Description |
 | ----- | ---- | ----------- |
-| module_len | `varuint32` | module string length |
-| module_str | `bytes` | module string of `module_len` bytes |
-| field_len | `varuint32` | field name length |
-| field_str | `bytes` | field name string of `field_len` bytes |
+| module_len | `varuint32` | length of `module_str` in bytes |
+| module_str | `bytes` | module name: valid UTF-8 byte sequence |
+| field_len | `varuint32` | length of `field_str` in bytes |
+| field_str | `bytes` | field name: valid UTF-8 byte sequence |
 | kind | `external_kind` | the kind of definition being imported |
 
 Followed by, if the `kind` is `Function`:
@@ -353,8 +355,8 @@ The encoding of the [Export section](Modules.md#exports):
 
 | Field | Type | Description |
 | ----- | ---- | ----------- |
-| field_len | `varuint32` | field name string length |
-| field_str | `bytes` | field name string of `field_len` bytes |
+| field_len | `varuint32` | length of `field_str` in bytes |
+| field_str | `bytes` | field name: valid UTF-8 byte sequence |
 | kind | `external_kind` | the kind of definition being exported |
 | index | `varuint32` | the index into the corresponding [index space](Modules.md) |
 
@@ -468,8 +470,8 @@ where a `naming` is encoded as:
 | Field | Type | Description |
 | ----- | ---- | ----------- |
 | index | `varuint32` | the index which is being named |
-| name_len | `varuint32` | number of bytes in name_str |
-| name_str | `bytes` | binary encoding of the name |
+| name_len | `varuint32` | length of `name_str` in bytes |
+| name_str | `bytes` | UTF-8 encoding of the name |
 
 #### Function names
 
@@ -626,7 +628,8 @@ As implied by the `log2(alignment)` encoding, the alignment must be a power of 2
 As an additional validation criteria, the alignment must be less or equal to 
 natural alignment. The bits after the
 `log(memory-access-size)` least-significant bits must be set to 0. These bits
-are reserved for use (e.g., for shared memory ordering requirements).
+are reserved for [future :unicorn:][future threads] use
+(e.g., for shared memory ordering requirements).
 
 The `reserved` immediate to the `current_memory` and `grow_memory` operators is
 for [future :unicorn:][future multiple tables] use and must be 0 in the MVP.
@@ -783,167 +786,9 @@ for [future :unicorn:][future multiple tables] use and must be 0 in the MVP.
 | `f32.reinterpret/i32` | `0xbe` | | |
 | `f64.reinterpret/i64` | `0xbf` | | |
 
-## Thread operators ([described here](Semantics.md#thread-operators))
-
-| Name | Opcode | Immediate | Description |
-| ---- | ---- | ---- | ---- |
-| `is_lock_free` | `0xff00` | | |
-| `wake` | `0xff01` | `memory immediate` | |
-| `i32.wait` | `0xff02` | `memory immediate` | |
-| `i64.wait` | `0xff03` | `memory immediate` | |
-
-The `memory_immediate` type is encoded as follows:
-
-| Name | Type | Description |
-| ---- | ---- | ---- |
-| flags | `varuint32` | a bitfield which currently must be 0 |
-| offset | `varuint32` | the value of the offset |
-
-## Atomic read-modify-write operators ([described here](Semantics.md#atomic-memory-accesses))
-
-| Name | Opcode | Immediates | Description |
-| ---- | ---- | ---- | ---- |
-| `i32.atomic.rmw.add` | `0xff10` | `0` `memory immediate` | atomic add to memory |
-| `i64.atomic.rmw.add` | `0xff11` | `0` `memory immediate` | atomic add to memory |
-| `i32.atomic.rmw8_s.add` | `0xff12` | `0` `memory immediate` | atomic add to memory |
-| `i32.atomic.rmw8_u.add` | `0xff13` | `0` `memory immediate` | atomic add to memory |
-| `i32.atomic.rmw16_s.add` | `0xff14` | `0` `memory immediate` | atomic add to memory |
-| `i32.atomic.rmw16_u.add` | `0xff15` | `0` `memory immediate` | atomic add to memory |
-| `i64.atomic.rmw8_s.add` | `0xff16` | `0` `memory immediate` | atomic add to memory |
-| `i64.atomic.rmw8_u.add` | `0xff17` | `0` `memory immediate` | atomic add to memory |
-| `i64.atomic.rmw16_s.add` | `0xff18` | `0` `memory immediate` | atomic add to memory |
-| `i64.atomic.rmw16_u.add` | `0xff19` | `0` `memory immediate` | atomic add to memory |
-| `i64.atomic.rmw32_s.add` | `0xff1a` | `0` `memory immediate` | atomic add to memory |
-| `i64.atomic.rmw32_u.add` | `0xff1b` | `0` `memory immediate` | atomic add to memory |
-| `i32.atomic.rmw.sub` | `0xff10` | `1` `memory immediate` | atomic sub from memory |
-| `i64.atomic.rmw.sub` | `0xff11` | `1` `memory immediate` | atomic sub from memory |
-| `i32.atomic.rmw8_s.sub` | `0xff12` | `1` `memory immediate` | atomic sub from memory |
-| `i32.atomic.rmw8_u.sub` | `0xff13` | `1` `memory immediate` | atomic sub from memory |
-| `i32.atomic.rmw16_s.sub` | `0xff14` | `1` `memory immediate` | atomic sub from memory |
-| `i32.atomic.rmw16_u.sub` | `0xff15` | `1` `memory immediate` | atomic sub from memory |
-| `i64.atomic.rmw8_s.sub` | `0xff16` | `1` `memory immediate` | atomic sub from memory |
-| `i64.atomic.rmw8_u.sub` | `0xff17` | `1` `memory immediate` | atomic sub from memory |
-| `i64.atomic.rmw16_s.sub` | `0xff18` | `1` `memory immediate` | atomic sub from memory |
-| `i64.atomic.rmw16_u.sub` | `0xff19` | `1` `memory immediate` | atomic sub from memory |
-| `i64.atomic.rmw32_s.sub` | `0xff1a` | `1` `memory immediate` | atomic sub from memory |
-| `i64.atomic.rmw32_u.sub` | `0xff1b` | `1` `memory immediate` | atomic sub from memory |
-| `i32.atomic.rmw.and` | `0xff10` | `2` `memory immediate` | atomic and with memory |
-| `i64.atomic.rmw.and` | `0xff11` | `2` `memory immediate` | atomic and with memory |
-| `i32.atomic.rmw8_s.and` | `0xff12` | `2` `memory immediate` | atomic and with memory |
-| `i32.atomic.rmw8_u.and` | `0xff13` | `2` `memory immediate` | atomic and with memory |
-| `i32.atomic.rmw16_s.and` | `0xff14` | `2` `memory immediate` | atomic and with memory |
-| `i32.atomic.rmw16_u.and` | `0xff15` | `2` `memory immediate` | atomic and with memory |
-| `i64.atomic.rmw8_s.and` | `0xff16` | `2` `memory immediate` | atomic and with memory |
-| `i64.atomic.rmw8_u.and` | `0xff17` | `2` `memory immediate` | atomic and with memory |
-| `i64.atomic.rmw16_s.and` | `0xff18` | `2` `memory immediate` | atomic and with memory |
-| `i64.atomic.rmw16_u.and` | `0xff19` | `2` `memory immediate` | atomic and with memory |
-| `i64.atomic.rmw32_s.and` | `0xff1a` | `2` `memory immediate` | atomic and with memory |
-| `i64.atomic.rmw32_u.and` | `0xff1b` | `2` `memory immediate` | atomic and with memory |
-| `i32.atomic.rmw.or` | `0xff10` | `3` `memory immediate` | atomic or with memory |
-| `i64.atomic.rmw.or` | `0xff11` | `3` `memory immediate` | atomic or with memory |
-| `i32.atomic.rmw8_s.or` | `0xff12` | `3` `memory immediate` | atomic or with memory |
-| `i32.atomic.rmw8_u.or` | `0xff13` | `3` `memory immediate` | atomic or with memory |
-| `i32.atomic.rmw16_s.or` | `0xff14` | `3` `memory immediate` | atomic or with memory |
-| `i32.atomic.rmw16_u.or` | `0xff15` | `3` `memory immediate` | atomic or with memory |
-| `i64.atomic.rmw8_s.or` | `0xff16` | `3` `memory immediate` | atomic or with memory |
-| `i64.atomic.rmw8_u.or` | `0xff17` | `3` `memory immediate` | atomic or with memory |
-| `i64.atomic.rmw16_s.or` | `0xff18` | `3` `memory immediate` | atomic or with memory |
-| `i64.atomic.rmw16_u.or` | `0xff19` | `3` `memory immediate` | atomic or with memory |
-| `i64.atomic.rmw32_s.or` | `0xff1a` | `3` `memory immediate` | atomic or with memory |
-| `i64.atomic.rmw32_u.or` | `0xff1b` | `3` `memory immediate` | atomic or with memory |
-| `i32.atomic.rmw.xor` | `0xff10` | `4` `memory immediate` | atomic xor with memory |
-| `i64.atomic.rmw.xor` | `0xff11` | `4` `memory immediate` | atomic xor with memory |
-| `i32.atomic.rmw8_s.xor` | `0xff12` | `4` `memory immediate` | atomic xor with memory |
-| `i32.atomic.rmw8_u.xor` | `0xff13` | `4` `memory immediate` | atomic xor with memory |
-| `i32.atomic.rmw16_s.xor` | `0xff14` | `4` `memory immediate` | atomic xor with memory |
-| `i32.atomic.rmw16_u.xor` | `0xff15` | `4` `memory immediate` | atomic xor with memory |
-| `i64.atomic.rmw8_s.xor` | `0xff16` | `4` `memory immediate` | atomic xor with memory |
-| `i64.atomic.rmw8_u.xor` | `0xff17` | `4` `memory immediate` | atomic xor with memory |
-| `i64.atomic.rmw16_s.xor` | `0xff18` | `4` `memory immediate` | atomic xor with memory |
-| `i64.atomic.rmw16_u.xor` | `0xff19` | `4` `memory immediate` | atomic xor with memory |
-| `i64.atomic.rmw32_s.xor` | `0xff1a` | `4` `memory immediate` | atomic xor with memory |
-| `i64.atomic.rmw32_u.xor` | `0xff1b` | `4` `memory immediate` | atomic xor with memory |
-| `i32.atomic.rmw.xchg` | `0xff10` | `5` `memory immediate` | atomic exchange with memory |
-| `i64.atomic.rmw.xchg` | `0xff11` | `5` `memory immediate` | atomic exchange with memory |
-| `i32.atomic.rmw8_s.xchg` | `0xff12` | `5` `memory immediate` | atomic exchange with memory |
-| `i32.atomic.rmw8_u.xchg` | `0xff13` | `5` `memory immediate` | atomic exchange with memory |
-| `i32.atomic.rmw16_s.xchg` | `0xff14` | `5` `memory immediate` | atomic exchange with memory |
-| `i32.atomic.rmw16_u.xchg` | `0xff15` | `5` `memory immediate` | atomic exchange with memory |
-| `i64.atomic.rmw8_s.xchg` | `0xff16` | `5` `memory immediate` | atomic exchange with memory |
-| `i64.atomic.rmw8_u.xchg` | `0xff17` | `5` `memory immediate` | atomic exchange with memory |
-| `i64.atomic.rmw16_s.xchg` | `0xff18` | `5` `memory immediate` | atomic exchange with memory |
-| `i64.atomic.rmw16_u.xchg` | `0xff19` | `5` `memory immediate` | atomic exchange with memory |
-| `i64.atomic.rmw32_s.xchg` | `0xff1a` | `5` `memory immediate` | atomic exchange with memory |
-| `i64.atomic.rmw32_u.xchg` | `0xff1b` | `5` `memory immediate` | atomic exchange with memory |
-
-The `memory_immediate` type is encoded as follows:
-
-| Name | Type | Description |
-| ---- | ---- | ---- |
-| flags | `varuint32` | a bitfield which currently must be 0 |
-| offset | `varuint32` | the value of the offset |
-
-## Atomic compare exchange operators ([described here](Semantics.md#atomic-memory-accesses))
-
-| Name | Opcode | Immediate | Description |
-| ---- | ---- | ---- | ---- |
-| `i32.atomic.rmw.cmpxchg` | `0xff1c` | `memory immediate` | atomic compare exchange with memory |
-| `i64.atomic.rmw.cmpxchg` | `0xff1d` | `memory immediate` | atomic compare exchange with memory |
-| `i32.atomic.rmw8_s.cmpxchg` | `0xff1e` | `memory immediate` | atomic compare exchange with memory |
-| `i32.atomic.rmw8_u.cmpxchg` | `0xff1f` | `memory immediate` | atomic compare exchange with memory |
-| `i32.atomic.rmw16_s.cmpxchg` | `0xff20` | `memory immediate` | atomic compare exchange with memory |
-| `i32.atomic.rmw16_u.cmpxchg` | `0xff21` | `memory immediate` | atomic compare exchange with memory |
-| `i64.atomic.rmw8_s.cmpxchg` | `0xff22` | `memory immediate` | atomic compare exchange with memory |
-| `i64.atomic.rmw8_u.cmpxchg` | `0xff23` | `memory immediate` | atomic compare exchange with memory |
-| `i64.atomic.rmw16_s.cmpxchg` | `0xff24` | `memory immediate` | atomic compare exchange with memory |
-| `i64.atomic.rmw16_u.cmpxchg` | `0xff25` | `memory immediate` | atomic compare exchange with memory |
-| `i64.atomic.rmw32_s.cmpxchg` | `0xff26` | `memory immediate` | atomic compare exchange with memory |
-| `i64.atomic.rmw32_u.cmpxchg` | `0xff27` | `memory immediate` | atomic compare exchange with memory |
-
-The `memory_immediate` type is encoded as follows:
-
-| Name | Type | Description |
-| ---- | ---- | ---- |
-| flags | `varuint32` | a bitfield which currently must be 0 |
-| offset | `varuint32` | the value of the offset |
-
-## Atomic load/store operators ([described here](Semantics.md#atomic-memory-accesses))
-
-| Name | Opcode | Immediate | Description |
-| ---- | ---- | ---- | ---- |
-| `i32.atomic.load` | `0xff28` | `memory immediate` | atomic load from memory |
-| `i64.atomic.load` | `0xff29` | `memory immediate` | atomic load from memory |
-| `f32.atomic.load` | `0xff2a` | `memory immediate` | atomic load from memory |
-| `f64.atomic.load` | `0xff2b` | `memory immediate` | atomic load from memory |
-| `i32.atomic.load8_s` | `0xff2c` | `memory immediate` | atomic load from memory |
-| `i32.atomic.load8_u` | `0xff2d` | `memory immediate` | atomic load from memory |
-| `i32.atomic.load16_s` | `0xff2e` | `memory immediate` | atomic load from memory |
-| `i32.atomic.load16_u` | `0xff2f` | `memory immediate` | atomic load from memory |
-| `i64.atomic.load8_s` | `0xff30` | `memory immediate` | atomic load from memory |
-| `i64.atomic.load8_u` | `0xff31` | `memory immediate` | atomic load from memory |
-| `i64.atomic.load16_s` | `0xff32` | `memory immediate` | atomic load from memory |
-| `i64.atomic.load16_u` | `0xff33` | `memory immediate` | atomic load from memory |
-| `i64.atomic.load32_s` | `0xff34` | `memory immediate` | atomic load from memory |
-| `i64.atomic.load32_u` | `0xff35` | `memory immediate` | atomic load from memory |
-| `i32.atomic.store` | `0xff36` | `memory immediate` | atomic store to memory |
-| `i64.atomic.store` | `0xff37` | `memory immediate` | atomic store to memory |
-| `f32.atomic.store` | `0xff38` | `memory immediate` | atomic store to memory |
-| `f64.atomic.store` | `0xff39` | `memory immediate` | atomic store to memory |
-| `i32.atomic.store8` | `0xff3a` | `memory immediate` | atomic store to memory |
-| `i32.atomic.store16` | `0xff3b` | `memory immediate` | atomic store to memory |
-| `i64.atomic.store8` | `0xff3c` | `memory immediate` | atomic store to memory |
-| `i64.atomic.store16` | `0xff3d` | `memory immediate` | atomic store to memory |
-| `i64.atomic.store32` | `0xff3e` | `memory immediate` | atomic store to memory |
-
-The `memory_immediate` type is encoded as follows:
-
-| Name | Type | Description |
-| ---- | ---- | ---- |
-| flags | `varuint32` | a bitfield which currently must be 0 |
-| offset | `varuint32` | the value of the offset |
-
 [future general]: FutureFeatures.md
 [future multiple return]: FutureFeatures.md#multiple-return
+[future threads]: FutureFeatures.md#threads
 [future types]: FutureFeatures.md#more-table-operators-and-types
 [future multiple tables]: FutureFeatures.md#multiple-tables-and-memories
 [future compression]: https://github.com/WebAssembly/decompressor-prototype/blob/master/CompressionLayer1.md
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -869,6 +869,9 @@ for [future :unicorn:][future multiple tables] use and must be 0 in the MVP.
 | `i64.atomic_xchg16_u` | `0xe619` | `5` `memory immediate` | atomic exchange with memory |
 | `i64.atomic_xchg32_s` | `0xe61a` | `5` `memory immediate` | atomic exchange with memory |
 | `i64.atomic_xchg32_u` | `0xe61b` | `5` `memory immediate` | atomic exchange with memory |
+
+## Atomic compare exchange operators ([described here](Semantics.md#atomic-memory-accesses))
+
 | `i32.atomic_cmpxchg` | `0xe61c` | `memory immediate` | atomic compare exchange with memory |
 | `i64.atomic_cmpxchg` | `0xe61d` | `memory immediate` | atomic compare exchange with memory |
 | `i32.atomic_cmpxchg8_s` | `0xe61e` | `memory immediate` | atomic compare exchange with memory |
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -64,9 +64,12 @@ opcodes.
 
 ## Language Types
 
-All types are distinguished by a negative `varint7` values that is the first byte of their encoding (representing a type constructor):
+All types are distinguished by a negative `varint7` values that is the first
+byte of their encoding (representing a type constructor).
+Some types have an additional operand indicating table number, encoded as a
+varuint32:
 
-| Opcode | Type constructor |
+| Encoding | Type constructor |
 |--------|------------------|
 | `-0x01` (i.e., the byte `0x7f`) | `i32` |
 | `-0x02` (i.e., the byte `0x7e`) | `i64` |
@@ -74,6 +77,7 @@ All types are distinguished by a negative `varint7` values that is the first byt
 | `-0x04` (i.e., the byte `0x7c`) | `f64` |
 | `-0x10` (i.e., the byte `0x70`) | `anyfunc` |
 | `-0x20` (i.e., the byte `0x60`) | `func` |
+| `-0x30` (i.e., the byte `0x50`) | `table` followed by varuint32 indicating a table number (currently only can be zero). This type is only allowed in import function signatures for parameter and return types. |
 | `-0x40` (i.e., the byte `0x40`) | pseudo type for representing an empty `block_type` |
 
 Some of these will be followed by additional fields, see below.
@@ -89,6 +93,20 @@ A `varint7` indicating a [value type](Semantics.md#types). One of:
 * `f64`
 
 as encoded above.
+
+### `table` types
+`table` types are indexed references into a statically specified type table
+(currently only table 0).
+Functions including these types in their signatures expect an i32 for inputs
+(for the table index). They expect an i32 operand, for the index of the return
+value destination.
+In general the behavior of functions including these table types as parameters
+is left to the embedder. For a JavaScript embedding, these will typically by
+JS Objects.
+Calls to imported JS Functions will fail when called outside the original
+Wasm thread, unless the function has been specifically designed to be called
+off the main thread. In a context without exceptions, such failures will
+exit the thread, or throw otherwise.
 
 ### `block_type`
 A `varint7` indicating a block signature. These types are encoded as:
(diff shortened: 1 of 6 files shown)


Semantics.md
```diff
@@ -42,7 +42,7 @@ environment such as a browser, a trap results in throwing a JavaScript exception
 If developer tools are active, attaching a debugger before the
 termination would be sensible.
 
-## Stack Overflow
+## Stack overflow
```
jfbastien: ?

Semantics.md
```diff
@@ -81,7 +81,7 @@ Note that the value types `i32` and `i64` are not inherently signed or
 unsigned. The interpretation of these types is determined by individual
 operators.
 
-## Linear Memory
+## Linear memory
```
jfbastien: ?

Semantics.md
```diff
@@ -112,10 +112,15 @@ or [defined inside the module](Modules.md#linear-memory-section). After import
...
+### Shared linear memory
 
-### Linear Memory Accesses
+A Linear memory can be marked as shared, which allows it to be shared between
+threads of execution. The shared memory can be imported or defined in the
+module. It is a validation error to attempt to import shared linear
+memory if the module's memory import doesn't specify that it allows shared
+memory.
+
+### Linear memory accesses
```
jfbastien: Capitalization.

Semantics.md
```diff
@@ -157,6 +162,146 @@ Store operators do not produce a value.
...
+| `i32.atomic_xchg` | 4 bytes as i32 | nop | 4 bytes |
+| `i64.atomic_xchg8_s` | 1 byte, sign-extended i8 to i64 | nop | 1 byte, wrapped from i64 to i8 |
+| `i64.atomic_xchg8_u` | 1 byte, zero-extended i8 to i64 | nop | 1 byte, wrapped from i64 to i8 |
+| `i64.atomic_xchg16_s` | 2 bytes, sign-extended i16 to i64 | nop | 2 bytes, wrapped from i64 to i16 |
+| `i64.atomic_xchg16_u` | 2 bytes, zero-extended i16 to i64 | nop | 2 bytes, wrapped from i64 to i16 |
+| `i64.atomic_xchg32_s` | 4 bytes, sign-extended i32 to i64 | nop | 4 bytes, wrapped from i64 to i32 |
+| `i64.atomic_xchg32_u` | 4 bytes, zero-extended i32 to i64 | nop | 4 bytes, wrapped from i64 to i32 |
+| `i64.atomic_xchg` | 8 bytes | nop | 8 bytes |
+
+The atomic compare exchange RMW operators take three operands: a linear memory
```
jfbastien: Kinda weird to classify cmpxchg as RMW. It's a 3rd category IMO.

Semantics.md
```diff
@@ -211,7 +356,7 @@ Thus, it is recommend that WebAssembly producers align frequently-used data to
 permit the use of natural alignment access, and use loads and stores with the
 greatest alignment values practical, while always avoiding misaligned accesses.
 
-### Out of Bounds
+### Out of bounds
```
jfbastien: ?

Semantics.md
```diff
@@ -461,7 +606,7 @@ supported (including NaN values of all possible bit patterns).
   * `f32.const`: produce the value of an f32 immediate
   * `f64.const`: produce the value of an f64 immediate
 
-## 32-bit Integer operators
+## 32-bit integer operators
```
jfbastien: ?

Semantics.md
```diff
@@ -678,6 +823,48 @@ outside the range which rounds to an integer in range) traps.
     This trap is intended to be impossible for user code to catch or handle, even in the future when it may be possible to
     handle some other kinds of traps or exceptions.
 
+## Thread operators
+
+  * `is_lock_free`: return 1 if an atomic memory access of a given size can be
+    completed without the execution engine acquiring a lock. If not, return `0`.
```
jfbastien: Can you import the wording from https://tc39.github.io/ecma262/ ?

Semantics.md
```diff
@@ -678,6 +823,48 @@ outside the range which rounds to an integer in range) traps.
     This trap is intended to be impossible for user code to catch or handle, even in the future when it may be possible to
     handle some other kinds of traps or exceptions.
 
+## Thread operators
+
+  * `is_lock_free`: return 1 if an atomic memory access of a given size can be
```
jfbastien: Code-quote `1`.

Semantics.md
```diff
@@ -678,6 +823,48 @@ outside the range which rounds to an integer in range) traps.
...
+    completed without the execution engine acquiring a lock. If not, return `0`.
+
+### Wait
+
+The wake and wait operators are optimizations over busy-waiting for a value to
+change. It is a validation error to use these operators on non-shared linear
+memory.
+
+The wait operators take three operands: a linear memory index, an
+expected value, and a relative timeout in milliseconds. The wait operation
```
jfbastien: Can it be negative?

Semantics.md
binji: Capitalization was inconsistent in the file, but seemed to be "Upper lower lower." Probably should be a different change though.

Semantics.md
binji: Described below, but I moved it up.

Semantics.md
binji: Done.

Semantics.md
binji: Done.

Semantics.md
binji: Done.

binji:
> We need a way to identify post-MVP features. :atom_symbol: is an easy on here! We then need a page that identifies all these symbols (MVP, each post-MVP, and :unicorn: as "future").

Good point, perhaps we should open a separate issue to discuss?

> This is also missing the specification of how non-atomic accesses behave. We could just wait for the JS work to finish thought?

Yeah, I was considering just referencing the ecma262 memory model, but it's not immediately obvious to a reader how it would map.

> How do we intend to move such proposals forward?

I seem to remember there being a plan about this. Little help, anyone? :-)

lukewagner:
I think we're still figuring out what, post-MVP, is the process for proposing extensions.  Also, with the concurrent emergence of a proper spec in the spec repo, the shape of a PR/proposal might be quite different.

jfbastien:
I'm not sure we want to block everything on the spec repo at this point in time.

lukewagner:
I guess it depends on how long each of these things (process for evolving post-MVP, spec repo, threads) take.

flagxor:
Agree we don't want to block if possible.
@lukewagner would pulling out the sections this adds into a separate Threads.md doc be easier in terms of weathering the shifting repos?

BinaryEncoding.md
```diff
@@ -786,9 +783,138 @@ for [future :unicorn:][future multiple tables] use and must be 0 in the MVP.
...
+| `i64.atomic_xchg8_s` | `0xe616` | `5` `memory immediate` | atomic exchange with memory |
+| `i64.atomic_xchg8_u` | `0xe617` | `5` `memory immediate` | atomic exchange with memory |
+| `i64.atomic_xchg16_s` | `0xe618` | `5` `memory immediate` | atomic exchange with memory |
+| `i64.atomic_xchg16_u` | `0xe619` | `5` `memory immediate` | atomic exchange with memory |
+| `i64.atomic_xchg32_s` | `0xe61a` | `5` `memory immediate` | atomic exchange with memory |
+| `i64.atomic_xchg32_u` | `0xe61b` | `5` `memory immediate` | atomic exchange with memory |
+
+## Atomic compare exchange operators ([described here](Semantics.md#atomic-memory-accesses))
+
+| `i32.atomic_cmpxchg` | `0xe61c` | `memory immediate` | atomic compare exchange with memory |
```
AndrewScheidecker: The formatting on this table is broken, possible because of the lack of a heading.

binji: Oops, thanks!

BinaryEncoding.md
```diff
@@ -786,9 +783,138 @@ for [future :unicorn:][future multiple tables] use and must be 0 in the MVP.
...
+| `i64.atomic_or16_s` | `0xe618` | `3` `memory immediate` | atomic or with memory |
+| `i64.atomic_or16_u` | `0xe619` | `3` `memory immediate` | atomic or with memory |
+| `i64.atomic_or32_s` | `0xe61a` | `3` `memory immediate` | atomic or with memory |
+| `i64.atomic_or32_u` | `0xe61b` | `3` `memory immediate` | atomic or with memory |
+| `i32.atomic_xor` | `0xe610` | `4` `memory immediate` | atomic xor with memory |
+| `i64.atomic_xor` | `0xe611` | `4` `memory immediate` | atomic xor with memory |
+| `i32.atomic_xor8_s` | `0xe612` | `4` `memory immediate` | atomic xor with memory |
+| `i32.atomic_xor8_u` | `0xe613` | `4` `memory immediate` | atomic xor with memory |
+| `i32.atomic_xor16_s` | `0xe614` | `4` `memory immediate` | atomic xor with memory |
+| `i32.atomic_xor16_u` | `0xe615` | `4` `memory immediate` | atomic xor with memory |
```
sunfishcode: Names like `i32.atomic_xor16_s` are a little awkward with `_s`/`_u` apparently associated with otherwise sign-agnostic operators. What would you think of any of the following conventions?

 - `i32.atomic16_s.xor`
 - `i32.atomic_rmw16_s.xor`
 - `i32.atomic.rmw16_s.xor`   (in this case, following atomic with `.` could be a broader convention)

These would make it clear that the xor isn't signed itself; it's just the loaded value from the rmw that's being sign-extended. This isn't super important with the current opcodes, but could be in the future if wasm ever wants atomic rotate, popcnt, or other operators which care about the number of bits in the actual arithmetic part.

Such a clarification would also align with an idea anticipated [here](https://github.com/WebAssembly/design/blob/master/FutureFeatures.md#integer-overflow-detection) concerning the possibility of adding a real `add_s` and similar operators (which could trap or do other things on overflow) in the future.

binji: I like `i32.atomic.rmw16_s.xor`, it nicely matches the proposed opcode layout too. Though I wonder about `xchg` and `cmpxchg`:
`i32.atomic.rmw16_s.xchg`? And as @jfbastien mentioned, cmpxchg isn't really RMW, so `i32.atomic.cmpxchg16_s`? And I suppose for the non-extending forms:
`i32.atomic.rmw.xor`?

lukewagner:
@flagxor I'm not sure what would be better without knowing what the intended process is for proposing and making changes.  Since threads are a high priority, we could just cheat and carry on like we did pre-MVP.

BinaryEncoding.md
```diff
@@ -786,9 +783,140 @@ for [future :unicorn:][future multiple tables] use and must be 0 in the MVP.
...
+| `i64.atomic_cmpxchg16_u` | `0xe625` | `memory immediate` | atomic compare exchange with memory |
+| `i64.atomic_cmpxchg32_s` | `0xe626` | `memory immediate` | atomic compare exchange with memory |
+| `i64.atomic_cmpxchg32_u` | `0xe627` | `memory immediate` | atomic compare exchange with memory |
+
+## Atomic load/store operators ([described here](Semantics.md#atomic-memory-accesses))
+
+| Name | Opcode | Immediate | Description |
+| ---- | ---- | ---- | ---- |
+| `i32.atomic_load` | `0xe628` | `memory immediate` | atomic load from memory |
+| `i64.atomic_load` | `0xe629` | `memory immediate` | atomic load from memory |
```
lukewagner: For all these load/store ops, could we reuse the existing load/store ops, adding a "consistency" annotation to the `memory_immediate.flags`?

Semantics.md
```diff
@@ -157,6 +162,146 @@ Store operators do not produce a value.
...
+| `i64.atomic_xchg` | 8 bytes | nop | 8 bytes |
+
+The atomic compare exchange operators take three operands: a linear memory
+index, an expected value, and a replacement value. The operators conditionally
+store the replacement value, but only if the loaded value matches the expected
+value.
+
+| Name | Read | Modify | Write |
+| ---- | ---- | ---- | ---- |
+| `i32.atomic_cmpxchg8_s` | 1 byte, sign-extended i8 to i32 | nop | (conditional) 1 byte, wrapped from i32 to i8 |
```
sunfishcode: One detail that's not clear from this presentation is whether the comparison is the sign-extended value with the full 32 bits of the expected value, or the expected value wrapped to 8 bits with the 8-bit loaded value.

BinaryEncoding.md
sunfishcode: `i32.atomic.cmpxchg16_s` and `i32.atomic.rmw.xor` seem to fit the pattern nicely.

AndrewScheidecker:
What about atomic operations on globals?

BinaryEncoding.md
binji: Yes, and I believe that's what was discussed before. But I had a thought it might be nice to have all atomic/threading operators use the same prefix, which led me down this route. One additional nicety is that the opcodes for atomic load/store operators match the non-atomic ones, e.g. `i32.atomic_load` is `0xe628` and `i32.load` is `0x28`, etc.

Since the opcode specifies that it is atomic, this also means that we don't need a special "non-atomic" value for the memory immediate for RMW ops too.

But I'm not married to the idea, just thought it was kinda neat. :-)

lukewagner: Hm, those are fair points.  I guess we are already adding a bunch of operators, so it's not a huge difference, and there would be a natural extension for using the atomic prefix on global ops (when we eventually allow shared globals).

lukewagner:
@AndrewScheidecker I think first we need to define a new `WebAssembly.Global` object for mutable globals that can be imported, exported and marked `shared` (symmetric to `Memory`); but after that it makes sense to have atomic global operations.

BinaryEncoding.md
jfbastien: I'm partial to calling the "atomic" prefix the "lock" prefix, and numbering it `0xF0`.

binji: @jfbastien can't tell if serious...

Semantics.md
binji: I tried to clarify, ptal

BinaryEncoding.md
binji: Ok, done

Semantics.md
```diff
@@ -157,6 +162,146 @@ Store operators do not produce a value.
...
+Atomic read-modify-write (RMW) operators atomically read a value from a memory
+index, modify the value, and store the resulting value to the same memory
+index. All RMW operators return the value read from memory before the modify
+operation was performed.
+
+The sign-agnostic operations are further described [below](#32-bit-integer-operators).
+
+| Name | Read | Modify | Write |
+| ---- | ---- | ---- | ---- |
+| `i32.atomic.rmw8_s.add` | 1 byte, sign-extended i8 to i32 | sign-agnostic addition | 1 byte, wrapped from i32 to i8 |
```
sunfishcode: The cmpxchg clarification looks good. So good, that I'm now wondering about a Return column for the rmw table too :-). This is admittedly a really minor nit at this point because none of the current rmw operators care whether their inputs are extended or wrapped, but if we add other rmw operators in the future, they could potentially care.

Semantics.md
```diff
@@ -678,6 +823,50 @@ outside the range which rounds to an integer in range) traps.
     This trap is intended to be impossible for user code to catch or handle, even in the future when it may be possible to
     handle some other kinds of traps or exceptions.
 
+## Thread operators
+
+  * `is_lock_free`: If the atomic step of an atomic primitive (see
+    [Atomic Memory Accesses](#atomic-memory-accesses)) on a datum of size N
```
rossberg-chromium: Where is N coming from, is that an i32 operand?

Semantics.md
```diff
@@ -157,6 +162,146 @@ Store operators do not produce a value.
...
+  * `i32.atomic.store16`: wrap i32 to i16 and atomically store 2 bytes
+  * `i32.atomic.store`: (no conversion) atomically store 4 bytes
+  * `i64.atomic.store8`: wrap i64 to i8 and atomically store 1 byte
+  * `i64.atomic.store16`: wrap i64 to i16 and atomically store 2 bytes
+  * `i64.atomic.store32`: wrap i64 to i32 and atomically store 4 bytes
+  * `i64.atomic.store`: (no conversion) atomically store 8 bytes
+  * `f32.atomic.store`: (no conversion) atomically store 4 bytes
+  * `f64.atomic.store`: (no conversion) atomically store 8 bytes
+
+Atomic read-modify-write (RMW) operators atomically read a value from a memory
```
rossberg-chromium: Can you specify in which order the operands are expected on the stack for these instructions?

Semantics.md
```diff
@@ -157,6 +162,146 @@ Store operators do not produce a value.
...
+| `i32.atomic.rmw.xchg` | 4 bytes | nop | 4 bytes | as i32 |
+| `i64.atomic.rmw8_s.xchg` | 1 byte | nop | 1 byte | sign-extended i8 to i64 |
+| `i64.atomic.rmw8_u.xchg` | 1 byte | nop | 1 byte | zero-extended i8 to i64 |
+| `i64.atomic.rmw16_s.xchg` | 2 bytes | nop | 2 bytes | sign-extended i16 to i64 |
+| `i64.atomic.rmw16_u.xchg` | 2 bytes | nop | 2 bytes | zero-extended i16 to i64 |
+| `i64.atomic.rmw32_s.xchg` | 4 bytes | nop | 4 bytes | sign-extended i32 to i64 |
+| `i64.atomic.rmw32_u.xchg` | 4 bytes | nop | 4 bytes | zero-extended i32 to i64 |
+| `i64.atomic.rmw.xchg` | 8 bytes | nop | 8 bytes | as i64 |
+
+The atomic compare exchange operators take three operands: a linear memory
```
rossberg-chromium: Nit: the linear memory is an immediate, not an operand.

Semantics.md
```diff
@@ -678,6 +823,50 @@ outside the range which rounds to an integer in range) traps.
...
+    bytes will be performed without the agent acquiring a lock outside the N
+    bytes comprising the datum, then return `1`. Otherwise, return `0`.
+
+### Wait
+
+The wake and wait operators are optimizations over busy-waiting for a value to
+change. It is a validation error to use these operators on non-shared linear
+memory.
+
+The wait operators take three operands: a linear memory index, an expected
```
rossberg-chromium: Nit: as above, I suppose the memory is an immediate.

BinaryEncoding.md
```diff
@@ -786,9 +783,140 @@ for [future :unicorn:][future multiple tables] use and must be 0 in the MVP.
 | `f32.reinterpret/i32` | `0xbe` | | |
 | `f64.reinterpret/i64` | `0xbf` | | |
 
+## Thread operators ([described here](Semantics.md#thread-operators))
+
+| Name | Opcode | Immediate | Description |
+| ---- | ---- | ---- | ---- |
+| `is_lock_free` | `0xe600` | | |
```
rossberg-chromium: Can we allocate prefix opcodes at the end of the opcode space instead of the middle? So `0xff` for threads being the first such proposal.

Semantics.md
```diff
@@ -678,6 +823,50 @@ outside the range which rounds to an integer in range) traps.
...
+| 0 | "ok", woken by another thread |
+| 1 | "not-equal", the loaded value did not match the expected value |
+| 2 | "timed-out", not woken by another thread before timeout expired |
+
+  * `i32.wait`: load i32 value, compare to expected, and wait for `i32.wake` at same memory index
+  * `i64.wait`: load i64 value, compare to expected, and wait for `i64.wake` at same memory index
+
+### Wake
+
+The wake operators take two operands: a linear memory index, and a wake count.
```
rossberg-chromium: Nit: same here.

BinaryEncoding.md
```diff
@@ -786,9 +783,140 @@ for [future :unicorn:][future multiple tables] use and must be 0 in the MVP.
...
+| `i32.wait` | `0xe601` | `offset` | |
+| `i64.wait` | `0xe602` | `offset` | |
+| `i32.wake` | `0xe603` | `offset` | |
+| `i64.wake` | `0xe604` | `offest` | |
```
rossberg-chromium: Don't these all have memory immediates?

BinaryEncoding.md
```diff
@@ -786,9 +783,140 @@ for [future :unicorn:][future multiple tables] use and must be 0 in the MVP.
...
+| `i32.atomic.rmw8_u.xor` | `0xe613` | `4` `memory immediate` | atomic xor with memory |
+| `i32.atomic.rmw16_s.xor` | `0xe614` | `4` `memory immediate` | atomic xor with memory |
+| `i32.atomic.rmw16_u.xor` | `0xe615` | `4` `memory immediate` | atomic xor with memory |
+| `i64.atomic.rmw8_s.xor` | `0xe616` | `4` `memory immediate` | atomic xor with memory |
+| `i64.atomic.rmw8_u.xor` | `0xe617` | `4` `memory immediate` | atomic xor with memory |
+| `i64.atomic.rmw16_s.xor` | `0xe618` | `4` `memory immediate` | atomic xor with memory |
+| `i64.atomic.rmw16_u.xor` | `0xe619` | `4` `memory immediate` | atomic xor with memory |
+| `i64.atomic.rmw32_s.xor` | `0xe61a` | `4` `memory immediate` | atomic xor with memory |
+| `i64.atomic.rmw32_u.xor` | `0xe61b` | `4` `memory immediate` | atomic xor with memory |
+| `i32.atomic.rmw.xchg` | `0xe610` | `5` `memory immediate` | atomic exchange with memory |
```
rossberg-chromium: Is xchg actually a RMW operation? Seems like it could be its own group, with names like `i32.atomic.xchg16_s`, in line with `cmpxchg`.

Semantics.md
```diff
@@ -157,6 +162,146 @@ Store operators do not produce a value.
...
+| `i32.atomic.rmw16_u.xor` | 2 bytes | 16-bit sign-agnostic bitwise exclusive or | 2 bytes | zero-extended i16 to i32 |
+| `i32.atomic.rmw.xor` | 4 bytes | 32-bit sign-agnostic bitwise exclusive or | 4 bytes | as i32 |
+| `i64.atomic.rmw8_s.xor` | 1 byte | 8-bit sign-agnostic bitwise exclusive or | 1 byte | sign-extended i8 to i64 |
+| `i64.atomic.rmw8_u.xor` | 1 byte | 8-bit sign-agnostic bitwise exclusive or | 1 byte | zero-extended i8 to i64 |
+| `i64.atomic.rmw16_s.xor` | 2 bytes | 16-bit sign-agnostic bitwise exclusive or | 2 bytes | sign-extended i16 to i64 |
+| `i64.atomic.rmw16_u.xor` | 2 bytes | 16-bit sign-agnostic bitwise exclusive or | 2 bytes | zero-extended i16 to i64 |
+| `i64.atomic.rmw32_s.xor` | 4 bytes | 32-bit sign-agnostic bitwise exclusive or | 4 bytes | sign-extended i32 to i64 |
+| `i64.atomic.rmw32_u.xor` | 4 bytes | 32-bit sign-agnostic bitwise exclusive or | 4 bytes | zero-extended i32 to i64 |
+| `i64.atomic.rmw.xor` | 8 bytes | 64-bit sign-agnostic bitwise exclusive or | 8 bytes | as i64 |
+| `i32.atomic.rmw8_s.xchg` | 1 byte | nop | 1 byte | sign-extended i8 to i32 |
```
rossberg-chromium: Is xchg actually a RMW operation? It seems like it would fit better with the `cmpxchg` group -- it's an unconditional version of it. With analogous names like `i32.atomic.xchg16_s`. Or as its own group.

jfbastien: xchg is usually considered an RMW operation.

Semantics.md
```diff
@@ -157,6 +162,146 @@ Store operators do not produce a value.
...
+| `i64.atomic.rmw16_s.xchg` | 2 bytes | nop | 2 bytes | sign-extended i16 to i64 |
+| `i64.atomic.rmw16_u.xchg` | 2 bytes | nop | 2 bytes | zero-extended i16 to i64 |
+| `i64.atomic.rmw32_s.xchg` | 4 bytes | nop | 4 bytes | sign-extended i32 to i64 |
+| `i64.atomic.rmw32_u.xchg` | 4 bytes | nop | 4 bytes | zero-extended i32 to i64 |
+| `i64.atomic.rmw.xchg` | 8 bytes | nop | 8 bytes | as i64 |
+
+The atomic compare exchange operators take three operands: a linear memory
+index, an `expected` value, and a `replacement` value. The operators
+conditionally store the `replacement` value, but only if the `loaded` value is
+equal to the `expected` value.
```
sbc100: The logic of this sentence seems a little strange to me.  "X conditionally does Y, but only when Z"..  I think what you are trying to say is that its always conditional, and then describe said condition.

Semantics.md
binji: Whoops, I meant address. :-)

Semantics.md
binji: Done.

BinaryEncoding.md
binji: Done.

BinaryEncoding.md
binji: Done.

Semantics.md
binji: Done.

Semantics.md
binji: I couldn't find another place in the doc to crib from, so I just specified it by describing the . Should I be more explicit?

Semantics.md
binji: I'm fine with it either way, but it probably shouldn't be grouped with `cmpxchg` since those have 3 operands.

jfbastien:
I think we've got a great opportunity to call the "atomic" prefix the "lock" prefix, and number it `0xF0`.

syg:
> Yeah, I was considering just referencing the ecma262 memory model, but it's not immediately obvious to a reader how it would map.

@binji Happy to help if you bring me up to speed about spec conventions used here.

BinaryEncoding.md
```diff
@@ -786,6 +950,18 @@ for [future :unicorn:][future multiple tables] use and must be 0 in the MVP.
...
+
+| Name | Opcode | Immediate | Description |
+| ---- | ---- | ---- | ---- |
+| `i32.wait` | `0xc0` | `offset` | SAB like wait, takes an index + `offset` and an f64 timeout |
+| `i64.wait` | `0xc1` | `offset` | SAB like wait, takes an index + `offset`and an f64 timeout |
+| `i32.wake` | `0xc2` | `offset` | SAB like wake, takes an index + `offset` |
+| `i64.wake` | `0xc3` | `offest` | SAB like wake, takes an index + `offset` |
+| `i32.thread_create` | `0xc4` | `table` | Creates a thread, stores thread in `table` at an i32 index. Returns 0 on success or non-zero otherwise. |
+| `i32.thread_join` | `0xc5` | `table` | Waits to join to thread stored at an i32 index in `table`. Returns 0 on success or non-zero otherwise. |
+| `thread_exit` | `0xc6` | | Halts current thread. |
```
jfbastien: It would be good to document this! Is the intent to start off with workers, mimic JS' API, and then add better capabilities later?

Semantics.md
rossberg-chromium: I guess I'm wondering by what criterion xchg is RMW but cmpxchg isn't.

Bikeshedding territory, but why the complicated names in the first place? Wouldn't i32.atomic.add8_s etc work just fine? Sure it's RMW, but that's kind of implied.

sunfishcode: `cmpxchg` (compare-and-swap) is fundamentally [more powerful](http://www.cs.yale.edu/homes/aspnes/pinewiki/WaitFreeHierarchy.html) than other RMW operations. Between that, and having a distinct type signature, it commonly gets its own category, even if it is also an RMW operation.

`cmpxchg` isn't just a conditional version of `xchg`. For example, an abstract formal semantics wouldn't be able to reduce `cmpxchg` to `xchg` when the compare succeeds; that would imply the comparison is a separate step from the swap, breaking atomicity.

`add8_s` under wasm's existing naming scheme would seem to imply that the `add`, `or`, `and`, `xor`, etc., are signed/unsigned, which is inconsistent with the existing operators.

rossberg-chromium: @sunfishcode, interesting, thanks!

Regarding the naming, I see your argument. My only remaining quibble is that, even being in different expressive categories, it would still be nice to have a somewhat more symmetric naming scheme between cmpxchg and the other RMW operators.

binji:
@jfbastien isn't that a bit biased? ;-)

@syg not 100% on conventions here, since it's informal. For memory model stuff, I'd guess it would be best to write a new markdown file that we could reference here.

sunfishcode:
Can we specify a set of sizes for which `is_lock_free` is guaranteed to return true? This was [discussed earlier](https://github.com/WebAssembly/design/issues/300), and it seems that 1, 2, and 4 should be lock free on all relevant platforms. Such a guarantee allows applications to avoid having fallback paths in common situations. And, these would support eg. C++'s `std::atomic::is_always_lock_free`.

Another topic specific to wasm is alignment for atomics. It seems that at least, `is_lock_free` doesn't describe misaligned atomic accesses, though one could contemplate stronger constraints as well, such as having misaligned atomics trap.

Is this the best place for these topics, or should these be addressed in the (upcoming) memory model document?

Semantics.md
sunfishcode: `i32.atomic.rmw8_s.cmpxchg` would be the maximally symmetric convention, it seems. I don't think it's critical, but I wouldn't be opposed to it.

jfbastien:
@sunfishcode 
> Can we specify a set of sizes for which `is_lock_free` is guaranteed to return true? This was [discussed earlier](https://github.com/WebAssembly/design/issues/300), and it seems that 1, 2, and 4 should be lock free on all relevant platforms. Such a guarantee allows applications to avoid having fallback paths in common situations. And, these would support eg. C++'s `std::atomic::is_always_lock_free`.

I'd like to match exactly what JavaScript guarantees. A mismatch would be akin to raw C++ programs talking through shared memory with different compiler flags (e.g. `is_always_lock_free` differing!). Fraught with peril because code will statically decide to behave differently (say, use a lock-free datastructure versus a plain old lock).

> Another topic specific to wasm is alignment for atomics. It seems that at least, `is_lock_free` doesn't describe misaligned atomic accesses, though one could contemplate stronger constraints as well, such as having misaligned atomics trap.

And SAB doesn't specify alignment of atomics because the array buffer accesses are aligned, right? We're in novel territory here, and we know there's significant hardware difference in what outcome can be observed.

It would be good to lay out different options and perform some Science. At the end of the day alignment of atomics is a non-feature which we *must* spec, but I don't really care *how* we spec it. Determinism is best, unless the cost is too high.

For now I'd lean towards trap as @sunfishcode suggests, but I can be moved away.

> Is this the best place for these topics, or should these be addressed in the (upcoming) memory model document?

Does the memory model address wasm-specific issues?

Agreed though, maybe we want this entire change to live in its own document for now, and merge its sub-parts once we've reached consensus.

binji:
> I'd like to match exactly what JavaScript guarantees.

JavaScript only guarantees that 4-byte is lock free.

> And SAB doesn't specify alignment of atomics because the array buffer accesses are aligned, right?

Right, and you can't use atomics on a `DataView`.

> For now I'd lean towards trap as @sunfishcode suggests, but I can be moved away.

+1

> Agreed though, maybe we want this entire change to live in its own document for now, and merge its sub-parts once we've reached consensus.

Not a bad idea, something like a "Threads-Proposal.md" that specifies how the rest of the files would change? Is that nicer than having a "thread-proposal" branch that we review PRs on?

Semantics.md
binji: Done.

jfbastien:
>> Agreed though, maybe we want this entire change to live in its own document for now, and merge its sub-parts once we've reached consensus.
>
> Not a bad idea, something like a "Threads-Proposal.md" that specifies how the rest of the files would change? Is that nicer than having a "thread-proposal" branch that we review PRs on?

Maybe something that won't change names once stabilized, so URLs on the internets don't break.

sunfishcode:
> I'd like to match exactly what JavaScript guarantees. A mismatch would be akin to raw C++ programs talking through shared memory with different compiler flags (e.g. is_always_lock_free differing!). Fraught with peril because code will statically decide to behave differently (say, use a lock-free datastructure versus a plain old lock).

If, within a given engine, JS `Atomics.isLockFree` agrees with WebAssembly `is_lock_free` for all values, then there is no problem. On the other hand, if, within a given engine, JS `Atomics.isLockFree` ever disagrees with WebAssembly `is_lock_free`, for any reason, it'd cause all the problems, even if without static guarantees, because JS code and WebAssembly code would *dynamically* pick different datastructures.

So adding a static guarantee doesn't seem to introduce any new peril.

lars-t-hansen:
@sunfishcode 

> Can we specify a set of sizes for which is_lock_free is guaranteed to return true? This was discussed earlier, and it seems that 1, 2, and 4 should be lock free on all relevant platforms. Such a guarantee allows applications to avoid having fallback paths in common situations. And, these would support eg. C++'s std::atomic::is_always_lock_free.

I think the matter has been resolved but for the record:

It is arguable that 1 and 2 are not lock free on all relevant platforms.  MIPS has only 4-byte LL/SC so if you want a 1-byte lock-free atomic you must use LL/SC on the word that contains that byte.  Now the other bytes of that word may be used for other, unrelated variables that are not necessarily atomic, and we don't control that.  In this situation we're only good if a concurrent non-atomic update to any of those other bytes will cause the SC to fail and the LL/SC loop to retry.  It's not obvious from the architecture manuals that I have seen that this is so (though ISTR that I found some hints in that direction).

> Another topic specific to wasm is alignment for atomics. It seems that at least, is_lock_free doesn't describe misaligned atomic accesses, though one could contemplate stronger constraints as well, such as having misaligned atomics trap.

Copy atomicity is also compromised by misaligned atomic accesses (but so is copy atomicity on regular accesses).  I don't know how much that matters.  The JS memory model goes out of its way to specify copy atomicity in some situations, but assumes alignment.

IMO unaligned atomics should just trap, until we know for sure that we can reasonably support atomics that don't.

sunfishcode:
@lars-t-hansen Thanks for the explanation. Clang is currently written in a way that does not permit it to consider this, and it therefore doesn't consider it on MIPS or any other of its targets. I have now filed https://reviews.llvm.org/D31516 with a patch to Clang which introduces logic in clang to allow it to consider this. Hopefully either the patch will be accepted, or we will learn more.

binji:
Thanks, @sunfishcode for looking into this. In the meantime, I've updated this PR to specify that 1- and 2-byte atomics are not guaranteed lock-free. I've also added some text about misaligned atomics.

Semantics.md
```diff
@@ -678,6 +829,51 @@ outside the range which rounds to an integer in range) traps.
...
+
+### Wake
+
+The wake operators take two operands: an address operand, and a wake count. The
+operation will wake a maximum of "wake count" threads that are waiting on the
+same address. The operator returns the number of thread that were
+woken.
+
+  * `i32.wake`: wake up to N threads waiting on the given address via `i32.wait`
+  * `i64.wake`: wake up to N threads waiting on the given address via `i64.wait`
```
jfbastien: I think we only want the 32-bit variants of wait / wake.

Semantics.md
```diff
@@ -678,6 +829,51 @@ outside the range which rounds to an integer in range) traps.
     This trap is intended to be impossible for user code to catch or handle, even in the future when it may be possible to
     handle some other kinds of traps or exceptions.
 
+## Thread operators
+
+  * `is_lock_free`: Given an operand `N`, if the atomic step of an atomic
+    primitive (see [Atomic Memory Accesses](#atomic-memory-accesses)) on a
+    datum of size `N` bytes will be performed without the agent acquiring a
+    lock outside the `N` bytes comprising the datum, then return `1`.
+    Otherwise, return `0`.
```
jfbastien: Is `N` an immediate? It's not clear.

What type?

Are there invalid values for `N`?

Can we express the `constexpr`-ness of this? SAB had some wording somewhere about not changing between visible agents or something.

Are `wait` and `wake` atomic primitives? Because this wording would affect them and they definitely aren't lock-free. Currently the headings seem to say "no", it should remain that way.

Semantics.md
```diff
@@ -678,6 +829,51 @@ outside the range which rounds to an integer in range) traps.
...
+### Wait
+
+The wake and wait operators are optimizations over busy-waiting for a value to
+change. It is a validation error to use these operators on non-shared linear
+memory.
+
+The wait operators take three operands: an address operand, an expected
+value, and a relative timeout in milliseconds as an f64. The timeout must be
+non-negative, and can also be positive infinity. If the timeout is positive
+infinity then the timeout will never expire.
```
jfbastien: Why disallow negative values? A timer would just be expired if it were negative. Otherwise, specify the effects of a negative timer.

What of NaN?

Semantics.md
```diff
@@ -678,6 +829,51 @@ outside the range which rounds to an integer in range) traps.
...
+
+The wait operation begins by performing an atomic load from the given address.
+If the loaded value is not equal to the expected value, the operator returns 1
+("not-equal"). If the values are equal, the thread is suspended. If the thread
+is woken (by execution of the corresponding "wake" operator on another thread),
```
jfbastien: What if the wake overlaps but isn't aligned?

Semantics.md
```diff
@@ -678,6 +829,51 @@ outside the range which rounds to an integer in range) traps.
...
+The wake and wait operators are optimizations over busy-waiting for a value to
+change. It is a validation error to use these operators on non-shared linear
+memory.
+
+The wait operators take three operands: an address operand, an expected
+value, and a relative timeout in milliseconds as an f64. The timeout must be
+non-negative, and can also be positive infinity. If the timeout is positive
+infinity then the timeout will never expire.
+
+The wait operation begins by performing an atomic load from the given address.
```
jfbastien: Implication here is it can fault if out-of-bounds, and can't be misaligned?

Semantics.md
```diff
@@ -678,6 +829,51 @@ outside the range which rounds to an integer in range) traps.
...
+| 1 | "not-equal", the loaded value did not match the expected value |
+| 2 | "timed-out", not woken by another thread before timeout expired |
+
+  * `i32.wait`: load i32 value, compare to expected, and wait for `i32.wake` at same address
+  * `i64.wait`: load i64 value, compare to expected, and wait for `i64.wake` at same address
+
+### Wake
+
+The wake operators take two operands: an address operand, and a wake count. The
+operation will wake a maximum of "wake count" threads that are waiting on the
```
jfbastien: The wording can be interpreted as saying: there can be 1000 waiters, it is valid to wake none of them, ever.

Semantics.md
jfbastien: What's the type of the load, and `expected`?

Semantics.md
```diff
@@ -678,6 +829,51 @@ outside the range which rounds to an integer in range) traps.
...
+change. It is a validation error to use these operators on non-shared linear
+memory.
+
+The wait operators take three operands: an address operand, an expected
+value, and a relative timeout in milliseconds as an f64. The timeout must be
+non-negative, and can also be positive infinity. If the timeout is positive
+infinity then the timeout will never expire.
+
+The wait operation begins by performing an atomic load from the given address.
+If the loaded value is not equal to the expected value, the operator returns 1
```
jfbastien: Return is `i32`?

Semantics.md
```diff
@@ -678,6 +829,51 @@ outside the range which rounds to an integer in range) traps.
...
+| 0 | "ok", woken by another thread |
+| 1 | "not-equal", the loaded value did not match the expected value |
+| 2 | "timed-out", not woken by another thread before timeout expired |
+
+  * `i32.wait`: load i32 value, compare to expected, and wait for `i32.wake` at same address
+  * `i64.wait`: load i64 value, compare to expected, and wait for `i64.wake` at same address
+
+### Wake
+
+The wake operators take two operands: an address operand, and a wake count. The
```
jfbastien: Wake count is `i32`, interpreted as unsigned?

BinaryEncoding.md
```diff
@@ -786,9 +783,140 @@ for [future :unicorn:][future multiple tables] use and must be 0 in the MVP.
...
 
+## Thread operators ([described here](Semantics.md#thread-operators))
+
+| Name | Opcode | Immediate | Description |
+| ---- | ---- | ---- | ---- |
+| `is_lock_free` | `0xff00` | | |
+| `i32.wait` | `0xff01` | `memory immediate` | |
+| `i64.wait` | `0xff02` | `memory immediate` | |
+| `i32.wake` | `0xff03` | `memory immediate` | |
+| `i64.wake` | `0xff04` | `memory immediate` | |
```
jfbastien: I think we only want the 32-bit variants of wait / wake.

jfbastien: What's the memory immediate?

jfbastien: Do we intend wait / wake to have sequentially-consistent semantics, or acquire (wait) / release (wake) sematics?

Semantics.md
jfbastien: How do I "wake all"? In JS I can pass infinity. It seems silly to pass UINT_MAX.

BinaryEncoding.md
```diff
@@ -786,9 +783,140 @@ for [future :unicorn:][future multiple tables] use and must be 0 in the MVP.
...
+
+## Atomic read-modify-write operators ([described here](Semantics.md#atomic-memory-accesses))
+
+| Name | Opcode | Immediates | Description |
+| ---- | ---- | ---- | ---- |
+| `i32.atomic.rmw.add` | `0xff10` | `0` `memory immediate` | atomic add to memory |
```
jfbastien: Why use `memory immediate` to denote the operation (add / sub / etc)? Why not use another opcode instead? It would be helpful to do a map of opcode extension space usage here, similar to what was done for the MVP opcodes. Is the intent that `0xff` is the "atomics / threads" prefix? I still think we should use `0xf0`, it's free!

Semantics.md
```diff
@@ -112,8 +112,13 @@ or [defined inside the module](Modules.md#linear-memory-section). After import
...
 
-In the MVP, linear memory cannot be shared between threads of execution.
-The addition of [threads :unicorn:][future threads] will allow this.
+### Shared Linear Memory
+
+A Linear memory can be marked as shared, which allows it to be shared between
+threads of execution. The shared memory can be imported or defined in the
+module. It is a validation error to attempt to import shared linear
+memory if the module's memory import doesn't specify that it allows shared
+memory.
```
jfbastien: There are two places where we can mark a memory as shared:

1. In the `WebAssembly.Memory` constructor. You need to update JS.md for this.
2. In [`memory_type`](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#memory_type) (for both import and memory sections). This looks like we don't have room for expansion though?

You also need to update JS.md to specify that `.buffer` will return a `SharedArrayBuffer`.

We also must specify what happens when you `.grow` or use `grow_memory`.

jfbastien:
How does one share a `WebAssembly.Memory`, which is marked as shared, across workers? Structured clone?

Semantics.md
binji: AIUI, we use `operand` to mean stack value, correct? I could be more explicit if desired, though. Tried to clear up the other stuff, PTAL.

Semantics.md
binji: Good points, I made this value match the behavior of JS's `Atomics.wait`.

Semantics.md
binji: Done.

binji: Done.

Semantics.md
binji: Done.

Semantics.md
binji: Done.

Semantics.md
binji: Done.

Semantics.md
binji: Good point, fixed.

Semantics.md
binji: Yeah, this is a little clunky. I made it a signed `i32`, where < 0 means wake all. Using an `f64` for the wake count seemed really odd. Perhaps it could be a separate opcode?

BinaryEncoding.md
binji: Done.

binji: The JS behavior is sequentially-consistent, so we should match, I think.

BinaryEncoding.md
binji: > Why use memory immediate to denote the operation (add / sub / etc)? Why not use another opcode instead?

I considered using separate opcodes, but it eats up the opcode space pretty quickly since we have 12 variants per operation. I also kind of like that all the rmw ops are grouped together, but I suppose that's just personal preference.

> It would be helpful to do a map of opcode extension space usage here, similar to what was done for the MVP opcodes.

Sorry, do you mean mapping the usage of the atomics prefix space? Or mapping the usage of the remaining opcode space?

> I still think we should use 0xf0, it's free!

I don't have a preference for "atomics/threads" prefix, `0xf0` is just as good as anything else as far as I'm concerned. But it does seem like adding opcode prefixes starting at the end and going backward is logical. I suppose we could argue for 0xf0 by saying all prefixes start with `0xf_` and this is the first one.

binji:
Yeah, I was hoping to get away with adding the easy stuff before diving in to all the other thread stuff. But it doesn't really make sense to add these primitives until we have a good base to build on top of. I'll work on a `Threads.md` next for this.

AndrewScheidecker:
I was trying to figure out if it's possible to implement the `wait`/`wake` ops using the Windows API's `WaitOnAddress`/`WakeByAddressSingle`/`WakeByAddressAll`, and found some interesting caveats in how they are currently implemented: [search on this page for WaitOnAddress](http://www.cbloom.com/rambles.html), which references [this post on the subject](https://blogs.msdn.microsoft.com/oldnewthing/20160826-00/?p=94185).

However, it seems like if `wait`/`wake` are specified loosely enough to be implemented using the system-level APIs in Windows, then the only robust way to use `wake` will be to wake all waiting threads. That's still useful, but maybe there should just be a `wake_all` op for now.

Semantics.md
jfbastien: Yes, separate `wakeAll` sounds good.

BinaryEncoding.md
jfbastien: >> It would be helpful to do a map of opcode extension space usage here, similar to what was done for the MVP opcodes.
>
> Sorry, do you mean mapping the usage of the atomics prefix space? Or mapping the usage of the remaining opcode space?

The former.

> I don't have a preference for "atomics/threads" prefix, 0xf0 is just as good as anything else as far as I'm concerned. But it does seem like adding opcode prefixes starting at the end and going backward is logical. I suppose we could argue for 0xf0 by saying all prefixes start with 0xf_ and this is the first one.

🎉

jfbastien:
> I was trying to figure out if it's possible to implement the `wait`/`wake` ops using the Windows API's `WaitOnAddress`/`WakeByAddressSingle`/`WakeByAddressAll`, and found some interesting caveats in how they are currently implemented: [search on this page for WaitOnAddress](http://www.cbloom.com/rambles.html), which references [this post on the subject](https://blogs.msdn.microsoft.com/oldnewthing/20160826-00/?p=94185).
>
> However, it seems like if `wait`/`wake` are specified loosely enough to be implemented using the system-level APIs in Windows, then the only robust way to use `wake` will be to wake all waiting threads. That's still useful, but maybe there should just be a `wake_all` op for now.

We should match the SharedArrayBuffer spec at a minimum, because JS and wasm can wait / wake each other, and write to the same atomic locations. I'll punt implementability concerns to SAB: if MS folks were OK with SAB they'll be OK with the same API here.

@binji This embedder/embedded wake/wait should be pointed out in the wording.

BinaryEncoding.md
```diff
@@ -789,9 +789,14 @@ for [future :unicorn:][future multiple tables] use and must be 0 in the MVP.
...
-| `i32.wake` | `0xff03` | `memory immediate` | |
-| `i64.wake` | `0xff04` | `memory immediate` | |
+| `i32.wake` | `0xff02` | `memory immediate` | |
+
+The `memory_immediate` type is encoded as follows:
+
+| Name | Type | Description |
+| ---- | ---- | ---- |
+| flags | `varuint32` | a bitfield which currently must be 0 |
+| offset | `varuint32` | the value of the offset |
```
jfbastien: Is it useful to have offsets on wake / wait? I'm not sure, but open to being convinced for / against. Seems to match with other memory ops?

Semantics.md
```diff
@@ -831,29 +831,39 @@ outside the range which rounds to an integer in range) traps.
...
-    datum of size `N` bytes will be performed without the agent acquiring a
-    lock outside the `N` bytes comprising the datum, then return `1`.
-    Otherwise, return `0`.
+  * `is_lock_free`: Given an operand `N` of type `i32`, if the atomic step of
+    an atomic primitive (see [Atomic Memory Accesses](#atomic-memory-accesses))
+    on a datum of size `N` bytes will be performed without the [agent](agent)
+    acquiring a lock outside the `N` bytes comprising the datum, then return
+    `1`. Otherwise, return `0`. Once the value of `is_lock_free` for a given
+    value `N` has been observed for any [agent](agent) in an
+    [agent cluster](agent cluster) , it cannot change.
```
jfbastien: Extra space here.

Semantics.md
```diff
@@ -831,29 +831,39 @@ outside the range which rounds to an integer in range) traps.
...
 
-The wait operation begins by performing an atomic load from the given address.
-If the loaded value is not equal to the expected value, the operator returns 1
-("not-equal"). If the values are equal, the thread is suspended. If the thread
-is woken (by execution of the corresponding "wake" operator on another thread),
-the wait operator returns 0 ("ok"). If the timeout expires before another
-thread wakes this one, this operator returns 2 ("timed-out").
+| `timeout` value | Behavior |
+| ---- | ---- |
+| `timeout` <= 0 | Expires immediately |
```
jfbastien: "immediately": is it guaranteed synchronous, or can it wait until the next runloop?

Semantics.md
```diff
@@ -862,18 +872,24 @@ thread wakes this one, this operator returns 2 ("timed-out").
...
+
+| `wake count` value | Behavior |
+| ---- | ---- |
+| `wake count` < 0 | Wake all waiters |
+| `wake count` == 0 | Wake no waiters |
+| `wake count` > 0 | Wake up to `wake count` waiters |
 
-  * `i32.wake`: wake up to N threads waiting on the given address via `i32.wait`
-  * `i64.wake`: wake up to N threads waiting on the given address via `i64.wait`
+  * `i32.wake`: wake up `wake count` threads waiting on the given address via `i32.wait`
```
jfbastien: I think you went a bit too far: if there aren't `wake count` waiters then you just wake all of them. The SAB algorithm does this IIUC. We could use `min(wake count , waiters)` or something?

jfbastien:
Actually I talked with @pizlonator and it may be useful to keep i32 + i64 variants of wake / wait: if we specify that only the address, not the size, participate in realizing wake / wait then it's super convenient to have both. I was wary of how we'd specify overlap, but based on address just works.

BinaryEncoding.md
binji: I figure it makes sense for consistency. And is convenient for the same reason that it is for other memory accesses, so you can have a pointer to a struct and use that for wait/wake instead of having to add it in yourself.

jfbastien: Yeah sounds fine. Just figured I'd ask :)

Semantics.md
binji: Good question, it's not clear to me from the ES spec what the behavior is. Seems like synchronous is useful, but it's also often good to have synchronous vs. asynchronous behavior be consistent for all values of an operator. What do you think?

Semantics.md
binji: Oops, I updated the text above but this one needs to be fixed too.

Semantics.md
jfbastien: I defer to @lars-t-hansen / @syg on this.

Semantics.md
```diff
@@ -848,6 +848,12 @@ memory. They are non-atomic and have sequentially consistent memory ordering.
 Both wake and wait operators trap if the effective address of either operator
 is misaligned or out-of-bounds.
 
+The embedder is also permitted to suspend or wake a thread. In those cases,
+these operators will behave as though the equivalent wake or wait operation was
+performed by the execution engine.
```
jfbastien: This goes both ways: a wake will also wake waiters from the embedder.

binji: Right, I guess that wasn't clear from the wording here. Let me try again :)

AndrewScheidecker:
@jfbastien I missed [the SAB wait/wake](https://tc39.github.io/ecma262/#sec-atomics.wait) somehow when I looked before.

Incidentally, I believe I found an erroneous assertion in the SAB spec. In [`Atomics.wait`](https://tc39.github.io/ecma262/#sec-atomics.wait), step 18 performs `RemoveWaiter` after `Suspend` completes. `RemoveWaiter` asserts that the agent being removed is in the `WaiterList`. However, if `Suspend` returned because a call to `Atomics.wake` called `WakeWaiter`, the agent will have been removed from the `WaiterList` already by step 11 of [`Atomics.wake`](https://tc39.github.io/ecma262/#sec-atomics.wake).

The SAB wait/wake seems to be specified in a way that can't use any of the existing OS-level APIs. e.g. it specifies wake order where Linux futexes do not, and it specifies that a `wait` call with infinite timeout will *only* be woken by a `wake` call that occurs after the `wait` call, which isn't necessarily the case on Windows.

My original thought was that it would be useful to spec the futex stuff in a way that can use the OS-level APIs to synchronize with native processes through shared memory, but I see now that's futile. However, it does suggest that some higher-level synchronization primitives should be exposed that *can* plug into the OS-level primitives. That is probably beyond the scope of the initial threading extensions.

binji:
@AndrewScheidecker I think the spec is correct; it looks like `RemoveWaiters` creates a new list with one of the waiters removed, but it doesn't modify the original list.

AndrewScheidecker:
@binji The RemoveWaiters spec says:
> 1. Assert: The calling agent is in the critical section for WL.
> 2. Let L be a new empty List.
> 3. Let S be a reference to the list of waiters in WL.
> 4. Repeat, while c > 0 and S is not an empty List,
>     a. Let W be the first waiter in S.
>     b. Add W to the end of L.
>     c. Remove W from S.
>     d. Subtract 1 from c.
> 5. Return L.

I would interpret step 3's "Let S be a reference to the list of waiters in WL" to mean that when step 4c removes W from S, it's removing W from WL's list of waiters.

binji:
@AndrewScheidecker Yes, I see what you mean. Filed a bug here: https://github.com/tc39/ecma262/issues/861

binji:
Moved everything to a single Threads.md file, curious what everyone thinks.

Threads.md
```diff
@@ -0,0 +1,403 @@
...
+    acquiring a lock outside the `N` bytes comprising the datum, then return
+    `1`. Otherwise, return `0`. Once the value of `is_lock_free` for a given
+    value `N` has been observed for any [agent](agent) in an
+    [agent cluster](agent cluster), it cannot change.
+
+#### Wait and Wake operators
+
+The wake and wait operators are optimizations over busy-waiting for a value to
+change. It is a validation error to use these operators on non-shared linear
+memory. They are non-atomic and have sequentially consistent memory ordering.
```
jfbastien: They're not in the "atomic" section of operators, but the "read' operation from wait is atomic.

Threads.md
```diff
@@ -0,0 +1,403 @@
...
+| Name | Type | Description |
+| ---- | ---- | ---- |
+| flags | `varuint32` | a bitfield which currently must be 0 |
+| offset | `varuint32` | the value of the offset |
+
+### Atomic read-modify-write operators ([described here](Semantics.md#atomic-memory-accesses))
+
+| Name | Opcode | Immediates | Description |
+| ---- | ---- | ---- | ---- |
+| `i32.atomic.rmw.add` | `0xff10` | `0` `memory immediate` | atomic add to memory |
```
jfbastien: I still find this kinda weird: there's the memory immediate as above, but here there's an opcode number as well. If we carve off all of `0xf0` for atomics then we've got plenty of space, and we'd use a byte less.

Threads.md
```diff
@@ -0,0 +1,403 @@
...
+The wait operator take three operands: an address operand, an expected value,
+and a relative timeout in milliseconds as an `f64`. The return value is `0`,
+`1`, or `2`, returned as an `i32`.
+
+| `timeout` value | Behavior |
+| ---- | ---- |
+| `timeout` <= 0 | Expires immediately |
+| 0 < `timeout` < Positive infinity | Expires after `timeout` milliseconds |
+| Positive infinity | Never expires |
+| NaN | Never expires |
```
lars-t-hansen: Given how heavyweight wait is, my slight preference would probably be for negative and NaN timeouts to trap, but I do see the significant appeal of making timeouts in the past expire immediately.

NaN though - surely NaN here would indicate an ill-computed timeout and a trap would be as meaningful as would a range error on a heap access?

rossberg-chromium:
@jfbastien, can you motivate why you insist on 0xf0? It seems strictly worse than starting from the end.

Threads.md
jfbastien: NaN is free to handle if you do: current < timeout. It'll be false if timeout is NaN.

jfbastien:
> @jfbastien, can you motivate why you insist on 0xf0? It seems strictly worse than starting from the end.

Because it's fun to match x86, and starting from `0xf0` and counting up is pretty neat for extensions.

Threads.md
lars-t-hansen: The cost of the operation seems like a non-argument for `wait` more than for almost any other functionlity I can imagine wasm having.  I would also expect implementations to convert an FP argument to some integer type and compute with that, so the NaN has to be handled specially in any case.

rossberg-chromium:
@jfbastien, I don't see the relevance of matching x86 -- if anything I'd avoid that for the sake of avoiding bias.

jfbastien:
> @jfbastien, I don't see the relevance of matching x86 -- if anything I'd avoid that for the sake of avoiding bias.

When picking arbitrary numbers, fun becomes the bikeshed differentiator. 🐴

rossberg-chromium:
@jfbastien, minimizing fragmentation in the opcode space > fun ;)

jfbastien:
> @jfbastien, minimizing fragmentation in the opcode space > fun ;)

That's already something that was given up on quite a while ago. Witness `0x06`–`0x0a` and other gaps.

This is a super simple, inconsequential, silly thing. There's goodness in designs having elegance, and there's also goodness in designs having some witticism. People in a few years will look at this silly thing, say "cute, designers knew how to have fun" and smile.

Threads.md
```diff
@@ -0,0 +1,403 @@
...
+
+## Additions to [Semantics.md](Semantics.md)
+
+### Shared Linear Memory
+
+A Linear memory can be marked as shared, which allows it to be shared between
+threads of execution. The shared memory can be imported or defined in the
+module. It is a validation error to attempt to import shared linear
+memory if the module's memory import doesn't specify that it allows shared
+memory.
```
lukewagner: It'd be good to flesh this out a bit more, including [these sorts of details](https://github.com/WebAssembly/design/issues/950#issuecomment-272360540).

jfbastien:
We need to specify:

* What happens with `grow_memory` and `WebAssembly.Memory.prototype.grow`.
* What happens to `instance.export.memory.buffer` that are in JS.
* Whether `current_memory` is sequentially-consistent.

Threads.md
binji: Maybe you're right, I made a gist mapping out the opcode space, and we still have quite a bit of room if we expand these out: https://gist.github.com/binji/0d498bbe16173e7fad18cfb39897e2b0

Threads.md
binji: I actually just copied the behavior from `Atomics.wait` for NaN. :-)

lars-t-hansen: Hrmpf.  Well, fair enough.  In my (feeble) defense the NaN behavior there was chosen to mimic existing APIs, and IMO we should do better, but I'll stop arguing :)

Threads.md
jfbastien: And then every op we encode saves a byte, right?

binji: Yes, although I'd guess that atomic RMW ops are relatively rare so that extra byte is probably not a big deal.

jfbastien: Yeah I agree that's not a big criteria. It just seems weird to have a modifier since we don't do this elsewhere. At that point all atomics could be a single op with a modifier, instead of using prefixes.

Threads.md
```diff
@@ -0,0 +1,403 @@
+# Threads
+
+This page describes a proposal for the post-MVP
+[threads feature][future threads].
```
binji: how about :atom_symbol: (`:atom_symbol:`)

Threads.md
```diff
@@ -0,0 +1,403 @@
...
+  * `i32.atomic.store`: (no conversion) atomically store 4 bytes
+  * `i64.atomic.store8`: wrap i64 to i8 and atomically store 1 byte
+  * `i64.atomic.store16`: wrap i64 to i16 and atomically store 2 bytes
+  * `i64.atomic.store32`: wrap i64 to i32 and atomically store 4 bytes
+  * `i64.atomic.store`: (no conversion) atomically store 8 bytes
+  * `f32.atomic.store`: (no conversion) atomically store 4 bytes
+  * `f64.atomic.store`: (no conversion) atomically store 8 bytes
+
+Atomic read-modify-write (RMW) operators atomically read a value from an
+address , modify the value, and store the resulting value to the same address.
```
binji: Done.

Threads.md
binji: OK, done.

Threads.md
binji: I've added some of the details, PTAL (still no JS.md stuff yet)

binji:
OK, I've added a little bit here about agents/agent clusters, mostly copied from https://tc39.github.io/ecma262/#sec-agents. Not sure how much detail we want to go into here. Gonna write up some JS.md stuff next.

Threads.md
```diff
@@ -31,15 +57,32 @@ Alignment of non-atomic accesses affects performance as follows:
 
 ...
 
-## Additions to [Semantics.md](Semantics.md)
+### Resizing
+
+...
+
+If linear memory is marked as shared, the
+[maximum memory size](Modules.md#linear-memory-section) must be specified.
```
jfbastien: Same restriction for JS.md WebAssembly.Memory.

Threads.md
```diff
@@ -31,15 +57,32 @@ Alignment of non-atomic accesses affects performance as follows:
...
+
+`grow_memory` and `current_memory` have sequentially consistent ordering.
+
+`grow_memory` of a shared linear memory is allowed, but may fail if the new
+size is greater than the specified maximum size, or if reserving the additional
+memory fails. All agents in the executing agent's cluster will have access to
+the additional linear memory.
```
jfbastien: Do we need to discuss neutering?

hemobo:
> @jfbastien, can you motivate why you insist on 0xf0? It seems strictly worse than starting from the end.

Yay, bikeshedding! Count me in!

Reserving the last row of 16 bytes for extension prefixes sounds like a good idea to me. In particular because the potential cost of some future version running out of possibilities for extension seems far higher than the cost of putting some of the more rare prospective one byte opcodes behind a prefix due to lack of space in the lower bytes.

Putting the first extension byte at 0xf0 would be a nice way to mark off that reserved range. Opcode space fragmentation would then be a feature here, not a bug, keeping room for (possibly unanticipated) future extensions.

The only freedom gained by going backwards from 0xff (i.e. defragmentation) seems to be to allow for the possibility of one byte opcodes 'bleeding' into that reserved range, which I think is exactly what one would like to prevent.

Also, extension opcodes could still grow from 0xf0 backwards if at some point in the future it becomes clear that more than 16 prefixes are needed and one byte opcodes haven't filled the lower bytes yet. So one doesn't give up any growth potential for extensions here compared to starting from 0xff and going backwards, the only difference is the explicit reservation of the last row of bytes.

It also feels kind of awkward to spec all the extension opcodes backwards, in a list of operations one would either end up with features running from most obscure to most common, or otherwise with opcodes being listed backwards for extension prefixes and the other way around for the rest... just feels icky :)

rossberg-chromium:
One question that also needs answering is how shared memories interact with instantiation. In particular, what is the amount of atomicity, if any, that is provided for writes performed for data sections?

jfbastien:
> One question that also needs answering is how shared memories interact with instantiation. In particular, what is the amount of atomicity, if any, that is provided for writes performed for data sections?

Good point. Ordering of these as well.

lukewagner:
w.r.t. misaligned atomic accesses trapping:
* Does this match the behavior of any of the major ISAs?  Or do they "succeed" but fail to provide their atomicity guarantees?
* Will trapping atomics end up leading to some of the same double-branching that we've seen with the trapping float operations (where user code branches on alignment to avoid the trap, and then the engine branches again to trap)?
* Each trapping instruction has some small amount of overhead (metadata/code) to record the (semi-)precise trapping bytecode offset (for error reporting).  For codes with millions of heap accesses, this adds up to several MB for the current out-of-bounds heap access checks.  Since separate instructions would check for misalignment vs. out-of-bounds, I expect trapping semantics will incur some amount of size overhead.

jfbastien:
> w.r.t. misaligned atomic accesses trapping:
> * Does this match the behavior of any of the major ISAs?  Or do they "succeed" but fail to provide their atomicity guarantees?

Some ISAs "just work", others tear (sometimes just when crossing cachelines), others silently align, and others trap. It's unclear which of these ISAs we care about.

> * Will trapping atomics end up leading to some of the same double-branching that we've seen with the trapping float operations (where user code branches on alignment to avoid the trap, and then the engine branches again to trap)?

No: there's no "polyfill" possible. having a side-lock for misaligned accesses won't do the right thing.

> * Each trapping instruction has some small amount of overhead (metadata/code) to record the (semi-)precise trapping bytecode offset (for error reporting).  For codes with millions of heap accesses, this adds up to several MB for the current out-of-bounds heap access checks.  Since separate instructions would check for misalignment vs. out-of-bounds, I expect trapping semantics will incur some amount of size overhead.

We'd need to quantify how many atomic instructions actually occur. I expect very few, so low cost. Would be good to confirm.

lukewagner:
> Some ISAs "just work", others tear (sometimes just when crossing cachelines), others silently align, and others trap. It's unclear which of these ISAs we care about.

What about just x86/x64/ARM32/ARM64/MIPS?

> No: there's no "polyfill" possible. having a side-lock for misaligned accesses won't do the right thing.

Well, not a polyfill that provides atomicity, just a path that does the load non-atomically w/o trapping.  Auto-aligning via mask (like asm.js+SAB would do) would be another option.

jfbastien:
>> Some ISAs "just work", others tear (sometimes just when crossing cachelines), others silently align, and others trap. It's unclear which of these ISAs we care about.
>
> What about just x86/x64/ARM32/ARM64/MIPS?

The behavior I described is covered by just x86 and ARM. *Which* ARM is another question. The auto-align and trapping ones aren't the mainstream ones AFAIK.

>> No: there's no "polyfill" possible. having a side-lock for misaligned accesses won't do the right thing.
>
> Well, not a polyfill that provides atomicity, just a path that does the load non-atomically w/o trapping.  Auto-aligning via mask (like asm.js+SAB would do) would be another option.

If the intent is to have atomicity, except the code is OK without atomicity, then... Why even bother with atomicity? The polyfill you describe makes no sense.

We could auto-align, but I don't see how that's useful compared to trapping. Trapping is much more consistent with what WebAssembly otherwise does, and as I mentioned above I think the cost is minimal given how few of these instructions I expect to see.

lukewagner:
Again, I'm not talking about a "polyfill"; I'm wondering whether we'd get into a situation (as we've seen with the trapping conversion ops) where a compiler is forced to keep code "working" (i.e, not trapping) by adding its own branches/masks so that unaligned atomic accesses don't trap.

> We could auto-align, but I don't see how that's useful compared to trapping. 

I'm totally not suggesting that; I'm anti-suggesting that, even.  I'm trying to determine whether it's better for us to not trap since, it sounds like, on all our tier 1 platforms, that's what the ISA gives us and trapping just adds several minor sources of overhead.

jfbastien:
> Again, I'm not talking about a "polyfill"; I'm wondering whether we'd get into a situation (as we've seen with the trapping conversion ops) where a compiler is forced to keep code "working" (i.e, not trapping) by adding its own branches/masks so that unaligned atomic accesses don't trap.

That's what I was calling a polyfill. I don't think it can be made to work.


>> We could auto-align, but I don't see how that's useful compared to trapping. 
>
> I'm totally not suggesting that; I'm anti-suggesting that, even. 

Sure, I'm listing possibilities.

> I'm trying to determine whether it's better for us to not trap since, it sounds like, on all our tier 1 platforms, that's what the ISA gives us and trapping just adds several minor sources of overhead.

I understand what you're asking. I just don't understand which alternative you prefer out of the ones I listed. All our tier 1 platforms *don't* have the same behavior.

lukewagner:
> That's what I was calling a polyfill. I don't think it can be made to work.

It depends what you mean by "work".  If "work" means "loads the bytes as if by a non-atomic load", then sure it can.

> I just don't understand which alternative you prefer out of the ones I listed. All our tier 1 platforms don't have the same behavior.

Ok, maybe I've misunderstood the cases here.  I assumed there are three: trap, auto-align (boo!), do the unaligned load (without atomicity guarantees).  Is that right?  If so, then when you said above "The auto-align and trapping ones aren't the mainstream ones AFAIK", I assumed that meant all the tier 1 platforms are in the last case.

jfbastien:
>> That's what I was calling a polyfill. I don't think it can be made to work.
>
> It depends what you mean by "work".  If "work" means "loads the bytes as if by a non-atomic load", then sure it can.

But then it's non-atomic. Why would a user bother with atomics if their code is OK not being atomic? I'm totally happy making that type of code slower.

>> I just don't understand which alternative you prefer out of the ones I listed. All our tier 1 platforms *don't* have the same behavior.
>
> Ok, maybe I've misunderstood the cases here.  I assumed there are three: trap, auto-align (boo!), do the unaligned load (without atomicity guarantees).  Is that right?  If so, then when you said above "The auto-align and trapping ones aren't the mainstream ones AFAIK", I assumed that meant all the tier 1 platforms are in the last case.

AFAIK x86 "just works" and will be atomic even across cache lines, whereas "mainstream" mobile ARMs will tear. These are "tier 1". Other ARMs (e.g. for embedded) will trap or auto-align.

I agree we should tune for tier 1. We shouldn't ignore tier 2 if it's easy. In this case tier 1 differs, and trapping is cheap enough that I'd go with that because it prevents "I tested on Chrome on x86 and it worked, ship it!".

lukewagner:
> But then it's non-atomic. Why would a user bother with atomics if their code is OK not being atomic? 
> I'm totally happy making that type of code slower.

If the compiler has to blindly work around it (like it had to do with float conversion traps), it would penalize the whole app, not just the dumb code.  Then again, maybe explicit atomics + unaligned bugs are a rare enough combination in practice that this situation wouldn't arise in practice.

> and trapping is cheap enough that I'd go with that because it prevents "I tested on Chrome on x86 
> and it worked, ship it!".

Yeah, I think that is the central tradeoff: making all tier-1 fast paths slower for some amount of bug-catching.  It'd certainly help to have some measurements here.

Threads.md
```diff
@@ -0,0 +1,471 @@
...
+| ---- | ---- |
+| `0` | "ok", woken by another agent in the cluster or the embedder |
+| `1` | "not-equal", the loaded value did not match the expected value |
+| `2` | "timed-out", not woken before timeout expired |
+
+The wait operation begins by performing an atomic load from the given address.
+If the loaded value is not equal to the expected value, the operator returns 1
+("not-equal"). If the values are equal, the agent is suspended. If the agent
+is woken, the wait operator returns 0 ("ok"). If the timeout expires before
+another agent wakes this one, this operator returns 2 ("timed-out").
```
lars-t-hansen: As you know, in the HTML embedding of ES2017 we do not allow Atomics.wait to wait on the main thread of a window (or, in Firefox at present, on the main thread of a SharedWorker).  In ES, a call to Atomics.wait on such a thread results in a TypeError being thrown.  Wasm can be used on the main thread, so is there going to be the same restriction there?

Threads.md
```diff
@@ -0,0 +1,471 @@
...
+The wake and wait operators are optimizations over busy-waiting for a value to
+change. It is a validation error to use these operators on non-shared linear
+memory. The operators have sequentially consistent ordering.
+
+Both wake and wait operators trap if the effective address of either operator
+is misaligned or out-of-bounds.
+
+The embedder is also permitted to suspend or wake an agent. A suspended agent
+can be woken by the embedder or the wake operator, regardless of how the agent
+was suspended (e.g. via the embedder or a wait operator).
```
lars-t-hansen: Is this attempting to formalize the push-into-history / restore-from-history mechanism of browsers, or to allow spurious wakeups (which are not in the ES version of Atomics.wait)?

Threads.md
lukewagner: I never really liked this [AgentCanSuspend](https://github.com/WebAssembly/design/pull/1019) compromise since it seems like it would just lead to spin-locking instead.  That being said, it seems like wasm should be symmetric with JS here, so either we make the restriction too or we try to remove the restriction from JS.  If we make the restriction, it seems like we'd throw a `RuntimeError`.

lars-t-hansen: It will lead to some spin-locking, but those that dabble in that will quickly discover that the browser locks up, as the main thread is not able to service requests made on behalf of other parts of the system.  (Though I sort of doubt that will be apparent to many of the people bitten by this problem.)

jfbastien:
I created a new repo for this: https://github.com/WebAssembly/threads
@binji can you move things there? It'll be easier to iterate.

binji:
Moved here: https://github.com/WebAssembly/threads/pull/1, I'll close this for now. Please move relevant comments there!

Threads.md
binji: Good point, I think we need to be symmetric with JS here.

Threads.md
binji: Oh, it was allowing spurious wakeups, but of course you're right, that's not allowed.

binji: Oh, sorry -- this is meant to allow the thread to be suspended/woken by JavaScript as well as WebAssembly. But it's a good point that it is similar to a spurious wakeup. I wonder if there's a good way to make a distinction.

syg: Can you reference the abstract [WakeWaiter](https://tc39.github.io/ecma262/#sec-wakewaiter) operation? If the wasm `wake` operator uses that, that'll unify wasm with how the embedder can explicitly wake up agents.

PR #1025, opened 2017-03-29
Clarify the behavior of partially out of bounds stores.

sunfishcode:
As requested in https://github.com/WebAssembly/spec/issues/438, this clarifies the behavior of partially out-of-bounds stores in the design.


sunfishcode:
Merging; I believe this is an uncontroversial clarification.

Issue #1024, opened 2017-03-29
WebAssembly.Memory information


jfbastien:
It seems useful and low-cost to add the following functionality to `WebAssembly.Memory`:

* `initial` size
* `maximum` size
* `current` (or `byteLength` to match `ArrayBuffer`?)

It would also be nice to have similar accessors on `WebAssembly.Instance`: an Instance's Memory isn't necessarily exported, and therefore isn't accessible. Introspecting its size would still be nice.

Thoughts? Name bikeshed?

lukewagner:
Adding `initial`, `maximum` and `byteLength` (or just `length`, to match `Table.prototype.length`) to `Memory` makes sense and seems easy.

I'm not so sure about adding properties relating to `Memory` to `Instance`.  For one thing, it feels like it breaks the encapsulation of the module (if the module chose not to export memory); for another it also wouldn't (easily) generalize to multi-memory.

jfbastien:
Ah yes, the multi-memory point is spot on! OK, so I'll whip up a PR soon if nobody has any other suggestions.

kmiller68:
Couldn't we just have each of these properties take a number index for the memory you want? Unless we were going to make them getters, in which case we could return an Array.

jfbastien:
> Couldn't we just have each of these properties take a number index for the memory you want?
 Unless we were going to make them getters, in which case we could return an Array.

Yeah that sounds good, would be similar to `Module.customSections`.

I'm not really worried about the encapsulation argument: the memory can't be modified, it can only be queried from this API. We're not breaking the encapsulation of it, we're just acknowledging that there's a memory in there.

lukewagner:
Hmm, it still doesn't seem right to have `Instance` be arbitrarily exposing these things but not others and using internal indices; that's... what exports are for.

jfbastien:
@lukewagner but what if you want to export everything *about* a memory, but not the memory itself? 🤔

lukewagner:
Haha, that's a great choice of emoji.

hvenev:
Actually, I think it would be nice for `WebAssembly.Module` to provide information about the defined (both exported and not) and imported memories and their initial/maximum sizes. This is useful if using WebAssembly for a sandbox.

rossberg:
The [JS API type reflection proposal](https://github.com/WebAssembly/js-types/blob/master/proposals/js-types/Overview.md) is meant to provide this information in a uniform manner.

[...]

PR #1027, opened 2017-03-30
WebAssembly.Memory: length, initial, maximum

jfbastien:
As suggested in #1024.

I'll follow-up with a similar suggestion for WebAssembly.Instance, which needs its own discussion about encapsulation as well as multiple memories. WebAssembly.Memory is simpler: there's only one Memory and it's not encapsulated.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -561,8 +561,11 @@ Any attempts to [`detach`](http://tc39.github.io/ecma262/#sec-detacharraybuffer)
 the detachment performed by [`m.grow`](#webassemblymemoryprototypegrow) shall throw a 
 [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
 
-Return a new `WebAssembly.Memory` instance with [[Memory]] set to `m` and
-[[BufferObject]] set to `buffer`.
+Return a new `WebAssembly.Memory` instance with
+[[Memory]] set to `m`,
+[[BufferObject]] set to `buffer`,
+[[Initial]] set to `initial`, and
+[[Maximum]] set to `undefined` if `maximum` is `None` or `maximum` otherwise.
 
 ### `WebAssembly.Memory.prototype [ @@toStringTag ]` Property
 
--- a/JS.md
+++ b/JS.md
@@ -619,7 +619,7 @@ accessor function performs the following steps:
 If `this` is not a `WebAssembly.Memory`, a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
 is thrown. Otherwise return `M.[[BufferObject]]`.
 
-### `WebAssembly.Memory.prototype.length`
+### `WebAssembly.Memory.prototype.byteLength`
 
 [⚜](#Post-MVP-JavaScript-API)
 
--- a/JS.md
+++ b/JS.md
@@ -510,11 +510,13 @@ call one with the `new` operator.
 
 A `WebAssembly.Memory` object contains a single [linear memory](Semantics.md#linear-memory)
 which can be simultaneously referenced by multiple `Instance` objects. Each
-`Memory` object has two internal slots:
+`Memory` object has four internal slots:
 
  * [[Memory]] : a [`Memory.memory`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/memory.mli)
  * [[BufferObject]] : the current `ArrayBuffer` whose [[ArrayBufferByteLength]]
    matches the current byte length of [[Memory]]
+ * [[Initial]] : the `initial` number of WebAssembly pages
+ * [[Maximum]] : the `maximum` number of WebAssembly pages, or `undefined` if `maximum` is `None`
 
 ### `WebAssembly.Memory` Constructor
 
@@ -607,6 +609,36 @@ accessor function performs the following steps:
 
 If `this` is not a `WebAssembly.Memory`, a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
 is thrown. Otherwise return `M.[[BufferObject]]`.
+
+### `WebAssembly.Memory.prototype.length`
+
+This is an accessor property whose [[Set]] is Undefined and whose [[Get]]
+accessor function performs the following steps:
+
+Let `T` be the `this` value. If `T` is not a `WebAssembly.Memory`, a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
+is thrown.
+
+Return `T.[[BufferObject]].[[ArrayBufferByteLength]]`.
+
+### `WebAssembly.Memory.prototype.initial`
+
+This is an accessor property whose [[Set]] is Undefined and whose [[Get]]
+accessor function performs the following steps:
+
+Let `T` be the `this` value. If `T` is not a `WebAssembly.Memory`, a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
+is thrown.
+
+Return `T.[[Initial]]`.
+
+### `WebAssembly.Memory.prototype.maximum`
+
+This is an accessor property whose [[Set]] is Undefined and whose [[Get]]
+accessor function performs the following steps:
+
+Let `T` be the `this` value. If `T` is not a `WebAssembly.Memory`, a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
+is thrown.
+
+Return `T.[[Maximum]]`.
 
 ## `WebAssembly.Table` Objects
 
--- a/JS.md
+++ b/JS.md
@@ -7,6 +7,14 @@ be loaded and run directly from an HTML `<script type='module'>` tag—and
 any other Web API that loads ES6 modules via URL—as part of 
 [ES6 Module integration](Modules.md#integration-with-es6-modules).)
 
+## Post-MVP JavaScript API
+
+Some JavaScript APIs were added post-MVP and can be
+[feature-tested](FeatureTest.md) together. These are grouped together,
+and each group is denoted with one of the following symbols:
+
+* ⚜
+
 ## Traps
 
 Whenever WebAssembly semantics specify a [trap](Semantics.md#traps),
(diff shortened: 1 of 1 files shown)


jfbastien:
One thing that remains outstanding is how to identify that these are post-MVP things. We use 🦄 for future features, maybe we could use another marker for post-MVP?

`length` is trivial to polyfill. `initial` and `maximum` aren't, unless one wraps `WebAssembly.Memory` **and** `WebAssembly.Instance` in proxies, as well as all the free functions.

jfbastien:
@domenic 
> This needs to initialize those slots, not just state they exist. (Or, maybe grab them from the internal [[Memory]].

Ah yes! Done.

> I don't think adding a shortcut for `.buffer.length` is a good idea. Objects with `.length` properties in JavaScript are uniformly indexed-access objects. The only reasonable name is `.bufferLength`, but then, why not `.buffer.length`?

That was discussed in #1024. I'm OK with any name that's consistent.

`.buffer.length` means we'd create an `ArrayBuffer` only to throw it away, no? Doesn't hurt much, but seems silly?

domenic:
> .buffer.length means we'd create an ArrayBuffer only to throw it away, no? Doesn't hurt much, but seems silly?

Ah, I didn't realize the ArrayBuffer would be created lazily by implementations. (The spec creates it at creation time, but that's the spec; it's observably equivalent to create it lazily.)

More accurately you'd be creating it, and never throwing it away, since you require `m.buffer === m.buffer`.

If it is indeed anticipated that people will want to know the buffer length but not ever access the buffer itself, then yeah, adding a `bufferByteLength` makes sense. (I realize now it should be `bufferByteLength` instead of `bufferLength` since `buffer` is an ArrayBuffer, not a Uint8Array.)

jfbastien:
> More accurately you'd be creating it, and never throwing it away, since you require `m.buffer === m.buffer`.

Until we detach due to `grow` or the `grow_memory` opcode, I believe that true.

> If it is indeed anticipated that people will want to know the buffer length but not ever access the buffer itself, then yeah, adding a `bufferByteLength` makes sense. (I realize now it should be `bufferByteLength` instead of `bufferLength` since `buffer` is an ArrayBuffer, not a Uint8Array.)

I was writing code to massage some Emscripten output and it would be useful to have. People aren't clamoring for it.

`bufferByteLength` SGTM. Will let @lukewagner chime in.

domenic:
Given that conceptual model, `byteLength` is probably OK; it does make this object kind of take over the territory of ArrayBuffers though. I wonder if it should have been a subclass or something.

jfbastien:
I've renamed to `byteLength` and added markers for post-MVP features similar to what we do for feature testing. WDYT?

rossberg-chromium:
`maximum` and `current`/`length` sound good to me, but I'm less convinced about `initial`. In V8 at least, we don't even remember that information. What would be the use case for "historic" information like that?

As for names, I would prefer either `length` or `currentLength`, and `maximumLength`. That's symmetric, and both those names would be analogously applicable to `Table`, whereas `byteLength` is not.

jfbastien:
> `maximum` and `current`/`length` sound good to me, but I'm less convinced about `initial`. In V8 at least, we don't even remember that information. What would be the use case for "historic" information like that?

It's symmetrical with the `WebAssembly.Memory` constructor, which I agree a user could just remember, but symmetry is nice. It is, however, not accessible for an instance's module's `Memory` section without disassembling (and holding onto the binary).

> As for names, I would prefer either `length` or `currentLength`, and `maximumLength`. That's symmetric, and both those names would be analogously applicable to `Table`, whereas `byteLength` is not.

If we want to be consistent I'd go with one of:

* `current_memory` in units of pages, which is what the binary format does
* `byteLength` which is what `ArrayBuffer` does

Going with the corresponding JS API seems more consistent IMO, but I'm happy painting the bikeshed any length.


`maximumLength` is different from the `WebAssembly.Memory` constructor, and I don't think it should differ.

rossberg-chromium:
@jfbastien, re consistency: the right feature to look at would be neither of the above IMO but the "resizable limits" in the import/export types of memories/tables, because that's pretty much exactly what the current/maximum methods would be reflecting on a memory/table object and what is relevant for linking.

Fair point about the different units, but that doesn't need to affect naming. The constructors and `grow` methods of tables/memories already use different units in the same way.

Not convinced by the symmetry argument regarding `initial`. It is a very strange method to have IMO -- you cannot ask a JS array what size it was created with either. Furthermore, the constructor does not take a current length, so I'm not even sure what consistency would be established there. (Maybe it was a mistake to name the constructor param `initial` instead of just `length`, but that should not be the sole motivation for adding weird API.)

flagxor:
I agree with @rossberg-chromium that including `initial` is odd. This seems a pretty unusual thing to track for a memory buffer.
I'm indifferent on naming choice of byteLength / length but buy the utility of having a way to check size that doesn't require creating an ArrayBuffer.

I think a way to access the initial and maximum size of a memory import on a Modle is definitely useful too (possibly more so). Should we consider that?

jfbastien:
> I agree with @rossberg-chromium that including `initial` is odd. This seems a pretty unusual thing to track for a memory buffer.

Yeah let's do it as part of #1046 in Module. Does this sound good?

> I'm indifferent on naming choice of byteLength / length but buy the utility of having a way to check size that doesn't require creating an ArrayBuffer.

Same. We're inconsistent with one thing or another.

> I think a way to access the initial and maximum size of a memory import on a Modle is definitely useful too (possibly more so). Should we consider that?

Yes, #1046.

jfbastien:
Updated, PTAL.

JS.md
```diff
@@ -608,6 +619,30 @@ accessor function performs the following steps:
 If `this` is not a `WebAssembly.Memory`, a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
 is thrown. Otherwise return `M.[[BufferObject]]`.
 
+### `WebAssembly.Memory.prototype.length`
```
domenic: I'm still quite opposed to `length` for things without indexed properties 0, 1, 2, ..., length - 1, in JavaScript APIs. If you're indifferent between length and byteLength, please pick byteLength :)

jfbastien: Sounds good. That's what I had but @rossberg-chromium disliked it. I'll put it back!

rossberg-chromium:
@domenic, we need a name that works consistently for both memories and tables, and unfortunately `byteLength` does not.

@jfbastien, how about `size`?

jfbastien:
> @domenic, we need a name that works consistently for both memories and tables, and unfortunately `byteLength` does not.

Why do we need this property?

> @jfbastien, how about `size`?

If we indeed need that property then `size` is as good as any.

rossberg-chromium:
@jfbastien, because tables share the exact same limits properties, so should expose them in a consistent manner. If `length` is verboten, then we'll have to find another name for it for tables anyway, so could just as well use that for memories, too.

jfbastien:
> @jfbastien, because tables share the exact same limits properties, so should expose them in a consistent manner.

Can you clarify what limit property you mean? I don't understand what the similarity is.

rossberg-chromium:
Tables can grow in the same way as memories, and they also have an initial size, a current size, and an optional maximum size (see e.g. the use of `resizable_limits` for both in the design docs). The only difference is that these sizes are measured in number of elements instead of pages.

jfbastien:
> Tables can grow in the same way as memories, and they also have an initial size, a current size, and an optional maximum size (see e.g. the use of `resizable_limits` for both in the design docs). The only difference is that these sizes are measured in number of elements instead of pages.

Ah OK, so IIUC your point is: presence of initial and maximum plus growth, regardless of unit, is what should guide naming. Unit is irrelevant.

Is this accurate?

Two questions I think will help resolve this:

1. Is this something we've done in the WebAssembly JS API elsewhere (either what you suggest or the opposite)?
2. What do other web platform APIs do in these circumstances?

rossberg-chromium:
Well, in the API, we already have the same symmetry in the Table vs Memory constructor arguments (both taking an object with `initial` and `maximum` properties. And it is present throughout the Wasm design and semantics, so I would expect that to be reflected accordingly.

jfbastien:
> Well, in the API, we already have the same symmetry in the Table vs Memory constructor arguments (both taking an object with `initial` and `maximum` properties. And it is present throughout the Wasm design and semantics, so I would expect that to be reflected accordingly.

True. Your argument is then: we've ignored unit before, and we should therefore continue ignoring it. Right?

Seems sensible.

@domenic thoughts?

lukewagner:
Table already has `Table.prototype.length` so, if we want to be symmetric (which I like), we should have `Memory.prototype.length` and, to be symmetric with `initial`, `maximum` and `grow`, this must return the current number of *pages*.  But of course it's quite useful to have number of bytes so I think it would be reasonable to *additionally* have `Memory.prototype.byteLength` that saved the user the `*64*1024` or `.buffer`.

domenic:
Oh... if we've already decided to break the rule on .length only being for indexed objects, then I guess consistency argues we should continue to do so... it's a shame that wasn't caught earlier :(

JS.md
```diff
@@ -559,8 +568,10 @@ Any attempts to [`detach`](http://tc39.github.io/ecma262/#sec-detacharraybuffer)
 the detachment performed by [`m.grow`](#webassemblymemoryprototypegrow) shall throw a 
 [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
 
-Return a new `WebAssembly.Memory` instance with [[Memory]] set to `m` and
-[[BufferObject]] set to `buffer`.
+Return a new `WebAssembly.Memory` instance with
+[[Memory]] set to `m`,
+[[BufferObject]] set to `buffer`, and
+[[Maximum]] set to `undefined` if `maximum` is `None` or `maximum` otherwise.
```
littledan: What's the motivation for storing the maximum here, rather than looking it up in the store?

JS.md
```diff
@@ -608,6 +619,30 @@ accessor function performs the following steps:
...
+
+[⚜](#Post-MVP-JavaScript-API)
+
+This is an accessor property whose [[Set]] is Undefined and whose [[Get]]
+accessor function performs the following steps:
+
+Let `T` be the `this` value. If `T` is not a `WebAssembly.Memory`, a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
+is thrown.
+
+Return `T.[[BufferObject]].[[ArrayBufferByteLength]]`.
```
littledan: What is the motivation for this feature, when a user can do `memory.buffer.byteLength`? Is it more optimizable if you never access the ArrayBuffer, or something like that?

jfbastien:
> What's the current state of this proposal? Should I add it to the JS API specification?

That would be great, though I'd advise bringing it up at a CG meeting (either in-person or VC).

JS.md
jfbastien: Yeah creating the array buffer is expensive-ish. We can optimize it away, but that's being extra silly IMO.

JS.md
jfbastien: What do you mean "in the store"?

rossberg: He means that the memory instance `m` (living in the Wasm abstract store) already has this information, so the internal slot is redundant.

rossberg:
We deliberately chose to make this post-MVP, so we should not rush it into the MVP spec now all of a sudden. There is a bit of a design space around how to provide this information and fit it nicely with #1046, which would expose equivalent information in other places. I think these features need to be designed together.

sunfishcode:
If anyone wishes to pursue this feature, please make a new proposal to the CG.

Issue #970, opened 2017-01-30
UTF-8 decoding of import/export names in JS


rossberg-chromium:
Related to #968, I noticed that https://github.com/WebAssembly/design/blob/master/Web.md#names says:

> Property names in JS are UTF-16 encoded strings. A WebAssembly module may fail validation on the Web if it imports or exports functions whose names do not transcode cleanly to UTF-16 according to the following conversion algorithm

There are at least two problems with this.

The first sentence is simply incorrect. JS property names can be arbitrary strings, and JS strings are arbitrary sequences of (unsigned) 16 bit values. They can happily contain zeros or lonely halves of what would be surrogate pairs. The only relevance of UTF-16 is that some ES6+ library functions assume UTF-16 inputs.

The second statement seems rather problematic. String formats are not prescribed or restricted by Wasm, validation explicitly allows any sequence. Hence I think we must not allow modules with malformed encodings to fail validation, no matter what the platform.

I see several options for resolving the latter (in decreasing order of preference):

1. Do not perform UTF-8 decoding (UTF-8 to UTF-16 transcoding, really) in the JS API. Instead, simply treat the strings as a sequence of (unsigned) 8 bit values extended to 16 bit code points pointwise.

2. Do not throw on invalid UTF-8 encodings. Import/export names that do not decode are merely inaccessible from JS.

3. Change the wording such that this is not described as a validation failure but an instantiation failure in the JS API. In particular, it does not cause `W.validate` to return false.

Kardax:
1. This would limit names to the Basic Latin (aka ASCII) and Latin-1 Supplement blocks.  Seems like an unnecessary restriction to me.  It disallows interesting scenarios that could leverage higher code points while still accepting code points not legal in names with many mainstream languages.
2. I like this.  Might be worth a single debug console warning that reports that one or more unusable names are present, but that doesn't have to be part of the spec.
3. I suppose this depends on the intended use of `validate`.  If it's to ensure spec conformance, I agree, but it doesn't provide a reliable way to check if instantiation will actually work before trying.  In the future, this could put an unnecessary burden on JS implementors if WASM gains features not relevant to JS.

rossberg-chromium:
[...]

PR #1028, opened 2017-03-30
Clarify import/export identifier validation on the Web.

sunfishcode:
https://github.com/WebAssembly/design/issues/970 observed that the mapping of WebAssembly import/export names into JS wasn't specified clearly. This PR clarifies and updates the relevant paragraph in light of https://github.com/WebAssembly/design/pull/1016 and other changes.

### DIFF
--- a/Web.md
+++ b/Web.md
@@ -130,14 +130,11 @@ the ES6 module system](Modules.md#integration-with-es6-modules).
 
 ### Names
 
-A WebAssembly module imports and exports functions. WebAssembly names functions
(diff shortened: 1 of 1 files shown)


rossberg-chromium:
Text sounds good to me, but perhaps this should in fact be moved to JS.md.

sunfishcode:
Merging, as this is an improvement over the status quo.

I agree that this text feels like it probably should be in JS.md rather than Web.md, but I don't know whether there's a reason it wasn't done this way originally.

PR #1045, opened 2017-04-19
Fix markup of algorithms

rossberg-chromium:
I don't know what happened, but it seems like GH markup rules changed and sub enumerations must now be indented by 3 spaces to be recognised as such. This PR fixes the layout of the algorithms accordingly.


lukewagner:
Thanks, lgtm

PR #1054, opened 2017-05-04
Throw `TypeError` on incorrect MIME type

mtrofin:
Proposing that we throw `TypeError` when the Response parameter has incorrect MIME type. This seems more in-line to what we do for:

`WebAssembly.{compile|instantiate}("hello!")`

### DIFF
--- a/Web.md
+++ b/Web.md
@@ -69,7 +69,7 @@ from the `Response`, WebAssembly `source` data must have a MIME type of `applica
 extra parameters are not allowed (including empty `application/wasm;`).
 MIME type mismatch or `opaque` response types
 [reject](http://tc39.github.io/ecma262/#sec-rejectpromise) the Promise with a
-`WebAssembly.CompileError`.
+`TypeError`.
 
 #### `WebAssembly.instantiate`
 


jfbastien:
What do other Response APIs usually do?

mtrofin:
@jfbastien I'm not sure there are other similar APIs (i.e. that take a Response object as parameter). Do you have an example in mind?

In absence of precedent, does the proposal here seem reasonable?

annevk:
This looks reasonable to me, FWIW. It does clash with #1060 a bit but I can rebase (or fold this in there).

jfbastien:
> This looks reasonable to me, FWIW. It does clash with #1060 a bit but I can rebase (or fold this in there).

Yeah it seems like differentiating actual compilation error from "your request was bad" or "ethernet cable was bent" or "you so teapot" is useful. #1060 adds another such throw condition which should probably be a `TypeError`.

So if that's the usual behavior for Response APIs then LGTM.

rossberg-chromium:
Agreed this is better. `CompileError` is supposed to specifically indicate a Wasm validation failure.

annevk:
@mtrofin this patch wasn't quite correct and you missed updating various places. I tried to account for them in an update to #1060.

Issue #977, opened 2017-02-02
Do we need a memcpy opcode?


juj:
In a number of game engines I've been optimizing and benchmarking, interestingly the performance of memcpy() does show up relatively high in profiles. (~2%-5% of total execution time)

I did a recent optimization to memcpy in Emscripten, and added some benchmarks. See https://github.com/kripken/emscripten/pull/4127 (especially http://clb.demon.fi/emcc/memcpy_test/results.html)

The result there was that asm.js+SIMD memcpys are at worst ~3x slower than native at aligned copy sizes around 4k - 16k bytes, and Wasm memcpys are at worst ~6x slower than native at aligned copy sizes around the same size.

It is not immediately obvious to me how much we can close this gap e.g. by further optimizing the memcpy implementation in Emscripten, and optimizing Wasm implementations in browsers, which raises the question of whether it would make sense to have a memcpy (and probably as extension, memset) opcode directly in wasm?

Currently in Emscripten very large memcpys round trip to HEAP8.set(), but it has a drawback of generating temporary garbage.

juj:
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1254573

lukewagner:
(Copying over from the bugzilla comment) I think it would make sense to add a page-unit `mempcy` operator to our [finer-grained control over memory list](https://github.com/WebAssembly/design/blob/master/FutureFeatures.md#finer-grained-control-over-memory).  Having page units means wasm can assume the caller is doing all the small-memcpy() optimizations (inlining, unrolling, peeling, etc) and then simply call out to the C++ `memcpy()` (which presumably then does all the fancy stuff).

jfbastien:
Dup #236 ?

lukewagner:
Hah, 2015... seems so long ago.  (It'd be nice to get it listed in FutureFeatures.md, though.)

PR #1057, opened 2017-05-09
Add Memset+Memcpy to FutureFeatures.md

binji:
See #236 and #977. I added both `set_memory` and `zero_memory` since there was discussion about this in #236.

### DIFF
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -572,8 +572,8 @@ performance:
 
 * `move_memory`: Copy data from a source memory region to destination region;
    these regions may overlap: the copy is performed as if the source region was 
-   first copied to a temporary buffer, then copied from this temporary buffer 
-   to the destination region
+   first copied to a temporary buffer, then the temporary buffer is copied to
+   the destination region
 * `set_memory`: Set all bytes in a memory region to a given byte
 
 We expect that WebAssembly producers will use these operations when the region
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -570,8 +570,10 @@ via `i32.load` and `i32.store`, this requires more bytes of code and forces VMs
 to recognize the loops as well. The following operators can be added to improve
 performance:
 
-* `move_memory`: Copy data from one memory region to another region, copying backward if the regions overlap
-* `zero_memory`: Set all bytes in a memory region to zero
+* `move_memory`: Copy data from a source memory region to destination region;
+   these regions may overlap: the copy is performed as if the source region was 
+   first copied to a temporary buffer, then copied from this temporary buffer 
+   to the destination region
 * `set_memory`: Set all bytes in a memory region to a given byte
 
 We expect that WebAssembly producers will use these operations when the region
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -578,3 +578,5 @@ performance:
 
 We expect that WebAssembly producers will use these operations when the region
 size is known to be large, and will use loads/stores otherwise.
+
+TODO: determine how these operations interact w/ shared memory.
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -561,3 +561,13 @@ static signature validation check. This could be improved by allowing:
   check of a call to a heterogeneous table;
 * any other specific GC reference type, effectively allowing WebAssembly code
   to implement a variety of rooting API schemes.
+
+### Memset and Memcpy Operators
+
+Copying and clearing large memory regions is very common. This can be done in
+the MVP via `i32.load` and `i32.store`, but this is not very fast. The
+following operators can be added to improve performance:
+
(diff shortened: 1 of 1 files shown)


FutureFeatures.md
```diff
@@ -561,3 +561,13 @@ static signature validation check. This could be improved by allowing:
...
+
+### Memset and Memcpy Operators
+
+Copying and clearing large memory regions is very common. This can be done in
+the MVP via `i32.load` and `i32.store`, but this is not very fast. The
+following operators can be added to improve performance:
+
+* `move_memory`: Copy data from one memory region to another region, even if overlapping
+* `zero_memory`: Set all bytes in a memory region to zero
+* `set_memory`: Set all bytes in a memory region to a given byte
```
jfbastien: Id it useful to distinguish zero?

FutureFeatures.md
```diff
@@ -561,3 +561,13 @@ static signature validation check. This could be improved by allowing:
   check of a call to a heterogeneous table;
 * any other specific GC reference type, effectively allowing WebAssembly code
   to implement a variety of rooting API schemes.
+
+### Memset and Memcpy Operators
+
+Copying and clearing large memory regions is very common. This can be done in
+the MVP via `i32.load` and `i32.store`, but this is not very fast. The
```
jfbastien: I'm not sure the speed argument is well stated: a fast `mem*` is dependent on architecture, and the load/store approach requires more bytecode while forcing VMs to recognize the loop.

Overall the reasons to have `mem*` aren't huge, but they seem at least worth considering.

I'd also mention that some things are expected from dev-side compilers such as LLVM. I don't expect to see small `mem*` operations in WebAssembly.

FutureFeatures.md
```diff
@@ -561,3 +561,13 @@ static signature validation check. This could be improved by allowing:
...
+following operators can be added to improve performance:
+
+* `move_memory`: Copy data from one memory region to another region, even if overlapping
```
jfbastien: "even if" doesn't state what happens if it is overlapping.

FutureFeatures.md
binji: Not sure, but I figured I'd include both since both were discussed.

FutureFeatures.md
binji: Done.

FutureFeatures.md
binji: Done.

RyanLamansky:
Would there be two variants of these: one which takes the size as an immediate, and another from the stack?  `move_memory` is likely to be used for moving constant-sized data structures around.  Loading a constant from the stack can be expected to be optimized by compilers, but it would save a byte to have its own opcode.

binji:
We haven't done that sort of thing with any other operators, so it feels a bit out of place to start doing it now. I'm guessing this wouldn't be a very common opcode anyway, so the byte savings may not make much of a difference.

FutureFeatures.md
```diff
@@ -561,3 +561,18 @@ static signature validation check. This could be improved by allowing:
...
+
+### Memset and Memcpy Operators
+
+Copying and clearing large memory regions is very common, and making these
+operations fast is architecture dependent. Although this can be done in the MVP
+via `i32.load` and `i32.store`, this requires more bytes of code and forces VMs
+to recognize the loops as well. The following operators can be added to improve
+performance:
+
+* `move_memory`: Copy data from one memory region to another region, copying backward if the regions overlap
```
lukewagner: Instead of saying "copying backward", I think we should say what memmove() docs say which is that it works as-if a temporary buffer was used.  (If there is an overlap and dst > src, that does entail copying backward, but if src > dst, then you want to copy forward.)

lukewagner:
@RyanLamansky We're expecting compilers to have optimized small/constant-sized memcpy/set (into unrolled loads/stores), saving the builtin operators only for the bulk operations.

FutureFeatures.md
```diff
@@ -561,3 +561,18 @@ static signature validation check. This could be improved by allowing:
...
+* `zero_memory`: Set all bytes in a memory region to zero
+* `set_memory`: Set all bytes in a memory region to a given byte
```
lukewagner: If we have `set_memory` (which I think we should), then I don't think there's much reason for a specialized zero form; we're all just going to be calling `memset()` anyhow.

lukewagner:
Thanks, lgtm

jfbastien:
I'm not sure this is good though... how it's implemented has observable effects with shared memory. There should be a note that we need to fix this, then lgtm.

lukewagner:
Since these operators would be performing a bunch of non-atomic reads and writes, it doesn't seem like specifying the order in more detail would allow one to write a correct program any more than specifying as if a temporary copy was made.

binji:
Well, this is for `FutureFeatures.md`, not an actual proposal just yet. So saying we're not sure here makes sense, I suppose.

Issue #903, opened 2016-12-13
OOM errors


jfbastien:
Forking from https://github.com/WebAssembly/design/pull/901#issuecomment-266523709

Should we call out out-of-memory errors separately, or do they cause Compile / Link errors? This seems similar to @gahaas' clarification w.r.t. stack error in #875.

flagxor:
We should do:
GrowMemory stays as is.
Everything else is up to the embedder, see Andreas' other verbiage.

jfbastien:
As a clarification: I think OOM on `WebAssembly.validate` should be an OOM exception, not a validation failure: we may never be able to compile the module, but it's not fundamentally invalid. Pretty much the only valid usage of `validate` is to do feature detection, it would be odd to "lie" on OOM.

Again, this is a corner-case.

titzer:
What JF says sounds reasonable to me.

lukewagner:
Agreed; that also matches what we've done.

sunfishcode:
This was fixed in #1059.

PR #1059, opened 2017-05-11
OOM

jfbastien:
Addresses #903.


jfbastien:
I guess I'd forgotten to merge this. Had discussed it with other folks and we had agreement.

Issue #1048, opened 2017-04-26
Formal description of serializing and deserializing a Module 


annevk:
I realize there's #956 and such, but currently I can't find any description of how Module integrates with the HTML Standard's StructuredSerialize and StructuredDeserialize algorithms and therefore how it's supposed to work with `postMessage()` and friends. Is this being worked on or am I not looking in the right place?

domenic:
I hear Memory also needs to work, as a thin wrapper around a SAB.

lukewagner:
This isn't being worked on afaik; we just have a [basic description in JS.md](https://github.com/WebAssembly/design/blob/master/JS.md#structured-clone-of-a-webassemblymodule).

jfbastien:
It would be great to improve this. I think @mtrofin wanted to help there.

annevk:
So I think you'll be defining your API in terms of the ECMAScript standard, not in terms of the IDL standard, right? Because if you don't define it in IDL, you'll have to effectively change the HTML Standard and define how your object is serialized and deserialized inline. (There's no extension hooks for objects defined in terms of the ECMAScript standard, only for objects defined in terms of the IDL standard.)

Either way, I'm happy to help out and I think @domenic will be too. Having a formal description might also make the security story easier to tackle.

mtrofin:
Re. @jfbastien's comment - while I am supportive of more rigor being brought into the API spec, unfortunately I have no expertise, and therefore, no plan, to do anything there.

jfbastien:
Assigning to @annevk.

annevk:
So I'm happy to help with this and draft some text, but someone first needs to explain if these new objects are going to be implemented as IDL objects or something directly added to the JavaScript engine (akin to https://streams.spec.whatwg.org/) as that has implications.

jfbastien:
What are the pros and cons of implementing as IDL objects versus direct implementation in JS engines?

annevk:
There are some subtle differences as outlined in https://github.com/heycam/webidl/issues/226. I suspect not using IDL might also mean more work for the eventual WebAssembly translation (if these objects are also meant to be exposed there at some point)?

In general using IDL means a lot of the argument handling is taken care of by the binding, which adds some robustness, but it also comes with the subtle differences which not everyone appreciates.

Having said all that, given that you're already taking IDL-defined objects as arguments (`Response` for instance) you might as well go down that path I think, but it's a decision everyone needs to agree on.

domenic:
[...]

PR #1060, opened 2017-05-12
Require an ok status for a response

### LINKED ISSUE #1039: Require an ok status for instantiate responses?
It seems that may you want to require response's status to be an https://fetch.spec.whatwg.org/#ok-status to avoid compiling a 404 or 500 as WebAssembly.

annevk:
This changes both WebAssembly.compile and WebAssembly.instantiate to be
both more clear about response requirements and require an ok status.

Fixes #1039.

### DIFF
--- a/Web.md
+++ b/Web.md
@@ -60,7 +60,7 @@ as described in the [`WebAssembly.Module` constructor](#webassemblymodule-constr
 On success, the `Promise` is [fulfilled](http://tc39.github.io/ecma262/#sec-fulfillpromise)
 with the resulting `WebAssembly.Module` object. On failure, the `Promise` is
 [rejected](http://tc39.github.io/ecma262/#sec-rejectpromise) with a
-`TypeError`.
+`WebAssembly.CompileError` or `TypeError`, depending on the type of failure.
 
 The `Promise<Response>` is used as the source of the bytes to compile.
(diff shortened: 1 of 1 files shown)


annevk:
Once there's a more formal specification we should deduplicate this, of course.

Web.md
```diff
@@ -60,15 +60,18 @@ as described in the [`WebAssembly.Module` constructor](#webassemblymodule-constr
 On success, the `Promise` is [fulfilled](http://tc39.github.io/ecma262/#sec-fulfillpromise)
 with the resulting `WebAssembly.Module` object. On failure, the `Promise` is
 [rejected](http://tc39.github.io/ecma262/#sec-rejectpromise) with a
-`WebAssembly.CompileError`.
+`WebAssembly.CompileError` or `TypeError`, depending on the type of failure.
```
mtrofin: Nit: should we say "depending on the type of failure, as it will be clarified further below"

... or something to that extent?

annevk: I went with what we had for instantiate, but happy to change both in this manner if others agree. Eventually we need to redo both these descriptions in a more formal matter once you all figure out how you want to define your objects (see #1048).

annevk:
Anyone else have any feedback?

Issue #960, opened 2017-01-18
why not java or .net?


kovalenko0:
Why not just run JVM or .net bytecode in browser?

In that case we could benefit from infrastructure of these platforms and already implemented interpreters

krux02:
I am new to WebAssembly too. But I guess this has already been been asked several times. Maybe it should be in the FAQ.

My guess is that it is because .net and jvm are not only bytecode, they are a virtual machine with their own garbage collection and class loader and reflection. WebAssembly is really just about low level operations platform independent in the browser. I think you could run a JVM in WebAssembly, but that would not give a result you would want to have. It would simply add the startup time of WebAssembly and the startuptime of the JVM. And you would have two different garbage collected heaps. The one from JavaScript, and the one from your new JVM. I think currently you are better with GWT.

Kardax:
It's more accurate to think of Web Assembly as an evolution of asm.js, rather than a completely new technology.  The only capability it really has over asm.js is 64-bit integers, which aren't even needed for most programs.  This limited scope allows it to be integrated with relative ease into existing JavaScript engines.

One of the veterans from the Web Assembly project will probably have a more complete answer, and I agree this would be a good FAQ item as this is likely to come up often when 1.0 ships.

distransient:
Jvm and .Net bytecodes are specialized for their relevant virtual machines, each of which have their own concerns and problem spaces they're trying to solve separate of the virtual machines ran in web browsers. Adapting an entire second virtual machine to run on web browsers would be cumbersome for users, and adapting Jvm and .Net vm apis to run within existing virtual machines would be implementationally cumbersome, and inevitably cumbersome for users, and possibly more error/bug prone as complexity grows. Web Assembly is designed as a simpler ISA that can easily be ran in the existing Javascript virtual machines already in browsers, with a focus on optimizing transport-time and load-time, a set of requirements existing bytecodes don't fulfill adequately.

jfbastien:
I was sure we had something about this in the FAQ or Rationale documents. We don't. It would be a great addition.

wllang:
[...]

PR #1061, opened 2017-05-12
Add FAQ on java and .net virtual machines

distransient:
Figured I'd have a go at #960.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -242,13 +242,13 @@ generate WebAssembly code from C and C++, and will use LLVM IR in their
 implementation similarly to how PNaCl and Emscripten do today.
 
 
-## Why not use the .net or jvm virtual machines and bytecodes?
+## Why not use the .NET or Java virtual machines and bytecodes?
 
 Simply, web browsers already have their own virtual machines for executing 
-platform-independent code, and integrating the existing infrastructure of .net 
-or the jvm into browser environments on top of these existing systems would 
-incur massive costs for end-users. Also, the design goals in jvm and .net's 
-runtimes and bytecodes are far from aligning with those of this project, 
+platform-independent code, and integrating the existing infrastructure of .NET 
+or the JVM into browser environments on top of these existing systems would 
+incur serious runtime costs for end-users. Also, the design goals in JVM and 
+.NET's runtimes and bytecodes are far from aligning with those of this project, 
 similar to the incongruencies in the above section on llvm bytecode.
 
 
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -242,6 +242,16 @@ generate WebAssembly code from C and C++, and will use LLVM IR in their
 implementation similarly to how PNaCl and Emscripten do today.
 
 
+## Why not use the .net or jvm virtual machines and bytecodes?
```
RyanLamansky: .NET should be all caps.

JVM virtual machines is redundant (java virtual machine virtual machines).  I suggest changing "jvm" to "Java".

FAQ.md
```diff
@@ -242,6 +242,16 @@ generate WebAssembly code from C and C++, and will use LLVM IR in their
...
+
+Simply, web browsers already have their own virtual machines for executing 
+platform-independent code, and integrating the existing infrastructure of .net 
+or the jvm into browser environments on top of these existing systems would 
+incur massive costs for end-users. Also, the design goals in jvm and .net's
```
RyanLamansky: "costs" could be misinterpreted as a monetary expense, which although being a legitimate fear given the history of both Java and .NET, would be unlikely.  Regardless, I'd like to see a less ambiguous word or phrase here, but not sure what it would be...

Also, JVM and .NET should be all caps.

chicoxyzzy: LLVM too

distransient:
Thanks @RyanLamansky, @chicoxyzzy; updated.

jfbastien:
One thing that's worth mentioning is the abstraction layer that WebAssembly operates at versus other VMs. The bytecode doesn't know about higher-level language constructs, and the hope is that these can be layered on top of a WebAssembly VM.

Mentioning that we already have JS VMs isn't super relevant: in all cases we don't reuse JS-isms in our bytecodes, we fast-forward to our static-typed optimizers.

rossberg-chromium:
Agreed with @jfbastien, existing JS VMs are not particularly relevant. The real reason is that other VMs and IRs do not meet the design goals defined for Wasm. You can find relevant discussion in our paper (Sections 1 and 8 in particular):

https://github.com/WebAssembly/spec/blob/master/papers/pldi2017.pdf

distransient:
Thanks for the help, I'll work to adjust the wording a bit.

jaredthirsk:
FWIW somebody got an .NET interpreter running on WebAssembly as an experiment: https://github.com/SteveSanderson/Blazor

distransient:
Hey, was just going through my open pull requests, is this outdated? I could make a patch for the current HEAD with any necessary changes if so. Otherwise we could close it.

distransient:
Closing due to age and lack of activity

Issue #1007, opened 2017-03-06
should Instance.exports be a getter on Instance.prototype instead of data property?


lukewagner:
Scanning the JS spec, it seems pretty irregular to have the `exports` property of instances be defined as an own data property directly on the `Instance` objects as opposed to a getter on `Instance.prototype` that reads an internal [[Exports]] property of `Instance` objects.  I'd be pretty surprised if making this change broke any code, so it seems like a minor tweak we could make after MVP shipping.

I also have a vague impression that having an internal immutable [[Exports]] property might be useful in the future if any JS API needed to find the exports object of an instance.

rossberg-chromium:
I thought the idea of the current design is to mirror a module namespace object. That also carries the exports as own properties (https://tc39.github.io/ecma262/#sec-module-namespace-exotic-objects).

jfbastien:
> I thought the idea of the current design is to mirror a module namespace object. That also carries the exports as own properties (https://tc39.github.io/ecma262/#sec-module-namespace-exotic-objects).

That's my understanding as well, and is exactly what we've implemented in JSC. On instance creation we literally do:

```cpp
putDirect(vm, Identifier::fromString(&vm, "exports"), moduleNamespaceObject, None);
```

I thought we'd try to move *closer* to JS modules, not further.

lukewagner:
@rossberg-chromium I think you're referring to the exports object; I'm talking about the `exports` *property* on `Instance` objects.  I didn't know there was an analogue to JS for `exports` property, but certainly happy to follow it if there is.

jfbastien:
@lukewagner do you think this can move forward with a PR? (and can you do that PR :wink:)

domenic:
Note that data properties are not possible in Web IDL, so if there is an idea to expose these APIs to wasm eventually via some Web IDL binding, the data property design makes it impossible for wasm to see `Instance` objects.

lukewagner:
Thanks @domenic!  That sounds like the strong reason to make it a getter that I was lacking.  I'm assuming the symmetry-with-module-objects concern is addressed by my previous comment, so I'll write up a PR in a bit.

rossberg-chromium:
@lukewagner, yes, thanks for clarifying.

lukewagner:
Resolved by #1062.

PR #1062, opened 2017-05-12
Make Instance.exports a getter

lukewagner:
As discussed in #1007.

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -283,6 +283,7 @@ internal slot:
 
 * [[Instance]] : an [`Instance.instance`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/instance.ml#L17)
   which is the WebAssembly spec definition of an instance
+ * [[Exports]] : the exports object created during instantiation
 
 ### `WebAssembly.Instance` Constructor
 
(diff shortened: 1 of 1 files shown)


saambarati:
Do we want to specify configurable/enumerable explicitly? Or do we want to use ECMA's defined values?

domenic:
It seems like the spec is assuming some fallback, either to the ES-spec defined values, or the Web IDL-spec defined ones. It's not clear which. But for example no methods have their property descriptors specified, so I am assuming there is some fallback.

saambarati:
@domenic The defaults are specified in tc39. I'm just wondering if we want to use the defaults or not.

domenic:
Yeah, one set of defaults is specified there, another is specified in Web IDL. I too am not sure which defaults WebAssembly wants to use.

saambarati:
Gotcha. I didn’t realize Web IDL had different defaults. I think it’s a good idea to define what we want in Wasm in the spec.

jfbastien:
Right, all accessors seem to not specify a default.

jfbastien:
This is unclear, needs discussion. Moving to #1069.

PR #1064, opened 2017-05-15
Expand binary name section to include type, table, memory, global, and label names.

### LINKED ISSUE #750: Include types, tables, labels, globals, and memories in binary format name section

AndrewScheidecker:
This is meant to resolve #750.

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -457,6 +457,11 @@ be skipped over by an engine. The current list of valid `name_type` codes are:
 | [Module](#module-name)      | `0`  | Assigns a name to the module         |
 | [Function](#function-names) | `1`  | Assigns names to functions           |
 | [Local](#local-names)       | `2`  | Assigns names to locals in functions |
+| [Labels](#label-names)      | `3`  | Assigns names to labels in functions |
+| [Type](#type-names)         | `4`  | Assigns names to types               |
+| [Table](#table-names)       | `5`  | Assigns names to tables              |
+| [Memory](#memory-names)     | `6`  | Assigns names to memories            |
+| [Global](#global-names)     | `7`  | Assigns names to globals             |
 
 
 When present, subsections must appear in this order and at most once. The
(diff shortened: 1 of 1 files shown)


BinaryEncoding.md
```diff
@@ -500,6 +504,47 @@ where a `local_name` is encoded as:
 | index | `varuint32` | the index of the function whose locals are being named |
 | local_map | `name_map` | assignment of names to local indices |
 
+#### Label names
+
+The label names subsection assigns `name_map`s to a subset of functions in the
+[function index space](Modules.md#function-index-space). This may include both
+module-defined or imported functions, but is only meaningful for module-defined
+functions. The `name_map` for a given function assigns names to labels within
+the function, with labels identified by the offset from the start of the function's
```
rossberg-chromium: Hm, using raw byte offsets pierces the abstraction level on which the rest of the binary format is defined and makes the mapping a bit less robust. I wonder if there is an alternative. For example, couldn't we just assume an implicit numbering of a function's structured control instructions in order of appearance?

AndrewScheidecker: I considered an implicit numbering of the control flow operators first, but the benefit to using a byte offset is that it doesn't require tracking additional state when decoding the instructions to compute some implicit label index for the instruction being decoded.

rossberg-chromium: That depends on the implementation. If you have already constructed an AST and thrown away your pointers to the binary then it actually requires keeping _more_ information around, and longer.

AndrewScheidecker: A third option would be to use an instruction index that can hopefully be reused for other debug information.

I'm fine with any of those options, but I think it should be consistent across the various types of debug metadata. #1051 (source maps) and #990 (standard call stack format) both use file offsets (though not necessarily function-relative). Perhaps there's an argument that those should use instruction indices instead?

AndrewScheidecker: I just noticed there's a PR for #990, #1053.

AndrewScheidecker: @dschuff Any thoughts on this (and its consistency with the source maps and call stack formatting PRs)?

hemobo: @AndrewScheidecker It feels like the purpose of the name section is assigning optional names to 'logical' entities in a module, which ought to work independent of the current representation of that module, be it binary encoding (with possibly different byte sequences encoding the same semantics), text format (where label names would also be used) or some data structure in memory (where the original encoding bytes might not be available anymore, or not yet, or possibly never at all). 

As such the name section should probably use something that has meaning independent of any concrete sequence of bytes, either the 'new index space for labels' idea (most consistent with what the name section does for all other kinds of entities) or using an instruction index.
Adding names to arbitrary positions in the instruction sequence is probably a different feature from adding names to control flow labels, though.

I think you're right, there are discussions to be had about whether, in the case of Webassembly, either one or both of stack traces and source maps should rather refer to positions in a concrete file (byte or text offsets, which is what they traditionally do) or to logical instructions in a function instead. I think there's arguments for both approaches. That however seems like a very different thing from adding names to specific entities in the name section, which probably should depend neither on the former two nor on the concrete binary encoding.

AndrewScheidecker:
I updated this to index labels sequentially according to the order of the operators that introduce them.

jfbastien:
This generally lgtm. Could you put it on the agenda for the next CG meeting (not tomorrow's since that's too tight an addition, but the next one when it's posted)? Would you also be able to attend and present?

AndrewScheidecker:
I will be traveling Dec 11-25, so I probably cannot attend the next video call. I can still add it to the agenda if you'd like.

jfbastien:
@AndrewScheidecker you can put it in a section of "items to discuss at next meeting", noting your preference to be present. That way we can punt the item until you can attend.

sbc100:
Can we name memory segments too?  This would be useful for simplifying the static linking which requires naming of segments and currently does this using its own custom section.

binji:
We don't currently reference memory or table segments by index, but the [conditional segment initializer proposal](https://github.com/WebAssembly/threads/issues/62) will change that with the addition of the `init_memory` and `init_table` instructions. So yeah, it will make sense to have names for those too, if we go forward with that proposal.

kumpera:
We're using twiggy to inspect our generated wasm images (<3 <3 to the rust team for writing it) and the non-code bits are a significant part of our image.

This would help us a lot figure out what's taking all the space.

AndrewScheidecker:
How should I proceed with this PR? I don't think it's controversial, but is very stale now.

FWIW, I have been maintaining [a fork of LLD](https://github.com/WAVM/Wavix/blob/master/lld/wasm/SyntheticSections.cpp#L479) that writes this extended name section, with the exception of the label names (I think getting that information in the linker requires some larger changes).

binji:
@AndrewScheidecker yes, this PR is too old to land as-is. I think it's useful and a reasonable extension of the current name section format. I think the right way to move forward is to make it a new proposal. Currently it would be phase 0, so we'd need to bring it to the CG to move it to phase 1. At that point we can fork the spec repo and make the change to appropriate place in the spec (https://github.com/WebAssembly/spec/blob/master/document/core/appendix/custom.rst)

AndrewScheidecker:
I've created a proposal repo here and hope to present it at the CG meeting tomorrow: https://github.com/AndrewScheidecker/wasm-extended-name-section

I'll close this PR now, but if anybody would like to provide feedback here on the proposal repo, I'd appreciate it!

nomeata:
https://github.com/AndrewScheidecker/wasm-extended-name-section says it is archived. What does that mean?

AndrewScheidecker:
> https://github.com/AndrewScheidecker/wasm-extended-name-section says it is archived. What does that mean?

The repo has moved here: https://github.com/WebAssembly/extended-name-section

PR #1065, opened 2017-05-15
Remove streaming compilation

jfbastien:
We already have promise / fetch APIs, and ES6 modules are documented elsewhere


jfbastien:
This should be uncontroversial. Fast-merge, happy to revisit if not.

domenic:
Not the best commit message/PR title, I must say ^_^

Issue #998, opened 2017-02-27
Add rationale for polymorphic typechecking to Rationale.md


pepyakin:
I have a few questions about the following paragraph in Validation section of [Semantics.md](https://github.com/WebAssembly/design/blob/master/Semantics.md).

> A special case is unconditional control transfers (br, br_table, return, unreachable) because execution never proceeds after them. The stack after such an instruction is unconstrained and thus said to be polymorphic. The following instructions still must type-check, but conceptually, values of any type can be popped off a polymorphic stack for the sake of checking consecutive instructions. A polymorphic stack also matches any possible signature at the end of a block or function. After the end of a block, the stack is determined by the block signature and the stack before the block.

1. Can somebody clarify what means "values of any type can be popped off a polymorphic stack", but "the following instructions still must type-check"? Sounds some like contradictory to me.

2. "A polymorphic stack also matches any possible signature at the end of a block or function."
So in code below (emscripten generated $stackAlloc)
```
func i32
  block i32
    ...
    get_local 0x1
    return
  end
end
```
`return` pops last value from the stack, so when we hitting `end` of `block i32` there are no more operands on the stack. But because the stack is polymorphic we accept that, pop control stack, and we should push i32. Then we are hitting `end` of `func i32`, and we are poping that i32 and return it (because of implicit `end` of `func` behavior). 
Am I get it right?

3. Why is this polymorphic stack thing even needed in a first place? Why just not throw validation error on an unreachable code?

rossberg-chromium:
On 27 February 2017 at 09:40, Sergey Pepyakin <notifications@github.com>
wrote:

> I have a few questions about the following paragraph in Validation section
> of Semantics.md
> <https://github.com/WebAssembly/design/blob/master/Semantics.md>.
>
> A special case is unconditional control transfers (br, br_table, return,
> unreachable) because execution never proceeds after them. The stack after
> such an instruction is unconstrained and thus said to be polymorphic. The
> following instructions still must type-check, but conceptually, values of
> any type can be popped off a polymorphic stack for the sake of checking
> consecutive instructions. A polymorphic stack also matches any possible
> signature at the end of a block or function. After the end of a block, the
> stack is determined by the block signature and the stack before the block.
>
>
>    1.
>
>    Can somebody clarify what means "values of any type can be popped off
>    a polymorphic stack", but "the following instructions still must
>    type-check"?
>
>
Other values may be pushed onto a polymorphic stack and must still be
consumed correctly. E.g.

[...]

PR #1067, opened 2017-05-17
Rationale for stack polymorphism

rossberg-chromium:
To address #998.

### DIFF
--- a/Rationale.md
+++ b/Rationale.md
@@ -136,7 +136,7 @@ some platforms for increased TLB efficiency.
 The `grow_memory` operator returns the old memory size. This is desirable for
 using `grow_memory` independently on multiple threads, so that each thread can
 know where the region it allocated starts. The obvious alternative would be for
-such threads to communicate manually, however wasm implementations will likely
+such threads to communicate manually, however WebAssembly implementations will likely
 already be communicating between threads in order to properly allocate the sum
 of the allocation requests, so it's expected that they can provide the needed
 information without significant extra effort.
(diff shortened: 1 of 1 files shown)


Rationale.md
```diff
@@ -566,3 +566,44 @@ Yes:
...
+
+There is no control edge from an unconditional control transfer instruction (like `br`, `br_table`, `return`, or `unreachable`) to the textually following instruction -- the next instruction is unreachable.
+Consequently, no constraint has to be imposed between the two regarding the stack.
+In a linear type checking algorithm this can be expressed by allowing to assume any type of stack.
+In type system terms, the instruction can thus be said to be _polymorphic_ in its stack type.
+
+This solution is canonical in the sense that it induces the minimal type constraints necessary to ensure soundness, while also maintaining the following useful structural properties about valid (i.e., well-typed) code:
+
+* It is composable: if statement sequences A and B are valid, then block(A,B) is valid.
+* It is decomposable: if block(A,B) is valid, then both A and B are valid.
```
lukewagner: Don't these two statements need some sort of qualification that the signature of A and B are `[]->[]`?

Rationale.md
```diff
@@ -566,3 +566,44 @@ Yes:
...
+Some of these properties are relevant to support certain compilation techniques without artificial complication for producers.
+As a small but representative example, consider a textbook one-pass compiler, which would often use something like the following scheme for compiling expressions:
+```
+compile(A + B) = compile(A); compile(B); emit(i32.add)
+compile(A / B) = compile(A); compile(B); emit(dup i32.eqz (br_if $div_zero) drop i32.div)
+compile(A / 0) = compile(A); emit(br $div_zero)
+```
+The third case is a specialisation of the second, a simple optimisation.
+Without polymorphic typing of the `br` instruction, however, this simple scheme would not work, since `compile((x/0)+y)` would result in invalid code. Worse, it can lead to subtle compiler bugs.
+Similar situations arise in production compilers, for example, in the asm.js-to-wasm compilers that some of the Wasm VMs/tools are implementing.
```
lukewagner: nit: s/Wasm/WebAssembly|wasm/ here and elsewhere

Rationale.md
rossberg-chromium: Yeah, at least the first one. Added.

Rationale.md
rossberg-chromium: Done.

Issue #1040, opened 2017-04-13
How do the Response overloads actually work?


domenic:
Putting together JS.md + Web.md, we have the overloads:

```
Promise<WebAssembly.Module> compile(BufferSource bytes)
Promise<WebAssembly.Module> compile(Response source)
Promise<WebAssembly.Module> compile(Promise<Response> source)
```

How does the overload logic actually work, given an arbitrary object X?

In the Web IDL sense of overloads, these overloads are impossible impossible, because nothing is distinguishable with promises. That is, in most web APIs, any promise-based API causes the processing model to immediately convert the argument to a promise with `Promise.resolve(arg)`, including BufferSources or Responses, so the `Promise<Response>` overload would always win.

My guess is that you are not using Web IDL's processing model for overloads, but it's not clear what processing model you are using.

I think the best processing model here that would fit with other web APIs would be the following:

```js
function compile(arg) {
  return Promise.resolve(arg).then(arg => {
    if (isResponse(arg)) {
      ...
    } else if (isBufferSource(arg)) {
      ...
    } else {
      throw new TypeError(...);
    }
  });
}
```

However, if this is the intent, it's definitely not clear from the overload definitions. I would probably spec it as saying something like

> On the web, the compile method is replaced with `Promise<WebAssembly.Module> compile(Promise<Response or BufferSource> source)`, with the semantics that any incoming value is cast to a promise immediately. (This allows passing in Responses or BufferSources directly, as well as promises for them.)

Some notable test cases for this:

- `{ then(f) { f(aResponseOrBufferSource); }` should work
- No sync errors are thrown ever; only rejected promises, including for invalid args like `5` or `Promise.resolve(5)`.
- If you have any way to observe whether compilation is sync or async, it'd be good to test that it's always async.

rossberg-chromium:
Good points, and part of the reason why I argued for making the Response
version a separate function.

lukewagner:
Would changing the current signature to `Promise<WebAssembly.Module> compile(Promise<Response or BufferSource> source)` match any existing Promise-taking APIs and is this signature expressible in WebIDL?  If not, or maybe regardless, since noone has shipped this particular API, I think we could change it without breaking anything.

domenic:
> Would changing the current signature to `Promise<WebAssembly.Module> compile(Promise<Response or BufferSource> source)` match any existing Promise-taking APIs and is this signature expressible in WebIDL?

Yes and yes, assuming you include my

[...]

PR #1068, opened 2017-05-18
Rename Response based compile + instantiate *Streaming.

flagxor:


lukewagner:
Oh right, we also need to (in this PR or another) address @domenic's root issue in #1040 which is that (`compile`|`instantiate`)`Streaming` should not be defined as having two overloads but rather taking some argument `v` and unconditionally `Promise.resolve(v)` and then throwing if the resolved value isn't a `Response`.

flagxor:
PTAL

Revised to address @domenic 's concern + using the syntax suggested by TAG.

JS.md
```diff
@@ -113,7 +113,12 @@ If neither of the following overloads match, then the returned `Promise` is
 with a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror).
 
 ```
-Promise<{module:WebAssembly.Module, instance:WebAssembly.Instance}>
```
domenic: Hmm, I guess the TAG suggested this change? I'm not sure it makes things clearer, given that you haven't switched wholesale to Web IDL (yet?), so it's just halfway between.

Probably WebAssembly.Module/Instance below, at the very least?

Web.md
```diff
@@ -31,25 +31,23 @@ Non-web embeddings are not required to support these additional methods.
...
 ```
-Promise<WebAssembly.Module> compile(Response source)
-
-Promise<WebAssembly.Module> compile(Promise<Response> source)
+Promise<WebAssembly.Module> compileStreaming(source)
 ```
 
-Developers can set the argument `source` with either a promise that resolves
-with a
+`source` is unconditionally passed through `Promise.resolve(source)`.
```
domenic: You'll want to add "using the built-in value of Promise.resolve" just so that nobody thinks that monkeypatching `Promise.resolve = function () { }` should affect this implementation.

mtrofin:
How about "{compile|instantiate}Stream" rather than "..Streaming"? 

"Streaming" suggests implementation detail, which may not be what necessarily happens. For example, if the engine doesn't AOT but only lazily JIT, "compileStreaming" would be incorrect.

"Stream", on the other hand, refers to the parameter, and is factually true regardless of the implementation.

flagxor:
Can folks please chime in on if we like compileStream + instantiateStream better.
(I think I like it)

domenic:
> "Stream", on the other hand, refers to the parameter, and is factually true regardless of the implementation.

Unfortunately it is not true. A Response is not a Stream. It contains a ReadableStream as its `body` in some implementations (eventually all).

But even if it were a Stream, we don't generally add the type name of arguments to the name of methods. E.g. we don't have `compileBufferSource`.

flagxor:
Dominic makes a good point on the name. Leaving that as is.

JS.md
flagxor: Done.

PR #1072, opened 2017-05-22
Fix missed replace in API name change.

flagxor:
From https://github.com/WebAssembly/design/pull/1068


jfbastien:
Edited the message to refer to https://github.com/WebAssembly/design/pull/1068

PR #1074, opened 2017-05-23
Clarify structured clone behavior for WebAssembly objects.

flagxor:
PTAL
(Particularly around the MAY for secure origins)

### DIFF
--- a/JS.md
+++ b/JS.md
@@ -169,7 +169,7 @@ with the resulting `WebAssembly.Instance` object. On failure, the `Promise` is
 A `WebAssembly.Module` object represents the stateless result of compiling a
 WebAssembly binary-format module and contains one internal slot:
 
- * [[Module]] : an [`Ast.module`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/ast.ml#L176)
+ * [[WebAssembly.Module]] : an [`Ast.module`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/ast.ml#L176)
    which is the spec definition of a module
 
 ### `WebAssembly.Module` Constructor
@@ -195,7 +195,7 @@ Otherwise, this function performs synchronous compilation of the `BufferSource`:
    according to the rules in [spec/valid.ml](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/valid.ml#L415).
 1. The spec `string` values inside `Ast.module` are decoded as UTF8 as described in 
    [Web.md](Web.md#names).
-1. On success, a new `WebAssembly.Module` object is returned with [[Module]] set to
+1. On success, a new `WebAssembly.Module` object is returned with [[WebAssembly.Module]] set to
    the validated `Ast.module`.
 1. On failure, a new `WebAssembly.CompileError` is thrown.
 
@@ -219,7 +219,7 @@ is thrown.
 
 This function returns a new `Array` every time it is called. Each such `Array` is produced by mapping each
 [`Ast.export`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/ast.ml#L152)
-`e` of [moduleObject.[[Module]].exports](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/ast.ml#L187)
+`e` of [moduleObject.[[WebAssemby.Module]].exports](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/ast.ml#L187)
 to the Object `{ name: String(e.name), kind: e.ekind }` where `e.name` is [decoded as UTF8](Web.md#names)
 and `e.ekind` is mapped to one of the String values `"function"`, `"table"`, `"memory"`, `"global"`.
 
@@ -240,7 +240,7 @@ is thrown.
 
 This function returns a new `Array` every time it is called. Each such `Array` is produced by mapping each
 [`Ast.import`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/ast.ml#L167)
-`i` of [moduleObject.[[Module]].imports](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/ast.ml#L203)
+`i` of [moduleObject.[[WebAssembly.Module]].imports](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/ast.ml#L203)
 to the Object `{ module: String(i.module_name), name: String(i.item_name), kind: i.ikind }` where
 `i.module_name` and `i.item_name` are  [decoded as UTF8](Web.md#names) and
 `i.ikind` is mapped to one of the String values `"function"`, `"table"`, `"memory"`, `"global"`.
@@ -270,22 +270,6 @@ This function returns a new `Array` every time it is called. Each such `Array` i
 
 The `Array` is populated in the same order custom sections appear in the WebAssembly binary.
 
-### Structured Clone of a `WebAssembly.Module`
-
-A `WebAssembly.Module` is a
-[cloneable object](https://html.spec.whatwg.org/multipage/infrastructure.html#cloneable-objects)
-which means it can be cloned between windows/workers and also
-stored/retrieved into/from an [IDBObjectStore](https://w3c.github.io/IndexedDB/#object-store).
-The semantics of a structured clone is as-if the binary source, from which the
-`WebAssembly.Module` was compiled, were cloned and recompiled into the target realm.
-Engines should attempt to share/reuse internal compiled code when performing
-a structured clone although, in corner cases like CPU upgrade or browser
-update, this may not be possible and full recompilation may be necessary.
-
-Given the above engine optimizations, structured cloning provides developers
-explicit control over both compiled-code caching and cross-window/worker code
-sharing.
-
 ## `WebAssembly.Instance` Objects
 
 A `WebAssembly.Instance` object represents the instantiation of a 
@@ -313,7 +297,7 @@ If `moduleObject` is not a `WebAssembly.Module`, a [`TypeError`](https://tc39.gi
 is thrown.
 
 Let `module` be the [`Ast.module`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/ast.ml#L176)
-`moduleObject.[[Module]]`.
+`moduleObject.[[WebAssembly.Module]]`.
 
 If the `importObject` parameter is not `undefined` and `Type(importObject)` is
 not Object, a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
@@ -371,7 +355,7 @@ For each [`import`](https://github.com/WebAssembly/spec/blob/master/interpreter/
       [`memory_type`](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#memory_type)
       by `Eval.init` below.)
    1. Append `v` to `memories`.
-   1. Append `v.[[Memory]]` to `imports`.
+   1. Append `v.[[WebAssembly.Memory]]` to `imports`.
 1. Otherwise (`i` is a table import):
    1. If `v` is not a [`WebAssembly.Table` object](#webassemblytable-objects),
       throw a `WebAssembly.LinkError`.
@@ -446,7 +430,7 @@ each [external](https://github.com/WebAssembly/spec/blob/master/interpreter/spec
    1. If `v` is an `i64`, throw a `WebAssembly.LinkError`.
    1. Return [`ToJSValue`](#tojsvalue)`(v)`.
 1. If `e` is a [memory](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/instance.ml#L14) `m`:
-   1. If there is an element `memory` in `memories` whose `memory.[[Memory]]` is `m`, then return `memory`.
+   1. If there is an element `memory` in `memories` whose `memory.[[WebAssembly.Memory]]` is `m`, then return `memory`.
    1. (Note: At most one `WebAssembly.Memory` object is created for any memory, so the above `memory` is unique, even if there are multiple occurrances in the list. Moreover, if the item was an import, the original object will be found.)
    1. Otherwise:
       1. Let `memory` be a new `WebAssembly.Memory` object created via [`CreateMemoryObject`](#creatememoryobject) from `m`.
@@ -537,9 +521,9 @@ A `WebAssembly.Memory` object contains a single [linear memory](Semantics.md#lin
 which can be simultaneously referenced by multiple `Instance` objects. Each
 `Memory` object has two internal slots:
 
- * [[Memory]] : a [`Memory.memory`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/memory.mli)
- * [[BufferObject]] : the current `ArrayBuffer` whose [[ArrayBufferByteLength]]
-   matches the current byte length of [[Memory]]
+ * [[WebAssembly.Memory]] : a [memory instance](https://webassembly.github.io/spec/exec/runtime.html#memory-instances)
+ * [[BufferObject]] : the current `ArrayBuffer` or `SharedArrayBuffer` whose [[ArrayBufferByteLength]]
+   matches the byte length of [[WebAssembly.Memory]].
 
 ### `WebAssembly.Memory` Constructor
 
@@ -571,7 +555,7 @@ Return the result of [`CreateMemoryObject`](#creatememoryobject)(`memory`).
 
 ### CreateMemoryObject
 
-Given a [`Memory.memory`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/memory.mli#L1)
+Given a [memory instance](https://webassembly.github.io/spec/exec/runtime.html#memory-instances)
 `m`, to create a `WebAssembly.Memory`:
 
 Let `buffer` be a new `ArrayBuffer` whose
@@ -584,7 +568,7 @@ Any attempts to [`detach`](http://tc39.github.io/ecma262/#sec-detacharraybuffer)
 the detachment performed by [`m.grow`](#webassemblymemoryprototypegrow) shall throw a 
 [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)
 
-Return a new `WebAssembly.Memory` instance with [[Memory]] set to `m` and
+Return a new `WebAssembly.Memory` instance with [[WebAssembly.Memory]] set to `m` and
 [[BufferObject]] set to `buffer`.
 
 ### `WebAssembly.Memory.prototype [ @@toStringTag ]` Property
@@ -615,13 +599,22 @@ with delta `d`. On failure, a
 [`RangeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-rangeerror)
 is thrown.
 
-Perform [`DetachArrayBuffer`](http://tc39.github.io/ecma262/#sec-detacharraybuffer)(`M.[[BufferObject]]`).
+If `M.[[BufferObject]]` is a not Shared Data Block:
+  1. Assign to `M.[[BufferObject]]` a new `ArrayBuffer` whose
+     [[[ArrayBufferData]]](http://tc39.github.io/ecma262/#sec-properties-of-the-arraybuffer-prototype-object)
+     aliases `M.[[WebAssembly.Memory]]` and whose
+     [[[ArrayBufferByteLength]]](http://tc39.github.io/ecma262/#sec-properties-of-the-arraybuffer-prototype-object)
+     is set to the new byte length of `M.[[WebAssembly.Memory]]`.
+  2. Do the same in all Shared Data Block which mirror `M.[[BufferObject]]`,
+     doing so atomically in relation to other updates of such blocks.
+Otherwise:
+  1. perform [`DetachArrayBuffer`](http://tc39.github.io/ecma262/#sec-detacharraybuffer)(`M.[[BufferObject]]`).
 
-Assign to `M.[[BufferObject]]` a new `ArrayBuffer` whose
-[[[ArrayBufferData]]](http://tc39.github.io/ecma262/#sec-properties-of-the-arraybuffer-prototype-object)
-aliases `M.[[Memory]]` and whose 
-[[[ArrayBufferByteLength]]](http://tc39.github.io/ecma262/#sec-properties-of-the-arraybuffer-prototype-object)
-is set to the new byte length of `M.[[Memory]]`.
+  2. Assign to `M.[[BufferObject]]` a new `ArrayBuffer` whose
+     [[[ArrayBufferData]]](http://tc39.github.io/ecma262/#sec-properties-of-the-arraybuffer-prototype-object)
+     aliases `M.[[WebAssembly.Memory]]` and whose
+     [[[ArrayBufferByteLength]]](http://tc39.github.io/ecma262/#sec-properties-of-the-arraybuffer-prototype-object)
+     is set to the new byte length of `M.[[WebAssembly.Memory]]`.
 
 Return `ret` as a Number value.
 
@@ -639,7 +632,7 @@ A `WebAssembly.Table` object contains a single [table](Semantics.md#table)
 which can be simultaneously referenced by multiple `Instance` objects. Each
 `Table` object has two internal slots:
 
- * [[Table]] : a [`Table.table`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/table.mli#L1)
+ * [[Table]] : a [`table instance`](https://webassembly.github.io/spec/exec/runtime.html#table-instances)
  * [[Values]] : an array whose elements are either `null` or [Exported Function Exotic Object](#exported-function-exotic-objects)
 
 ### `WebAssembly.Table` Constructor
--- a/Web.md
+++ b/Web.md
@@ -202,6 +202,95 @@ Given these values, to process a potential WebAssembly response:
       _reason_.
 1. Return _returnValue_.
 
+### Structured serialization of a `WebAssembly.Memory`
+
+A `WebAssembly.Memory` is a
+[serializable object](https://html.spec.whatwg.org/multipage/infrastructure.html#serializable-objects), and will require changes to the
+[HTML Standard](https://html.spec.whatwg.org/multipage/infrastructure.html#serialization-steps) as follows:
+
+[StructuredSerializeInternal](https://html.spec.whatwg.org/multipage/infrastructure.html#structuredserializeinternal) adds a new step after step 12:
+
+* Otherwise, if *value* has a [[WebAssembly.Memory]] internal slot, then:
+  1. Let *memory* be *value*.[[WebAssembly.Memory]], which is a
+     [`memory instance`](https://webassembly.github.io/spec/exec/runtime.html#memory-instances).
+  2. Let *data* be *memory*'s `data` field, which is a sizable vector of bytes.
+  3. Let *maximum* be the value of *memory*'s `max` field, which is an optional u32.
+  4. If *data* is a Shared Data Block, then:
+     1. If *forStorage* is true, then throw a
+        "[`DataCloneError`](https://heycam.github.io/webidl/#datacloneerror)"
+        [`DOMException`](https://heycam.github.io/webidl/#dfn-DOMException).
+     2. Otherwise,
+        set *serialized* to { [[Type]]: "WebAssembly.Memory",
+        [[MemoryData]]: *data*,
+        [[MemoryMaximum]]: *maximum*,
+        [[AgentCluster]]: the [current Realm Record](https://tc39.github.io/ecma262/#current-realm)'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) }.
+  5. Otherwise throw a
+     "[`DataCloneError`](https://heycam.github.io/webidl/#datacloneerror)"
+     [`DOMException`](https://heycam.github.io/webidl/#dfn-DOMException).
+
+[StructuredDeserialize](https://html.spec.whatwg.org/multipage/infrastructure.html#structureddeserialize) adds a new step after step 12:
+
+* Otherwise, if *serialized*.[[Type]] is "WebAssembly.Memory", then:
+  1. Assert *serialized*.[[MemoryData]] is a Shared Data Block.
+  2. If *targetRealm*'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) is not
+     *serialized*.[[AgentCluster]], then throw a
+     "[`DataCloneError`](https://heycam.github.io/webidl/#datacloneerror)"
+     [`DOMException`](https://heycam.github.io/webidl/#dfn-DOMException).
+  3. Set *value* to a new WebAssembly.Memory object in *targetRealm*, where:
+     1. *value*'s internal slot [[WebAssembly.Memory]] is set to a new
+        [`memory instance`](https://webassembly.github.io/spec/exec/runtime.html#memory-instances) *mem*, where:
+        1. *mem*'s `data` field is set to *serialized*.[[MemoryData]].
+        2. *mem*'s `max` size is set to *serialized*.[[MemoryMaximum]].
+     2. *value*'s internal slot [[BufferObject]] is set to a new `SharedArrayBuffer` *buf*, where:
+        1. *buf*'s internal slot [[ArrayBufferData]]
+           aliases *mem*'s `data` field, which is a sizable vector of bytes.
+        2. *buf*'s internal slot [[ArrayBufferByteLength]]
+           is set to the byte length of *mem*'s linear memory.
+        3. The above is done atomically in relation to other updates to all
+           aliases of [[BufferObject]].
+
+
+### Structured serialization of a `WebAssembly.Module`
+
+A `WebAssembly.Module` is a
+[serializable object](https://html.spec.whatwg.org/multipage/infrastructure.html#serializable-objects), and will require changes to the
+[HTML Standard](https://html.spec.whatwg.org/multipage/infrastructure.html#serialization-steps) as follows:
+
+[StructuredSerializeInternal](https://html.spec.whatwg.org/multipage/infrastructure.html#structuredserializeinternal) adds a new step after step 12:
+
+* Otherwise, if *value* has a [[WebAssembly.Module]] internal slot, then:
+  1. If *forStorage* is true and the relevant settings
+     object of *value* is not "Secure" per the
+     ["Is the environment settings object a secure context?"](https://w3c.github.io/webappsec-secure-contexts/#settings-object)
+     algorithm, throw a
+     "[`DataCloneError`](https://heycam.github.io/webidl/#datacloneerror)"
+     [`DOMException`](https://heycam.github.io/webidl/#dfn-DOMException).
+  2. Otherwise,
+     set *serialized* to { [[Type]]: "WebAssembly.Module",
+     [[WebAssembly.Module]]: *value*.[[WebAssembly.Module]],
+     [[AgentCluster]]: the [current Realm Record](https://tc39.github.io/ecma262/#current-realm)'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) }.
+
+[StructuredDeserialize](https://html.spec.whatwg.org/multipage/infrastructure.html#structureddeserialize) adds a new step after step 12:
+
+* Otherwise, if *serialized*.[[Type]] is "WebAssembly.Module", then:
+  1. If *targetRealm*'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) is not
+     *serialized*.[[AgentCluster]], then throw a
+     "[`DataCloneError`](https://heycam.github.io/webidl/#datacloneerror)"
+     [`DOMException`](https://heycam.github.io/webidl/#dfn-DOMException).
+  2. Otherwise, set *value* to a new WebAssembly.Module object in *targetRealm*
+     whose [[WebAssembly.Module]] internal slot *value* is *serialized*.[[WebAssembly.Module]].
+
+The semantics of a structured serialization is as-if the binary source, from which the
+`WebAssembly.Module` was compiled, is serialized, then deserialized, and recompiled into the target realm.
+Engines should attempt to share/reuse internal compiled code when performing
+a structured serialization, although in corner cases like CPU upgrade or browser
+update, this might not be possible and full recompilation may be necessary.
+
+Given the above engine optimizations, structured serialization provides developers
+explicit control over both compiled-code caching and cross-window/worker code
+sharing.
+
+
 ## Developer-facing display conventions
 
 Browsers, JavaScript engines, and offline tools have common ways of referring to


Web.md
```diff
@@ -124,6 +124,81 @@ MIME type mismatch or `opaque` response types
 [reject](http://tc39.github.io/ecma262/#sec-rejectpromise) the Promise with a
 `WebAssembly.CompileError`.
 
+### Structured Clone of a `WebAssembly.Memory`
+
+A `WebAssembly.Memory` is under some circumstances a
+[cloneable object](https://html.spec.whatwg.org/multipage/infrastructure.html#cloneable-objects), and will impact the
```
jfbastien: "will impact"? How solid is this prediction? Reference?

Web.md
```diff
@@ -124,6 +124,81 @@ MIME type mismatch or `opaque` response types
...
+[HTML Standard](https://html.spec.whatwg.org/multipage/infrastructure.html#serialization-steps) as follows:
+
+[StructuredSerializeInternal](https://html.spec.whatwg.org/multipage/infrastructure.html#structuredserializeinternal) adds a new step after step 12:
+
+* Otherwise, if `value` has an [[WebAssembly.Memory]] internal slot, then:
```
jfbastien: "has a" slot, no?

Web.md
```diff
@@ -124,6 +124,81 @@ MIME type mismatch or `opaque` response types
...
+  1. Let `size` be value.[[BufferObject].
```
jfbastien: Brackets mismatch.

jfbastien: Code quotes mismatch on `value` compared to above? Or is that on purpose?

jfbastien: There's no such slot according to this snippet from above:

> A `WebAssembly.Module` object represents the stateless result of compiling a WebAssembly binary-format module and contains one internal slot: [...]

Web.md
```diff
@@ -124,6 +124,81 @@ MIME type mismatch or `opaque` response types
...
+  2. If !  [IsSharedArrayBuffer](https://tc39.github.io/ecma262/#sec-issharedarraybuffer) is true, then:
```
jfbastien: What's with the bang? "If it's not shared is true"?

Web.md
```diff
@@ -124,6 +124,81 @@ MIME type mismatch or `opaque` response types
...
+    1. If `forStorage` is true, then throw a
+       ["DataCloneError" DomException](https://heycam.github.io/webidl/#datacloneerror).
+    2. Set `serialized` to { [[Type]]: "WebAssembly.Memory",
+       [[ArrayBufferData]]: value.[[BufferObject]].[[ArrayBufferData],
+       [[ArrayBufferByteLength]]: size,
+       [[AgentCluster]]: the [current Realm Record](https://tc39.github.io/ecma262/#current-realm)'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) }.
+    3. Otherwise:
+       1. Let `dataCopy` be [CreateByteDataBlock](https://tc39.github.io/ecma262/#sec-createbytedatablock)(size).
+          NOTE: This can throw a RangeError exception on allocation failure.
```
jfbastien: RangeError? Ew. That's standard?

Web.md
```diff
@@ -124,6 +124,81 @@ MIME type mismatch or `opaque` response types
...
+
+[StructuredSerializeInternal](https://html.spec.whatwg.org/multipage/infrastructure.html#structuredserializeinternal) adds a new step after step 12:
+
+* Otherwise, if `value` has an [[WebAssembly.Memory]] internal slot, then:
+  1. Let `size` be value.[[BufferObject].
+  2. If !  [IsSharedArrayBuffer](https://tc39.github.io/ecma262/#sec-issharedarraybuffer) is true, then:
+    1. If `forStorage` is true, then throw a
+       ["DataCloneError" DomException](https://heycam.github.io/webidl/#datacloneerror).
+    2. Set `serialized` to { [[Type]]: "WebAssembly.Memory",
+       [[ArrayBufferData]]: value.[[BufferObject]].[[ArrayBufferData],
```
jfbastien: Mismatched bracket.

Web.md
```diff
@@ -124,6 +124,81 @@ MIME type mismatch or `opaque` response types
...
+       [[AgentCluster]]: the [current Realm Record](https://tc39.github.io/ecma262/#current-realm)'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) }.
+    3. Otherwise:
+       1. Let `dataCopy` be [CreateByteDataBlock](https://tc39.github.io/ecma262/#sec-createbytedatablock)(size).
+          NOTE: This can throw a RangeError exception on allocation failure.
+       2. Perform `CopyDataBlockBytes`(https://tc39.github.io/ecma262/#sec-copydatablockbytes)
+          (dataCopy, 0,
+          value.[[BufferObject]].[[ArrayBufferData]], 0, size).
+       3. Set `serialized` to { [[Type]]: "WebAssembly.Memory",
+          [[ArrayBufferData]]: dataCopy, [[ArrayBufferByteLength]]: size }.
+
```
jfbastien: What if it's the reverse of 2. above with the weird bang? Doesn't seem to have a sequel for the opposite.

Web.md
annevk: The bang means not throwing. It's a convention from the JavaScript Standard.

Web.md
annevk: Yeah, see the link.

Web.md
```diff
@@ -124,6 +124,81 @@ MIME type mismatch or `opaque` response types
 [reject](http://tc39.github.io/ecma262/#sec-rejectpromise) the Promise with a
 `WebAssembly.CompileError`.
 
+### Structured Clone of a `WebAssembly.Memory`
+
+A `WebAssembly.Memory` is under some circumstances a
```
annevk: "Under some circumstances"?

annevk:
I'm a bit confused. This seems to address #1048 to some extent, yet you didn't address the design questions raised there.

Web.md
```diff
@@ -124,6 +124,81 @@ MIME type mismatch or `opaque` response types
...
+A `WebAssembly.Module` is under some circumstances a
+[cloneable object](https://html.spec.whatwg.org/multipage/infrastructure.html#cloneable-objects), and will impact the
+[HTML Standard](https://html.spec.whatwg.org/multipage/infrastructure.html#serialization-steps) as follows:
+
+[StructuredSerializeInternal](https://html.spec.whatwg.org/multipage/infrastructure.html#structuredserializeinternal) adds a new step after step 12:
+
+* Otherwise, if `value` has an [[WebAssembly.Module]] internal slot, then:
+  1. If `forStorage` is true, implementations MAY chose to exclude
+     "insecure origins", in which case they should throw a
+     ["DataCloneError" DomException](https://heycam.github.io/webidl/#datacloneerror).
```
annevk: This isn't precise enough.

Web.md
```diff
@@ -124,6 +124,81 @@ MIME type mismatch or `opaque` response types
 [reject](http://tc39.github.io/ecma262/#sec-rejectpromise) the Promise with a
 `WebAssembly.CompileError`.
 
+### Structured Clone of a `WebAssembly.Memory`
```
domenic: This should be Structured serialization; structured cloning is no longer a thing at the spec level (although it can be used as an informal term outside of specs).

Web.md
domenic: Should be "serializable object" and link to https://html.spec.whatwg.org/multipage/infrastructure.html#serializable-objects

Web.md
domenic: Right, it's always a serializable object. But sometimes that serialization throws.

Web.md
domenic: Nit: generally we use _value_ for variables, not `value`.

Web.md
domenic: Is size a number or an object?

Web.md
domenic: IsSharedArrayBuffer is an abstract op. It needs to take an argument. Probably _value_? Or maybe _value_.[[BufferObject]], which is called for some reason _size_?

Web.md
```diff
@@ -124,6 +124,81 @@ MIME type mismatch or `opaque` response types
...
+[cloneable object](https://html.spec.whatwg.org/multipage/infrastructure.html#cloneable-objects), and will impact the
+[HTML Standard](https://html.spec.whatwg.org/multipage/infrastructure.html#serialization-steps) as follows:
+
+[StructuredSerializeInternal](https://html.spec.whatwg.org/multipage/infrastructure.html#structuredserializeinternal) adds a new step after step 12:
+
+* Otherwise, if `value` has an [[WebAssembly.Memory]] internal slot, then:
+  1. Let `size` be value.[[BufferObject].
+  2. If !  [IsSharedArrayBuffer](https://tc39.github.io/ecma262/#sec-issharedarraybuffer) is true, then:
+    1. If `forStorage` is true, then throw a
+       ["DataCloneError" DomException](https://heycam.github.io/webidl/#datacloneerror).
```
domenic: Formatting:

```md
"`[DataCloneError](https://heycam.github.io/webidl/#datacloneerror)" `[DOMException](https://heycam.github.io/webidl/#dfn-DOMException)`
```

Web.md
```diff
@@ -124,6 +124,81 @@ MIME type mismatch or `opaque` response types
...
+    2. Set `serialized` to { [[Type]]: "WebAssembly.Memory",
+       [[ArrayBufferData]]: value.[[BufferObject]].[[ArrayBufferData],
+       [[ArrayBufferByteLength]]: size,
+       [[AgentCluster]]: the [current Realm Record](https://tc39.github.io/ecma262/#current-realm)'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) }.
+    3. Otherwise:
```
domenic: So WebAssembly.Memory objects can be backed by non-shared array buffers!? Wow, interesting.

Web.md
domenic: Yes. This needs to decide which object's origin is being tested, per https://github.com/whatwg/html/pull/2518#discussion_r112337423, and needs to link to https://w3c.github.io/webappsec-secure-contexts/#settings-object applied to one of the settings objects involved (depending on which one you choose).

Web.md
```diff
@@ -124,6 +124,81 @@ MIME type mismatch or `opaque` response types
...
+
+* Otherwise, if `serialized`.[[Type]] is "WebAssembly.Module", then:
+  1. If `targetRealm`'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) is not
+     `serialized`.[[AgentCluster]], then throw a
+     ["DataCloneError" DomException](https://heycam.github.io/webidl/#dfn-DOMException).
+  2. Otherwise, set `value` to a new WebAssembly.Module object in `targetRealm`
+     whose [[Module]] internal slot value is `serialized`.[[Module]].
+
+The semantics of a structured clone is as-if the binary source, from which the
+`WebAssembly.Module` was compiled, were cloned and recompiled into the target realm.
```
domenic: "serialized, then deserialized and recompiled the target realm"

Web.md
```diff
@@ -124,6 +124,81 @@ MIME type mismatch or `opaque` response types
...
+Engines should attempt to share/reuse internal compiled code when performing
+a structured clone although, in corner cases like CPU upgrade or browser
```
domenic: Nit: comma should be maybe "when performing structured serialization, although in corner cases..."

domenic:
> This seems to address #1048 to some extent, yet you didn't address the design questions raised there.

Right, it seems to go the route of treating these as non-Web IDL-defined objects. Which is fine for now, I guess, since they aren't defined using Web IDL yet.

annevk:
Well, #1068 added some IDL syntax. It would be a lot better to first make a choice there as the current mismatch will just end up being super confusing the moment anyone outside the people closely involved starts looking at this.

Web.md

Web.md

Web.md

Web.md
flagxor: Revised to to say "will require changes to"
We could structure these as changes against said standards?
Domenic / Anne, WDYT?

Web.md
jfbastien: Yeah memories and modules need to opt in to being shared,

Web.md
flagxor: Shuffled round.

Web.md

Web.md

Web.md

Web.md
flagxor: number, changed (was the wrong one).

Web.md
flagxor: Bad definition for size.

Web.md

Web.md
flagxor: Yep.

Web.md

Web.md

Web.md
flagxor: Revised.
JF + others. Are other browsers willing to require the more strict behavior on IDB + secure origins?

flagxor:
PTAL
JF + others, are you ok with the secure origin requirement around forStorage?

Not sure how to resolve the syntax concern, advice appreciated.

jfbastien:
Let me ping others in my team about the details.

Web.md
```diff
@@ -124,6 +124,86 @@ MIME type mismatch or `opaque` response types
 [reject](http://tc39.github.io/ecma262/#sec-rejectpromise) the Promise with a
 `WebAssembly.CompileError`.
 
+### Structured serialization of a `WebAssembly.Memory`
+
+A `WebAssembly.Memory` is under some circumstances a
```
domenic: Remove "under some circumstances", here and below. It's always a serializable object; sometimes serialization throws.

Web.md
```diff
@@ -124,6 +124,86 @@ MIME type mismatch or `opaque` response types
...
+[serializable object](https://html.spec.whatwg.org/multipage/infrastructure.html#serializable-objects), and will require changes to the
+[HTML Standard](https://html.spec.whatwg.org/multipage/infrastructure.html#serialization-steps) as follows:
+
+[StructuredSerializeInternal](https://html.spec.whatwg.org/multipage/infrastructure.html#structuredserializeinternal) adds a new step after step 12:
+
+* Otherwise, if *value* has a [[WebAssembly.Memory]] internal slot, then:
+  1. Let `size` be the number value.[[BufferObject]].[[ArrayBufferByteLength]].
```
domenic: Nit: _size_, _forStorage_ should all be italicized, not code-ified.

Web.md
```diff
@@ -124,6 +124,86 @@ MIME type mismatch or `opaque` response types
...
+       `[DOMException](https://heycam.github.io/webidl/#dfn-DOMException)`.
+    2. Otherwise:
+       1. Set *serialized* to { [[Type]]: "WebAssembly.Memory",
+          [[ArrayBufferData]]: value.[[BufferObject]].[[ArrayBufferData]],
+          [[ArrayBufferByteLength]]: size,
+          [[AgentCluster]]: the [current Realm Record](https://tc39.github.io/ecma262/#current-realm)'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) }.
+       2. Otherwise:
+         1. Let *dataCopy* be [CreateByteDataBlock](https://tc39.github.io/ecma262/#sec-createbytedatablock)(size).
+            NOTE: This can throw a RangeError exception on allocation failure.
+         2. Perform `CopyDataBlockBytes`(https://tc39.github.io/ecma262/#sec-copydatablockbytes)
```
domenic: Link, not code, for CopyDataBlockBytes

Web.md
```diff
@@ -124,6 +124,86 @@ MIME type mismatch or `opaque` response types
...
+[serializable object](https://html.spec.whatwg.org/multipage/infrastructure.html#serializable-objects), and will require changes to the
+[HTML Standard](https://html.spec.whatwg.org/multipage/infrastructure.html#serialization-steps) as follows:
+
+[StructuredSerializeInternal](https://html.spec.whatwg.org/multipage/infrastructure.html#structuredserializeinternal) adds a new step after step 12:
+
+* Otherwise, if *value* has a [[WebAssembly.Memory]] internal slot, then:
+  1. Let `size` be the number value.[[BufferObject]].[[ArrayBufferByteLength]].
+  2. If !  [IsSharedArrayBuffer](https://tc39.github.io/ecma262/#sec-issharedarraybuffer)(*value*) is true, then:
+    1. If `forStorage` is true, then throw a
+       "`[DataCloneError](https://heycam.github.io/webidl/#datacloneerror)"
```
domenic: Missing closing backtick

Web.md
```diff
@@ -124,6 +124,86 @@ MIME type mismatch or `opaque` response types
...
+            value.[[BufferObject]].[[ArrayBufferData]], 0, size).
+         3. Set *serialized* to { [[Type]]: "WebAssembly.Memory",
+            [[ArrayBufferData]]: dataCopy, [[ArrayBufferByteLength]]: size }.
+
+[StructuredDeserialize](https://html.spec.whatwg.org/multipage/infrastructure.html#structureddeserialize) adds a new step after step 12:
+
+* Otherwise, if *serialized*.[[Type]] is "WebAssembly.Memory", then:
+  1. If *targetRealm*'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) is not
+     *serialized*.[[AgentCluster]], then throw a
+     ["DataCloneError" DomException](https://heycam.github.io/webidl/#dfn-DOMException).
```
domenic: Use the above formatting for the DOMException.

Web.md
```diff
@@ -124,6 +124,86 @@ MIME type mismatch or `opaque` response types
...
+  2. Otherwise, set *value* to a new WebAssembly.Memory object in `targetRealm`
```
domenic: _targetRealm_ italicized not code-ified

Web.md
```diff
@@ -124,6 +124,86 @@ MIME type mismatch or `opaque` response types
...
+     whose [[ArrayBufferData]] internal slot value is
```
domenic: WebAssembly.Memory does not have [[ArrayBufferData]] and [[ArrayBufferByteLength]] internal slots. I think you need to create a new ArrayBuffer, or SharedArrayBuffer (branching on whether _serialized_.[[ArrayBufferData]] is a Shared Data Block or not). Then create a new WebAssembly.Memory object whose [[BufferObject]] is the new AB/SAB object.

Web.md
```diff
@@ -124,6 +124,86 @@ MIME type mismatch or `opaque` response types
...
+### Structured serialization of a `WebAssembly.Module`
+
+A `WebAssembly.Module` is under some circumstances a
+[serializable object](https://html.spec.whatwg.org/multipage/infrastructure.html#serializable-objects), and will require changes to the
+[HTML Standard](https://html.spec.whatwg.org/multipage/infrastructure.html#serialization-steps) as follows:
+
+[StructuredSerializeInternal](https://html.spec.whatwg.org/multipage/infrastructure.html#structuredserializeinternal) adds a new step after step 12:
+
+* Otherwise, if *value* has a [[WebAssembly.Module]] internal slot, then:
+  1. If *forStorage* is true and the environmental settings
```
domenic: s/environmental settings object of/relevant settings object of; the latter is the correct jargon. Sorry about how confusing this stuff is :-/.

Web.md
```diff
@@ -124,6 +124,86 @@ MIME type mismatch or `opaque` response types
...
+     object of *value* is not "Secure" per
+     [this](https://w3c.github.io/webappsec-secure-contexts/#settings-object)
+     algorithm, throw a
```
domenic: It'd be more conventional to say

> per the "Is the environment settings object a secure context?" algorithm

with a link, instead of "this".

Web.md
```diff
@@ -124,6 +124,86 @@ MIME type mismatch or `opaque` response types
...
+     ["DataCloneError" DomException](https://heycam.github.io/webidl/#datacloneerror).
```
domenic: Copy DOMException formatting from above.

Web.md
```diff
@@ -124,6 +124,86 @@ MIME type mismatch or `opaque` response types
...
+     Set *serialized* to { [[Type]]: "WebAssembly.Module",
+     [[Module]]: value.[[Module]],
+     [[AgentCluster]]: the [current Realm Record](https://tc39.github.io/ecma262/#current-realm)'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) }.
+
+[StructuredDeserialize](https://html.spec.whatwg.org/multipage/infrastructure.html#structureddeserialize) adds a new step after step 12:
+
+* Otherwise, if *serialized*.[[Type]] is "WebAssembly.Module", then:
+  1. If *targetRealm*'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) is not
+     *serialized*.[[AgentCluster]], then throw a
+     ["DataCloneError" DomException](https://heycam.github.io/webidl/#dfn-DOMException).
```
domenic: Copy DOMException formatting from above.

Web.md
```diff
@@ -124,6 +124,86 @@ MIME type mismatch or `opaque` response types
...
+  2. Otherwise, set *value* to a new WebAssembly.Module object in `targetRealm`
+     whose [[Module]] internal slot value is *serialized*.[[Module]].
+
+The semantics of a structured serialization is as-if the binary source, from which the
+`WebAssembly.Module` was compiled, serialized, then deserialized, and recompiled into the target realm.
```
domenic: Missing word in this sentence? I think it should be "is serialized"

Web.md
```diff
@@ -124,6 +124,86 @@ MIME type mismatch or `opaque` response types
...
+Engines should attempt to share/reuse internal compiled code when performing
+a structured seraialization, although in corner cases like CPU upgrade or browser
+update, this may not be possible and full recompilation may be necessary.
+
+Given the above engine optimizations, structured cloning provides developers
```
domenic: structured serialization

Web.md

Web.md

Web.md

Web.md

Web.md

Web.md

Web.md
flagxor: thanks, done

Web.md

Web.md

Web.md

Web.md

Web.md

Web.md

flagxor:
PTAL

annevk:
>  We could structure these as changes against said standards?

Sounds good to me. We usually call these kind of things "monkey patches". It would be good to file an issue upstream (against HTML at least) to make them aware of this. Having said that, this again somewhat depends on your strategy. If you do go with IDL you don't need to change HTML. IDL objects are more extensible than objects that are defined using more primitive means.

Web.md
```diff
@@ -124,6 +124,103 @@ MIME type mismatch or `opaque` response types
...
+### Structured serialization of a `WebAssembly.Memory`
+
+A `WebAssembly.Memory` is a
+[serializable object](https://html.spec.whatwg.org/multipage/infrastructure.html#serializable-objects), and will require changes to the
+[HTML Standard](https://html.spec.whatwg.org/multipage/infrastructure.html#serialization-steps) as follows:
+
+[StructuredSerializeInternal](https://html.spec.whatwg.org/multipage/infrastructure.html#structuredserializeinternal) adds a new step after step 12:
+
+* Otherwise, if *value* has a [[WebAssembly.Memory]] internal slot, then:
+  1. Let *size* be the number value.[[BufferObject]].[[ArrayBufferByteLength]].
```
domenic: Nit: remove "the number"; italicize _value_ and _size_ everywhere throughout.

Web.md
```diff
@@ -124,6 +124,103 @@ MIME type mismatch or `opaque` response types
...
+  2. If ! [IsSharedArrayBuffer](https://tc39.github.io/ecma262/#sec-issharedarraybuffer)(*value*.[[BufferObject]]) is true, then:
+    1. If *forStorage* is true, then throw a
+       "`[DataCloneError](https://heycam.github.io/webidl/#datacloneerror)`"
```
domenic: It appears I gave bad advice here, viewing the output: https://github.com/flagxor/design/blob/5fa0fbf29437688f86fcff1efeaabcff0770a0bc/Web.md

I think the backticks need to be inside the brackets, e.g.

```
"[`DataCloneError`](...)"
```

Here and everywhere.

Web.md
```diff
@@ -124,6 +124,103 @@ MIME type mismatch or `opaque` response types
...
+    2. Otherwise:
+       1. Set *serialized* to { [[Type]]: "WebAssembly.Memory",
+          [[ArrayBufferData]]: value.[[BufferObject]].[[ArrayBufferData]],
+          [[ArrayBufferByteLength]]: size,
+          [[AgentCluster]]: the [current Realm Record](https://tc39.github.io/ecma262/#current-realm)'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) }.
+       2. Otherwise:
+         1. Let *dataCopy* be [CreateByteDataBlock](https://tc39.github.io/ecma262/#sec-createbytedatablock)(size).
+            NOTE: This can throw a RangeError exception on allocation failure.
+         2. Perform [CopyDataBlockBytes](https://tc39.github.io/ecma262/#sec-copydatablockbytes)
+            (dataCopy, 0,
```
domenic: Nits: no space before the open paren; _dataCopy_ should be italicized here and everywhere

Web.md
```diff
@@ -124,6 +124,103 @@ MIME type mismatch or `opaque` response types
...
+[StructuredDeserialize](https://html.spec.whatwg.org/multipage/infrastructure.html#structureddeserialize) adds a new step after step 12:
+
+* Otherwise, if *serialized*.[[Type]] is "WebAssembly.Memory", then:
+  1. If *targetRealm*'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) is not
+     *serialized*.[[AgentCluster]], then throw a
+     "`[DataCloneError](https://heycam.github.io/webidl/#datacloneerror)`"
+     `[DOMException](https://heycam.github.io/webidl/#dfn-DOMException)`.
+  2. Otherwise,
+     if ! [IsSharedArrayBuffer](https://tc39.github.io/ecma262/#sec-issharedarraybuffer)(*serialized*.[[ArrayBufferData]]) is true, then:
+     Set *value* to a new WebAssembly.Memory object in *targetRealm*
```
domenic: Nit: just say "then set" instead of "then: Set"

Web.md
```diff
@@ -124,6 +124,103 @@ MIME type mismatch or `opaque` response types
...
+
+[StructuredDeserialize](https://html.spec.whatwg.org/multipage/infrastructure.html#structureddeserialize) adds a new step after step 12:
+
+* Otherwise, if *serialized*.[[Type]] is "WebAssembly.Memory", then:
+  1. If *targetRealm*'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) is not
+     *serialized*.[[AgentCluster]], then throw a
+     "`[DataCloneError](https://heycam.github.io/webidl/#datacloneerror)`"
+     `[DOMException](https://heycam.github.io/webidl/#dfn-DOMException)`.
+  2. Otherwise,
+     if ! [IsSharedArrayBuffer](https://tc39.github.io/ecma262/#sec-issharedarraybuffer)(*serialized*.[[ArrayBufferData]]) is true, then:
```
domenic: You can't use IsSharedArrayBuffer on a data block (which is the type of [[ArrayBufferData]]). You can only say "is a Shared Data Block"

Web.md
```diff
@@ -124,6 +124,103 @@ MIME type mismatch or `opaque` response types
...
+     Set *value* to a new WebAssembly.Memory object in *targetRealm*
+     whose [[BufferObject]] internal slot is set to a new
+     [[ArrayBufferData]] internal slot value is
+     set to a new `SharedArrayBuffer`, where:
+     1. The `SharedArrayBuffer`'s internal slot [[ArrayBufferData]]
+        if set to *serialized*.[[BufferObject]].[[ArrayBufferData]].
+     2. The `SharedArrayBuffer`'s internal slot [[ArrayBufferByteLength]]
+        is set to *serialized*.[[BufferObject]].[[ArrayBufferByteLength]].
+  3. Otherwise,
+     Set *value* to a new WebAssembly.Memory object in *targetRealm*
```
domenic: Nit: lowercase "set"

Web.md
```diff
@@ -124,6 +124,103 @@ MIME type mismatch or `opaque` response types
...
+     *serialized*.[[AgentCluster]], then throw a
+     "`[DataCloneError](https://heycam.github.io/webidl/#datacloneerror)`"
+     `[DOMException](https://heycam.github.io/webidl/#dfn-DOMException)`.
+  2. Otherwise,
+     if ! [IsSharedArrayBuffer](https://tc39.github.io/ecma262/#sec-issharedarraybuffer)(*serialized*.[[ArrayBufferData]]) is true, then:
+     Set *value* to a new WebAssembly.Memory object in *targetRealm*
+     whose [[BufferObject]] internal slot is set to a new
+     [[ArrayBufferData]] internal slot value is
+     set to a new `SharedArrayBuffer`, where:
+     1. The `SharedArrayBuffer`'s internal slot [[ArrayBufferData]]
```
domenic: Nit: this should be consistified with the below; this says "The `SharedArrayBuffer`'s internal slot" whereas the below gives a name to it and says "_mem_'s internal slot"

Web.md
```diff
@@ -124,6 +124,103 @@ MIME type mismatch or `opaque` response types
...
+  2. If ! [IsSharedArrayBuffer](https://tc39.github.io/ecma262/#sec-issharedarraybuffer)(*value*.[[BufferObject]]) is true, then:
+    1. If *forStorage* is true, then throw a
+       "`[DataCloneError](https://heycam.github.io/webidl/#datacloneerror)`"
+       `[DOMException](https://heycam.github.io/webidl/#dfn-DOMException)`.
+    2. Otherwise:
+       1. Set *serialized* to { [[Type]]: "WebAssembly.Memory",
+          [[ArrayBufferData]]: value.[[BufferObject]].[[ArrayBufferData]],
+          [[ArrayBufferByteLength]]: size,
+          [[AgentCluster]]: the [current Realm Record](https://tc39.github.io/ecma262/#current-realm)'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) }.
+       2. Otherwise:
```
lukewagner: I think this "Otherwise:" step is meant to apply to the not-SAB case in which case it should be unindented.  Also, the same comment about losing 'maximum' applies.  Moreover, looking at the rich preview I think the indenting is messed up in a few other places as well.

Web.md
```diff
@@ -124,6 +124,103 @@ MIME type mismatch or `opaque` response types
...
+         3. Set *serialized* to { [[Type]]: "WebAssembly.Memory",
+            [[ArrayBufferData]]: dataCopy, [[ArrayBufferByteLength]]: size }.
+
+[StructuredDeserialize](https://html.spec.whatwg.org/multipage/infrastructure.html#structureddeserialize) adds a new step after step 12:
+
+* Otherwise, if *serialized*.[[Type]] is "WebAssembly.Memory", then:
+  1. If *targetRealm*'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) is not
+     *serialized*.[[AgentCluster]], then throw a
+     "`[DataCloneError](https://heycam.github.io/webidl/#datacloneerror)`"
+     `[DOMException](https://heycam.github.io/webidl/#dfn-DOMException)`.
```
lukewagner: If we want to be symmetric with AB/SAB, I think we only need this AgentCluster check for the SAB case.

Web.md
```diff
@@ -124,6 +124,103 @@ MIME type mismatch or `opaque` response types
...
+  2. Otherwise,
+     if ! [IsSharedArrayBuffer](https://tc39.github.io/ecma262/#sec-issharedarraybuffer)(*serialized*.[[ArrayBufferData]]) is true, then:
+     Set *value* to a new WebAssembly.Memory object in *targetRealm*
+     whose [[BufferObject]] internal slot is set to a new
+     [[ArrayBufferData]] internal slot value is
+     set to a new `SharedArrayBuffer`, where:
```
lukewagner: The deserialization of shared and unshared `Memory` seems to shear off the `maximum` property.

Web.md

Web.md

Web.md

Web.md

Web.md

Web.md

Web.md

Web.md
flagxor: Oops. Done.

Web.md
flagxor: Ah yes. Done.

Web.md
flagxor: Good point!. Fixing.

flagxor:
Long delay on this one, sorry about that.

PTAL

annevk:
So you decided not to go with IDL? That will make it harder once you eventually get to wasm bindings.

Web.md
```diff
@@ -133,6 +133,107 @@ A MIME type mismatch, a response whose
...
+        set *serialized* to { [[Type]]: "WebAssembly.Memory",
+        [[ArrayBufferData]]: *value*.[[BufferObject]].[[ArrayBufferData]],
+        [[ArrayBufferByteLength]]: *size*,
+        [[MemoryMaximum]]: *maximum*,
+        [[AgentCluster]]: the [current Realm Record](https://tc39.github.io/ecma262/#current-realm)'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) }.
+  4. Otherwise:
+     1. Let *dataCopy* be [CreateByteDataBlock](https://tc39.github.io/ecma262/#sec-createbytedatablock)(*size*).
+        NOTE: This can throw a RangeError exception on allocation failure.
+     2. Perform [CopyDataBlockBytes](https://tc39.github.io/ecma262/#sec-copydatablockbytes)
+        (*dataCopy*, 0,
```
domenic: Nit: there shouldn't be a linebreak before the open paren

Web.md
```diff
@@ -133,6 +133,107 @@ A MIME type mismatch, a response whose
...
+
+* Otherwise, if *serialized*.[[Type]] is "WebAssembly.Memory", then:
+  1. If ! [IsSharedArrayBuffer](https://tc39.github.io/ecma262/#sec-issharedarraybuffer)(*serialized*.[[ArrayBufferData]]) is true:
+     1. If *targetRealm*'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) is not
+        *serialized*.[[AgentCluster]], then throw a
+        "[`DataCloneError`](https://heycam.github.io/webidl/#datacloneerror)"
+        [`DOMException`](https://heycam.github.io/webidl/#dfn-DOMException).
+     2. Set *value* to a new WebAssembly.Memory object in *targetRealm*
+        whose [[BufferObject]] internal slot is set to a new
+        [[ArrayBufferData]] internal slot *value* is
```
domenic: It looks like "a new [[ArrayBufferData]] internal slot *value* is set to a new" should be deleted; this sentence is a bit jumbled.

Web.md
```diff
@@ -133,6 +133,107 @@ A MIME type mismatch, a response whose
...
+        set to a new `SharedArrayBuffer` *mem*, where:
+        1. *mem*'s internal slot [[ArrayBufferData]]
+           if set to *serialized*.[[BufferObject]].[[ArrayBufferData]].
```
domenic: s/if/is

Web.md
```diff
@@ -133,6 +133,107 @@ A MIME type mismatch, a response whose
...
+        2. *mem*'s internal slot [[ArrayBufferByteLength]]
+           is set to *serialized*.[[BufferObject]].[[ArrayBufferByteLength]].
+        3. *mem*'s internal slot [[WebAssembly.Memory]].[[maximum]]
+           is set to *serialized*.[[MemoryMaximum]].
```
domenic: _mem_ does not have a [[WebAssembly.Memory]] internal slot; *value* does.

Web.md
```diff
@@ -133,6 +133,107 @@ A MIME type mismatch, a response whose
...
+
+[StructuredDeserialize](https://html.spec.whatwg.org/multipage/infrastructure.html#structureddeserialize) adds a new step after step 12:
+
+* Otherwise, if *serialized*.[[Type]] is "WebAssembly.Memory", then:
+  1. If ! [IsSharedArrayBuffer](https://tc39.github.io/ecma262/#sec-issharedarraybuffer)(*serialized*.[[ArrayBufferData]]) is true:
+     1. If *targetRealm*'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) is not
+        *serialized*.[[AgentCluster]], then throw a
+        "[`DataCloneError`](https://heycam.github.io/webidl/#datacloneerror)"
+        [`DOMException`](https://heycam.github.io/webidl/#dfn-DOMException).
+     2. Set *value* to a new WebAssembly.Memory object in *targetRealm*
```
domenic: In both branches you need to set _value_'s [[WebAssembly.Memory]] to a new [`Memory.memory`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/memory.mli) object before you can set properties on [[WebAssembly.Memory]].

It seems like you also need to create some kind of alias between the Memory.memory and the [[ArrayBufferData]]? Like CreateMemoryObject does?

Web.md
```diff
@@ -133,6 +133,107 @@ A MIME type mismatch, a response whose
...
+
+[StructuredDeserialize](https://html.spec.whatwg.org/multipage/infrastructure.html#structureddeserialize) adds a new step after step 12:
+
+* Otherwise, if *serialized*.[[Type]] is "WebAssembly.Module", then:
+  1. If *targetRealm*'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) is not
+     *serialized*.[[AgentCluster]], then throw a
+     "[`DataCloneError`](https://heycam.github.io/webidl/#datacloneerror)"
+     [`DOMException`](https://heycam.github.io/webidl/#dfn-DOMException).
+  2. Otherwise, set *value* to a new WebAssembly.Module object in *targetRealm*
+     whose [[Module]] internal slot *value* is *serialized*.[[Module]].
```
domenic: s/Module/WebAssembly.Module

Web.md
```diff
@@ -133,6 +133,107 @@ A MIME type mismatch, a response whose
...
+
+The semantics of a structured serialization is as-if the binary source, from which the
+`WebAssembly.Module` was compiled, is serialized, then deserialized, and recompiled into the target realm.
+Engines should attempt to share/reuse internal compiled code when performing
+a structured seraialization, although in corner cases like CPU upgrade or browser
```
domenic: Typo "seraialization"

Web.md
```diff
@@ -133,6 +133,107 @@ A MIME type mismatch, a response whose
...
+     1. If *forStorage* is true, then throw a
+        "[`DataCloneError`](https://heycam.github.io/webidl/#datacloneerror)"
+        [`DOMException`](https://heycam.github.io/webidl/#dfn-DOMException).
+     2. Otherwise,
+        set *serialized* to { [[Type]]: "WebAssembly.Memory",
+        [[ArrayBufferData]]: *value*.[[BufferObject]].[[ArrayBufferData]],
+        [[ArrayBufferByteLength]]: *size*,
+        [[MemoryMaximum]]: *maximum*,
+        [[AgentCluster]]: the [current Realm Record](https://tc39.github.io/ecma262/#current-realm)'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) }.
+  4. Otherwise:
```
lukewagner: Based on today's poll, I think this case would throw (and be asserted not to happen on the deserialize side).

Web.md
```diff
@@ -133,6 +133,107 @@ A MIME type mismatch, a response whose
...
+
+A `WebAssembly.Memory` is a
+[serializable object](https://html.spec.whatwg.org/multipage/infrastructure.html#serializable-objects), and will require changes to the
+[HTML Standard](https://html.spec.whatwg.org/multipage/infrastructure.html#serialization-steps) as follows:
+
+[StructuredSerializeInternal](https://html.spec.whatwg.org/multipage/infrastructure.html#structuredserializeinternal) adds a new step after step 12:
+
+* Otherwise, if *value* has a [[WebAssembly.Memory]] internal slot, then:
+  1. Let *size* be *value*.[[BufferObject]].[[ArrayBufferByteLength]].
+  2. Let *maximum* be *value*.[[WebAssembly.Memory]].[[maximum]].
```
lukewagner: IIUC, [[maximum]] is "reaching into" the memory defined by the core wasm spec.  In that case, could you add a link to https://webassembly.github.io/spec/exec/runtime.html#memory-instances, rename to `max` and note that the value is a `u32?` (optional `u32`)?

Web.md
binji: I'm curious if this should throw or be a messageerror instead.

lukewagner: IIUC, messageerrors only pop out when the error condition is dependent on the recipient; if the error is only depending on the argument's type (and we could consider shared-ness to be part of the memory's type), then I think this would throw just as if you tried to pass a `Function`.

binji: I see, makes sense.

Web.md

Web.md

Web.md

Web.md

Web.md

Web.md
flagxor: oops, fixed.

Web.md

Web.md
flagxor: changed to throw in the non-shared case as per our poll.

Web.md
flagxor: Reworked to lean on that.

flagxor:
PTAL

annevk@ Unsure on the whether WebIDL makes more sense for the Web embedding, JS.md likely needs this notation to match up with the ES spec. Will talk more to domenic next week on notation preferences. Mainly trying to narrow what's currently here (as before the PR it just says structured clone).
Do I understand correctly your preference is WebIDL here, or throughout?

annevk:
@flagxor based on the discussion in #1048 it's my understanding that eventually these objects need to be exposed inside wasm (Wasm?) as well. It's also my understanding that there will be a wasm bindings for Web IDL. It would therefore be much "cheaper" to define these in IDL (and also provide a more consistent API from the eventual wasm side of things).

And as a lesser concern, it matters for how we organize the specifications. If you define these without IDL we'll need to define the serialization/deserialization in HTML, but if you define them with IDL you can define everything in the wasm standard (or wasm API standard or whatever it is).

domenic:
Although I think it's good to have a discussion about IDL vs. not, I think this PR can proceed without resolving that. This PR does an admiral job of formalizing the semantics of structured serialization, which is sorely needed. It provides a precise specification implementations can use to get interoperable behavior. The spec structuring (IDL vs. not) can be fixed later, without changing any observable behavior.

Now, to continue the review efforts...

Web.md
```diff
@@ -202,6 +202,92 @@ Given these values, to process a potential WebAssembly response:
...
+        [[MemoryMaximum]]: *maximum*,
+        [[AgentCluster]]: the [current Realm Record](https://tc39.github.io/ecma262/#current-realm)'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) }.
+  4. Otherwise throw a
+     "[`DataCloneError`](https://heycam.github.io/webidl/#datacloneerror)"
+     [`DOMException`](https://heycam.github.io/webidl/#dfn-DOMException).
+
+[StructuredDeserialize](https://html.spec.whatwg.org/multipage/infrastructure.html#structureddeserialize) adds a new step after step 12:
+
+* Otherwise, if *serialized*.[[Type]] is "WebAssembly.Memory", then:
+  1. Assert [IsSharedArrayBuffer](https://tc39.github.io/ecma262/#sec-issharedarraybuffer)(*serialized*.[[ArrayBufferData]]) is true.
```
domenic: You can't use IsSharedArrayBuffer on a data block, only on JavaScript objects. Instead, this should be "Assert: *serialized*.[[ArrayBufferData]] is a Shared Data Block.

Web.md
```diff
@@ -202,6 +202,92 @@ Given these values, to process a potential WebAssembly response:
...
+     "[`DataCloneError`](https://heycam.github.io/webidl/#datacloneerror)"
+     [`DOMException`](https://heycam.github.io/webidl/#dfn-DOMException).
+  2. Otherwise, set *value* to a new WebAssembly.Module object in *targetRealm*
+     whose [[WebAssembly.Module]] internal slot *value* is *serialized*.[[WebAssembly.Module]].
+
+The semantics of a structured serialization is as-if the binary source, from which the
+`WebAssembly.Module` was compiled, is serialized, then deserialized, and recompiled into the target realm.
+Engines should attempt to share/reuse internal compiled code when performing
+a structured serialization, although in corner cases like CPU upgrade or browser
+update, this may not be possible and full recompilation may be necessary.
```
domenic: Nit: s/may/might

Web.md
```diff
@@ -202,6 +202,92 @@ Given these values, to process a potential WebAssembly response:
...
+  3. Set *value* to a new WebAssembly.Memory object in *targetRealm*, where:
+     1. *value*'s internal slot [[WebAssembly.Memory]] is set to a new `Memory.memory` *mem*, where:
+        1. *mem*'s linear memory is set to *serialized*.[[BufferObject]].[[ArrayBufferData]].
+        2. *mem*'s linear memory has a byte length of *serialized*.[[BufferObject]].[[ArrayBufferByteLength]].
+        3. *mem*'s `max` size is set tof *serialized*.[[MemoryMaximum]].
+     2. *value*'s internal slot [[BufferObject]] is set to a new `SharedArrayBuffer` *buf*, where:
+        1. *buf*'s internal slot [[ArrayBufferData]]
+           aliases *mem*'s linear memory.
+        2. *buf*'s internal slot [[ArrayBufferByteLength]]
+           is set to the byte length of *mem*'s linear memory.
```
lukewagner: I think there is a spec-level "race" here with concurrent `grow_memory`: the byte length can be different than when it was tupled on the serialization side.  I think the fix is to pull length from the core spec memory instance so it's "fresh".  Here (and above, where you have `Memory.memory`), it'd be good to link to https://webassembly.github.io/spec/exec/runtime.html#memory-instances to say exactly what you're referring to.

There's also a question of whether `grow_memory` invalidates the tupled [[ArrayBufferData]].  Actually, I'm not clear what the relationship is to the `data : vec(byte)` field of `meminst` and [[ArrayBufferData]].  I think there's not a real question here; it's just a matter of getting the spec concepts to line up with reality.

Web.md
```diff
@@ -202,6 +202,92 @@ Given these values, to process a potential WebAssembly response:
...
+     [[WebAssembly.Module]]: *value*.[[WebAssembly.Module]],
+     [[AgentCluster]]: the [current Realm Record](https://tc39.github.io/ecma262/#current-realm)'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) }.
+
+[StructuredDeserialize](https://html.spec.whatwg.org/multipage/infrastructure.html#structureddeserialize) adds a new step after step 12:
+
+* Otherwise, if *serialized*.[[Type]] is "WebAssembly.Module", then:
+  1. If *targetRealm*'s corresponding [agent cluster](https://tc39.github.io/ecma262/#sec-agent-clusters) is not
+     *serialized*.[[AgentCluster]], then throw a
+     "[`DataCloneError`](https://heycam.github.io/webidl/#datacloneerror)"
+     [`DOMException`](https://heycam.github.io/webidl/#dfn-DOMException).
```
lukewagner: @domenic Just to be clear, this means a `Module` can be post-messaged between some windows (linked by owner and same-origin, same-origin and iframe, and *maybe* even cross-origin; I'm not sure how to read the *could* in "could be same-origin" in [8.1.3.10](https://html.spec.whatwg.org/multipage/webappapis.html#integration-with-the-javascript-agent-cluster-formalism)).  Is that right?

annevk: Basically windows that can document.domain each other are in the same agent (similar-origin window agent). And therefore also in the same agent cluster.

domenic: Yes, sorry for not getting back to you earlier, but @annevk covered it. Also note that "same origin" and "same origin-domain" are distinct concepts; whether a Window is same origin with another is immutable, but whether it's same origin-domain can change (by changing document.domain). Thus the "could be".

JS.md
```diff
@@ -537,9 +521,9 @@ A `WebAssembly.Memory` object contains a single [linear memory](Semantics.md#lin
 which can be simultaneously referenced by multiple `Instance` objects. Each
 `Memory` object has two internal slots:
 
- * [[Memory]] : a [`Memory.memory`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/memory.mli)
- * [[BufferObject]] : the current `ArrayBuffer` whose [[ArrayBufferByteLength]]
-   matches the current byte length of [[Memory]]
+ * [[WebAssembly.Memory]] : a [`Memory.memory`](https://webassembly.github.io/spec/exec/runtime.html#memory-instances)
```
rossberg-chromium: s/`Memory.memory`/memory instance/

JS.md
```diff
@@ -571,7 +555,7 @@ Return the result of [`CreateMemoryObject`](#creatememoryobject)(`memory`).
 
 ### CreateMemoryObject
 
-Given a [`Memory.memory`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/memory.mli#L1)
+Given a [`Memory.memory`](https://webassembly.github.io/spec/exec/runtime.html#memory-instances)
```
rossberg-chromium: Same here

JS.md
```diff
@@ -639,7 +623,7 @@ A `WebAssembly.Table` object contains a single [table](Semantics.md#table)
 which can be simultaneously referenced by multiple `Instance` objects. Each
 `Table` object has two internal slots:
 
- * [[Table]] : a [`Table.table`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/table.mli#L1)
+ * [[Table]] : a [`Table.table`](https://webassembly.github.io/spec/exec/runtime.html#table-instances)
```
rossberg-chromium: s/`Table.table`/table instance/

Web.md
rossberg-chromium: Thinking about this, it probably would make sense if the core spec provided an *allochostmem* function to delegate to for creating that memory instance, just like it does for functions. (And similarly for tables and globals.) I can add that.

Web.md
lukewagner: Gotcha.  So my root question was whether this definition was permissive enough to allow cross-origin code caching (e.g., in the local IDB of a CDN origin), bit it looks like 'no', that'll have to wait for foreign-fetch...

Web.md
lukewagner: Yeah, I was thinking something like that might be necessary.  You also have to think about growing and avoiding (at a spec level) leaving "stale" pointers to old buffers around in extant SABs.

Web.md

Web.md

JS.md
flagxor: done, yeah, that was goofy.

JS.md
flagxor: done

JS.md
flagxor: done

Web.md
flagxor: Yep, that matches my understanding too.

Web.md
flagxor: @lukewagner Agreed there is a spec race in terms of getting to the other side of serialization, I think I've now fixed that. Additionally (just to make sure we're on the same page), there's an actually race around two parties accessing the buffer (and then later its size). Or said another way: once you've fished out the SAB, it's size stays fixed even if wasm grows the memory.

@rossberg-chromium Adding the op seems to make sense, but trying to clarify here without it for now.

flagxor:
PTAL

JS.md
```diff
@@ -615,13 +599,22 @@ with delta `d`. On failure, a
...
 is thrown.
 
-Perform [`DetachArrayBuffer`](http://tc39.github.io/ecma262/#sec-detacharraybuffer)(`M.[[BufferObject]]`).
-
-Assign to `M.[[BufferObject]]` a new `ArrayBuffer` whose
-[[[ArrayBufferData]]](http://tc39.github.io/ecma262/#sec-properties-of-the-arraybuffer-prototype-object)
-aliases `M.[[Memory]]` and whose 
-[[[ArrayBufferByteLength]]](http://tc39.github.io/ecma262/#sec-properties-of-the-arraybuffer-prototype-object)
-is set to the new byte length of `M.[[Memory]]`.
+If `M.[[BufferObject]]` is a not Shared Data Block:
```
domenic: Should be M.[[BufferObject]].[[ArrayBufferData]]

JS.md
```diff
@@ -615,13 +599,22 @@ with delta `d`. On failure, a
...
+  1. Assign to `M.[[BufferObject]]` a new `ArrayBuffer` whose
+     [[[ArrayBufferData]]](http://tc39.github.io/ecma262/#sec-properties-of-the-arraybuffer-prototype-object)
+     aliases `M.[[WebAssembly.Memory]]` and whose
+     [[[ArrayBufferByteLength]]](http://tc39.github.io/ecma262/#sec-properties-of-the-arraybuffer-prototype-object)
+     is set to the new byte length of `M.[[WebAssembly.Memory]]`.
+  2. Do the same in all Shared Data Block which mirror `M.[[BufferObject]]`,
```
domenic: "Blocks", not "Block", I guess

JS.md
```diff
@@ -615,13 +599,22 @@ with delta `d`. On failure, a
...
+     doing so atomically in relation to other updates of such blocks.
+Otherwise:
+  1. perform [`DetachArrayBuffer`](http://tc39.github.io/ecma262/#sec-detacharraybuffer)(`M.[[BufferObject]]`).
```
domenic: Uppercase "P"

Web.md
```diff
@@ -202,6 +202,95 @@ Given these values, to process a potential WebAssembly response:
...
+A `WebAssembly.Memory` is a
+[serializable object](https://html.spec.whatwg.org/multipage/infrastructure.html#serializable-objects), and will require changes to the
+[HTML Standard](https://html.spec.whatwg.org/multipage/infrastructure.html#serialization-steps) as follows:
+
+[StructuredSerializeInternal](https://html.spec.whatwg.org/multipage/infrastructure.html#structuredserializeinternal) adds a new step after step 12:
+
+* Otherwise, if *value* has a [[WebAssembly.Memory]] internal slot, then:
+  1. Let *memory* be *value*.[[WebAssembly.Memory]], which is a
+     [`memory instance`](https://webassembly.github.io/spec/exec/runtime.html#memory-instances).
+  2. Let *data* be *memory*'s `data` field, which is a sizable vector of bytes.
```
domenic: Is "sizable" standard terminology here? If not, "resizable" is probably more clear, as I usually think of "sizable" as a synonym for "large"

domenic: I'm confused. Is _data_ a Shared Data Block, or a sizable vector of bytes? Are the two concepts the same somehow?

JS.md
```diff
@@ -615,13 +599,22 @@ with delta `d`. On failure, a
...
-is set to the new byte length of `M.[[Memory]]`.
+If `M.[[BufferObject]]` is a not Shared Data Block:
+  1. Assign to `M.[[BufferObject]]` a new `ArrayBuffer` whose
+     [[[ArrayBufferData]]](http://tc39.github.io/ecma262/#sec-properties-of-the-arraybuffer-prototype-object)
+     aliases `M.[[WebAssembly.Memory]]` and whose
+     [[[ArrayBufferByteLength]]](http://tc39.github.io/ecma262/#sec-properties-of-the-arraybuffer-prototype-object)
+     is set to the new byte length of `M.[[WebAssembly.Memory]]`.
+  2. Do the same in all Shared Data Block which mirror `M.[[BufferObject]]`,
+     doing so atomically in relation to other updates of such blocks.
+Otherwise:
```
lukewagner: nit: markdown puts this at the end of item 2, I think you need a \n or something to render correctly

JS.md
lukewagner: I think you need to drop the "not"

JS.md
```diff
@@ -571,7 +555,7 @@ Return the result of [`CreateMemoryObject`](#creatememoryobject)(`memory`).
 
 ### CreateMemoryObject
 
-Given a [`Memory.memory`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/memory.mli#L1)
+Given a [memory instance](https://webassembly.github.io/spec/exec/runtime.html#memory-instances)
 `m`, to create a `WebAssembly.Memory`:
 
 Let `buffer` be a new `ArrayBuffer` whose
```
lukewagner: Should this be changed to create a new `SharedArrayBuffer` depending on whether `m` is shared?  Or is that a separate proposal?

JS.md
lukewagner: I was expecting this to look more like: "Perform step 1 to all `WebAssembly.Memory` objects `N` in the same Agent Cluster as `M` where `N.[[BufferObject]].[[ArrayBufferData]] == M.[[BufferObject]].[[ArrayBufferData]]`..."

JS.md
binji: I probably should just update it in the [threads repo](https://github.com/WebAssembly/threads/blob/master/proposals/threads/Overview.md#creatememoryobject), I think.

Web.md
lukewagner: Actually, it [looks like foreign fetch is no longer being proposed](https://github.com/w3c/ServiceWorker/issues/1188).  So what's the recommended mechanism for explicit sharing of compiled code that doesn't rely on pure HTTP cache magic?  I think this is a really important use case for frameworks/middleware in the future.

littledan: What was the reasoning for restricting Modules from being postMessage'd across agent clusters?

It's hard for me to understand how this squares with serializing Modules in IndexedDB (which this patch seems to allow AFAICT, though I don't see how it specifies the sub-serialization of the agent cluster).

littledan:
I've put a version of this patch which just handles Modules and not Memory out for review at https://github.com/WebAssembly/spec/pull/711 .

littledan:
For the SharedArrayBuffer side, I put up a PR in the threads fork at https://github.com/WebAssembly/threads/pull/90 . Is there anything important in this PR which is missing from these patches? If not, maybe this thread could be closed. Thanks for the great work on this thread which made my job in these two PRs straightforward.

JS.md
```diff
@@ -219,7 +219,7 @@ is thrown.
 
 This function returns a new `Array` every time it is called. Each such `Array` is produced by mapping each
 [`Ast.export`](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/ast.ml#L152)
-`e` of [moduleObject.[[Module]].exports](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/ast.ml#L187)
+`e` of [moduleObject.[[WebAssemby.Module]].exports](https://github.com/WebAssembly/spec/blob/master/interpreter/spec/ast.ml#L187)
```
numbermaniac: `WebAssembly` is spelt incorrectly here.

littledan:
This has landed in the specification; I think we can close this PR.

binji:
> This has landed in the specification; I think we can close this PR.

Agreed.

Issue #1079, opened 2017-05-25
🛤 Garbage collection


jfbastien:
**This is a tracking issue for a post-MVP feature**
It will be updated as the issue progresses.

**Topic** Garbage collection
**Champion** Andreas Rossberg (@rossberg-chromium)
**Status** In progress
**Linked issues** #985 #919 #733 #715 #484 #289 #238
**Linked repositories** [github.com/WebAssembly/gc/](https://github.com/WebAssembly/gc/blob/master/proposals/gc/Overview.md)

# Details

To realize the [high-level goals](HighLevelGoals.md) of (1) integrating well with the existing Web platform and (2) supporting languages other than C++, WebAssembly needs to be able to:

* reference DOM and other Web API objects directly from WebAssembly code;
* call Web APIs (passing primitives or DOM/GC/Web API objects) directly from WebAssembly without calling through JavaScript; and
* efficiently allocate and manipulate GC objects directly from WebAssembly code.

An important constraint is that, while WebAssembly should allow tight integration with the [Web](Web.md), it should not bake in details or Web standards dependencies that prevent execution in a 
[non-Web embedding](NonWeb.md). This suggests a design (called [opaque reference types](GC.md#opaque-reference-types) below) that hides the details of JavaScript and WebIDL behind Web-embedding-specific builtin modules. On the other hand, WebAssembly can define a set of native GC primitives that allowed portable GC code to be written regardless of the host environment.

Issue #1066, opened 2017-05-15
Proposal: reorganize feature planning


jfbastien:
We currently track future features in a super ad-hoc manner:

* In FutureFeatures.md
* In random other files
* In issues
* In individual repos
* In our heads

I'm volunteering to organize all of this, because it's madness. The current arrangement is impossible for casual WebAssemblers to understand, and I forget things all the time.

I propose:

* File one issue per future feature.
* Unless there's a repo for bigger features that are well along (or close the issue if so).
* List prior discussions.
* Rewrite FututreFeatures.md to point at the issue / repo, document the champion, and the Unicode identifier we're using to refer to the feature.

Importantly, I'd keep all post-MVP features which have been accepted, and document then date when added. Maybe at some point we'd do something like caniuse and document which engine implemented each feature fully, and when.

Thoughts?

RyanLamansky:
I suggest one issue for future feature, even if it has it's own repo (in which case, the issue would have a link to the repo).  Each future feature should be labeled as such.  With this, you can have a direct link to a complete list of all future features.  This link can be added to the site for easy discoverability.

domenic:
To me the most important artifact here is a single document listing all in-development features, and linking to the place where they are being worked on. As long as that is present, everything else is just bonus.

jgravelle-google:
What @RyanLamansky said, but with the requirement that any external artifacts (e.g. work repos) be linked in the top comment for easier searchability.

jfbastien:
I opened #1073 and #1075 to track threads and SIMD, and I've removed them from [FutureFeatures.md](https://github.com/WebAssembly/design/blob/master/FutureFeatures.md). Please take a look and let me know if the general structure is sensible. If so I'll proceed with all the other features that need to be tracked.

Once that's done I should also add an explanation of feature tracking somewhere... Probably the readme? Or FutureFeatures.md (which would be empty by then)? Or FeatureDetection.md?

domenic:
> and I've removed them from FutureFeatures.md. 

Wait, but where do I go now to find a list of all future/in-development features?

jfbastien:
I'll have a list there. Should have started already! Will do, sorry!

jfbastien:
@domenic added a table: https://github.com/WebAssembly/design/commit/74676255fbc2d482d404cfe66c9ce4f072cc57a0
Which other fields do you think would be useful there?

RyanLamansky:
[...]

PR #1080, opened 2017-05-25
Remove GC.md

jfbastien:
We should removing GC.md as part of #1066. It's more aspirational that concrete, and we're now tracking GC work with #1079 as well as in https://github.com/WebAssembly/gc/.

Caveat before removing this:

* Other pages link to it. I'll clean up.
* Some of the content may still be valid. We should move it to the GC proposal instead so it doesn't become stale.


jfbastien:
Luke / Andreas: what do you think should be moved to the GC proposal from this page?

lukewagner:
Yeah, GC.md was just a stopgap early on when the question kept coming up.

jfbastien:
Is there anything you think should be migrated to https://github.com/WebAssembly/gc/blob/master/proposals/gc/Overview.md ?

lukewagner:
I guess, now that you mention it, the "JS integration" and "WebIDL integration" subsections would be separate proposals (that build on top of gc/Overview.md).  They're just sketches though (and a bit antiquated by now), so out of place w.r.t the rest of the design repo, which describes the shipping status quo.  So I'm fine removing them.

jfbastien:
I can move them to a tracking issue in the design repo.

rektide:
Even if it's a place-holder, having some kind of information available that links WebAssembly to the rest of the Web Platform seems like a really good healthy thing. I'd expect that not having this would cause a lot of silent head scratching and create a support burden for randos like me wandering in wondering what the story is.

I'd like to see a very minimal gc/JS.md and/or gc/WebIDL.md created before dropping GC.md, even if just a placeholder, to prevent confusion from spreading. It's important to leave some kind of stake in the ground here. Links in these documents to a tracking issue would be appropriate and helpful.

PR #1086, opened 2017-06-01
Clarify support == presence for web APIs

mtrofin:
Clarified that "support" means "presence", rather than (for example) throwing exceptions when not present; also indicated that the 2 streaming APIs should be present/absent together: it simplifies developer experience.


rossberg (approved):

Web.md
bjornalbertramstein20-ctrl, line 57: New app and a new webpage available

bjornalbertramstein20-ctrl (commented):
Approved and updated

PR #1088, opened 2017-06-09
Acknolwedge IEEE 754-2018's changes concerning `minNum`/`maxNum`.

sunfishcode:
This just adds a minor note to FutureFeatures.md noting that the IEEE 754-2018 draft has changed with respect to the `minNum` and `maxNum` operations, and noting that it's advisable to follow these changes.


sunfishcode:
This is just adding more information to an existing future-feature idea, and I don't expect it's controversial. Merging with the approval above.

PR #1089, opened 2017-06-09
Define non-trapping float-to-int conversions.

sunfishcode:
This also introduces the concept of prefix bytes, and defines the "numeric" prefix byte, used for encoding the new conversion instructions.

As far as I'm aware, this proposal is fully in accord with the [CG guidance](https://github.com/WebAssembly/meetings/blob/master/2017/CG-05.md#non-trapping-float-to-int) expressed at the recent CG meeting.

### DIFF
--- a/Semantics.md
+++ b/Semantics.md
@@ -648,14 +648,14 @@ is NaN, and *ordered* otherwise.
   * `f64.convert_u/i32`: convert an unsigned 32-bit integer to a 64-bit float
   * `f64.convert_u/i64`: convert an unsigned 64-bit integer to a 64-bit float
   * `f64.reinterpret/i64`: reinterpret the bits of a 64-bit integer as a 64-bit float
-  * `i32.trunc_s:sat/f32`: :bowling: truncate a 32-bit float to a signed 32-bit integer with saturation
-  * `i32.trunc_s:sat/f64`: :bowling: truncate a 64-bit float to a signed 32-bit integer with saturation
-  * `i32.trunc_u:sat/f32`: :bowling: truncate a 32-bit float to an unsigned 32-bit integer with saturation
-  * `i32.trunc_u:sat/f64`: :bowling: truncate a 64-bit float to an unsigned 32-bit integer with saturation
-  * `i64.trunc_s:sat/f32`: :bowling: truncate a 32-bit float to a signed 64-bit integer with saturation
-  * `i64.trunc_s:sat/f64`: :bowling: truncate a 64-bit float to a signed 64-bit integer with saturation
-  * `i64.trunc_u:sat/f32`: :bowling: truncate a 32-bit float to an unsigned 64-bit integer with saturation
-  * `i64.trunc_u:sat/f64`: :bowling: truncate a 64-bit float to an unsigned 64-bit integer with saturation
+  * `i32.trunc_sat_f32_s`: :bowling: truncate a 32-bit float to a signed 32-bit integer with saturation
+  * `i32.trunc_sat_f64_s`: :bowling: truncate a 64-bit float to a signed 32-bit integer with saturation
+  * `i32.trunc_sat_f32_u`: :bowling: truncate a 32-bit float to an unsigned 32-bit integer with saturation
+  * `i32.trunc_sat_f64_u`: :bowling: truncate a 64-bit float to an unsigned 32-bit integer with saturation
+  * `i64.trunc_sat_f32_s`: :bowling: truncate a 32-bit float to a signed 64-bit integer with saturation
+  * `i64.trunc_sat_f64_s`: :bowling: truncate a 64-bit float to a signed 64-bit integer with saturation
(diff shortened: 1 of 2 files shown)


Semantics.md
```diff
@@ -663,7 +671,12 @@ round-to-nearest ties-to-even rounding.
 
 Truncation from floating point to integer where IEEE 754-2008 would specify an
 invalid operator exception (e.g. when the floating point value is NaN or
-outside the range which rounds to an integer in range) traps.
+outside the range which rounds to an integer in range) is handled as follows:
+ - For instructions with no exceptional behavior specified, a trap is produced.
+ - For instructions with the `:sat` modifier, no trap is produced, and:
```
RyanLamansky: Exposing myself as not having a strong C/C++ background, is this min/max/zero functionality available as an intrinsic on mainstream processors?

sunfishcode: No, I'm not aware of any such intrinsics in popular C/C++ compilers.

sunfishcode:
Ok, I added :bowling: as feature marker for this feature. Does this look right?

jfbastien:
> Ok, I added :bowling: as feature marker for this feature. Does this look right?

Does it looks good?!?! IT LOOKS AMAZING.

Still lgtm, but it would be great to get Goog / MS to sign off on this. @flagxor / @MikeHolman ?

juj:
This is superb!

How does one feature test for the presence of additions like this?

sunfishcode:
One can construct minimal WebAssembly modules and test whether they validate, for example using [`WebAssembly.validate`](https://github.com/WebAssembly/design/blob/master/JS.md#webassemblyvalidate) in the JS API.

For example:

```js
var t = new Uint8Array([0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00,
                        0x01, 0x06, 0x01, 0x60, 0x01, 0x7d, 0x01, 0x7f,
                        0x03, 0x02, 0x01, 0x00, 0x0a, 0x08, 0x01, 0x06,
                        0x00, 0x20, 0x00, 0xfc, 0x00, 0x0b]);

if (WebAssembly.validate(t)) {
  // i32.trunc_s:sat/f32 is supported.
}
```

juj:
That makes sense, thanks!

lars-t-hansen:
@sunfishcode, somehow that proposal does not seem quite satisfactory to me.  I know the version number is to stay at 1 "forever" but if we could guarantee that then there would be no version number to begin with.  We should therefore expect that the version number will eventually have to change, notably if a feature has to be changed or removed because it turns out to be unsafe; eventually implementations may then stop accepting old content.  Ergo we may in time see a situation where the above code no longer validates, not because the feature is not supported, but because the version is not accepted.

Technically, the snipped above tests not whether the feature is available, but whether it's available in a version 1 module.

Presumably the test above could attempt to validate two modules, one empty module with version one and then the module with the feature.  It might be useful to add a method, `WebAssembly.acceptsVersion(k)`, to highlight this issue?

lukewagner:
@lars-t-hansen The version is indeed intended to stay at 1 "forever".  Instead of bumping version, we can continue to add new: section ids (and thus section-body formats), import and export kinds (to the existing import/export sections), and types and operators (to the existing code section).  All of these can be added by just accepting new bit patterns to existing fields that were previously rejected.  This is not fundamentally different than JS adding new syntactic forms over time that you have to test via `try { eval("function*(){}") }`.

> We should therefore expect that the version number will eventually have to change, notably if a feature has to be changed or removed because it turns out to be unsafe; eventually implementations may then stop accepting old content.

If a feature had to be removed, the associated opcodes/typecodes/sectionids/etc would just begin to fail validation; that doesn't require a version bump.  (JS and the Web do this all the time without bumping versions.)  The only question is if the removed bit patterns stay reserved forever or one day get "recycled".

Really, the version word is just an escape hatch if it becomes evident in some number of years that there is a *much* better binary format that we'd like to support (in addition to version=1 format, of course) or just some terribly fundamental flaw we need to fix.  The bar for this is extremely high, though.

jfbastien:
It would be helpful if someone volunteered to update https://github.com/WebAssembly/design/blob/master/FeatureTest.md 😀

jfbastien:
Can someone from Google please confirm that this is good to go?

jfbastien:
No, I just want to make sure all implementors sign off. It's between you, @flagxor, and friends who between you can sign off 😄

sunfishcode:
Rebased on master, updated opcodes to reflect the new names, renamed "numeric" to "misc".

sunfishcode:
Non-trapping float-to-int conversions have progressed to [phase 4](https://github.com/WebAssembly/proposals/#phase-4---standardize-the-feature-wg), and this design-repo PR has multiple approvals and no outstanding objections, so let's merge this!

I know some of these design-repo docs aren't active in general, but I mainly just want to merge this to close out this PR :-).

Abbesvensk:
hi

PR #1093, opened 2017-06-14
Rewrite compileStreaming/instantiateStreaming to be more rigorous

### LINKED ISSUE #1081: Unclear how to resolve/reject promises for response based compile API
Right now, the spec says that the API will run an asynchronous task that eventually resolves the promise. It's a bit unclear to me what an asynchronous task should be, it would be great to explicitly reference an event loop model to explain what this is.

Furthermore, in HTML's event loop model, only regular tasks enter a microtask checkpoint. Without doing this, resolved promise will be left hanging until something else spins the event loop again.

I'd expect that the spec here uses similar terms as e.g. fetch, i.e, when a response promise is passed in and the method doesn't immediately abort, it will finish the compilation asynchronously and when the module is ready create a task on the main event loop and that task resolves the promise.

domenic:
Fixes #1081 by being very clear about the interaction with the event loop, i.e. which portions of the tasks happen in parallel and how they get back to the main thread, where objects can actually be created. In general this also makes the specification algorithmic, links it to appropriate definitions, and makes various failure cases (such as being passed a rejected promise, or attempting to read from a locked/disturbed body) clearer.

/cc @jeisinger. @annevk may also enjoy this when he's back from vacation.

### DIFF
--- a/Web.md
+++ b/Web.md
@@ -27,59 +27,43 @@ In non-web embeddings, these APIs may not be present.
 
 ### Additional Web Embedding API
 
+In Web embeddings, the following methods are added.
+
+Note that it is expected that `compileStreaming` and `instantiateStreaming` be either both present or both absent.
+
 #### `WebAssembly.compileStreaming`
 
 :cyclone: Added for milestone 2, developers must feature detect.
-
-In Web embeddings, the following methods are added. 
-
-Note that it is expected that `compileStreaming` and `instantiateStreaming` be either both present or both absent.
 
 ```
 Promise<WebAssembly.Module> compileStreaming(source)
 ```
 
-`source` is unconditionally passed through the built-in value
-of `Promise.resolve`.
-If the result is not a `Response` object, then the returned `Promise` is
-[rejected](http://tc39.github.io/ecma262/#sec-rejectpromise)
-with a [`TypeError`](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror).
-This allows developers to pass either a promise that resolves
-to a
-[`Response`](https://fetch.spec.whatwg.org/#response-class)
-object or a
-[`Response`](https://fetch.spec.whatwg.org/#response-class)
-object (which is automatically cast to a
-promise) for the `source`.
-Renderer-side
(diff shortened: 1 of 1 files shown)


jfbastien:
Is this intended to be an editorial clarification of something we'd agreed on, but which wasn't specified tightly enough?

domenic:
Pretty much... One could argue it's a normative change because previously it wasn't really specified how some of this stuff worked (e.g. the previous spec kind of implied by omission that locked body streams would just be magically read by the UA, instead of throwing as this PR specifies). But IMO there should be nothing controversial in here.

jfbastien:
OK, I think this can be reviewed as an editorial change since (I think?) only Chrome implements this API and it's clarifying what was ill-specified. If others discover issues after it's merged, we should be willing to revisit specific issues.

Does that sound right to you?

domenic:
That does sound reasonable. I'd appreciate a check on the semantics from @jeisinger and probably @mtrofin though, in addition to any editorial review.

jeisinger:
the changes lgtm

Web.md
```diff
@@ -90,48 +74,133 @@ dictionary WebAssemblyInstantiatedSource {
...
+- _successSteps_ and _failureSteps_ will occur in tasks posted back to the main event loop; they
+  accept the Realm-agnostic values returned from _processingSteps_, and must transform them into
+  JavaScript objects that can be used to appropriately fulfill or reject the returned promise.
+
+Given these values, to process a potential WebAssembly response:
+
+1. Let _returnValue_ be [a new
+   promise](https://www.w3.org/2001/tag/doc/promises-guide#a-new-promise).
+1. Let _sourceAsPromise_ be [a promise resolved
+   with](https://www.w3.org/2001/tag/doc/promises-guide#a-promise-resolved-with) _source_.
```
equalsJeffH: given the input description above on new line 129, should "source" actually be "argument" in these variable names?

domenic: Great catch, thanks; fixing and force-pushing.

domenic:
Let's merge, as this has @jeisinger's signoff and @mtrofin seems to be busy. He can follow up later if he has time to review...

Issue #1076, opened 2017-05-23
"Name Section" doesn't specify whether items can be named more than once


sbc100:
The spec says the names must be "sorted by index" but doesn't say if indexes can be repeated.   I guess the intention is that they cannot but that isn't clear.

We had a discussion around this when implementing the name section parser in wabt:
https://github.com/WebAssembly/wabt/pull/426

lukewagner:
Agreed we should specifically say no repeats.

PR #1096, opened 2017-06-17
Clarify function names subsection

### LINKED ISSUE #1094: Do names in the names section need to be unique?
Right now a lot of the wasm tooling won't handle multiple functions with the same name.   e.g. wasm2wast will generate an invalid wast file with multiple functions with the same label (which is not valid wast) and wasm-dis will error out.

I'm working on a port of lld it seems quite common to link objects that contain the same local symbol.  This result in a binary which has many duplicate local symbols.

Up until now most of our tooling is based around llvm-link which seems to rename duplicate locals by adding a number to the end, so the final binary has foo.1, foo.2 etc, so this issue has been hidden.

sbc100:
Fixes #1094, #1076

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -496,6 +496,13 @@ The function names subsection is a `name_map` which assigns names to
 a subset of the [function index space](Modules.md#function-index-space)
 (both imports and module-defined).
 
+Each function may be named at most once.  Naming a function more than once
+results in the section being malformed.
+
+However, names need not be unique.  The same name may be given for multiple
+functions. This is common for C++ programs where the multiple compilation units
+that comprise a binary can contain local functions with the same name.
+
 #### Local names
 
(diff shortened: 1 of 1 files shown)


BinaryEncoding.md
```diff
@@ -481,6 +481,12 @@ The function names subsection is a `name_map` which assigns names to
 a subset of the [function index space](Modules.md#function-index-space)
 (both imports and module-defined).
 
+Each function may be named at most once.
```
jfbastien: What's the outcome if this is broken?

sbc100: I was thinking it would result in a malformed section.  Similar to other restrictions such as that functions must be named in index order.

jfbastien: I would state it here.

sbc100:
Done

PR #1154, opened 2017-11-13
Make it obvious that arguments aren't even coerced when calling i64 exports/imports in JS

bnjbvr:


littledan:
I don't think this is necessary anymore with the Bikeshed conversion complete.

Issue #1144, opened 2017-11-03
🛤 Tail call


jfbastien:
**This is a tracking issue for a post-MVP feature**
It will be updated as the issue progresses.

**Topic** Tail call
**Champion** Andreas Rossberg @rossberg 
**Status** in progress
**[Phase](https://github.com/WebAssembly/meetings/blob/master/process/phases.md)** Feature proposal
**Linked issues** #936, #189, #484, 
**Linked repositories** [github.com/WebAssembly/tail-call](https://github.com/WebAssembly/tail-call)

# Details

* Currently, the Wasm design explicitly forbids tail call optimisations

* Want support to enable
  - the correct and efficient implementations of languages that require tail call elimination
  - the compilation of control constructs that can be implemented with it (e.g., forms of coroutines, continuations)
  - compilation and optimization techniques that require it (e.g., dynamic recompilation, tracing, CPS)
  - other sorts of computation being expressed as Wasm functions, e.g., FSMs

PR #1158, opened 2017-11-15
Delete unnecessary sections

josephlr:
As garbage collection, multiple return, and Memset/Memcpy Operators have
dedicated tracking issues and all the links have been fixed, we can
remove their sections. Most of the data being removed is still present
in the tracking issues.

The two sections on tail calls are not removed as they contain some
information not in the tail calls tracking issue (#1144).

### DIFF
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -133,10 +133,6 @@ Options under consideration:
   making it easier to support other languages, especially functional programming
   languages.
 
-### GC/DOM Integration
-
-See issue [1079][].
-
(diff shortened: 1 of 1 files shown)


jfbastien:
I think it would be better to move them to the bottom, header names untouched, and have a redirect to the right tracking issue. Reason being that the header correspond to anchor links, which would break if any website links to them 😞

josephlr:
@jfbastien I created a section at the bottom for old headers. I also moved all the tail call stuff to this new section and did some other minor cleanup.

PR #1170, opened 2018-01-05
【翻译】design/Portability.md

h476982844:
[issue 27](https://github.com/WebAssembly-cn/website/issues/27)


jfbastien:
It looks like this translates English pages? We need to keep English as well :)

PR #1199, opened 2018-04-08
Update BinaryEncoding.md

SpyderTL:
Clarify description of function_body length field.


sunfishcode:
In general, BinaryEncoding.md and other files in this design repo are no longer being updated. However, I think it's reasonable to merge this one PR, since it was filed in 2018 and it is just a simple clarification.

PR #1209, opened 2018-05-17
Fix broken links to GC.md

kenrick95:
Since GC.md doesn't exist, I link these to the [GitHub issue](https://github.com/WebAssembly/design/issues/1079)

### DIFF
--- a/Semantics.md
+++ b/Semantics.md
@@ -328,7 +328,7 @@ In the MVP, the primary use case of global variables is to represent
 instantiation-time immutable values as a useful building block for
 [dynamic linking](DynamicLinking.md).
 
-After the MVP, when [reference types](GC.md) are added to the set of [value types](#types),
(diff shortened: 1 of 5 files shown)


Semantics.md
```diff
@@ -326,7 +326,7 @@ In the MVP, the primary use case of global variables is to represent
 instantiation-time immutable values as a useful building block for
 [dynamic linking](DynamicLinking.md).
 
-After the MVP, when [reference types](GC.md) are added to the set of [value types](#types),
+After the MVP, when [reference types][future garbage collection] are added to the set of [value types](#types),
```
dtig: This should point to the reference types proposal as this was split out from the GC proposal after this text was added.
https://github.com/WebAssembly/reference-types

kenrick95: Updated! Thanks!

PR #1223, opened 2018-08-03
Updated conversion anchor links to correct id

hansoksendahl:
These two anchor links were referencing the wrong URL.

Updated `#datatype-conversions` to `#datatype-conversions-truncations-reinterpretations-promotions-and-demotions`


binji:
Thanks again!

PR #1230, opened 2018-09-10
fixed typo hex magic number

devwebcl:


BinaryEncoding.md
```diff
@@ -182,7 +182,7 @@ The module starts with a preamble of two fields:
 
 | Field | Type | Description |
 | ----- |  ----- | ----- |
-| magic number | `uint32` |  Magic number `0x6d736100` (i.e., '\0asm') |
+| magic number | `uint32` |  Magic number `0x61736d00` (i.e., '\0asm') |
```
yurydelendik: if it is big-endian, it needs to be `0x0061736d`. But looks like entire spec in in little-endian format.

binji: Perhaps a bit misleading, but this document does say that the value is a `uint32`, which is described as "An unsigned integer of N bits, represented in N/8 bytes in little endian order.", so the original value is correct.

The spec describes it using raw bytes: http://webassembly.github.io/spec/core/bikeshed/index.html#binary-magic

In general we haven't been making changes to this document unless there are errors, so I'd leave it as-is.

sunfishcode:
As noted above, the document isn't incorrect here.

Issue #1266, opened 2019-03-12
Emoji in design docs are broken


maxwell-clarke-intranel:
Expected behaviour:
![image](https://user-images.githubusercontent.com/39687678/54170916-71d8b700-44dd-11e9-9aeb-7d8aae8d0c96.png)
Observed behaviour:
![image](https://user-images.githubusercontent.com/39687678/54170850-1c9ca580-44dd-11e9-9fde-6140e7cacd0c.png)

GitHub has changed the location of the emoji images, so they 404 when viewed on the webassembly.org website.
When the .md files are viewed here on github, the correct behaviour can be seen.

The new image location is:
https://github.githubassets.com/images/icons/emoji/unicode/2b50.png
The old image location is:
https://assets-cdn.github.com/images/icons/emoji/unicode/2b50.png

xtuc:
I would rather replace them with UTF-8 chars

binji:
I updated the website bundles and it seems to have fixed the issue.

PR #1267, opened 2019-03-12
Fix broken GitHub emoji.

maxeonyx:
Fix broken GitHub emoji by changing them to roughly equivalent Unicode emoji.

I found the following broken emoji using the regex /:\w+?:/

- `unicorn` :unicorn:. Replaced with 🦄
- `star` :star:. Replaced with ⭐
- `cyclone` :cyclone:. Replaced with 🌀

These all render the same on Firefox/Ubuntu 18.04

### DIFF
--- a/BinaryEncoding.md
+++ b/BinaryEncoding.md
@@ -8,19 +8,19 @@ The binary encoding is a dense representation of module information that enables
 small files, fast decoding, and reduced memory usage.
 See the [rationale document](Rationale.md#why-a-binary-encoding) for more detail.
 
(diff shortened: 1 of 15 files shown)


maxeonyx:
#1266

binji:
Thanks for contributing! We haven't been updating the website for a while (honestly, I thought it wasn't working properly!) But it seems that after updating the website bundles it works now, so I think we don't need to replace these.

maxeonyx:
All good, glad it's fixed!

PR #1270, opened 2019-03-25
Fix LLVM official repo link

dduan:
In my attempt to link to the official LLVM repo, a mistake was made in
the new URL (I wrongly assumed the old mirror and the official mono-repo
has the same structure). This should fix that mistake.


binji:
Thanks! I should've checked the URL too :)

PR #1302, opened 2019-09-24
Host bindings => Interface types

pygy:


binji:
This document hasn't been kept up to date, it's probably better to redirect to https://github.com/WebAssembly/proposals, which has been.

pygy:
So you'd suggest scraping `FutureFeatures.md` (replacing it with a stub, because LinkRotSucks<sup>TM</sup>), and updating the corresponding links elsewhere in these docs?

binji:
Yes, I think that would be better. Thanks!

sunfishcode:
The table in question has now been moved to https://github.com/WebAssembly/proposals.

PR #1305, opened 2019-10-25
Added few comments on your very known project

FA-CZ:
Tried to add commenting expression.
Thank you for the project.


sunfishcode:
The proposed idea here is already covered in the existing text&mdash;programs "can't escape the sandbox", and it's important for that section to avoid ambiguous phrasing such as "basic security". And while it's true that the performance iplications of virtual memory are important, this isn't essential to understanding the security model.

PR #1311, opened 2019-11-15
Update the text on wasm64.

sunfishcode:
Discussions of wasm64 have settled on thinking of it as a per-memory flag,
rather than a per-module flag, so update this documentation accordingly.

### DIFF
--- a/FutureFeatures.md
+++ b/FutureFeatures.md
@@ -145,24 +145,13 @@ Options under consideration:
 
 ### Linear memory bigger than 4 GiB
 
-The WebAssembly MVP will support the wasm32 mode of WebAssembly, with linear
-memory sizes up to 4 GiB using 32-bit linear memory indices. To support larger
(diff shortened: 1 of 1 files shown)


aardappel:
Maybe mention something on how this would require load instructions to optionally also accept `i64` args? And once that is legal, what happens when a module containing a load from one type of index gets combined with a memory from the other? Or are we getting ahead of ourselves here? :)

sunfishcode:
I think load argument types are implied by saying the memory uses 64-bit indices. And I think there are straightforward ways to make linking work, but I'm not trying to write a full specification here; this is just FutureFeatures.md :).

binji:
This lgtm, anyone opposed to merging?

binji:
OK, merging. I started diving into more details [here](https://github.com/WebAssembly/design/issues/1325).

PR #1330, opened 2020-03-31
Create w3spec.bib

NANASHI0X74:
originally by @jendrikw, I just submitted the pull request. Does the bibtex item look fine? should the file be named w3spec.bib or maybe WASMStandard.bib or WASMSpec.bib or something like that?

### DIFF
--- a/w3spec.bib
+++ b/w3spec.bib
@@ -1,0 +1,8 @@
+@report{RossbergWebAssemblyCoreSpecification,
+  title = {{{WebAssembly Core Specification}}},
+  author = {Rossberg, Andreas},
+  date = {2019-12-05},
+  institution = {{W3C}},
+  url = {https://www.w3.org/TR/wasm-core-1/},
+  langid = {english}
+}


binji:
cc @ericprud

rossberg:
Having this makes sense, although the spec repo would be the more appropriate place for it, especially so that we can update it along with the spec itself.

Also, I should probably be listed as editor rather than author, since future editions will incorporate other people's contributions.

As for file name, perhaps `wasm-specs.bib`? Plural because we may add more versions or API specs in the future. Certainly no name misspelling Wasm in all-caps. :)

NANASHI0X74:
Is there an author then? I have the impression that the academic people I know usually want an author for sources...

rossberg:
Not necessarily. Books may either have an author or an editor, proceedings have only editors. See [here](https://www.openoffice.org/bibliographic/bibtex-defs.html).

FWIW, `report` is not an official bibtex entry type. Maybe use `misc` instead, or `book`.

Would also make sense to add a `year` entry, since `date` also isn't standard.

jendrikw:
We may want to consider having two files, one that is strictcly conforming to bibtex (using `misc` and `year`), and one for biblatex (which can use features like `date`).

NANASHI0X74:
maybe it can be one file with comments? Most people I know have a single file for bibliography anyway where they paste all their bibtex items into (including me). Not sure if that's according to best practice, if there are any.
Personally, after my first experience with Latex, I get the impression that the whole ecosystem is kind of a mess and that for many things there aren't any solid best practices 😅 That may just be me though.

sunfishcode:
This PR was replaced by https://github.com/WebAssembly/spec/pull/1204, which is now merged.

PR #1342, opened 2020-05-04
Revise  FutureFeatures.md

lygstate:


FutureFeatures.md
```diff
@@ -273,6 +273,7 @@ use cases:
   * `i32.abs_s`: signed absolute value (traps on `INT32_MIN`)
   * `i32.bswap`: sign-agnostic reverse bytes (endian conversion)
   * `i32.bswap16`: sign-agnostic, `bswap16(x)` is `((x>>8)&255)|((x&255)<<8)`
+  * `i64.bswap`: sign-agnostic reverse bytes (endian conversion, for 64bit host)
```
sunfishcode, line 276: The other opcodes (`min_s` etc.) are also listed only in `i32` form here; it seems we should either list all of them as both `i32` and `i64`, or just have the `i32` versions and say "and corresponding `i64` versions").

lygstate: Fixes with full i64 alternative

lygstate:
Any suggestion?

tlively:
I'm not sure how valuable it is to list both the i32 and i64 versions of the instructions since this whole document is mostly just food for thought, but on the other hand I don't see much of a downside either. This seems fine to me 👍 @binji

binji:
Seems OK to me as-is. As @tlively said, this document is mostly for ideas.

PR #1372, opened 2020-09-08
Update FutureFeatures to point to proposals repo

binji:


binji:
Oops, just saw that Conrad already did this :-)

PR #1403, opened 2021-03-04
FAQ: Change asm2wasm to past tense

Artoria2e5:
binaryen has removed asm2wasm in version 97 (WebAssembly/binaryen#3042), since Emscripten stopped using it. Use a fixed-version link to minimize confusion, and change to past tense for good measure.

The https://webassembly.org/getting-started/advanced-tools/ also has a reference to the same link. I think that's the [website repo](https://github.com/WebAssembly/website).

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -70,8 +70,8 @@ polyfill.
 It is also the case that polyfilling WebAssembly to asm.js is less urgent
 because of the existence of alternatives, for example, a reverse polyfill -
 compiling
-[asm.js to WebAssembly](https://github.com/WebAssembly/binaryen/blob/master/src/asm2wasm.h) -
-exists, and it allows shipping a single build that can run as either
+[asm.js to WebAssembly](https://github.com/WebAssembly/binaryen/blob/version_96/src/asm2wasm.h) -
+existed, and it allows shipping a single build that can run as either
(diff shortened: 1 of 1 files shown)


kripken (commented):
How about mentioning `wasm2js`, which has replaced `asm2wasm` in binaryen? It is basically a polyfill for wasm, and currently maintained & used in production. I'd suggest replacing the section on `asm2wasm` with that (unless we want to keep a mention of `asm2wasm` for historical reasons too).

Letsgoque (commented):
Done

mb98ooutlook-ctrl (approved):

Artoria2e5:
aaaaaaaa i have no idea why i did nothing about the wasm2js comment but i am looking into it right n...

hmm. very much still alive, so i do want to mention it now. the question is how.

1. I can just replace the "experiments" (wasm2asm.h) link with wasm2js.h. This would be consistent with the existing change, since the wasm2asm link is ALSO dead.
2. I can alternatively rewrite the whole section to say: "Yes. It exists. It is called wasm2js, is actively maintained and of production quality." There will be no need for any of the existing text really.

any idea on which to choose? i'm not in a making-choices state of consciousness right now.

Artoria2e5:
ah what the heck i can just do (1) in this PR and do (2) in a new PR. y'all can decide later.

PR #1461, opened 2022-09-27
FAQ: Fix Assorted Typos

TheConner:
- Remove duplicated "it" when talking about LLVM
- Correct "but which can..." to "but can..."
- "...this issue at point" to "this issue at some point"

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -185,12 +185,12 @@ together in a number of configurations:
 
 ## Why not just use LLVM bitcode as a binary format?
 
-The [LLVM](https://llvm.org/) compiler infrastructure has a lot to recommend it:
+The [LLVM](https://llvm.org/) compiler infrastructure has a lot to recommend:
 it has an existing intermediate representation (LLVM IR) and binary encoding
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -185,12 +185,12 @@ together in a number of configurations:
 
 ## Why not just use LLVM bitcode as a binary format?
 
-The [LLVM](https://llvm.org/) compiler infrastructure has a lot to recommend it:
+The [LLVM](https://llvm.org/) compiler infrastructure has a lot to recommend:
```
tlively, line 188: The original text here was correct. It was using the definition of "recommend"  that appears fourth here: https://www.dictionary.com/browse/recommend. This could still be clarified by rewording it to something like "has attractive qualities," though.,

tlively (approved):
LGTM % comment.

FAQ.md
TheConner: Huh I never knew that, it sounded wrong to me but as per the dictionary it's right. Rephrased as per your suggestion.

tlively (approved):

PR #1515, opened 2024-04-23
Remove non-canonical CodeOfConduct.md

tlively:
The canonical code of conduct has moved to the meetings repo. Although we generally keep the docs in this repo around to preserve history, remove the non-canonical code of conduct document to avoid confusion if it drifts out of sync with the canonical document.


tlively:
Actually, it would be safer to replace the body with a link to the new canonical COC to avoid broken links.

